PR #13, opened 2016-06-16
PEP 522: BlockingIOError in security sensitive APIs on Linux

ncoghlan:


ncoghlan:
@dstufft @njsmith Any comments before I check in this initial draft?

pep-0522.txt
```diff
@@ -0,0 +1,239 @@
...
+========
+
+This PEP proposes that in Python 3.6+, ``os.urandom()`` be updated to call
+the new Linux ``getrandom()``` syscall in non-blocking mode if available and
+raise ``BlockingIOError: system random number generator is not ready`` if
+the kernel reports that the call would block.
+
+No changes are proposed for Windows or Mac OS X systems, as neither of those
+platforms provides any mechanism to run Python code before the operating
+system random number generator has been initialised.
```
dstufft: It might be reasonable to mention somewhere that OS X finds a secure sense of random so important, that it will kernel panic rather than continue to boot if it can't get enough random during it's boot process.

pep-0522.txt
```diff
@@ -0,0 +1,239 @@
...
+                pass
+
+
+Affected Linux specific non-security sensitive applications
+-----------------------------------------------------------
+
+Non-security sensitive applications that don't need to worry about cross
+platform compatibility can be updated to access ``/dev/urandom`` directly::
+
+    def blocking_urandom(num_bytes):
```
dstufft: ``def nonblocking_urandom(num_bytes):`` I assume you meant here?

pep-0522.txt
```diff
@@ -0,0 +1,239 @@
...
+==========
+
+On operating systems other than Linux, ``os.urandom()`` may already block
+waiting for the operating system's random number generator to be ready.
+
+On Linux, even when the operating system's random number generator doesn't
+consider itself ready for use in security sensitive operations, it will return
+random values based on the entropy it as available.
+
+This behaviour is potentially problematic, so Linux 3.17 added a new
```
dstufft: Notably the new interface to the urandom pool did *not* include any way to get the old "get predictable numbers instead of block" functionality. It is not possible to insecurely use the API, you either:

* Use urandom, blocking until the pool is initialized (default)
* Use random, blocking whenever the pool thinks it's entropy is low (``GRND_RANDOM``).
* Use urandom, returning a ``-1`` and setting eerno to ``EAGAIN`` if the pool isn't initialized (``GRND_NONBLOCK``).

dstufft:
A few small nits, but 👍

ncoghlan:
I merged this to explicitly reserve the PEP number. I'll still give @njsmith a chance to comment before posting to python-dev about it.

njsmith:
I read it, and thought "hmm, I think this could flow better to make a more compelling argument", and then I started poking at the text and uh. Is a PR tomorrow good? :-)

pep-0522.txt
holdenweb: E_NOTENOUGHENTROPY? Wouldn't want the heat death of the universe to break the build ...

ncoghlan:
@njsmith I also didn't find a good place to explicitly state the assumption that random module initialisation and SipHash initialisation will be decoupled from reliance on the behaviour of os.urandom() (However, they could be incorporated into the PEP by saying that they will *try* os.urandom, and only use their fallbacks if that raises BlockingIOError)

ncoghlan:
@njsmith Just a heads up that I'm going to be working on major rewrite today - with an extra couple of days to ponder the topic, I have a different overall framing that I think will better sell the idea (although the core proposal remains the same)

njsmith:
Ah ok I'll at least throw what I've got up in a branch

njsmith:
#19

PR #23, opened 2016-06-20
PEP 522: Rewrite accounting for Nathaniel's feedback

ncoghlan:
The core proposal (raising BlockingIOError) remains the same,
but the rationale and background sections now go into much more
depth for folks that want it.

The PEP now also explicitly covers the related changes to interpreter
and random module initialization, as well as providing references
to most of the related bugs.python.org messages.

### DIFF
--- a/pep-0522.txt
+++ b/pep-0522.txt
@@ -1,8 +1,8 @@
 PEP: 522
-Title: Raise BlockingIOError in security sensitive APIs on Linux
+Title: Allow BlockingIOError in security sensitive APIs on Linux
 Version: $Revision$
 Last-Modified: $Date$
-Author: Nick Coghlan <ncoghlan@gmail.com>
+Author: Nick Coghlan <ncoghlan@gmail.com>, Nathaniel J. Smith <njs@pobox.com>
 Status: Draft
 Type: Standards Track
 Content-Type: text/x-rst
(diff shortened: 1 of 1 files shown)


pep-0522.txt
```diff
@@ -76,78 +125,84 @@ However, this guidance has also come with a longstanding caveat: developers
...
-and it takes extensive statistical analysis to show that a security
-vulnerability exists.
+operations. This generally only occurs if ``os.urandom()`` is read very
+early in the system initialization process, or on systems with few sources of
+available entropy (e.g. some kinds of virtualized or embedded systems), but
+unfortunately the exact conditions that trigger this are difficult to predict,
+and when it occurs then there is no direct way for userspace to tell it has
+happened without querying operating system specific interfaces.
+
+On \*BSD systems, encountering this situation means ``os.urandom()`` will block
```
njsmith: This is a minor nitpick, but AFAICT (could easily be misinformed on one piece of another but):
* OpenBSD: requires entropy be present before spawning the first userspace process
* FreeBSD: reads from `/dev/urandom` and `/dev/random` block if the RNG is uninitialized, and never afterwards (as described here)
* NetBSD: acts like Linux

This is why my draft avoided saying anything about `*BSD` :-)

PR #30, opened 2016-06-22
Convert pep 20 to Rst

Carreau:
I was torn between converting the Zen of Python into a list, or monospaced. It looks better monospaced.


brettcannon:
Thanks!

PR #31, opened 2016-06-22
Convert pep-0042 to Rst

Carreau:
git diff seem to get confused again, `--word-diff --ignore-all-space` is again useful but will remove most of the nesting conveying content. 

I did a manual check of the rendered version, but may have missed  somethings.


brettcannon:
Thanks!

PR #32, opened 2016-06-23
Handle pep files with `.rst` extensions.

Carreau:
Not completely sure how that would interact with deploying pep on pythondotorg and other services.

### DIFF
--- a/genpepindex.py
+++ b/genpepindex.py
@@ -41,7 +41,7 @@ def main(argv):
             abs_file_path = os.path.join(path, file_path)
             if not os.path.isfile(abs_file_path):
                 continue
-            if file_path.startswith("pep-") and file_path.endswith(".txt"):
+            if file_path.startswith("pep-") and file_path.endswith((".txt", "rst")):
                 with codecs.open(abs_file_path, 'r', encoding='UTF-8') as pep_file:
                     try:
                         pep = PEP(pep_file)
--- a/pep2html.py
+++ b/pep2html.py
@@ -31,7 +31,7 @@ Options:
 -h, --help
     Print this help message and exit.
 
-The optional arguments ``peps`` are either pep numbers or .txt files.
+The optional arguments ``peps`` are either pep numbers, .rst or .txt files.
 """
 
 from __future__ import print_function, unicode_literals
@@ -39,7 +39,6 @@ from __future__ import print_function, unicode_literals
 import sys
 import os
 import re
-import cgi
 import glob
 import getopt
 import errno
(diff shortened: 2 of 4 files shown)


pep2html.py
```diff
@@ -469,7 +471,7 @@ def pep_type_error(inpath, pep_type):
 def browse_file(pep):
     import webbrowser
     file = find_pep(pep)
-    if file.endswith(".txt"):
+    if (file.startswith('pep-') and (file.endswith(".txt") or file.endswith('.rst'))):
```
berkerpeksag: `file.endswith((".txt", ".rst"))` would be clearer.

pep2html.py
```diff
@@ -478,7 +480,7 @@ def browse_file(pep):
 def browse_remote(pep):
     import webbrowser
     file = find_pep(pep)
-    if file.endswith(".txt"):
+    if (file.startswith('pep-') and (file.endswith(".txt") or file.endswith('.rst'))):
```
berkerpeksag: `file.endswith((".txt", ".rst"))` would be clearer.

berkerpeksag:
> Not completely sure how that would interact with deploying pep on pythondotorg and other services.

I don't have time to test it this week, but it looks like a backward compatible change to me.

Thanks!

Carreau:
The only undefined behavior is if txt and rst are present.

I'll update with your suggestions later.

Carreau:
Updated as per comment.

Carreau:
You can see the difference in the generated html [here](https://github.com/python/peps/compare/c03a521ed103defbd1b5da50170cfb81105b6e59...Carreau:regen-html) , which is mostly due to the `random.choice` of banner, and (strangely) a couple of author that are not sorted in the same order in pep0. (investigating, but might be unrelated to that).

genpepindex.py
```diff
@@ -41,7 +41,7 @@ def main(argv):
             abs_file_path = os.path.join(path, file_path)
             if not os.path.isfile(abs_file_path):
                 continue
-            if file_path.startswith("pep-") and file_path.endswith(".txt"):
+            if file_path.startswith("pep-") and file_path[-4:] in (".txt", "rst"):
```
berkerpeksag: `file_path.endswith((".txt", ".rst"))` would work too.

pep2html.py
```diff
@@ -367,10 +366,13 @@ def get_input_lines(inpath):
...
 

 def find_pep(pep_str):
-    """Find the .txt file indicated by a cmd line argument"""
+    """Find the .rst or .txt file indicated by a cmd line argument"""
     if os.path.exists(pep_str):
         return pep_str
     num = int(pep_str)
+    rstpath = "pep-%04d.rst"
+    if os.path.exists(rstpath):
```
berkerpeksag: I think you forgot to pass `num` to `rstpath` here.

pep2html.py
```diff
@@ -520,32 +522,32 @@ def main(argv=None):
             browse = 1
 
     if args:
-        peptxt = []
+        peptxtrst = []
```
berkerpeksag: I'd rename it to `peps`, `pep_list` or something else (`peptxtrst` won't make sense once we convert all peps to reST).

ncoghlan:
Big +1 for the general idea, since setting the file extension will let GitHub render them nicely.

At the detailed level, I didn't have anything to add to Berker's existing comments.

ncoghlan:
That said, would it make sense to update https://www.python.org/dev/peps/pep-0012/ at the same time as making the changes to support the .rst extension?

It turns out that needs an update anyway to refer to the GitHub repo rather than hg.python.org

Carreau:
Will do both early next week.

pep2html.py
IanLee1521: Personally I prefer `pep_list`.

Carreau:
Updated as well as pep12.

brettcannon:
Thanks for this, @Carreau !

PR #42, opened 2016-06-30
PEP 287: fix typo in word frangible

hexum:
Spelling `frungible` changed to `frangible`.

### DIFF
--- a/pep-0287.txt
+++ b/pep-0287.txt
@@ -554,7 +554,7 @@ Questions & Answers
 
        Abstract
 
-           This PEP proposes adding frungible doodads [1] to the core.
+           This PEP proposes adding frangible doodads [1] to the core.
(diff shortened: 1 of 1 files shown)


hexum:
`Fungible` word could be meant here. It's not clean for me, but there is a mistake anyway.

Rosuav:
Or was it intentionally a nonsense word? Is there any particular indication that the original author intended any given meaning?

hexum:
May be.
It's not critical part of document. But readers like me stops and start trying to understand this strange word =)

brettcannon:
Why don't we ask @goodger what he meant to write? (my suspicion is it was meant to be a nonsense word).

goodger:
You are correct: it's just a nonsense word in an example.

DG

hexum:
Thanks for your answers!
Next time I will not be too serious while reading peps =)

PR #55, opened 2016-07-14
Fix spelling of "explicitly" in PEP 423

### LINKED ISSUE #54: Typo
https://github.com/python/peps/blob/master/pep-0423.txt

explicitely -> explicitly

Rosuav:
Closes #54.


Rosuav:
Already fixed.

PR #59, opened 2016-07-18
[WIP] PEP-514: Integrates feedback from python-dev.

zooba:


pep-0514.txt
```diff
@@ -13,8 +13,8 @@ Abstract
 ========
 
 This PEP defines a schema for the Python registry key to allow third-party
-installers to register their installation, and to allow applications to detect
-and correctly display all Python environments on a user's machine. No
+installers to register their installation, and to allow tools and applications
+to detect and correctly display all Python environments on a user's machine. No
 implementation changes to Python are proposed with this PEP.
```
pfmoore: I'm still struggling to understand what (in general) tools that consume this metadata would look like. There's a lack of motivating examples here. The only examples I can think of are ```virtualenv``` and "something that lists all the available installations" (but I'm not clear what it's then going to do with the user's selection). You mention Visual Studio as another consumer of this data. It would be really useful to have a description of which data VS uses, and how it exposes it to the user. And any other motivating cases you're aware of. The other use case I can think of, unfortunately obsolete, is the ```bdist_wininst``` installer, which presented a "which Python installation do you want" dialog. But I don't recall the details of that, and it's basically obsolete now anyway. As the PEP stands, it feels like a collection of things that we think would be "nice to have", but we're not sure enough of ourselves to make them mandatory.

pep-0514.txt
```diff
@@ -118,17 +138,22 @@ It is not possible to detect side-by-side installations of both 64-bit and
...
+than by preferring the user's setting.
 
 Company
 -------
 
 The Company part of the key is intended to group related environments and to
 ensure that Tags are namespaced appropriately. The key name should be
 alphanumeric without spaces and likely to be unique. For example, a trademarked
-name, a UUID, or a hostname would be appropriate::
+name (preferred), a UUID, or a hostname would be appropriate::
```
pfmoore: I'm still strongly -1 on a UUID. The ```virtualenv``` use case implies that this tag is expected to be used by the user, and UUIDs don't work in that context. (Nor do hostnames, frankly).

I'd suggest requiring this to be a short user-friendly abbreviation or name (ideally without spaces, as it's to be used on the command line). Examples could be "conda", "enthought", "jython", "pypy", "ironpython". If you want a longer form, such as your full company name, then the ```DisplayName``` value is what you should use.

I don't see name squatting as a real issue. If "Fred's Distribution" chooses to use "conda" as a company name, it simply won't work if installed alongside Anaconda. People will simply stop using it, because it "breaks their Anaconda installation" and the problem will be self-solving.

It's not like there will be so many distributions that this is going to be a big issue. Heck, the list I gave above is probably the majority of cases we'll see in reality.

pep-0514.txt
```diff
@@ -138,11 +163,13 @@ The company name ``PyLauncher`` is reserved for the PEP 397 launcher
...
-the environment category to users. Otherwise, the name of the key should be
-used.
+the environment manufacturer/developer/destributor to users. Otherwise, the name
+of the key should be used. (For ``PythonCore``, the default display name is
+"Python Software Foundation".)
 
 If a string value named ``SupportUrl`` exists, it may be displayed or otherwise
-used to direct users to a web site related to the environment.
+used to direct users to a web site related to the environment. (For
+``PythonCore``, the default support URL is "http://www.python.org/".)
```
pfmoore: While I have no objection to including SupportURL, I'd like to see an example of the intended use. At the moment, there's little obvious motivation for vendors to include this.

pep-0514.txt
```diff
@@ -158,25 +185,37 @@ The Tag part of the key is intended to uniquely identify an environment within
...
 version, a UUID or a partial/complete hash would be appropriate; an integer
-counter that increases for each new environment may not::
+counter that increases for each new environment may not. It is expected that
+some tools will require users to type the Tag into a command line, and that the
+Company may be optional provided the Tag is unique across all Python
+installations::
 
+    HKEY_CURRENT_USER\Software\Python\ExampleCorp\examplepy
     HKEY_CURRENT_USER\Software\Python\ExampleCorp\3.6
     HKEY_CURRENT_USER\Software\Python\ExampleCorp\6C465E66
```
pfmoore: Same objection here to opaque IDs like ```6C465E66``` as with the Company. No user's ever going to type ```-p 6C465E66-5A8C-4942-9E6A-D29159480C60/6C465E66```, or even ```-p examplecorp.org/6C465E66``` as an argument to virtualenv.

pep-0514.txt
```diff
@@ -187,6 +226,7 @@ may look like this::
         SupportUrl = "http://www.example.com/distro-3"
         Version = "3.0.12345.0"
         SysVersion = "3.6.0"
+        SysArchitecture = "64bit"
```
pfmoore: So a virtualenv user would use ```-p ExampleCorp\6C465E66``` to specify this distribution? That seems pretty user-hostile, and I'd expect them to just use ```-p <path to interpreter>``` instead, which pretty much defeats the purpose.

For an IDE, the above is probably OK. You present a "Select active Python" dialog, and display some of this data (Company DisplayName, tag DisplayName, Version, SysVersion and SysArchitecture would be about what I'd want to see) and the user selects the one they want.

pep-0514.txt
```diff
@@ -200,12 +240,16 @@ always named ``InstallPath``, and the default value must match ``sys.prefix``::
...
 
 If a string value named ``WindowedExecutablePath`` exists, it must be a path to
 the ``pythonw.exe`` (or equivalent) executable. Otherwise, the windowed
 interpreter executable is assumed to be called ``pythonw.exe`` and exist in the
-directory referenced by the default value.
+directory referenced by the ``(Default)`` value.
+
+Tools should be able to handle the case where the referenced executables do not
+actually exist, particularly when values for ``ExecutablePath`` and
+``WindowedExecutablePath`` are inferred from the ``(Default)`` value.
```
pfmoore: I honestly don't understand what use there is for an installation that doesn't include information on how to start the Python interpreter. I guess it's true that there isn't always a ```python.exe``` - Jython, for example, has ``jython.jar```. But tools need *some* way of knowing how to start the Python interpreter.

Maybe the two values should be ```InterpreterCommand``` and ```WindowedInterpreterCommand```. If they are not present, then ```<InstallPath>\python.exe``` and ```<InstallPath>\pythonw.exe``` should be assumed. The windowed interpreter is optional, but the "normal" interpreter command *must* be available (either as python.exe or in an explicit form). Tools shouldn't be expected to deal with a registered interpreter that they can't invoke.

Do you have an example of a use case where you need to introspect the installed Pythons, but you never actually need to run the interpreter?

pep-0514.txt
pfmoore: I should point out that I don't want to get too hung up on the specifics of how virtualenv might theoretically implement this option. It would be just as easy for us to add a user config file that allowed the user to manually specify "alias -> Python interpreter path" mappings (much like py.exe has in its ini file). And the fact that we've not done that indicates that the whole issue is low-priority. But essentially, this PEP doesn't add any value for virtualenv unless it's better than a "user maintained alias" approach.

(In practice, I don't think it's likely that this will ever be implemented for virtualenv in its current form - it's not clear that the gain is worth it - but virtualenv remains a simple concrete example of a command line interface).

pep-0514.txt
```diff
@@ -200,12 +240,16 @@ always named ``InstallPath``, and the default value must match ``sys.prefix``::
 If a string value named ``ExecutablePath`` exists, it must be a path to the
 ``python.exe`` (or equivalent) executable. Otherwise, the interpreter executable
 is assumed to be called ``python.exe`` and exist in the directory referenced by
-the default value.
+the ``(Default)`` value.
```
pfmoore: To confirm - the absolute minimum a correctly registered installation must provide is:

```
    HKEY_XXX\Software\Python\<Company>\<Tag>\InstallPath
        (Default) = "<sys.prefix>"
```

Everything else is optional?

jayvdb:
For the values of some of the keys, it would make sense to map them to PEP 496 identifiers. It would be odd if the values in the registry differed from the equivalent in the pep about environment markers.

If the registry had a standardised list of key names for all environment markers, non-python (native) apps could implement pep 496 mini language to download and install an appropriate wheel into a user selected environment that provided values for those keys, without needing to invoke the python interpreter for that environment. It might also need to know where to put the package, but that can be derived from `platform_python_implementation` and specified in a separate pep.

Rather than use virtualenv to think about this pep as an example use, even implicitly, a better example might be [pyenv](https://github.com/yyuu/pyenv) which attempts to do the same without invoking python wherever possible.  Theoretically it could be ported to a bash on windows with a few gnu binaries added.

pfmoore:
> If the registry had a standardised list of key names for all environment markers, non-python (native) apps could implement pep 496 mini language to download and install an appropriate wheel into a user selected environment that provided values for those keys, without needing to invoke the python interpreter for that environment.

Thanks. This is a good example of a non-IDE use case.

> a better example might be pyenv

I guess the key point here is ```pyenv versions``` which lists the available versions to select from. I don't know where the Unix pyenv gets its version names from, but the list shown in the example screenshot on the page you linked to:

```
2.7.10
3.5.0
miniconda3-3.16.0
pypy-2.6.0
````

is the sort of naming I'd expect in a command line utility (as opposed to the ```ExampleCorp\distro-3``` style, with no explicit version, that the PEP is currently implying).

**Comment by jayvdb** [comment]
The names are defined by the filenames in https://github.com/yyuu/pyenv/tree/master/plugins/python-build/share/python-build

**Comment by pfmoore** [comment]
OK, cool, so by pyvenv itself. That's similar to my thought about a "short alias to Company/Key mapping", and so tends to imply that it's not actually unreasonable for tools to manage their own short names.

I'm becoming more and more inclined *not* to worry about getting hung up on the format of "Company" and "Key", beyond saying that we recommend human-readable, short, descriptive values, and I'd like to establish the principle that Key should typically be a version. Ideally the PythonCore values should provide the example to follow - I'm just a little concerned that the PEP examples imply "anything goes" a bit too much.

File: pep-0514.txt (hunk 1 above)
Comment (zooba) [inline pep-0514.txt]: There's a few screenshots of Visual Studio [here](https://github.com/Microsoft/PTVS/wiki/Python-Environments), a sample of some of the extra info being shown [here](https://github.com/Microsoft/PTVS/pull/1437). We use everything in the PEP, since I fleshed out the PEP based on what we wanted to use (mostly - there are a few things we didn't bother with, such as localized text). The other IDEs have similar functionality, I just don't have handy screenshots of it..

FWIW, >90% of Windows users of Python don't seem to be using the command line ever...

Comment (pfmoore) [inline pep-0514.txt]: Cool, thanks. Maybe add those links as footnotes in the PEP?

In that "add an environment" dialog, how much is required? I ask because I'd argue that if any of the fields are mandatory, then the corresponding data item in the PEP should probably be mandatory (not "unknown").

I'm not sure about your >90% figure, we get a lot of bug reports for pip and virtualenv from Windows users. I can't imagine those typically come from IDE users. Like everything, I suspect it depends on what particular interest area you deal with, though.

Comment (zooba) [inline pep-0514.txt]: Maybe 90% is too many, but there's a *big* iceberg of Windows users who have their own internal support teams, aren't using pip/virtualenv, or aren't allowed to comment publicly on _anything_ even to report bugs. And I believe all the IDEs just act as thin wrappers around those tools and will pass bug reports on.

None of the information in that dialog is required, but if parts are missing you'll get a substandard experience. For example, we base some editor features on the language version, so if you pick the wrong one, you'll get the wrong squiggles/etc. If we don't have an executable, we can't run it. They're all fairly obvious IMO.

File: pep-0514.txt (hunk 2 above)
Comment (zooba) [inline pep-0514.txt]: For console users, a full file path (even one inside `%LocalAppData%`) is reasonable, but most Windows users don't have a clue where it's installed. I don't really expect this pep to add much value for virtualenv directly - the value really is for point-and-click types (but I also didn't want to exclude console users, hence I wasn't totally explicit about this).

I don't think malicious name squatting will be an issue, but accidental name squatting is already an issue. This PEP is meant to be guidance to avoid accidental collisions in the registry, so recommending unique and/or defensible names is more important than convenient user abbreviations.

I can see a use for having a hash of a repository URL and a commit ID as a Company/Tag pair in a distributed build system, for example. In this case, you'd almost certainly be passing around the values automatically.

File: pep-0514.txt (hunk 3 above)
Comment (zooba) [inline pep-0514.txt]: Something like "A console app may print this to the console while a GUI app may display it on the screen"? Seems too obvious to be worth the text :)

File: pep-0514.txt (hunk 4 above)
Comment (zooba) [inline pep-0514.txt]: See above - there may be reasons to pass it around automatically. (And now twenty seconds later I can think of two large companies with internal Python app distribution systems that might be able to take advantage of this...)

File: pep-0514.txt (hunk 5 above)
Comment (zooba) [inline pep-0514.txt]: Alternatively, a virtualenv user might be able to provide the index into a named list. But if all virtualenv needs is the path and the user is expected to figure that out on their own, that's also fairly hostile.

(Or virtualenv could display a list of all known paths to installed Python interpreters, which would not be reliable without this PEP, regardless of the format of the IDs)

File: pep-0514.txt (hunk 6 above)
Comment (zooba) [inline pep-0514.txt]: That suggestion is basically identical to what is there now, except it adds an extra burden on a Python distribution to include a command-line entry point.

The Python distro for [Windows IoT Core](https://github.com/ms-iot/python/releases) doesn't have a normal entry point, as it is supposed to be deployed to a remote device. Given the choice between "I can register this without an InterpreterCommand" and "I have to invent some new way to register this", I'd rather people chose the first and tools learned to deal with it. (This becomes more important when there are multiple side-by-side versions of such a package.)

You may also want to register the install location of Python embedded in Blender or another app, since there's a working standard library and (IIRC) a python**.dll that could be used.

Perhaps InterpreterCommand should not have a fallback (except for PythonCore <= 3.5). Tools that need to be able to launch a process can simply ignore those. (I don't want to get into narrow feature registration -- that's a PEP of its own.)

File: pep-0514.txt (hunk 7 above)
Comment (zooba) [inline pep-0514.txt]: Yep!

**Comment by zooba** [comment]
I would rather Key not include a specific version unless it can be installed side-by-side with others. e.g. `miniconda3` is sufficient, because nobody should have `miniconda3-3.16.0` and `miniconda3-3.17.0` simultaneously. We don't have `PythonCore\3.5.1` and `PythonCore\3.5.2` - it's just `3.5`.

Not sure if it was missed (or if I edited it out), but I suggested at one point that tools like pyenv/virtualenv should probably allow use of just the Tag to identify an installation when it's unambiguous. So given:

```
PythonCore\2.7
PythonCore\3.5
WinPython\3.5
Continuum\anaconda3
```

You could do:

```
virtualenv -p 2.7
# success!

virtualenv -p anaconda3
# success!

virtualenv -p 3.5
# failure, probably with a printout of
Did you mean:
  PythonCore\3.5
  WinPython\3.5
```

I could add this as an example, but I'm really hesitant to add anything that appears to dictate UI. If a tool wants to resolve in some other way, that's up to them and their users (bearing in mind that most tools will be internal to various companies and never see the light of day).

File: pep-0514.txt (hunk 3 above)
Comment (zooba) [inline pep-0514.txt]: I'll add a note to the abstract instead discussing intended use of the information in general, and pointing out why I don't refer to any specific UI throughout the PEP.

File: pep-0514.txt [hunk 8]
```diff
@@ -30,7 +30,9 @@ Motivation
 When installed on Windows, the official Python installer creates a registry key
 for discovery and detection by other applications. This allows tools such as
 installers or IDEs to automatically detect and display a user's Python
-installations.
+installations. For example, ``virtualenv`` supports a ``-p`` argument to use an
+alternative interpreter, but is unable to provide a list of all values that may
+be provided by a user.
```
zooba: Just noticed that this hypothetical motivation isn't as motivating as the fact that py.exe and some IDEs are already trusting the information, despite the potential for collisions. So I'm changing this around slightly to emphasise that we're fixing a problem here, not adding a new feature.

zooba:
I'm +0 right now on adding an "Illustrative Examples" section, both with some registry layouts and code for reading them (and also a `GetBinaryType` call to handle 32-bit vs 64-bit detection without running the interpreter).

My main hesitation is that I don't want examples to become binding (i.e. "you're only valid if the code in the PEP displays everything"), especially since (by design), practically everything in the PEP is optional.

pfmoore:
OK. Thanks for the updates and clarifications. I think they help a lot.

I don't honestly think there's much problem with adding examples, and I think it would help a lot. By all means add a disclaimer that the example code is for illustration only. I do think that an example should cater for missing data (run the interpreter -> look for the ExecutablePath value, if not, try InstallPath\python.exe, but allow for that exe not to exist), as IMO an example that shows what proper error handling looks like will stop people from making unwarranted assumptions in *their* production code. (If the resulting example is ugly and over-complex, that may say something about the spec, of course :-)) Personally, I'd like to see some code in here.

At the moment, I'd characterise this PEP as doing a few things:

1. Giving a clear message to non-python.org distributions that they should not use our registry space, and providing an alternative for them,
2. Agreeing the basic ```Company\Key``` structure for finding ```sys.prefix```.
3. Suggesting some informational data that distributions can optionally supply, and tools can optionally use (the intended use being for information display, likely in a GUI).
4. Providing a little more detail on historically how the python.org installers used the registry.

Within that scope, I'm happy with the PEP. I think I initially expected more from it - and I still think it's a shame that it doesn't go further in terms of "introspection without needing to run the interpreter" - but it achieves the above goals, which is great.

If you can decide on whether you're willing to add the example code, then I think we're there. I'll post a last call for comments on python-dev, for people who haven't been following the tracker discussion, and then we can sign this off.

zooba:
Added hopefully the final textual clarifications and the first example. It's a shame that dealing with the registry is so messy in Python - there's far more indentation there than I'd like, and it doesn't even include the backwards compatibility handling or most of the metadata.

Will try and get to a couple more examples over the coming week, but wanted to push it online in case I change machine before I next get a chance to work on it.

pfmoore:
Thanks! This looks pretty good now. Once you've completed the changes you want to make, I'm happy to sign this off (on a procedural note, I'm not sure if it needs a posting of the final version to python-dev, or if we can just refer back to here. I'll check on that.)

zooba:
I just finished another couple of examples, including the values that will be written for a 3.6 install (once the PEP is accepted and I update the installer, that is).

I think we should merge the changes and post one more time on python-dev. That way we'll have the mailing list trail of acceptance and I can do the final update here with the usual links.

pfmoore:
Sounds fine to me. Let's do that and unless anyone comes up with a last-minute objection (!), I'll formally accept the PEP on python-dev.

I'm also in the process of posting a note to python-committers suggesting that we update PEP 1 to cover this process, so it's formally documented.

PR #63, opened 2016-07-28
[WIP] Add scoping rules for type variables to PEP 484

ilevkivskyi:
Fixes https://github.com/python/typing/issues/249 
@gvanrossum please review

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -448,6 +448,63 @@ is not generic but implicitly inherits from ``Iterable[Any]``::
 Generic metaclasses are not supported.
 
 
+Scoping rules for type variables
+--------------------------------
+
+Type variables follow normal name resolution rules.
+However, there are some special cases in the static typechecking context:
+
+* A type variable could be inferred to be equal to different types in a block,
+  unless this block is a generic class definition block. In such case,
+  all uses of a type variable are bound to the type that
+  parameterizes the generic class. Example::
+
+    from typing import TypeVar, Generic
+
+    T = TypeVar('T')
+
+    def fun_1(x: T) -> T: ... # T here
+    def fun_2(x: T) -> T: ... # and here could be different, see below
+
+    class MyClass(Generic[T]):
+        def meth_1(self, x: T) -> T: ... # but T here
+        def meth_2(self, x: T) -> T: ... # and here are always the same
+
+    fun_1(1)   # This is OK, T is inferred to be int
+    fun_2('a') # This is aslo OK, now T is str
+
+    a = MyClass() # type: MyClass[int]
+    a.meth_1(1)   # OK
+    a.meth_2('a') # This is an error!
+
+* Free type variables should not appear in the bodies of generic functions.
+  As well, types of class attributes should not contain free type variables::
+
+    T = TypeVar('T')
+    S = TypeVar('S')
+    class Foo(Generic[T]):
+        def method(self, x: T, y: S) -> S: # This is an error!
+            ...
+
+* A generic class definition that appears inside a generic function
+  should not use type variables that parametereize the generic function::
+
+    from typing import List
+
+    def a_fun(x: T) -> None:
+
+        # This is OK
+        a_list = [] # type: List[T]
+        ...
+
+        # This is however illegal
+        class MyGeneric(Generic[T]):
+            ...
+
+* A generic class nested in another generic class cannot use the same
+  type variables, unless the inner class definition is inside a function.
+
+
 Instantiating generic classes and type erasure
 ----------------------------------------------
 
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -497,7 +497,7 @@ However, there are some special cases in the static typechecking context:
     y = x.method(0, "abc") # inferred type of y is str
 
 * Unbound type variables should not appear in the bodies of generic functions,
-  and in the class bodies appart from method definitions::
+  or in the class bodies apart from method definitions::
 
     T = TypeVar('T')
     S = TypeVar('S')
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username at [bugs.python.org](http://bugs.python.org/) (b.p.o). If you don't already have an account at b.p.o, please [create one](http://bugs.python.org/user?@template=register) and make sure to add your GitHub username. If you do already have an account at b.p.o then please go there and under "Your Details" add your GitHub username.

And in case you haven't already, please make sure to sign the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA); we can't legally look at your contribution until you have signed the CLA.

ilevkivskyi:
@the-knights-who-say-ni I have added my github username on bugs.python.org

pep-0484.txt
```diff
@@ -448,6 +448,63 @@ is not generic but implicitly inherits from ``Iterable[Any]``::
 Generic metaclasses are not supported.
 
 
+Scoping rules for type variables
+--------------------------------
+
+Type variables follow normal name resolution rules.
+However, there are some special cases in the static typechecking context:
+
+* A type variable could be inferred to be equal to different types in a block,
```
gvanrossum: This section feels unclear, though I'm not sure how to say it right myself.

The key point is that the occurrence of a type variable in a method signature can mean one of two things:

1. a generic function;
2. a method of a generic class.

The disambiguating rule is that if the class of which this is a method is generic over a type variable, then the type variable refers to the class's type parameter; if not, it is a "free" type variable, making the function generic. (I'm making up the term "free" here.)

The other rule, which you are describing in this same section, is that free type variables occurring in different functions are unconnected. (I think the example showing this should be shown first, separately, before discussing type variables belonging to a generic class.)

pep-0484.txt
```diff
@@ -448,6 +448,63 @@ is not generic but implicitly inherits from ``Iterable[Any]``::
...
+  As well, types of class attributes should not contain free type variables::
+
+    T = TypeVar('T')
+    S = TypeVar('S')
+    class Foo(Generic[T]):
+        def method(self, x: T, y: S) -> S: # This is an error!
+            ...
+
+* A generic class definition that appears inside a generic function
+  should not use type variables that parametereize the generic function::
```
gvanrossum: parameterize

pep-0484.txt
```diff
@@ -448,6 +448,63 @@ is not generic but implicitly inherits from ``Iterable[Any]``::
...
+    a.meth_1(1)   # OK
+    a.meth_2('a') # This is an error!
+
+* Free type variables should not appear in the bodies of generic functions.
+  As well, types of class attributes should not contain free type variables::
+
+    T = TypeVar('T')
+    S = TypeVar('S')
+    class Foo(Generic[T]):
+        def method(self, x: T, y: S) -> S: # This is an error!
```
gvanrossum: Is it really? I tried this and it seems fine. E.g. if we have
```python
x = Foo()  # type: Foo[int]
y = x.method(0, "abc")
```
The inferred type of y is `str`.

pep-0484.txt
gvanrossum: parameterize

pep-0484.txt
ilevkivskyi: I tried to reformulate this by splitting it into three separate rules. Please check if it is more clear now.

pep-0484.txt
ilevkivskyi: Actually, I also tend to think that it is OK, but @rwbarton in https://github.com/python/mypy/issues/1903 says that it should be prohibited (as I understand, the reason is to avoid errors with forgotten type variables in generic class declaration). Maybe I misunderstood him.
I removed this part almost completely, if necessary we could add it later.

I only left a restriction that prohibits "free" type variables in the bodies (see new version).

ilevkivskyi:
@gvanrossum The new version is ready for review now

pep-0484.txt
```diff
@@ -448,6 +448,92 @@ is not generic but implicitly inherits from ``Iterable[Any]``::
...
+    S = TypeVar('S')
+    class Foo(Generic[T]):
+        def method(self, x: T, y: S) -> S:
+            ...
+
+    x = Foo() # type: Foo[int]
+    y = x.method(0, "abc") # inferred type of y is str
+
+* Unbound type variables should not appear in the bodies of generic functions,
+  and in the class bodies appart from method definitions::
```
gvanrossum: "or". Also, "apart".

pep-0484.txt
```diff
@@ -448,6 +448,92 @@ is not generic but implicitly inherits from ``Iterable[Any]``::
...
+        # this is OK
+        y = [] # type: List[T]
+        # but below is an error!
+        y = [] # type: List[S]
+
+    class Bar(Generic[T]):
+        # this is also an error
+        an_attr = [] # type: List[S]
+
+        def do_something(x: S) -> S: # this is although OK
```
gvanrossum: "this is OK" (or "this is OK though")

pep-0484.txt
```diff
@@ -1297,7 +1383,7 @@ from ``UserId`` where ``int`` is expected. Examples::
 
         num = UserId(5) + 1     # type: int
 
-``NewType`` accepts only one argument that shoud be a proper class,
+``NewType`` accepts only one argument that should be a proper class,
```
gvanrossum: Hm, while you're at it, can you rephrase this some more? NewType takes 2 args, so this sentence is confusing.

gvanrossum:
That's all I have!

ilevkivskyi:
@gvanrossum I implemented all the comments from the second review in the new version.

PR #77, opened 2016-08-28
Update pep 345 to reference pep 440 that supersede pep 386.

Carreau:
Pep 386 - "Changing the version comparison module in Distutils" has been
replaced by Pep 440 - "Version Identification and Dependency
Specification"

Pep 345 still refer to pep 386, I suppose it is meant to reference pep
440 now.

--- 

The only "gotcha" is that pep 345 says that `Requires-Python = 3` is valid: 

> Requires-Python: 3 : Any Python 3 version, no matter wich one, excluding post or pre-releases.

That it must comply to the version specifiers:

> Version numbers must be in the format specified in Version Specifiers .

Which should match pep 386, now superseded by 440, in which `3` alone is not valid. 

So peps are contradicting themselves.

### DIFF
--- a/pep-0345.txt
+++ b/pep-0345.txt
@@ -68,7 +68,7 @@ Version
 :::::::
 
 A string containing the distribution's version number.  This
-field  must be in the format specified in PEP 386.
+field  must be in the format specified in PEP 440.
 
 Example::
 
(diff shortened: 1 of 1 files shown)


Rosuav:
Looks fine to me. The other part can be fixed separately if needed.

PR #78, opened 2016-08-28
'Fix typo: wich -> which'

Carreau:
---

In pep 345; I could have changed it into [a witch](https://www.youtube.com/watch?v=zrzMhU_4m-g) but I think it was not the intent.


berkerpeksag:
Thanks!

PR #96, opened 2016-09-07
[WIP] Minor additions to PEP 526 following discussions on python-dev

ilevkivskyi:
Attn @gvanrossum 
In this PR:
* Additions proposed on python-dev
* Replaced word "declaration" with "annotation" everywhere (and added a note in non-goals), it seems that some people are really annoyed by this word.


gvanrossum:
Thanks! Small enough that I don't feel the need for another official "publish" round.

PR #106, opened 2016-09-19
Change emoji used in PEP 505

rowillia:
The ✊🍆 emoji has an offensive connotation that would not be very welcoming to new developers.

### DIFF
--- a/pep-0505.txt
+++ b/pep-0505.txt
@@ -466,25 +466,25 @@ improved.
 
 This code *might* be improved, though, if there was a syntactic shortcut for
 this common need to supply a default value. We'll assume the fictitious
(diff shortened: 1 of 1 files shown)


Rosuav:
Heh. Isn't the whole point of that choice to *avoid* discussions about the spelling of the operator?

gvanrossum:
Thanks for fixing this! Nick C and I had noticed the same thing. (@Rosuav: no worries, the PR doesn't commit to a viable spelling -- it just chooses a less offensive emoji.)

Rosuav:
Fair enough. Hopefully this one will _really_ be non-controversial :)

PR #111, opened 2016-09-30
Updates for PEPs 484 and 483

ilevkivskyi:
@gvanrossum Please take a look. In this PR:
* More precise description of ``Any`` following recent discussion on ``Union[Any, X]``
* Updated properties of ``Union`` following same discussion
* Added an exception that unparameterized generics still could be used with instance and class checks (basically as a drop-in replacement for ``collections.abc``)


gvanrossum:
Thanks as always!

PR #120, opened 2016-10-21
Soften restriction for runtime generics in PEP 484

ilevkivskyi:
Fixes https://github.com/python/typing/issues/303

@gvanrossum Please take a look. I made a PR to allow this in mypy https://github.com/python/mypy/pull/2302 . Here I just propose to soften the restriction, i.e., avoid "must".

As a motivation, in Python one always can substitute expressions, so that if ``IntNode = Node[int]; IntNode()`` works, then it is reasonable to also allow ``Node[int]``, but say that the first way is preferred.


gvanrossum:
LGTM, I'll merge this once the mypy PR is merged.

gvanrossum:
Yay!

PR #121, opened 2016-10-23
Convert PEP 202 to reST

Mariatta:
as per #4 
Let me know if this works.
Thanks 😃

### DIFF
--- a/pep-0202.txt
+++ b/pep-0202.txt
@@ -5,32 +5,38 @@ Last-Modified: $Date$
 Author: barry@python.org (Barry Warsaw)
 Status: Final
 Type: Standards Track
+Content-Type: text/x-rst
 Created: 13-Jul-2000
 Python-Version: 2.0
 Post-History:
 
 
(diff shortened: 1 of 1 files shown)


gvanrossum:
Let's see if this shows up right on the website! Give it a few minutes.

gvanrossum:
Whee!

Mariatta:
:tada: Thanks so much, Guido! :smile:

gvanrossum:
Did you write a script to do this or did you do this one by hand?

Mariatta:
I did it the low-tech way (manual conversion) :sweat_smile: 
Let me see what I can do about writing a conversion script :thinking: 
PEP 3144 can be a good test case https://www.python.org/dev/peps/pep-3144/
What do you think?

brettcannon:
I think PEP 3144 seems reasonable.

gvanrossum:
+1. I'd be happy to review!

PR #140, opened 2016-11-24
517: Add user parameter to install_editable hook (Option 1a)

takluyver:
This is one option for fixing the problem I believe exists with `--user` editable installs.

### DIFF
--- a/pep-0517.txt
+++ b/pep-0517.txt
@@ -214,12 +214,11 @@ Mandatory.
 
 ::
 
-  def install_editable(prefix, config_settings, metadata_directory=None):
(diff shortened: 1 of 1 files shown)


brettcannon:
There's a conflict now, but I merged #141 without realizing you had multiple PRs in-progress (there was no `WIP` in the title). Should this just be closed?

takluyver:
Yep, this and #141 were alternatives, and I think we're in agreement that #141 was the preferable option: I don't think we can specify editable installs to everyone's satisfaction without a fair bit of discussion.

PR #147, opened 2016-12-02
Document `__foo` syntax for positional-only args

sixolet:
This resolves https://github.com/python/typing/issues/336

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -1268,6 +1268,21 @@ In such cases the default value may be specified as a literal
 ellipsis, i.e. the above example is literally what you would write.
 
 
+Positional-only arguments
+-------------------------
+
+Some functions are designed to take their arguments only positionally,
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Unfortunately our records indicate you have not signed a [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA). For legal reasons we need you to sign this before we can look at your contribution.

sixolet:
I have done the necessary!  But only a few minutes ago, maybe systems have not caught up.

gvanrossum:
LGTM, can be merged once the CLA signing has been recorded (I can't recall if there's a manual step somewhere in the process still).

brettcannon:
@gvanrossum There's a manual step where @ejodlowska logs the fact that the CLA was received on bugs.python.org, so you should usually give it a day before checking.

PR #162, opened 2016-12-20
PEP 487: fixed typo

amPerl:
Found a typo in PEP 487. Not sure if this is the proper way to go about reporting/fixing it, but here it is regardless.


the-knights-who-say-ni:

Unfortunately our records indicate you have not signed a [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA). For legal reasons we need you to sign this before we can look at your contribution.

Rosuav:
Trivial change, so I'm overriding the CLA bot. Thanks amPerl!

PR #166, opened 2016-12-24
Mandate that manylinux1 wheels be built without --with-fpectl

njsmith:


Rosuav:
Note that this relates to a python-dev post. Read that for more context.

njsmith:
Right, that thread starts here: https://mail.python.org/pipermail/python-dev/2016-December/147065.html

The thread is more about what to do in the future, though. For PEP 513 purposes, our mandate to maximize compatibility with already-released CPythons means that there's only one workable option, which happens to match what everyone is already doing, so this text just formalizes that.

brettcannon:
Merged!

Issue #168, opened 2016-12-27
Broken links


sofigern:
link there is returning 404 https://github.com/python/peps/blob/master/pep-3107.txt#L278

sofigern:
The same about https://github.com/python/peps/blob/master/pep-3107.txt#L281

brettcannon:
Any interest in trying to find out if the relevant pages live somewhere else and submit a PR to update the links?

brettcannon:
If there's no live site that they were moved to then archived pages are fine.

PR #169, opened 2017-01-01
Update outdated links in PEP 3107 with archived pages

Mariatta:
Fixes https://github.com/python/peps/issues/168

### DIFF
--- a/pep-3107.txt
+++ b/pep-3107.txt
@@ -275,10 +275,10 @@ References and Footnotes
    http://mail.python.org/pipermail/python-3000/2006-May/002103.html
 
 .. [#typecheck]
-   http://oakwinter.com/code/typecheck/
+   http://archive.is/dcuD
 
 .. [#maxime]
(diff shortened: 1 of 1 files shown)


Rosuav:
Good point. If it dies, people can come look through the git/hg history to find the actual URL - won't be too hard.

:+1:

brettcannon:
Thanks!

gvanrossum:
Actually I would have recommend using www.archive.org which *is* a
respected site (assuming it does have the desired pages archived).

Mariatta:
😞 :woman_facepalming: Eh .... so ...  maybe earlier I wasn't trying hard enough to look for the archived version for the first link on archive.org 😓  🙇‍♀️ 
sorry for that 😅 
I found it now 🎉 
http://web.archive.org/web/20070730120117/http://oakwinter.com/code/typecheck/

fixed in https://github.com/python/peps/pull/170

Issue #4, opened 2016-06-15
Convert text-based PEPs to reST


brettcannon:
Not really important, but it does unify things.

Carreau:
Stupid existential question, should pep-9 (Sample Plaintext PEP Template) be converted to rst ?

Rosuav:
Haha, good question. Maybe keep it as text and identify it as deprecated?

Carreau:
> Haha, good question. Maybe keep it as text and identify it as deprecated?

It is already. But if not converted, that would make it trickier to automatize the Sphinx build process. 

One possibility is to put the all content of the pep in code block. Indented by 4 spaces, and add an warning on top.

brettcannon:
That's a possibility. I would leave that PEP for last for conversion as it's obviously the least important one to convert.

brettcannon:
#142, #143, #144, and #148 are conversions from text to reST.

Mariatta:
Not including PEP 9, there are 133 more PEPs still in plain text :P

gvanrossum:
So my strategy would be to take each next conversion as an opportunity to
refine your conversion script. After 33 improvements the rest should not
need manual tweaking.

brettcannon:
Once you're comfortable that your script isn't going to break anything, @Mariatta , then I would be fine with one big PR or batches of PRs that converts several of the text-only PEPs at once. We have version control for a reason. 😄

Carreau:
For the few peps I made into rst I [used this](https://gist.github.com/Carreau/b287bd42405875ec1d30db94ecf7c45e), if it can be of help. I found that a couple of peps/peps author were using different style and the you regularly need to make manual check/fixes after auto transform.

brettcannon:
I just ran `grep -i -L "^Content-Type: text/x-rst" *.txt` over repo and this is what I got:

pep-0000.txt
pep-0009.txt
pep-0100.txt
pep-0101.txt
pep-0102.txt
pep-0205.txt
pep-0207.txt
pep-0209.txt
pep-0214.txt
pep-0225.txt
pep-0227.txt
pep-0230.txt
pep-0231.txt
pep-0234.txt
pep-0242.txt
pep-0245.txt
pep-0246.txt
pep-0252.txt
pep-0253.txt
pep-0255.txt
pep-0262.txt
pep-0275.txt
pep-0276.txt
pep-0280.txt
pep-0282.txt
pep-0285.txt
pep-0293.txt
pep-0296.txt
[...]

PR #180, opened 2017-01-14
Convert 10 PEPs to reSt

Mariatta:
PEP 212
PEP 216
PEP 219
PEP 228
PEP 235

PEP 269
PEP 288
PEP 312
PEP 315
PEP 3102

https://github.com/python/peps/issues/4

### DIFF
--- a/pep-0212.txt
+++ b/pep-0212.txt
@@ -38,13 +38,13 @@ Loop counter iteration
 ======================
 
 The current idiom for looping over the indices makes use of the
-built-in 'range' function::
+built-in ``range`` function::
 
     for i in range(len(sequence)):
         # work with index i
 
 Looping over both elements and indices can be achieved either by the
-old idiom or by using the new 'zip' built-in function [2]_::
+old idiom or by using the new ``zip`` built-in function [2]_::
 
     for i in range(len(sequence)):
         e = sequence[i]
@@ -64,12 +64,12 @@ non-reserved keyword, the other adds two built-in functions.
 A third solution adds methods to sequence objects.
 
 
-Non-reserved keyword 'indexing'
-===============================
+Non-reserved keyword ``indexing``
+=================================
 
 This solution would extend the syntax of the for-loop by adding
-an optional '<variable> indexing' clause which can also be used
-instead of the '<variable> in' clause.
+an optional ``<variable> indexing`` clause which can also be used
+instead of the ``<variable> in`` clause.
 
 Looping over the indices of a sequence would thus become::
 
@@ -82,10 +82,10 @@ Looping over both indices and elements would similarly be::
         # work with index i and element e
 
 
-Built-in functions 'indices' and 'irange'
-=========================================
+Built-in functions ``indices`` and ``irange``
+=============================================
 
-This solution adds two built-in functions 'indices' and 'irange'.
+This solution adds two built-in functions ``indices`` and ``irange``.
 The semantics of these can be described as follows::
 
     def indices(sequence):
@@ -111,8 +111,8 @@ over the indices and over both elements and indices::
 Methods for sequence objects
 ============================
 
-This solution proposes the addition of 'indices', 'items'
-and 'values' methods to sequences, which enable looping over
+This solution proposes the addition of ``indices``, ``items``
+and ``values`` methods to sequences, which enable looping over
 indices only, both indices and elements, and elements only
 respectively.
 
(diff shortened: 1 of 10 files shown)


pep-0212.txt
```diff
@@ -5,163 +5,180 @@ Last-Modified: $Date$
...
-    For all three solutions some more or less rough patches exist
-    as patches at SourceForge:
+For all three solutions some more or less rough patches exist
+as patches at SourceForge:
 
-        'for i indexing a in l': exposing the for-loop counter[3]
-        add indices() and irange() to built-ins[4]
-        add items() method to listobject[5]
+- 'for i indexing a in l': exposing the for-loop counter [3]_
+- add ``indices()`` and ``irange()`` to built-ins [4]_
```
berkerpeksag: I suggest using \`\`...\`\` formatting consistently (or don't use it at all) CONSTANTs and function()s aren't the only things that can be wrapped by \`\`...\`\`. If you change irange() to \`\`irange()\`\`, I'd expect the following sentence to be formatted too:

```
This solution adds two built-in functions 'indices' and 'irange'.
```

I guess you have a script that converts these PEPs to reST format, but It would be nice to do some proofreading before sending a PR.

brettcannon: Yep, @Mariatta has written a script to do the conversion.

gvanrossum: While I think a moderate amount of proofreading and manual fix-up is warranted, I don't think the formatting needs to be perfect. We're mostly trying to improve the rendering of old historical documents, and if occasionally something isn't using a code font that is code, that's fine with me. Someone else can do some proofreading. The most important thing is that we don't accidentally break the meaning of a PEP.

Mariatta:
Thanks for the review, @berkerpeksag @brettcannon and @gvanrossum :)

I've gone through these again, and made these changes:

- PEP 212:
  replace the single quotes (') with backticks (``)  for:
  'range', 'zip', 'indices', 'irange'

  Fix typo: dicitionaries -> dictionaries

- PEP 216:
  poplib module -> ``poplib`` module

- PEP 228:
  math module -> ``math`` module
  cmath module -> ``cmath`` module

- PEP 312:
  Inline "if" -> Inline ``if``

  Add reference url to a python-dev discussion about PEP 308.

Please let me know if you notice anything else that needs adjustment.

brettcannon:
Since you already address the backtick request from @berkerpeksag and I'm with Guido that as long as the content's meaning doesn't shift I'm not too worried about minor formatting, I'm merging this. Thanks @Mariatta !

Mariatta:
Thanks :)

berkerpeksag:
Here are some additional comments:

* \`something\` is not a valid markup for making something italic. We have a rule for this in the linter we use for CPython documentation
* In PEP 235, the paragraph starts with

  ```
  then if Python finds ``FiLe.py`` first [...]
  ```

  is part of the first item of the list.
* docutils will convert the following bullet points to numbered ones:

  ```
  a. foo
  b. bar

  A. spam
  B. eggs
  ```

  will be rendered as:

  ```
  1. foo
  2. bar

  1. spam
  2. eggs
  ```

  This might look harmless most of the time, but for example, in PEP 235,
  the following sentences refer to them as #A and #B:

  ```
  #B is the same rule as is used on Unix, so this will improve cross-
  platform portability.
  ```

  and

  ```
  #A is needed to cater to case-destroying filesystems mounted on Windows,
  and *may* also be used by people so enamored of "natural" Windows
  behavior that they're willing to set an environment variable to
  get it.
  ```

  See https://www.python.org/dev/peps/pep-0235/#proposed-semantics

https://github.com/python/peps/commit/d5583179f488478404f3d7c62ef85eb0c571e2be and https://github.com/python/peps/commit/e511fb5c14096fad6c927fec03ffb109bbbda932 should address some of these comments, but the last one still needs to be fixed.

Mariatta:
Sorry that I did not notice these mistakes before, and thanks for fixing them @berkerpeksag 

> but the last one still needs to be fixed.

Any suggestion on how to fix this? 
My initial thought is to replace #A and #B with #1 and #2. But then the proceeding paragraph is making a reference to a different #2.

i.e. 
> The potential damage is here: #2 (matching on ALLCAPS.PY ) is proposed to be dropped.

brettcannon:
I don't think keeping the references for pointing to specific bullet points is critical. If you would use an explanatory shorthand that does the same thing, e.g. "The second rule is the same rule ..."

PR #190, opened 2017-01-27
PEP 484: add an example of iterable unpacking with type comments

ilevkivskyi:
Fixes https://github.com/python/typing/issues/363

@gvanrossum Please, take a look.


gvanrossum:
Thanks! (BTW I am recovering from the flu and passing on most code reviews.)

ilevkivskyi:
Great! I am glad to hear that you feel better.

PR #202, opened 2017-02-09
Fix typos and add backticks for PEP-3156.

zhangyangyu:


gvanrossum:
Thanks!

--Guido (mobile)

PR #209, opened 2017-02-13
Fix the 'venv' target

gvanrossum:
Install virtualenv if needed (assuming pip exists), and use virtualenv, not venv.


zware:
I'd suggest to instead just change `PYTHON=python` to `PYTHON=python3`.

gvanrossum:
OK, see #210 instead.

PR #211, opened 2017-02-13
Clarify my preferences around trailing commas

gvanrossum:
I realized I have some opinions and guidance around trailing commas. Here they are. I guess the next flake8 can add some rules around this.


brettcannon:
LGTM

PR #218, opened 2017-03-04
PEP 484: Add NoReturn type

ilevkivskyi:
As discussed in https://github.com/python/typing/issues/165, it is time to document the ``NoReturn`` type.

@gvanrossum Please take a look when you will have time.

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -1063,6 +1063,54 @@ in turn, to ``collections.abc.Callable``::
       ...
 
 
+The ``NoReturn`` type
+---------------------
+
+``typing`` module provides a special type ``NoReturn`` to annotate functions
+that never return normally. For example, a function that unconditionally
+raises an exception::
+
+  from typing import NoReturn
+
+  def stop() -> NoReturn:
+      raise Exception('no way')
+
+Such return annotation is used in ``typeshed`` stubs for functions
+such as ``sys.exit``. Static type checkers will ensure that functions
+annotated as returning ``NoReturn`` truly never return, either implicitly
+or explicitly::
+
+  import sys
+  from typing import NoReturn
+
+    def f(x: int) -> NoReturn:  # Error, f(0) implicitly returns None
+        if x != 0:
+            sys.exit(1)
+
+The checkers will also recognize that the code after calls to such functions
+is unreachable and will behave accordingly::
+
+  # continue from first example
+  def g(x: int) -> int:
+      if x > 0:
+          return x
+      stop()
+      return 'whatever works'  # No error reported by checkers that
+                               # ignore errors in unreachable blocks
+
+The ``NoReturn`` type is only valid as a return annotation of functions,
+and considered an error if appears in other positions::
+
+  from typing import List, NoReturn
+
+  # All of the following are errors
+  def bad1(x: Noreturn) -> int:
+      ...
+  bad2 = None  # type: NoReturn
+  def bad3() -> List[NoReturn]:
+      ...
+
+
 The type of class objects
 -------------------------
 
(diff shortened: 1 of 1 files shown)


pep-0484.txt
```diff
@@ -1063,6 +1063,54 @@ in turn, to ``collections.abc.Callable``::
       ...
 
 
+The ``NoReturn`` type
+---------------------
+
+``typing`` module provides a special type ``NoReturn`` to annotate functions
```
DimitrisJim: Missing a "The" in the beginning here.

pep-0484.txt
```diff
@@ -1063,6 +1063,54 @@ in turn, to ``collections.abc.Callable``::
...
+  def g(x: int) -> int:
+      if x > 0:
+          return x
+      stop()
+      return 'whatever works'  # No error reported by checkers that
+                               # ignore errors in unreachable blocks
+
+The ``NoReturn`` type is only valid as a return annotation of functions,
+and considered an error if appears in other positions::
+
```
DimitrisJim: if appears -> if it appears

pep-0484.txt
```diff
@@ -1063,6 +1063,54 @@ in turn, to ``collections.abc.Callable``::
...
+  from typing import List, NoReturn
+
+  # All of the following are errors
+  def bad1(x: Noreturn) -> int:
```
DimitrisJim: Noreturn -> NoReturn

ilevkivskyi:
@DimitrisJim Thanks, fixed!

pep-0484.txt
```diff
@@ -1063,6 +1063,54 @@ in turn, to ``collections.abc.Callable``::
...
+---------------------
+
+The ``typing`` module provides a special type ``NoReturn`` to annotate functions
+that never return normally. For example, a function that unconditionally
+raises an exception::
+
+  from typing import NoReturn
+
+  def stop() -> NoReturn:
+      raise Exception('no way')
```
gvanrossum: Can you make this RuntimeException? `raise Exception` is such an anti-pattern I don't ever want to see it. :-)

pep-0484.txt
```diff
@@ -1063,6 +1063,54 @@ in turn, to ``collections.abc.Callable``::
...
+
+Such return annotation is used in ``typeshed`` stubs for functions
```
gvanrossum: Such return -> The ``NoReturn`` (mostly to avoid using "such" twice in one sentence :-).

Also, drop "in typeshed".

pep-0484.txt
```diff
@@ -1063,6 +1063,54 @@ in turn, to ``collections.abc.Callable``::
...
+
+The checkers will also recognize that the code after calls to such functions
+is unreachable and will behave accordingly::
+
+  # continue from first example
+  def g(x: int) -> int:
+      if x > 0:
+          return x
+      stop()
+      return 'whatever works'  # No error reported by checkers that
```
gvanrossum: I find this debatable. Why would you have dead code that doesn't type-check? (Though maybe we've had this discussion already? I don't feel strongly about it, just wonder if maybe PEP 484 isn't the place to require this.)

ilevkivskyi: This is how mypy behaves (in fact, and in documentation). I softened the formulation in the new commit (as well as fixed two other things).

I don't have a strong opinion here either, but I think it maybe worth keeping mypy and PEP "in sync".

gvanrossum:
LGTM. Let's see if @markshannon has anything to say.

ambv:
@markshannon, calling twice?

gvanrossum:
I'll juts merge it.

PR #224, opened 2017-03-11
PEP 544: Protocols

### LINKED ISSUE #222: Reserve PEP 544 for Protocols
The PEP is currently pre-drafted here:
https://github.com/ilevkivskyi/peps/blob/protocols/pep-0544.txt

ilevkivskyi:
This adds static support for structural subtyping. Previous discussion is here https://github.com/python/typing/issues/11

Fixes #222

### DIFF
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -593,7 +593,7 @@ In addition, we propose to add another special type construct
 it is not very useful, there are many situation where a variable should
 implement more than one protocol. Annotation by ``All[Proto1, Proto2, ...]``
 means that a given variable or parameter must implement all protocols
-``Ptoro1``, ``Proto2``, etc. either implicitly or explicitly. Example::
+``Proto1``, ``Proto2``, etc. either implicitly or explicitly. Example::
 
   from typing import Sequence, Hashable, All
 
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -691,7 +691,7 @@ for class and instance checks as for ``collections.abc`` classes::
   assert isinstance(open('some/file'), Closeable)
 
 Static type checkers will understand ``isinstance(x, Proto)`` and
-``issublclass(C, Proto)`` for protocols (as they already do
+``issubclass(C, Proto)`` for protocols (as they already do
 for ``Iterable`` etc.). Static type checkers will narrow types after such
 checks by the type erased ``Proto`` (i.e. with all variables having type
 ``Any`` and all methods having type ``Callable[..., Any]``).
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -280,6 +280,7 @@ at the beginning of the list. Here is a simple example::
   from typing import Protocol
 
   class SupportsClose(Protocol):
+      @abstractmethod
       def close(self) -> None:
           ...
 
@@ -486,6 +487,7 @@ having ``typing.Protocol`` as an immediate base class::
   from typing import Sized, Protocol
 
   class SizedAndCloseable(Sized, Protocol):
+      @abstractmethod
       def close(self) -> None:
           ...
 
@@ -526,6 +528,7 @@ non-protocol generic types::
   T = TypeVar('T', covariant=True)
 
   class Iterable(Protocol[T]):
+      @abstractmethod
       def __iter__(self) -> Iterator[T]:
           ...
 
@@ -585,9 +588,11 @@ classes. For example::
   from typing import Union, Optional, Protocol
 
   class Exitable(Protocol):
+      @abstracrmethod
       def exit(self) -> int:
           ...
   class Quitable(Protocol):
+      @abstractmethod
       def quit(self) -> Optional[int]:
           ...
 
@@ -623,6 +628,7 @@ Variables and parameters annotated with ``Type[Proto]`` accept only concrete
 (non-protocol) subtypes of ``Proto``. For example::
 
   class Proto(Protocol):
+      @abstractmethod
       def meth(self) -> int:
           ...
   class Concrete:
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -188,10 +188,9 @@ approaches related to structural subtyping in Python and other languages:
 
   This seems to be a perfect fit for both runtime and static behavior of
   protocols. The main goal of this proposal is to support such behavior
-  statically. In addition, to allow users achieving (if necessary) such
-  runtime behavior for user defined protocols, a simple class decorator
-  ``@runtime`` can be provided that automatically adds a
-  ``__subclasshook__()`` that mimics the one in ``collections.abc``,
+  statically. In addition, to allow users achieving such runtime behavior
+  for user defined protocols a ``__subclasshook__()`` that mimics the one in
+  ``collections.abc`` can be automatically added,
   see detailed `discussion`_ below.
 
 * TypeScript [typescript]_ provides support for user defined classes and
@@ -295,8 +294,8 @@ protocol types::
           self.file.close()
           self.lock.release()
 
-Protocol types can be used in function annotations, variable annotations,
-and for type checking::
+Apart from few restrictions explicitly mentioned below, protocol types can
+be used in every context where a normal types can::
 
   def close_all(things: Iterable[SupportsClose]) -> None:
       for t in things:
@@ -316,17 +315,15 @@ a compatible type signature.
 Protocol members
 ----------------
 
-There are two ways to define methods that are protocol members:
+The methods that are protocol members are defined using the
+``@abstractmethod`` decorator. Such methods can have a default
+implementation that will be type checked. Method body can be left empty,
+i.e. be either only a docstring or a single ``...`` ellipsis literal.
+In this case type checkers will recognize that there is no default
+implementation, i.e. it can't be used via ``super()`` in explicit
+subclasses.
 
-* Use the ``@abstractmethod`` decorator
-* Leave the method body empty, i.e. it should be either docstring,
-  a single ``pass`` statement, a single ``...`` ellipsis literal,
-  or a single ``raise NotImplementedError`` statement.
-
-A combination of both may be used if necessary for runtime purposes.
-If the method body is not empty but decorated by the ``@abstractmethod``
-decorator, then the corresponding definition serves as a default
-method implementation. If some or all parameters are not annotated
+If some or all parameters of protocol method are not annotated
 their types are assumed to be ``Any`` (see PEP 484). All other
 methods defined in the protocol class are considered non-protocol members.
 Example::
@@ -337,24 +334,21 @@ Example::
       @abstractmethod
       def first(self) -> int:     # This is a protocol member
           return 1
+
+      @abstractmethod
       def second(self) -> int:    # This one also
           ...
-      def third(self) -> int:     # And this one too
-          raise NotImplementedError('Implement this!')
 
       def auxiliary(self) -> str: # This is NOT a protocol member
           x = 1
 
-Having an empty method body in a protocol type makes the method
-implicitly abstract only statically, and there won't be any runtime checking.
-The rationale for this is that most methods in a protocol will be abstract,
-and having to always use ``@abstractmethod`` is a little verbose and ugly.
-Leaving the method body empty is the recommended way of defining protocol
-members that don't have a default implementation. Such syntax mirrors how
-methods are defined in stub files. Abstract staticmethods and abstract
-classmethods are equally supported.
+Note that although formally the implicit return type of a method with
+an empty body is ``None``, type checker will not warn about above example,
+such convention mirrors how methods are defined in stub files.
+Abstract staticmethods, abstract classmethods, and abstract properties are
+equally supported.
 
-To define a protocol variable, one should use PEP 526 variable
+To define a protocol variable, one must use PEP 526 variable
 annotations in the class body. Attributes defined in the body of a method
 by assignment via ``self`` are considered non-protocol members. The rationale
 for this is that the protocol class implementation is often not shared by
@@ -373,11 +367,9 @@ non-protocol member. Examples::
       def method(self) -> None:
           self.temp: List[int] = [] # And this one neither
 
-Note that unannotated variables are always assumed to have type ``Any``,
-so that a definition ``x = 0`` in a protocol body will create a protocol
-member ``x`` with type ``Any``, *not* ``int``. To distinguish between protocol
-class variables and protocol instance variables, the special ``ClassVar``
-annotation should be used as specified by PEP 526.
+To distinguish between protocol class variables and protocol instance
+variables, the special ``ClassVar`` annotation should be used as specified
+by PEP 526.
 
 
 Explicitly declaring implementation
@@ -391,46 +383,49 @@ can be used as regular base classes. In this case:
   (``typing.Sequence`` is an example with useful default methods).
   In particular, a protocol could provide a default implementation of
   ``__init__`` that could be used in explicit subclasses.
-  The default implementations will not be used if the subtype relationship
+  The default implementations can not be used if the subtype relationship
   is implicit and only via structural subtyping -- the semantics of
   inheritance is not changed. Examples::
 
     from typing import Protocol
 
     class PColor(Protocol):
-        def description(self) -> str:
-            ...
-        def simple_description(self) -> str:  # this is a non-protocol member
-            rerun 'nice color'
         @abstractmethod
-        def add_int(self, x: int) -> int:     # protocol method with default
+        def draw(self) -> str:
+            ...
+        @abstractmethod
+        def with_int(self, x: int) -> int:
             return 42
 
-    class SimpleColor(PColor):
-        def description(self) -> str:
-            return super().simple_description()
-        def add_int(self, x: int) -> int:
-            return super().add_int(x) + x
+        def complex_method(self) -> int:
+            ...
+            # some complex code here
 
-    class BadColor:
-        def description(self) -> str:
-            return 'not so nice'
+    class SimpleColor(PColor):
+        def draw(self) -> str:
+            return super().draw()  # Error, no default implementation
+
+        def with_int(self, x: int) -> int:
+            return super().with_int(x) + x
+
+        def other_method(self) -> None:
+            x = super().complex_method()
+            ...
 
     simple: SimpleColor
     def represent(c: PColor) -> None:
         ...
     represent(simple) # OK
-    represent(BadColor()) # Error, 'add_int' is not implemented implicitly
 
 * Type checkers will statically enforce that the class
-  actually implements the protocol correctly. In particular, type checkers
-  will always require members without default implementation to be defined in
-  subclasses that explicitly subclass the protocol::
+  actually implements the protocol correctly::
 
     from typing import Protocol, Tuple
 
     class RGB(Protocol):
         rgb: Tuple[int, int, int]
+
+        @abstractmethod
         def intensity(self) -> int:
             ...
         @abstractmethod
@@ -440,24 +435,32 @@ can be used as regular base classes. In this case:
     class Point(RGB):
         def __init__(self, red: int, green: int, blue: str) -> None:
             self.rgb = red, green, blue  # Error, 'blue' must be 'int'
+
+        def to_bytes(self) -> bytes:
+            return super().to_bytes()
+
         # Error, 'intensity' is not defined!
 
-        # Note that 'to_bytes' is not required, but this is safe
-        # since the default implementation is actually in the MRO.
-
 The general philosophy is that protocols are mostly like regular ABCs,
-but a static type checker will handle them specially.
-In particular, a class can explicitly inherit from multiple protocols.
+but a static type checker will handle them specially. Namely, type checkers
+will enforce that the implementation does not have abstract members.
+
+A class can explicitly inherit from multiple protocols.
 In this case methods are resolved using normal MRO and the type checker
 verifies that all subtyping are correct. Note that a concrete method with
 appropriate type signature can override a protocol method but not vice versa.
 Examples::
 
   class A(Protocol):
-      def m(self) -> None: ...
+      @abstractmethod
+      def m(self) -> str: ...
+      @abstractmethod
       def foo(self) -> int: ...
+
   class B(Protocol):
-      def m(self) -> None: print('Hi')
+      def m(self) -> str:
+         return 'non abstract'
+      @abstractmethod
       def bar(self) -> str: ...
 
   class C(A, B): # Fails typecheck, protocol member 'A.m' can't override 'B.m'
@@ -506,6 +509,10 @@ If one omits ``Protocol`` in the base class list, this would be a regular
 is included in the base class list, all the other base classes
 must be protocols. A protocol can't extend a regular class.
 
+Note that rules around explicit subclassing are different from regular ABCs,
+where abstractness is simply defined by having at least one abstract method
+being unimplemented. Protocol classes must be marked *explicitly*.
+
 
 Generic and recursive protocols
 -------------------------------
@@ -521,10 +528,12 @@ non-protocol generic types::
           ...
 
 Note that ``Protocol[T, S, ...]`` is allowed as a shorthand for
-``Protocol, Generic[T, S, ...]``. Recursive protocols are also supported.
-Forward references to the protocol class names can be given as strings as
-specified by PEP 484. Recursive protocols will be useful for representing
-self-referential data structures like trees in an abstract fashion::
+``Protocol, Generic[T, S, ...]``.
+
+Recursive protocols are also supported. Forward references to the protocol
+class names can be given as strings as specified by PEP 484. Recursive
+protocols will be useful for representing self-referential data structures
+like trees in an abstract fashion::
 
   class Traversable(Protocol):
       leafs: Iterable['Traversable']
@@ -545,8 +554,8 @@ relationships are subject to the following rules:
   if and only if ``X`` implements all protocol members of ``P``. In other
   words, subtyping with respect to a protocol is always structural.
 * Edge case: for recursive protocols, structural subtyping is decided
-  positively for all situations that are type safe. Continuing the
-  previous example::
+  positively for situations where such decision depends on itself. Continuing
+  the previous example::
 
     class Tree(Generic[T]):
         def __init__(self, value: T,
@@ -661,26 +670,19 @@ aliases::
 
 .. _discussion:
 
-``@runtime`` decorator and narrowing types by ``isinstance()``
---------------------------------------------------------------
+Narrowing types by ``isinstance()``
+-----------------------------------
 
-The default semantics is that ``isinstance()`` and ``issubclass()`` fail
-for protocol types. This is in the spirit of duck typing -- protocols
-basically would be used to model duck typing statically, not explicitly
-at runtime.
-
-However, it should be possible for protocol types to implement custom
-instance and class checks when this makes sense, similar to how ``Iterable``
-and other ABCs in ``collections.abc`` and ``typing`` already do it,
-but this is limited to non-generic and unsubscripted generic protocols
+Protocol types will implement special instance and class checks, similar to
+how ``Iterable`` and other ABCs in ``collections.abc`` and ``typing`` already
+do it, but this is limited to non-generic and unsubscripted generic protocols
 (``Iterable`` is statically equivalent to ``Iterable[Any]`).
-The ``typing`` module will define a special ``@runtime`` class decorator that
-automatically adds a ``__subclasshook__()`` that provides the same semantics
+The metaclass of ``typing.Protocol`` will automatically add
+a ``__subclasshook__()`` that provides the same semantics
 for class and instance checks as for ``collections.abc`` classes::
 
   from typing import runtime, Protocol
 
-  @runtime
   class Closeable(Protocol):
       @abstractmethod
       def close(self):
@@ -688,16 +690,13 @@ for class and instance checks as for ``collections.abc`` classes::
 
   assert isinstance(open('some/file'), Closeable)
 
-A static type checker will ensure that this decorator is only applied to
-classes where all protocol members are defined by ``@abstractmethod`` for
-methods and by a variable annotation for class and instance variables.
 Static type checkers will understand ``isinstance(x, Proto)`` and
-``issublclass(C, Proto)`` for protocols decorated with such decorator
-(as they already do for ``Iterable`` etc.). Static type checkers will
-narrow types after such checks by the type erased ``Proto`` (i.e. with all
-variables having type ``Any`` and all methods having type
-``Callable[..., Any]``). Note that ``isinstance(x, Proto[int])`` etc. will
-always fail in agreement with PEP 484. Examples::
+``issublclass(C, Proto)`` for protocols (as they already do
+for ``Iterable`` etc.). Static type checkers will narrow types after such
+checks by the type erased ``Proto`` (i.e. with all variables having type
+``Any`` and all methods having type ``Callable[..., Any]``).
+Note that ``isinstance(x, Proto[int])`` etc. will always fail in agreement
+with PEP 484. Examples::
 
   from typing import Iterable, Iterator, Sequence
 
@@ -713,17 +712,11 @@ Using Protocols in Python 3.5 and Earlier
 
 Variable annotation syntax was added in Python 3.6, so that the syntax
 for defining protocol variables proposed in `specification`_ section can't
-be used in earlier versions. To define these in earlier versions of Python we
-propose to either use type comments (and ellipsis literals if there is no
-default value), or use abstract properties if it is necessary to indicate
-abstract status at runtime. All of these are valid::
+be used in earlier versions. To define these in earlier versions of Python
+one can use abstract properties::
 
   class Foo(Protocol):
-
-      x = ... # type: int
-      s = 'abc' # type: str # Protocol instance variable with default.
-
-      @property
+      @abstractproperty
       def c(self) -> int: ...
 
       @abstractproperty
@@ -731,26 +724,10 @@ abstract status at runtime. All of these are valid::
           return 0
 
       @property
-      def e(self) -> int:   # This is not a protocol member.
+      def e(self) -> int:   # Note, this is not a protocol member.
           return 0
 
-The ellipsis ``...`` initializer can leak into subclasses at
-runtime, which is unfortunate::
-
-  class A(Protocol):
-      x = ...  # type: int
-
-  class B(A):
-      def __init__(self) -> None:
-          self.x = 1
-
-  b = B()
-  print(b.x)  # 1
-  print(B.x)  # Ellipsis
-
-For this reason we propose to make the ``Protocol`` metaclass recognize
-``...`` initializers and translate them to a special property that can't
-be read. The ``typing`` module changes proposed in this PEP will be also
+The ``typing`` module changes proposed in this PEP will be also
 backported to earlier versions via the backport currently available on PyPI.
 
 
@@ -769,9 +746,9 @@ effects on the core interpreter and standard library except in the
   a protocol or not. Add a class attribute ``__protocol__ = True``
   if that is the case. Verify that a protocol class only has protocol
   base classes in the MRO (except for object).
-* Implement ``@runtime`` class decorator that makes ``isinstance()`` behave
-  as for ``collections.abc`` by adding all abstract methods and all variables
-  from ``__annotations__`` to ``__subclsshook__()``.
+* Makes ``isinstance()`` behave as for ``collections.abc`` by adding all
+  abstract methods and all variables from ``__annotations__`` to
+  ``__subclsshook__()``.
 * In the backported version of ``typing``, translate ``...`` class attribute
   values to properties.
 * All structural subtyping checks will be performed by static type checkers,
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -50,8 +50,8 @@ of the library. Moreover, extensive use of ABCs might impose additional
 runtime costs.
 
 The intention of this PEP is to solve all these problems
-by allowing to write the above code without explicit base classes in
-the class definition, and ``Bucket`` would still be implicitly considered
+by allowing users to write the above code without explicit base classes in
+the class definition, allowing ``Bucket`` to be implicitly considered
 a subtype of both ``Sized`` and ``Iterable[int]`` by static type checkers
 using structural [wiki-structural]_ subtyping::
 
@@ -66,7 +66,7 @@ using structural [wiki-structural]_ subtyping::
   result: int = collect(Bucket())  # Passes type check
 
 The same functionality will be provided for user-defined protocols, as
-specified below. The code with a protocol class matches common Python
+specified below. The above code with a protocol class matches common Python
 conventions much better. It is also automatically extensible and works
 with additional, unrelated classes that happen to implement
 the required protocol.
@@ -158,9 +158,9 @@ approaches related to structural subtyping in Python and other languages:
   The drawback of this approach is the necessity to either subclass
   the abstract class or register an implementation explicitly::
 
-    from abc import ABCMeta
+    from abc import ABC
 
-    class MyTuple(metaclass=ABCMeta):
+    class MyTuple(ABC):
         pass
 
     MyTuple.register(tuple)
@@ -168,8 +168,8 @@ approaches related to structural subtyping in Python and other languages:
     assert issubclass(tuple, MyTuple)
     assert isinstance((), MyTuple)
 
-  As mentioned in `rationale`_, we want to avoid such necessity, especially
-  in static context. However, in runtime context, ABCs are good candidates for
+  As mentioned in the `rationale`_, we want to avoid such necessity, especially
+  in static context. However, in a runtime context, ABCs are good candidates for
   protocol classes and they are already used extensively in
   the ``typing`` module.
 
@@ -319,9 +319,9 @@ Protocol members
 There are two ways to define methods that are protocol members:
 
 * Use the ``@abstractmethod`` decorator
-* Leave the method body empty, i.e. it should be either a single ``pass``
-  statement, a single ``...`` ellipsis literal, or a single
-  ``raise NotImplementedError`` statement.
+* Leave the method body empty, i.e. it should be either docstring,
+  a single ``pass`` statement, a single ``...`` ellipsis literal,
+  or a single ``raise NotImplementedError`` statement.
 
 A combination of both may be used if necessary for runtime purposes.
 If the method body is not empty but decorated by the ``@abstractmethod``
@@ -383,7 +383,7 @@ annotation should be used as specified by PEP 526.
 Explicitly declaring implementation
 -----------------------------------
 
-To explicitly declare that a certain class implements a given protocols, they
+To explicitly declare that a certain class implements the given protocols, they
 can be used as regular base classes. In this case:
 
 * A class could use implementations of some methods in protocols, both
@@ -545,7 +545,7 @@ relationships are subject to the following rules:
   if and only if ``X`` implements all protocol members of ``P``. In other
   words, subtyping with respect to a protocol is always structural.
 * Edge case: for recursive protocols, structural subtyping is decided
-  positively for all situations that are type safe. Continuing
+  positively for all situations that are type safe. Continuing the
   previous example::
 
     class Tree(Generic[T]):
@@ -590,7 +590,7 @@ classes. For example::
 
 In addition, we propose to add another special type construct
 ``All`` that represents intersection types. Although for normal types
-it is not very useful, there are many situation where a variable should
+it is not very useful, there are many situations where a variable should
 implement more than one protocol. Annotation by ``All[Proto1, Proto2, ...]``
 means that a given variable or parameter must implement all protocols
 ``Proto1``, ``Proto2``, etc. either implicitly or explicitly. Example::
@@ -630,7 +630,7 @@ The same rule applies to variables::
   var = Concrete # OK
   var().meth()   # OK
 
-Assigning protocol class to a variable is allowed if it is not explicitly
+Assigning a protocol class to a variable is allowed if it is not explicitly
 typed, and such assignment creates a type alias. For non-protocol classes,
 the behavior of ``Type[]`` is not changed.
 
@@ -672,8 +672,8 @@ at runtime.
 However, it should be possible for protocol types to implement custom
 instance and class checks when this makes sense, similar to how ``Iterable``
 and other ABCs in ``collections.abc`` and ``typing`` already do it,
-but this is limited to non-generic and to unsubscripted generic protocols
-(statically ``Iterable`` is equivalent to ``Iterable[Any]`).
+but this is limited to non-generic and unsubscripted generic protocols
+(``Iterable`` is statically equivalent to ``Iterable[Any]`).
 The ``typing`` module will define a special ``@runtime`` class decorator that
 automatically adds a ``__subclasshook__()`` that provides the same semantics
 for class and instance checks as for ``collections.abc`` classes::
@@ -761,7 +761,7 @@ Implementation details
 ----------------------
 
 The runtime implementation could be done in pure Python without any
-effects on the core interpreter and standard library except in
+effects on the core interpreter and standard library except in the
 ``typing`` module:
 
 * Define class ``typing.Protocol`` similar to ``typing.Generic``.
@@ -779,8 +779,8 @@ effects on the core interpreter and standard library except in
   be provided at runtime.
 
 
-Changes in typing module
-------------------------
+Changes in the typing module
+----------------------------
 
 The following classes in ``typing`` module will be protocols:
 
@@ -896,8 +896,8 @@ Use assignments to check explicitly that a class implements a protocol
 ----------------------------------------------------------------------
 
 In Go language the explicit checks for implementation are performed
-via dummy assignments [golang]_. Such way is also possible with the current
-proposal. Example::
+via dummy assignments [golang]_. Such a way is also possible with the
+current proposal. Example::
 
   class A:
       def __len__(self) -> float:
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -575,8 +575,9 @@ Example::
       ...
   cached_func((1, 2, 3)) # OK, tuple is both hashable and sequence
 
-If the this will prove to be a widely used scenario. A special ``Intersection``
-type may be added in future as specified by PEP 483.
+If this will prove to be a widely used scenario, then a special
+intersection type construct may be added in future as specified by PEP 483,
+see `rejected`_ ideas for more details.
 
 
 ``Type[]`` with protocols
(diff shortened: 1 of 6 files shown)


pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+  cached_func((1, 2, 3)) # OK, tuple is hashable and sequence
+
+The interaction between union and intersection types is specified by PEP 483,
+and basically reflects the corresponding interactions for sets.
+
+
+``Type[]`` with protocols
+-------------------------
+
+Variables and parameters annotated with ``Type[Proto]`` accept only concrete
```
JukkaL: How is `Type[C]` defined to work with a non-protocol ABCs? It seems that they should both work the same.

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+typed, and such assignment creates a type alias. For non-protocol classes,
+the behavior of ``Type[]`` is not changed.
+
+
+``NewType()`` and type aliases
+------------------------------
+
+Protocols are essentially anonymous. To emphasize this point, static type
+checkers might refuse protocol classes inside ``NewType()`` to avoid an
+illusion that a distinct type is provided::
```
JukkaL: This seems like an unnecessary restriction. `NewType` could define a non-protocol subtype of the protocol, so that compatibility only goes from the new type to the protocol but not the other way around. It would define a distinct type and would be consistent with how `NewType` works in general.

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+Narrowing types by ``isinstance()``
+-----------------------------------
+
+Protocol types will implement special instance and class checks, similar to
+how ``Iterable`` and other ABCs in ``collections.abc`` and ``typing`` already
+do it, but this is limited to non-generic and unsubscripted generic protocols
+(``Iterable`` is statically equivalent to ``Iterable[Any]`).
+The metaclass of ``typing.Protocol`` will automatically add
+a ``__subclasshook__()`` that provides the same semantics
+for class and instance checks as for ``collections.abc`` classes::
```
JukkaL: I don't love the automatic `isinstance` check support, even though that's how the mypy `Protocol` class used to work. I don't think that they can be made to work reliably for arbitrary ABCs. Consider this example:

```py
class P(Protocol):
    @abstractmethod
    def common_method_name(self, x: int) -> int: ... 

class X:
    <a bunch of methods>
    def common_method_name(self) -> None: ...

def do_stuff(o: Union[P, X]) -> int:
    if isinstance(o, P):
        return o.common_method_name(1)  # oops what if it's an X instance instead?
    else:
        return (do something with an X)
```

Maybe a type checker can warn about this example, but more generally the `isinstance` check could be unreliable, except for things where there is a common signature convention such as `Iterable`.

Another potentially problematic case:

```py
class P(Protocol):
    x: int

class C:
    def initialize(self) -> None:
        self.x = 0

c1 = C()
c2 = C()
c2.initialize()
isinstance(c1, P)  # False
isinstance(c2, P)  # True

def f(x: Union[P, int]) -> None:
    if isinstance(x, P):
        # type of x is P here
        ...
    else:
        # type of x is int here?
        print(x + 1)

f(C())   # oops
```

Again, maybe we can live with this, but this could be pretty surprising. I'd argue that requiring an explicit class decorator would be better, since we can then attach warnings about problems like this in the documentation. The programmer would be able to evaluate whether the benefits outweigh the potential for confusion for each protocol and explicitly opt in -- but the default behavior would be safer.

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+  from typing import Iterable, Iterator, Sequence
+
+  def process(items: Iterable[int]) -> None:
+      if isinstance(items, Iterator):
+          # 'items' have type 'Iterator[int]' here
+      elif isinstance(items, Sequence[int]):
+          # Error! Can't use 'isinstance()' with subscripted protocols
+
+
+Using Protocols in Python 3.5 and Earlier
```
JukkaL: What about Python 2.7?

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+--------------------------------------
+
+Some languages such as Go make structural subtyping the only or the primary
+form of subtyping. We could achieve a similar result by making all classes
+protocols by default (or even always). However we believe that it is better
+to require classes to be explicitly marked as protocols, for the following
+reasons:
+
+* Protocols don't have some properties of regular classes. In particular,
+  ``isinstance()`` is not always well-defined for protocols, whereas it is
```
JukkaL: A better argument would be something like this: `isinstance`, as defined for classes that don't explicitly extend `Protocol`, is based on the nominal hierarchy. In order to make everything a protocol by default *and* have `isinstance` work would require changing the semantics of `isinstance`, which won't happen.

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+  well-defined (and quite commonly used) for regular classes.
+* Protocol classes should generally not have many method implementations,
+  as they describe an interface, not an implementation.
+  Most classes have many implementations, making them bad protocol classes.
+* Experience suggests that many classes are not practical as protocols anyway,
+  mainly because their interfaces are too large, complex or
+  implementation-oriented (for example, they may include de facto
+  private attributes and methods without a ``__`` prefix).
+* Most actually useful protocols in existing Python code seem to be implicit.
```
JukkaL: Additional reason: Many built-in functions only accept concrete instances of `int` (and subclass instances), and similarly for other built-in classes. Making `int` a structural type wouldn't be safe without major changes to the Python runtime, which won't happen.

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+to construct an instance of ``A``, which could be problematic if this requires
+accessing or allocating some resources such as files or sockets.
+We could work around the latter by using a cast, for example, but then
+the code would be ugly. Therefore we discourage the use of this pattern.
+
+
+References
+==========
+
+.. [typing]
```
JukkaL: Is it customary to include references to discussions that have happened about the PEP, such as in the typing issue tracker?

pep-0544.txt
ilevkivskyi: > I'd argue that requiring an explicit class decorator would be better, since we can then attach warnings about problems like this in the documentation.

This is how I initially wrote this, since I intuitively felt that this could not be 100% reliable. Guido wanted to make it default, by now we have good examples to make this opt-in, so that I will revert this to class decorator.

pep-0544.txt
ilevkivskyi: I think this could be used the same way for 2.7 - 3.5, modulo the fact that function type comments should be used in 2.7.

pep-0544.txt
ilevkivskyi: Yes, this is a good idea.

pep-0544.txt
ilevkivskyi: OK, I agree. The only limitation would be that ``NewType`` cannot provide a distinct protocol.

pep-0544.txt
ilevkivskyi: I think it could work like this for all ABCs. But it does not work like this in ``mypy `` now, and it is not specified in PEP 484. I feel like this should be specified here rather that in PEP 484.

By the way, there is an old high-priority issue for this https://github.com/python/mypy/issues/1843 and my PR https://github.com/python/mypy/pull/2853 that didn't receive a review for a month :-)

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+  def finish(task: Union[Exitable, Quitable]) -> int:
+      ...
+  class GoodJob:
+      ...
+      def quit(self) -> int:
+          return 0
+  finish(GoodJob()) # OK
+
+In addition, we propose to add another special type construct
+``All`` that represents intersection types. Although for normal types
```
ilevkivskyi: I think we could include this for three reasons:
* I could easily imagine a function parameter that requires more than one protocol, especially if the preferred style is to have several simple protocols, rather than few complex. By the way, some classes in ``collections.abc`` are simple intersections, for example ``Collection == All[Sized, Iterable, Container]``.
* It could be easily implemented for protocols. I also propose to use ``All`` instead of ``Intersection`` to emphasize it only works for protocols (mnemonics: variable implements _all_ these protocols). On the contrary, I think that intersections of nominal classes would be _very_ rarely used.
* If we decide to include this later, then people might simply be not aware of this feature (this is a minor reason, but still we need to take this into account).

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+  close_all([f, r])  # OK!
+  close_all([1])     # Error: 'int' has no 'close' method
+
+Note that both the user-defined class ``Resource`` and the built-in
+``IO`` type (the return type of ``open()``) are considered subtypes of
+``SupportsClose``, because they provide a ``close()`` method with
+a compatible type signature.
+
+
+Protocol members
```
ilevkivskyi: > This seems to imply that sequences no longer support iteration, since Sequence.__iter__ is not abstract and thus is not a protocol member. Is this intentional?

``Sequence`` _implements_ ``__iter__`` therefore it will be accepted wherever ``Iterable`` is expected. But asking for ``__iter__`` on ``Sequence`` indeed will fail.

This indeed looks problematic. Here is a solution that I could propose: we do not have such thing as non-protocol members. However, there will be two kinds of members, abstract and non-abstract. _All_ of them should be implemented in order to consider a class as structural (implicit) subtype.

The reason to have non-abstract methods is that one can get them "for free" using explicit subclassing provided abstract methods are implemented in subclass. So that for larger protocols like ``Sequence`` or ``Mapping`` one doesn't need to do
```python
def method(self):
    return super().method()
```
for every method.

This coincides with how ABCs work at runtime, so that this will be intuitive for someone who already worked with ABCs.

ilevkivskyi:
@JukkaL Thank you for helpful comments! I agree with almost all things. I will make corresponding changes in the text soon. The only thing where we do not agree is ``All``, I still think we should have it.

Also we probably need to find a way how to indicate whether an abstract method has a default implementation or not _in stubs_.

ddfisher:
As a general comment, I don't think it's a good idea to standardize this before having an implementation. An implementation would allow us to understand what's difficult to implement, what sorts of unexpected interactions it has with other type-system features, what parts are confusing, and what parts are practically useful.  I think it will be very difficult to get an understanding of those things otherwise.

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+To distinguish between protocol class variables and protocol instance
+variables, the special ``ClassVar`` annotation should be used as specified
+by PEP 526.
+
+
+Explicitly declaring implementation
+-----------------------------------
+
+To explicitly declare that a certain class implements the given protocols, they
+can be used as regular base classes. In this case:
```
ddfisher: I'm worried this could be pretty confusing: you could have a class that explicitly implements some protocols and implicitly implements others.  I don't know of any other language that allows this.

pep-0544.txt
ddfisher: I agree with Jukka -- I think it's confusing for `All` to work only with protocols, extending `All` to work with all types would add considerable complexity, and I'm not convinced there's a pressing need.

pep-0544.txt
ddfisher: I think this restriction makes sense, actually -- trying to construct a `NewType` from a protocol isn't cogent.  You probably want an alias instead.  Having `NewType` create a non-protocol subtype is surprising behavior and is *not* consistent with how `NewType` works in general, in my opinion.  We shouldn't support this without a clear need.

pep-0544.txt
ddfisher: I'm not sure I understand how a class decorator would solve the problem.  Wouldn't that mean that some classes that implement the protocol (i.e. the decorated ones) will work properly with `isinstance` at runtime but others will not?

pep-0544.txt
```diff
@@ -0,0 +1,949 @@
...
+  abstract methods and all variables from ``__annotations__`` to
+  ``__subclsshook__()``.
+* In the backported version of ``typing``, translate ``...`` class attribute
+  values to properties.
+* All structural subtyping checks will be performed by static type checkers,
+  such as ``mypy`` [mypy]_. No additional support for protocol validation will
+  be provided at runtime.
+
+
+Changes in the typing module
```
ddfisher: How do you propose to handle the backwards-incompatibility problem for all currently released versions of Python?

ilevkivskyi:
@ddfisher Of course I am not going to wait until this PEP is accepted :-)
I think implementation and discussion of specification should go in parallel. This is a big feature, but I hope I will make a first PR quite soon.

pep-0544.txt
ilevkivskyi: > classes that implement the protocol (i.e. the decorated ones)

The decorator is applied to protocol, not to implementation. This was in a previous draft. At runtime the decorator will add same ``__subclasshook__()`` to a protocol, that ``collections.abc`` classes have now. As Jukka pointed out, there are pros and cons of such approach, and a user could decide.

ddfisher: How would the `__subclasshook__` work?

pep-0544.txt
ilevkivskyi: The idea is that there should be no backward incompatibility. The classes in ``typing`` module (as well as in ``collections.abc``) already behave like protocols at runtime:
```python
class Whatever:
    def __iter__(self): return []
assert isinstance(Whatever(), Iterable)
```
as I mention below "Practically, few changes will be needed in ``typing`` since some of these classes already behave the necessary way at runtime", most of these will be changed only in typeshed.

Also, after I implement the changes proposed by Jukka, all code that passes mypy now, will still pass (unless I don't see some subtleties).

brettcannon:
@ilevkivskyi when you're ready for the first version to go in just let us know and we will commit it. Then you  can have discussions go to your personal fork and just send PRs for updates as needed.

pep-0544.txt
ilevkivskyi: Basically, if a class has all necessary attributes, then it is a subclass. This is why I mention below that "type checkers will narrow types after such checks by the type __erased__ ``Proto`` (i.e. with all variables having type ``Any`` and all methods having type ``Callable[..., Any]``)" because we cannot reliably check the types of attributes at runtime.

This is how it is done in ``collections.abc`` see https://github.com/python/cpython/blob/master/Lib/_collections_abc.py#L72 and e.g. https://github.com/python/cpython/blob/master/Lib/_collections_abc.py#L93

ilevkivskyi:
@brettcannon I implemented most comments that appeared so far, but I would prefer to wait for input from @ambv

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+            self.rgb = red, green, blue  # Error, 'blue' must be 'int'
+
+        # Type checker might warn that 'intensity' is not defined
+
+The general philosophy is that protocols are mostly like regular ABCs,
+but a static type checker will handle them specially. Subclassing a protocol
+class would not turn the subclass into a protocol unless it also has
+``typing.Protocol`` as an explicit base class. Without this base, the class
+is "downgraded" to a regular ABC that cannot be used with structural
+subtyping. See section on `extending`_ for details of defining subprotocols.
```
ambv: This paragraph belongs to the next section (`_extending`).

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+* A concrete type or a protocol ``X`` is a subtype of another protocol ``P``
+  if and only if ``X`` implements all protocol members of ``P``. In other
+  words, subtyping with respect to a protocol is always structural.
+* Edge case: for recursive protocols, structural subtyping is decided
+  positively for situations where such decision depends on itself. Continuing
+  the previous example::
+
+    class Tree(Generic[T]):
+        def __init__(self, value: T,
+                           leafs: 'List[Tree[T]]') -> None:
```
ambv: s/leafs/leaves/

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+Note that ``Protocol[T, S, ...]`` is allowed as a shorthand for
+``Protocol, Generic[T, S, ...]``.
+
+Recursive protocols are also supported. Forward references to the protocol
+class names can be given as strings as specified by PEP 484. Recursive
+protocols will be useful for representing self-referential data structures
+like trees in an abstract fashion::
+
+  class Traversable(Protocol):
+      leafs: Iterable['Traversable']
```
ambv: s/leafs/leaves/

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+Protocols cannot be instantiated, so there are no values with
+protocol types. For variables and parameters with protocol types, subtyping
+relationships are subject to the following rules:
+
+* A protocol is never a subtype of a concrete type.
+* A concrete type or a protocol ``X`` is a subtype of another protocol ``P``
+  if and only if ``X`` implements all protocol members of ``P``. In other
+  words, subtyping with respect to a protocol is always structural.
+* Edge case: for recursive protocols, structural subtyping is decided
+  positively for situations where such decision depends on itself. Continuing
```
ambv: > for recursive protocols, structural subtyping is decided positively for situations where such decision depends on itself.

I see what you mean here but the wording is a little confusing.

pep-0544.txt
ambv: I was pushing for `Intersection` for the original PEP because I saw the need for being able to intersect ABCs. Given that protocols themselves solve this, and we can use multiple inheritance as Jukka is suggesting, I agree that it's wiser to leave this out.

I would still *like* to have it, but that can be a separate PEP. I agree it's complex to implement for concrete types, and better yet, for combinations of both.

pep-0544.txt
ambv: I agree this PEP is the right place to finish specifying it. I find the existing explanation missing the "why". It is neatly demonstrated in https://github.com/python/mypy/issues/1843:
- we don't care for the additional *error* around passing a Protocol or an ABC
- we **do** care that treating `Type[P]` as any *concrete* subtype of `P` **silences** errors that signify invalid behavior only if ABCs or Protocols are passed

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+The same rule applies to variables::
+
+  var: Type[Proto]
+  var = Proto    # Error
+  var = Concrete # OK
+  var().meth()   # OK
+
+Assigning a protocol class to a variable is allowed if it is not explicitly
+typed, and such assignment creates a type alias. For non-protocol classes,
+the behavior of ``Type[]`` is not changed.
```
ambv: > For non-protocol classes, the behavior of ``Type[]`` is not changed.

No, we want to make ABCs behave the same way.

pep-0544.txt
ambv: I agree with @ddfisher, it doesn't fit my brain to have a NewType that is distinct from another Protocol, given that both are abstract and structural (and thus equivalent). To explain this differently:
- `What = NewType('What', SomeProtocol)` doesn't make sense because any concrete implementation of a protocol should both conform to SomeProtocol **and** What.

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+The runtime implementation could be done in pure Python without any
+effects on the core interpreter and standard library except in the
+``typing`` module:
+
+* Define class ``typing.Protocol`` similar to ``typing.Generic``.
+* Implement metaclass functionality to detect whether a class is
+  a protocol or not. Add a class attribute ``__protocol__ = True``
+  if that is the case. Verify that a protocol class only has protocol
+  base classes in the MRO (except for object).
+* Implement ``@runtime`` that adds all attributes to ``__subclsshook__()``.
```
ambv: `s/__subclsshook__/__subclasshook__/`

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+The existing class introspection machinery (``dir``, ``__annotations__`` etc)
+can be used with protocols. In addition, all introspection tools implemented
+in the ``typing`` module will support protocols. Since all attributes need
+to be defined in the class body based on this proposal, protocol classes will
+have even better perspective for introspection than regular classes where
+attributes can be defined implicitly -- protocol attributes can't be
+initialized in ways that are not visible to introspection
+(using ``setattr()``, assignment via ``self``, etc.). Still, some things like
+types of attributes will not be visible at runtime in Python 3.5 and earlier,
+but this looks like a reasonable limitation.
```
ambv: Yes.

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+We can always revisit this later if there is an actual need.
+
+
+Make protocols interoperable with other approaches
+--------------------------------------------------
+
+The protocols as described here are basically a minimal extension to
+the existing concept of ABCs. We argue that this is the way they should
+be understood, instead of as something that replaces Zope interfaces,
+for example. Attempting such interoperabilities will significantly
```
ambv: Zope interfaces are a *superset* of Protocols defined here, just using an incompatible syntax to define them. They necessarily use Attributes because pre-PEP 526 there was no other way to annotate this. Arguably in the 3.6+ world `zope.interface` could (optionally?) adopt the Protocol syntax.

In which case to achieve interoperability we'd only need to figure out a way to inform the type checker that "Interface" (or rather "InterfaceClass") is essentially "Protocol". This way we could type check successfully (pretty much ignoring `implements()` and treating those classes as *implicit* implementations. I think this is perfectly fine as the goal is **gradual typing**: if the type checker stays silent in some cases, that's OK. It just never should raise false positives.

Adapters, invariants, and runtime implements/provides checkers are obviously left out. This is a non-goal.

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+          self.x = 0
+
+  c = C()
+  isinstance(c1, P)  # False
+  c.initialize()
+  isinstance(c, P)  # True
+
+  def f(x: Union[P, int]) -> None:
+      if isinstance(x, P):
+          # type of x is P here
```
ambv: > type of x is P here

or rather a concrete type that implements P ;-)

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+accessing or allocating some resources such as files or sockets.
+We could work around the latter by using a cast, for example, but then
+the code would be ugly. Therefore we discourage the use of this pattern.
+
+
+Support ``isinstance()`` checks by default
+------------------------------------------
+
+The problem with this is instance checks could be unreliable, except for
+situations where there is a common signature convention such as ``Iterable``.
```
ambv: Can you give an example of this unreliability? This sounds like a much stronger argument than the attribute assignment presented below (read comment about it).

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+      x: int
+
+  class C:
+      def initialize(self) -> None:
+          self.x = 0
+
+  c = C()
+  isinstance(c1, P)  # False
+  c.initialize()
+  isinstance(c, P)  # True
```
ambv: This is not surprising to me and at runtime I can see this as being a useful feature. You can imagine:
```
class Client(Protocol):
    ...

class Connected(Protocol):
    connection: socket


...  # later in the code
isinstance(someClient, Connected)
```

Maybe we don't want to encourage this programming style but it definitely descibes some real world code, sigh.

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+      if isinstance(x, P):
+          # type of x is P here
+          ...
+      else:
+          # type of x is int here?
+          print(x + 1)
+
+  f(C())   # oops
+
+We argue that requiring an explicit class decorator would be better, since
```
ambv: That's the one part I'm not sure about. Having every Protocol behave consistently at runtime would make it easier to use.

pep-0544.txt
JukkaL: The problem here is that the runtime and static view of the type of `someClient` differ, potentially resulting in runtime type errors uncaught by type checkers, as shown by the code below this line. Since protocols are mostly about static type checking, I think it's reasonable to make restrictions that are biased towards letting static checkers do a better job.

pep-0544.txt
JukkaL: The intention probably was that the _static type_ of `x` is `P`.

pep-0544.txt
JukkaL: As discussed in the examples I gave, it would also make protocols more fragile and potentially surprising. My argument is that explicit here is better than implicit, even if it adds a bit of work for the programmer. Also, currently my working hypothesis is that protocols won't be very common in typical statically typed code -- they will be used every now and then, but not very frequently. If correct, the extra effort needed for adding the decorator would very little and basically insignificant. If this proves to be an incorrect hypothesis, it's much easier to add the missing implicit runtime check in the future than to remove an existing implicit runtime check, since the latter would break existing code. Basically, my reasoning is that if in doubt, leave it out from the first version until we have evidence that it would be useful.

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+  class Bucket(Sized, Iterable[int]):
+      ...
+      def __len__(self) -> int: ...
+      def __iter__(self) -> Iterator[int]: ...
+
+The same problem appears with user-defined ABCs: they must be explicitly
+subclassed or registered. This is particularly difficult to do with library
+types as the type objects may be hidden deep in the implementation
+of the library. Moreover, extensive use of ABCs might impose additional
+runtime costs.
```
ilevkivskyi:
```python
class C(collections.abc.Iterable):
    def __iter__(self):
        return []
```
is 2x slower than
```python
class C:
    def __iter__(self):
        return []
```
The numbers for ``typing.Iterable`` are even worse, but I have some ideas on how to improve those. The 2x slowdown is still not much, so I use "extensive" and "might". If you think it is not necessary we could remove this altogether.

pep-0544.txt
```diff
@@ -0,0 +1,968 @@
...
+
+    class MyIterable:
+        def __iter__(self):
+            return []
+
+    assert isinstance(MyIterable(), Iterable)
+
+  Such behavior seems to be a perfect fit for both runtime and static behavior
+  of protocols. The main goal of this proposal is to support such behavior
+  statically. In addition, to allow users achieving such runtime behavior
```
ilevkivskyi: http://ell.stackexchange.com/questions/11193/allow-to-infinitive-substantive-verb-ing says it probably should be "to achieve" here.

pep-0544.txt
ilevkivskyi: I added a shorter version of Jukka's example here.

ilevkivskyi:
@brettcannon I think this could be merged now, unless there are some objections. I hope I addressed all points that were discussed. There may be some things that could be ironed out, but we can do this is additional PRs later.

gvanrossum:
Let's merge it. That does not mean I am accepting the PEP but is ready for
pyrhon-dev discussion.

--Guido (mobile)

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+      def first(self) -> int:     # This is a protocol member
+          return 42
+
+      @abstractmethod
+      def second(self) -> int:    # Method without a default implementation
+          """Some method."""
+
+Note that although formally the implicit return type of a method with
+a trivial body is ``None``, type checker will not warn about above example,
+such convention is similar to how methods are defined in stub files.
```
gvanrossum: Hmm, I don't really like the idea of "trivial bodies". Remember that when explicitly subclassing an ABC, abstract methods can still be called, through `super()`. An abstract method that shouldn't be called ought to raise `NotImplementedError`.

In stubs, mypy already doesn't type-check trivial bodies (though I don't recall its exact definition of "trivial" -- it probably doesn't exactly match what you have here). If we want to specify that behavior for all checkers reading stubs (which would seem to make sense) it should go in PEP 484, not here.

Or perhaps I am missing a use case for these that you've found in real code?

pep-0544.txt
gvanrossum: > Here is a solution that I could propose: we do not have such thing as non-protocol members.

That's now specified below in the PEP right?

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+Static methods, class methods, and properties are equally allowed
+in protocols.
+
+To define a protocol variable, one must use PEP 526 variable
+annotations in the class body. Additional attributes *only* defined in
+the body of a method by assignment via ``self`` are not allowed. The rationale
+for this is that the protocol class implementation is often not shared by
+subtypes, so the interface should not depend on the default implementation.
```
gvanrossum: Well in some sense that same rationale would apply to default (i.e. non-abstract) implementations.

A defensible position could be that as long as a type checker can verify that an implementation that doesn't explicitly inherit from the protocol class still defines the variable, there's no great reason to disallow variables, since they could just be interpreted as a shorthand for a setter and a getter method.

Also note that mypy (to take one example) doesn't actually check whether an attribute is always set -- it only checks that *if* it is used it has the right type.

All this gets pretty messy though and I am leaning towards not allowing protocol variables at all, other than read-only properties (possibly abstract). I don't recall having seen real-world examples of them (but if there are examples that might sway me).

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+
+To distinguish between protocol class variables and protocol instance
+variables, the special ``ClassVar`` annotation should be used as specified
+by PEP 526.
+
+
+Explicitly declaring implementation
+-----------------------------------
+
+To explicitly declare that a certain class implements the given protocols,
```
gvanrossum: I'd use a singular protocol here and below.

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+they can be used as regular base classes. In this case a class could use
+default implementations of protocol members. ``typing.Sequence`` is a good
+example of a protocol with useful default methods.
+
+Abstract methods with trivial bodies are recognized by type checkers as
+having no default implementation and can't be used via ``super()`` in
+explicit subclasses. The default implementations can not be used if
```
gvanrossum: can not -> cannot

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+    nice: NiceColor
+    another: ImplicitColor
+
+    def represent(c: PColor) -> None:
+        print(c.draw(), c.complex_method())
+
+    represent(nice) # OK
+    represent(another) # Also OK
+
+Note that there is no conceptual difference between explicit and implicit
```
gvanrossum: Well, hm, there seem to be plenty of subtle differences, e.g. the requirement to implement everything only in an implicit subclass.

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+        def __init__(self, red: int, green: int, blue: str) -> None:
+            self.rgb = red, green, blue  # Error, 'blue' must be 'int'
+
+        # Type checker might warn that 'intensity' is not defined
+
+A class can explicitly inherit from multiple protocols and also form normal
+classes. In this case methods are resolved using normal MRO and a type checker
+verifies that all subtyping are correct. The semantics of ``@abstractmethod``
+is not changed, all of them must be implemented by an explicit subclass
+before it could be instantiated.
```
gvanrossum: could -> can

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+
+
+Merging and extending protocols
+-------------------------------
+
+The general philosophy is that protocols are mostly like regular ABCs,
+but a static type checker will handle them specially. Subclassing a protocol
```
gvanrossum: (At runtime maybe we could say that they behave like an implicit `register()` call exists?)

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+class would not turn the subclass into a protocol unless it also has
+``typing.Protocol`` as an explicit base class. Without this base, the class
+is "downgraded" to a regular ABC that cannot be used with structural
+subtyping.
```
gvanrossum: I think we need a rationale for this rule. I think there probably is one, but it needs to be stated clearly, and I'm not sure what it is -- perhaps that we don't want to accidentally have some class act as a protocol just because one of its base classes happens to be one. That is, we still (slightly) prefer nominal subtyping over structural subtyping (in the static typing world).

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+
+  from typing import Sized, Protocol
+
+  class SizedAndCloseable(Sized, Protocol):
+      def close(self) -> None:
+          ...
+
+Now the protocol ``SizedAndCloseable`` is a protocol with two methods,
+``__len__`` and ``close``. If one omits ``Protocol`` in the base class list,
+this would be a regular (non-protocol) class that must implement ``Sized``.
```
gvanrossum: Hm, this example makes this requirement feel like a (slight) nuisance -- in this particular example it's pretty clear that it's meant to define a protocol?

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+If ``Protocol`` is included in the base class list, all the other base classes
+must be protocols. A protocol can't extend a regular class.
```
gvanrossum: This seems to make sense, but does it really? There's no such rule for ABCs. What harm would there be in a protocol inheriting from a regular class? Hm, I guess there's ambiguity over whether the methods in that regular class all become protocol methods with default implementations. Hm... I guess it does make sense, but maybe an explicit example showing the rationale would help.

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+
+Alternatively, one can implement ``SizedAndCloseable`` like this, assuming
+the existence of ``SupportsClose`` from the example in `definition`_ section::
+
+  from typing import Sized
+
+  class SupportsClose(...): ...  # Like above
```
gvanrossum: I would just write this out rather than letting the reader guess to where exactly this is referring.

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+      @abstractmethod
+      def __iter__(self) -> Iterator[T]:
+          ...
+
+Note that ``Protocol[T, S, ...]`` is allowed as a shorthand for
+``Protocol, Generic[T, S, ...]``.
+
+Recursive protocols are also supported. Forward references to the protocol
+class names can be given as strings as specified by PEP 484. Recursive
+protocols will be useful for representing self-referential data structures
```
gvanrossum: will be -> are

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+      leaves: Iterable['Traversable']
+
+
+Using Protocols
+===============
+
+Subtyping relationships with other types
+----------------------------------------
+
+Protocols cannot be instantiated, so there are no values with
```
gvanrossum: That sounds a little weird -- certainly an instance of a class that implements a protocol is a value of that protocol type? Maybe all you need is to add "exactly" somewhere to the sentence.

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+non-protocol generic types::
+
+  T = TypeVar('T', covariant=True)
+
+  class Iterable(Protocol[T]):
+      @abstractmethod
+      def __iter__(self) -> Iterator[T]:
+          ...
+
+Note that ``Protocol[T, S, ...]`` is allowed as a shorthand for
```
gvanrossum: I think this deserves calling out, rather than just "Note".

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+    class Tree(Generic[T]):
+        def __init__(self, value: T,
+                           leaves: 'List[Tree[T]]') -> None:
+            self.value = value
+            self.leafs = leafs
+
+    def walk(graph: Traversable) -> None:
+        ...
+    tree: Tree[float] = Tree(0, [])
+    walk(tree) # OK, 'Tree[float]' is a subtype of 'Traversable'
```
gvanrossum: Hm, I'm not sure I like this example. Isn't the problem more that `Traversible` ought to be generic but isn't in this example?

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+
+``Union`` of protocol classes behaves the same way as for non-protocol
+classes. For example::
+
+  from typing import Union, Optional, Protocol
+
+  class Exitable(Protocol):
+      def exit(self) -> int:
+          ...
+  class Quitable(Protocol):
```
gvanrossum: I'd write "Quittable" -- just like it's "quitting" vs. "exiting". (English is never regular. :-)

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+      def quit(self) -> Optional[int]:
+          ...
+
+  def finish(task: Union[Exitable, Quitable]) -> int:
+      ...
+  class GoodJob:
```
gvanrossum: Can you find a different name for this? "Good job" is such an incredibly boring cliché...

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+
+  class HashableFloats(Sequence[float], Hashable, Protocol):
+      pass
+
+  def cached_func(args: HashableFloats) -> float:
+      ...
+  cached_func((1, 2, 3)) # OK, tuple is both hashable and sequence
+
+If this will prove to be a widely used scenario, then a special
+intersection type construct may be added in future as specified by PEP 483,
```
gvanrossum: may -> could

pep-0544.txt
gvanrossum: Hm, I'm not sure I agree that this should be specified in this PEP for cases other than protocols (though it's fine to discuss it in this thread). I'm also not sure about the sense of @ambv's second bullet, since "we care" is somehow ambiguous in my mind (do we find it a significant mistake or a significant feature?).

IIRC it's pretty subtle to prohibit passing an abstract class object (or a protocol object) to a parameter that requires something annotated with `Type[]`, since the thing you are passing might itself have such a type (e.g. `cls` in a class method).

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+
+  form typing import NewType , Protocol, Iterator
+
+  class Id(Protocol):
+      code: int
+      secrets: Iterator[bytes]
+
+  UserId = NewType('UserId', Id)  # Error, can't provide distinct type
+
+On the contrary, type aliases are fully supported, including generic type
```
gvanrossum: On the contrary -> In contrast

pep-0544.txt
gvanrossum: I'm with @ddfisher and @ambv here.

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+but this is limited to non-generic and unsubscripted generic protocols
+(``Iterable`` is statically equivalent to ``Iterable[Any]`).
+The ``typing`` module will define a special ``@runtime`` class decorator
+that provides the same semantics for class and instance checks as for
+``collections.abc`` classes, essentially making them "runtime protocols"::
+
+  from typing import runtime, Protocol
+
+  @runtime
+  class Closeable(Protocol):
```
gvanrossum: (Oh, I think I'd prefer Closable -- no 'e'.)

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+Note that ``isinstance(x, Proto[int])`` etc. will always fail in agreement
+with PEP 484. Examples::
+
+  from typing import Iterable, Iterator, Sequence
+
+  def process(items: Iterable[int]) -> None:
+      if isinstance(items, Iterator):
+          # 'items' have type 'Iterator[int]' here
+      elif isinstance(items, Sequence[int]):
+          # Error! Can't use 'isinstance()' with subscripted protocols
```
gvanrossum: I believe that's not unique to protocols -- it's the same for other generic classes.

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+  class Foo(Protocol):
+      @property
+      def c(self) -> int:
+          return 42         # Default value can be provided for property...
+
+      @abstractproperty
+      def d(self) -> int:   # ... or it can be abstract
+          return 0
+
+In Python 2.7 the function type comments should be used as per PEP 484.
```
gvanrossum: I think they should also be allowed for Python 3 (even 3.6+) and the text earlier that introduces these should mention this. (It's fine of course to recommend PEP 526 syntax and to use it in all examples except one showing the fallback syntax.)

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+because before PEP 526 there was no straightforward way to annotate attributes.
+In the 3.6+ world, ``zope.interface`` might potentially adopt the ``Protocol``
+syntax. In this case, type checkers could be taught to recognize interfaces
+as protocols and make simple structural checks with respect to them.
+
+
+Use assignments to check explicitly that a class implements a protocol
+----------------------------------------------------------------------
+
+In Go language the explicit checks for implementation are performed
```
gvanrossum: "In Go language" -> either "In Go" or "In the Go language"

pep-0544.txt
```diff
@@ -0,0 +1,1006 @@
...
+  class P(Protocol):
+      def common_method_name(self, x: int) -> int: ...
+
+  class X:
+      <a bunch of methods>
+      def common_method_name(self) -> None: ... # Note different signature
+
+  def do_stuff(o: Union[P, X]) -> int:
+      if isinstance(o, P):
+          return o.common_method_name(1)  # oops, what if it's an X instance?
```
gvanrossum: So this more or less assumes that the runtime isinstance check doesn't compare the arguments. But what if it did?

Or are you saying that in this case if `P` was decorated with `@runtime` the isinstance check would still be broken?

pep-0544.txt
gvanrossum: OK, I am actually convinced by @JukkaL's argument here. (But we should probably apply a similar reasoning to cut out some other esoteric ideas, like variables.)

pep-0544.txt
gvanrossum: I wonder if we could do more deep linking to specific comments or subthreads? (No big deal if you think it's too much effort, but it might save some readers some time finding the right part of the long discussion in e.g. python/typing#11 -- including myself :-).

gvanrossum:
There's a couple more things that I should bring up.

It looks like the mypy core team (all Dropboxers) is going to be very busy with other features, to the point where we probably won't have time to even review an implementation of this PEP for the foreseeable future. Now, this is all subject to various priorities that may still change, and perhaps the implementation is super simple, but if it's anything like I expect, even a complete, finished implementation (and one that allows importing `Protocol` from `mypy_extensions` so it can be made to work with the stdlib versions of typing in earlier Python 3.5 and 3.6 releases) would still take up quite some time to review, time which we just may not have.

There's also the absolute requirement to send this PEP to python-dev for discussion. Since I'm not an author we don't need a BDFL-delegate -- we just need to be careful to listen to feedback. And while I am aware that my own time is a *particularly* scarce resource, I still don't want to delegate this particular topic.

ilevkivskyi:
@gvanrossum 
> There's also the absolute requirement to send this PEP to python-dev for discussion.

OK, I will do this now.

pep-0544.txt
ilevkivskyi: > An abstract method that shouldn't be called ought to raise ``NotImplementedError``.

OK, I think this is a reasonable requirement, I will update this part.

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+The distinction is not important most of the time, and in other
+cases we propose to just add a qualifier such as *protocol classes*
+when referring to the static type concept.
+
+If a class includes a protocol in its MRO, the class is called
+an *explicit* subclass of the protocol. If a class is a structural subtype
+of a protocol, it is said to implement the protocol and to be compatible
+with a protocol. If a class is compatible with a protocol but the protocol
+is not included in the MRO, the class is an *implicit* subtype
+of the protocol.
```
ilevkivskyi: Yes. I will add a note that type checkers should understand setting attributes to ``None`` in the sense of [data model docs](https://docs.python.org/3/reference/datamodel.html#special-method-names): "Setting a special method to ``None`` indicates that the corresponding operation is not available."

pep-0544.txt
ilevkivskyi: OK, will do this.

pep-0544.txt
ilevkivskyi: > That's now specified below in the PEP right?

Yes.

pep-0544.txt
ilevkivskyi: @gvanrossum 
> Hm, I'm not sure I agree that this should be specified in this PEP for cases other than protocols.

As I understand, there are only two options:
* this behavior only applies to protocols
* it applies to all ABCs (including protocols)

I think the simplest solution is to say here than this behavior applies to protocols, but _don't_ 
say that it doesn't apply to ABCs. Then, when we decide on ABCs, we could update PEP 484 independently.

> IIRC it's pretty subtle to prohibit passing an abstract class object (or a protocol object) to a parameter that requires something annotated with ``Type[]``, since the thing you are passing might itself have such a type (e.g. ``cls`` in a class method).

My PR, that I mentioned above, actually takes care of this and few other edge cases, but it has never been reviewed...

pep-0544.txt
ilevkivskyi: Maybe this point needs some clarifications. First, I don't see why we need to force people to write:
```python
class Point2D(Protocol):
    @property
    def x(self) -> int:
        ...
    @property
    def y(self) -> int:
        ...
```
instead of just
```python
class Point2D(Protocol):
    x: int
    y: int
```
I think almost every function parameter annotated with a named tuple or a similar struct-like class could be replaced with a protocol annotation. (Also, this will be a natural counterpart of ``TypedDict``.) Second, I don't think we need to specify how "smart" type checkers should be. Maybe we could only mention that it is reasonable to expect that a type checker _will_ recognize
```python
class Coordinates:
    def __init__(self, x: int, y: int) -> None:
        self.x = x
        self.y = y
```
as implicitly implementing ``Point2D``.

pep-0544.txt
ilevkivskyi: > e.g. the requirement to implement everything only in an implicit subclass

This is exactly what I mean on the next line.

pep-0544.txt
ilevkivskyi: It may be clear for a human (taking into account the name of subclass). But for a type checker there are two equally possible options: a user actually wanted a protocol, or just forgot to implement a protocol method. As you said above:
> we don't want to accidentally have some class act as a protocol just because one of its base classes happens to be one

I think explicit is better than implicit applies here.

pep-0544.txt
```diff
@@ -0,0 +1,967 @@
...
+----------------------------------------
+
+Protocols cannot be instantiated, so there are no values with
+protocol types. For variables and parameters with protocol types, subtyping
+relationships are subject to the following rules:
+
+* A protocol is never a subtype of a concrete type.
+* A concrete type or a protocol ``X`` is a subtype of another protocol ``P``
+  if and only if ``X`` implements all protocol members of ``P``. In other
+  words, subtyping with respect to a protocol is always structural.
```
ilevkivskyi: > Maybe it would be less ambiguous if you separated the two cases into separate bullets.

I agree.

pep-0544.txt
ilevkivskyi: > Isn't the problem more that ``Traversible`` ought to be generic but isn't in this example?

No, this "problem" (I would rather say "rule") applies to all recursive protocols. For example:
```python
class Linked(Protocol):
    child: 'Linked'

class C:
    child: 'C'

x: Linked
x = C() # This should be OK, according to the "rule"
```
I think we just need a simpler example here.

pep-0544.txt
ilevkivskyi: > Or are you saying that in this case if ``P`` was decorated with ``@runtime`` the isinstance check would still be broken?

I think yes, for example now ``isinstance(obj, Iterable)`` returns ``True`` for a class with ``__iter__ = 42``. And we can't make ``@runtime`` much better because not all information is available statically. As Jukka said, it is not a big problem since static type checker will warn about this, but big enough to make support for ``isinstance()`` with protocols opt-in.

pep-0544.txt
ambv: > All this gets pretty messy though and I am leaning towards not allowing protocol variables at all, other than read-only properties

Disallowing protocol variables entirely significantly limits the scope of structural typing. There's many examples in the stdlib alone (like the `stdout.buffer` field, named tuples and enums used structurally, etc.).

Only allowing read-only properties is less limiting but still so. More importantly, it is confusing to the user, like Ivan's example shows above:
* if we specify Point2D as a read-only property, should the type checker reject a writable implementation?
* if not, it feels strange to use a property to define something that is just an attribute in the implementation.

I can imagine users of this rewriting their attributes to be properties just to satisfy the protocol specification.

ilevkivskyi:
@gvanrossum 
> we probably won't have time to even review an implementation of this PEP for the foreseeable future.

This sounds a bit frustrating, but what do you mean by "foreseeable future", few years or few days?
Also, @ddfisher said above in https://github.com/python/peps/pull/224#issuecomment-286907763 that "I don't think it's a good idea to standardize this before having an implementation", so that I thought that I should finish an implementation soon (even if not 100% complete).
I am a bit lost here.

pep-0544.txt
gvanrossum: I'm not adamantly against protocol variables, but I think we should at least settle what to do about read-only attributes. When you use `@property` (not using the `@f.setter` syntax) or use PEP 526 syntax inside a`NamedTuple` you get a read-only attribute; otherwise PEP 526 syntax gets you a writable attribute. Which should be the default for Protocols, and how should you be able to declare the other kind?

Note that this is *not* about immutability -- that's in the programmer's head (or perhaps guided by whether `__hash__` is defined). This is purely about whether attribute assignment should be allowed.

pep-0544.txt
gvanrossum: Still, starting with "there is no conceptual difference..." seems too strong.

pep-0544.txt
gvanrossum: Fair enough. Objection withdrawn.

pep-0544.txt
gvanrossum: Agreed. The generics just distract. (Maybe we need a simple example first and then one showing how to combine this with generics?)

pep-0544.txt
gvanrossum: OK, I'm convinced (also this review UI is very confusing, it seems we are having discussion on different subcommits of the same PR).

I still don't like the name `@runtime` though -- it's too generic. Can we brainstorm about a better name?

gvanrossum:
> This sounds a bit frustrating, but what do you mean by "foreseeable future", few years or few days?

A few months at least, is my best estimate based on discussions we had in our project planning (not all [arts of the project are open source).

I agree it's frustrating -- it is for us too, but we need to acknowledge that we have limited velocity.

> Also, @ddfisher said above in #224 (comment) that "I don't think it's a good idea to standardize this before having an implementation", so that I thought that I should finish an implementation soon (even if not 100% complete).

I believe David is more skeptical of this PEP than Jukka and I. But even so, the extra frustrating thing is that, indeed, the PEP process (by convention at least) prefers to have a reference implementation to support a proposal, without any guarantees that the reference implementation will ever be deployed. This is similar to the RFC-based IETF standardization process for internet protocols -- and the similarity is not coincidental, we modeled Python's PEP process after the RFC process (with which several core Python devs were highly familiar between 1995-2000).

> I am a bit lost here.

I'm truly sorry. It is frustrating for all of us (as are my continued health issues that reduce my productivity dramatically).

ambv:
@gvanrossum

> we need to acknowledge that we have limited velocity.

What would it take for the project to nominate a few non-Dropbox core contributors?

> A few months at least

This is not a huge obstacle for inclusion in Python 3.7, as long as we get to it sometime this year. Also, a reference implementation for the purpose of this PEP can be on a "branch" of mypy. In other words, for the PEP reviewers to consider it, it doesn't have to be merged. This was pretty common in the past.

@ilevkivskyi, I will provide preliminary review and use it internally before it gets merged (we are living on master anyway).

ilevkivskyi:
OK, what about this way: I will have a branch in my fork of mypy, where I will build a reference implementation. People will be able to play with it (I will take care to merge mypy master frequently, so that the branch will be fully functional). And it is fine with me if it will be not merged for few months. To avoid a potential huge (and impossible to review) PR if/when the PEP is accepted, I will cherry-pick few smaller branches and make few sequential PRs.

@gvanrossum @ambv @JukkaL does this sound reasonable?

gvanrossum:
It sounds reasonable (though I cannot make any hard promises).

CC @gnprice.

PR #225, opened 2017-03-14
PEP 7: add curly braces to if block

orenmn:
according to PEP 7, "braces are strongly preferred but may be omitted where C permits".
thus, IMHO, the examples in the PEP should have curly braces too.

### DIFF
--- a/pep-0007.txt
+++ b/pep-0007.txt
@@ -127,8 +127,9 @@ Code lay-out
 
       if (type->tp_dictoffset != 0 && base->tp_dictoffset == 0 &&
           type->tp_dictoffset == b_size &&
-          (size_t)t_size == b_size + sizeof(PyObject *))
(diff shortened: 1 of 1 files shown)


DimitrisJim:
I really have no opinion on this :-) , with or without braces, the snippet is still fine and correct according to the style rules. 

Looking at the codebase, one can see how different developers choose a different style so, unfortunately, there's really no uniformity on which to base the change. 

I just think someone will come along in a couple of months and try to to change it back :-) Either way, let's see what other people say.

brettcannon:
I say "merge", and so I did. 😉 Basically the thinking is right that since we are strongly encouraging moving towards curly braces then it makes sense to update the comment. You're right, @DimitrisJim , that it's technically acceptable, but the example should reflect what we are promoting.

PR #231, opened 2017-03-23
PEP 484: Clarify scoping of type variables

ilevkivskyi:
This question appeared recently on mypy's gitter channel. Here I clarify scoping rules for type variables in inner classes and add an example.

It was discussed, that these rules are maybe too restrictive. But this is necessary if we want to keep
```python
class Bad(Iterable[T]):
    ...
```
and
```python
class Bad(Iterable[T], Generic[T]):
    ...
```
equivalent. If people will ask, then we could later relax the rule and allow binding of inner class type variables to outer class type variables (however, I think this is highly unlikely).

@gvanrossum Please take a look.


gvanrossum:
Thanks! Agreed.

Issue #223, opened 2017-03-10
Suggest people use the PEP # 9999 when writing their own PEPs


brettcannon:
That way people can still pass Travis testing but won't be using a number that might get used before their PEP is accepted into the repo.

PR #234, opened 2017-03-31
Use number 9999 in a draft PEP

Mariatta:
Update PEP 1 and PEP 12 with instruction to use the number 9999.

Closes https://github.com/python/peps/issues/223

### DIFF
--- a/pep-0001.txt
+++ b/pep-0001.txt
@@ -147,7 +147,7 @@ errors may be corrected by the editors).
 The standard PEP workflow is:
 
 * You, the PEP author, fork the `PEP repository`_, and create a file named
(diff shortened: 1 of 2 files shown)


brettcannon:
I like it, but maybe @warsaw has an opinion?

brettcannon:
Do any @python/pep-editors have an opinion on this? I say we make the change.

Rosuav:
I'm in favour.

Mariatta:
🤔 Do I need to email python-dev about this change?

brettcannon:
OK, I say merge this and a quick note to python-dev would be good.

PR #236, opened 2017-04-01
Convert PEPs 519, 528 and 529 from CRLF to LF line endings.

serhiy-storchaka:


serhiy-storchaka:
Also remove trailing spaces in PEP 529 (were missed in #232 due to CRLF).

PR #238, opened 2017-04-02
Miscellaneous fixes and formatting enhancements.

serhiy-storchaka:


pep-0261.txt
```diff
@@ -74,26 +74,26 @@ user, Python 2.2 will allow the 4-byte implementation as a
 build-time option. Users can choose whether they care about
 wide characters or prefer to preserve memory.
 
-The 4-byte option is called ``wide Py_UNICODE``. The 2-byte option
-is called ``narrow Py_UNICODE``.
+The 4-byte option is called "wide ``Py_UNICODE``"". The 2-byte option
```
serhiy-storchaka: Typo: a double closing quote.

pep-0315.txt
```diff
@@ -102,7 +102,7 @@ belongs, and does not require code to be duplicated.
 Syntax
 ======
 
-The syntax of the while statement::
+The syntax of the while statement ::
```
brettcannon: Does the added space do anything?

serhiy-storchaka: Yes, it removes a colon from rendered text (but still makes following indented text a code block). The original text/plain version didn't have a colon, and this looks correct to me. "The syntax of the while statement ... is extended as follows: ..." (where "..." are code blocks) is a single sentence, and it shouldn't have a colon after "statement".

brettcannon: Ah, I'm used to seeing that done by putting the `::` on its own line.

pep-0215.txt
```diff
@@ -69,7 +69,7 @@ Examples
 ========
 
 Here is an example of an interactive session exhibiting the
-expected behaviour of this feature::
+expected behaviour of this feature. ::
```
DimitrisJim: It might be just me but won't that look odd after the formatting? With a dot appearing right before the colon? I unfortunately can't find any authoritative reference on this but it strikes me as odd because I don't think I've seen it elsewhere. (Similarly for the comma version.)

methane: No, since there is a space between `.` and `::`, colon not rendered.
But I prefer put `::` in own line, as @brettcannon's comment.

DimitrisJim: Ah, I see, thanks for clearing that up. 

A dot instead of a comma seems like a good replacement. The comma version, at least for me, not so much. I think it would out of place before a code block (while `:` seems natural).

serhiy-storchaka:
Thank you for your review @brettcannon.

PR #239, opened 2017-04-07
Update generic class attribute and type variable usage

pkch:
Clarification based on python/mypy#2878

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -288,7 +288,7 @@ elements.  Example::
 
   def notify_by_email(employees: Set[Employee], overrides: Mapping[str, str]) -> None: ...
 
-Generics can be parametrized by using a new factory available in
+Generics can be parameterized by using a new factory available in
 ``typing`` called ``TypeVar``.  Example::
 
   from typing import Sequence, TypeVar
@@ -583,7 +583,7 @@ argument(s) is substituted.  Otherwise, ``Any`` is assumed.  Example::
   class Node(Generic[T]):
       def __init__(self, label: T = None) -> None:
           ...
-      x: None  # Type: T
+      x = None  # Type: T
 
   x = Node('')  # Inferred type is Node[str]
   y = Node(0)   # Inferred type is Node[int]
@@ -611,9 +611,11 @@ the runtime class of the objects created by instantiating them doesn't
 record the distinction. This behavior is called "type erasure"; it is
 common practice in languages with generics (e.g. Java, TypeScript).
 
-Class attribute of a generic class can be looked up only through the instance
-of the class; any attempt to use generic classes (parametrized or not) to
-access attributes will result in both type check failure and runtime error::
+Using generic classes (parameterized or not) to access attributes will result
+in both type check failure and runtime error. Outside the class definition
+body, a class attribute cannot be assigned, and can only be looked up by
+accessing it through the class instance that does not have same-named
+instance attribute::
 
   # (continued from previous example)
   Node[int].x = 1  # Error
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username on [bugs.python.org](http://bugs.python.org/) (b.p.o) to verify you have signed the CLA. This is necessary for legal reasons before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://cpython-devguide.readthedocs.io/pullrequest.html#licensing) to rectify this issue.


pep-0484.txt
```diff
@@ -610,6 +611,21 @@ the runtime class of the objects created by instantiating them doesn't
...
+  # (continued from previous example)
+  Node[int].x = 1  # Error
+  Node[int].x      # Error
+  Node.x = 1       # Error
+  Node.x           # Error
+  type(p).x        # Error
+  p.x              # Ok
+  Node[int]().x    # Ok
+
+In addition, ``ClassVar`` cannot be parametrized.
```
JelleZijlstra: parameterized

pep-0484.txt
```diff
@@ -610,6 +611,21 @@ the runtime class of the objects created by instantiating them doesn't
 record the distinction. This behavior is called "type erasure"; it is
 common practice in languages with generics (e.g. Java, TypeScript).
 
+Class attribute of a generic class can be looked up only through the instance
```
DimitrisJim: "attributes" is probably what you meant here.

pkch: Hmm, I am not sure.

Somehow I would interpret `instance attribute` as referring to the physical object stored on an instance, and of course if there is one, we can't actually refer to the class attribute at all.

pep-0484.txt
pkch: Oh I thought so too, but I searched the text and found it uses `parametrized`. Now I searched again, and I found it uses both, 2 times each (not counting my own usage) 😆 I guess I should have checked dictionary instead ...

pkch:
I added another clarification: that `TypeVar` cannot contain type variables (based on the discussion in https://github.com/python/mypy/issues/2756#issuecomment-293109162). If needed, I can separate it into a separate PR.

pkch:
The last commit is as discussed [here](https://github.com/python/mypy/issues/3148#issuecomment-292834046)

ilevkivskyi:
@gvanrossum mentioning you just in case (this is not urgent).

pep-0484.txt
```diff
@@ -610,6 +612,24 @@ the runtime class of the objects created by instantiating them doesn't
 record the distinction. This behavior is called "type erasure"; it is
 common practice in languages with generics (e.g. Java, TypeScript).
 
+Using generic classes (parameterized or not) to access attributes will result
+in both type check failure and runtime error. Outside the class definition
```
ilevkivskyi: > and runtime error.

Is it really a runtime error for examples below? I just tried it and all worked fine _at runtime_.

pep-0484.txt
```diff
@@ -610,6 +612,24 @@ the runtime class of the objects created by instantiating them doesn't
...
+  Node[int].x = 1  # Error
+  Node[int].x      # Error
+  Node.x = 1       # Error
+  Node.x           # Error
+  type(p).x        # Error
+  p.x              # Ok (evaluates to None)
+  Node[int]().x    # Ok (evaluates to None)
+  p.x = 1          # Ok, but assigning to instance attribute
+
+In addition, ``ClassVar`` cannot be parameterized.
```
ilevkivskyi: ``ClassVar`` is not a part of PEP 484.

pep-0484.txt
```diff
@@ -701,11 +721,11 @@ classes without a metaclass conflict.
...
-A type variable may specify an upper bound using ``bound=<type>``.
-This means that an actual type substituted (explicitly or implicitly)
-for the type variable must be a subtype of the boundary type.  A
-common example is the definition of a Comparable type that works well
-enough to catch the most common errors::
+A type variable may specify an upper bound using ``bound=<type>`` (note:
+<type> itself cannot be parametrized by type variables). This means that an
+actual type substituted (explicitly or implicitly) for the type variable must
+be a subtype of the boundary type. A common example is the definition of a
+Comparable type that works well enough to catch the most common errors::
```
ilevkivskyi: While yo at it, can you please make this
```
``Comparable``
```
i.e. use backquotes.

pep-0484.txt
pkch: Ah right. I misunderstood your comment that [it's fixed](https://github.com/python/mypy/issues/2878#issuecomment-282302705) in `typing`. I thought you meant it now causes runtime error, but actually what you meant is that it no longer returns different values depending on how it's accessed.

gvanrossum:
Thanks!

ilevkivskyi:
@pkch Would you like to also make a small PR updating the PEP 526 and saying that things like ``ClassVar[T]`` are prohibited?

PR #243, opened 2017-04-19
Updates for PEP 544: Protocols

ilevkivskyi:
This implements two sets of comments by Guido, comments from python-dev, and adds links to working implementation.

Attn: @JukkaL @ambv @gvanrossum

### DIFF
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -263,7 +263,9 @@ an *explicit* subclass of the protocol. If a class is a structural subtype
 of a protocol, it is said to implement the protocol and to be compatible
 with a protocol. If a class is compatible with a protocol but the protocol
 is not included in the MRO, the class is an *implicit* subtype
-of the protocol.
+of the protocol. (Note that one could explicitly subclass a protocol and
+still not implement it if a protocol attribute is set to ``None``
+in the subclass, see Python [data model]_ for details.)
 
 The attributes (variables and methods) of a protocol that are mandatory
 for other class in order to be considered a structural subtype are called
@@ -276,7 +278,7 @@ Defining a protocol
 -------------------
 
 Protocols are defined by including a special new class ``typing.Protocol``
-(an instance of ``abc.ABCMeta``) in the base classes list, preferably
+(an instance of ``abc.ABCMeta``) in the base classes list, typically
 at the end of the list. Here is a simple example::
 
   from typing import Protocol
@@ -318,11 +320,11 @@ Protocol members
 ----------------
 
 All methods defined in the protocol class body are protocol members, both
-normal and decorated with ``@abstractmethod``. If some or all parameters of
+normal and decorated with ``@abstractmethod``. If any parameters of a
 protocol method are not annotated, then their types are assumed to be ``Any``
-(see PEP 484). Bodies of protocol methods are type checked, except for methods
-decorated with ``@abstractmethod`` with trivial bodies. A trivial body can
-contain a docstring. Example::
+(see PEP 484). Bodies of protocol methods are type checked.
+An abstract method that should not be called via ``super()`` ought to raise
+``NotImplementedError``. Example::
 
   from typing import Protocol
   from abc import abstractmethod
@@ -333,15 +335,12 @@ contain a docstring. Example::
 
       @abstractmethod
       def second(self) -> int:    # Method without a default implementation
-          """Some method."""
+          raise NotImplementedError
 
-Note that although formally the implicit return type of a method with
-a trivial body is ``None``, type checker will not warn about above example,
-such convention is similar to how methods are defined in stub files.
 Static methods, class methods, and properties are equally allowed
 in protocols.
 
-To define a protocol variable, one must use PEP 526 variable
+To define a protocol variable, one could use PEP 526 variable
 annotations in the class body. Additional attributes *only* defined in
 the body of a method by assignment via ``self`` are not allowed. The rationale
 for this is that the protocol class implementation is often not shared by
@@ -357,22 +356,29 @@ Examples::
       def method(self) -> None:
           self.temp: List[int] = [] # Error in type checker
 
+  class Concrete:
+      def __init__(self, name: str, value: int) -> None:
+          self.name = name
+          self.value = value
+
+  var: Template = Concrete('value', 42)  # OK
+
 To distinguish between protocol class variables and protocol instance
 variables, the special ``ClassVar`` annotation should be used as specified
-by PEP 526.
+by PEP 526. By default, protocol variables as defined above are considered
+readable and writable. To define a read-only protocol variable, one can use
+an (abstract) property.
 
 
 Explicitly declaring implementation
 -----------------------------------
 
-To explicitly declare that a certain class implements the given protocols,
-they can be used as regular base classes. In this case a class could use
+To explicitly declare that a certain class implements a given protocol,
+it can be used as a regular base class. In this case a class could use
 default implementations of protocol members. ``typing.Sequence`` is a good
 example of a protocol with useful default methods.
 
-Abstract methods with trivial bodies are recognized by type checkers as
-having no default implementation and can't be used via ``super()`` in
-explicit subclasses. The default implementations can not be used if
+The default implementations cannot be used if
 the subtype relationship is implicit and only via structural
 subtyping -- the semantics of inheritance is not changed. Examples::
 
@@ -406,7 +412,7 @@ subtyping -- the semantics of inheritance is not changed. Examples::
     represent(nice) # OK
     represent(another) # Also OK
 
-Note that there is no conceptual difference between explicit and implicit
+Note that there is little difference between explicit and implicit
 subtypes, the main benefit of explicit subclassing is to get some protocol
 methods "for free". In addition, type checkers can statically verify that
 the class actually implements the protocol correctly::
@@ -428,7 +434,7 @@ A class can explicitly inherit from multiple protocols and also form normal
 classes. In this case methods are resolved using normal MRO and a type checker
 verifies that all subtyping are correct. The semantics of ``@abstractmethod``
 is not changed, all of them must be implemented by an explicit subclass
-before it could be instantiated.
+before it can be instantiated.
 
 
 Merging and extending protocols
@@ -439,7 +445,10 @@ but a static type checker will handle them specially. Subclassing a protocol
 class would not turn the subclass into a protocol unless it also has
 ``typing.Protocol`` as an explicit base class. Without this base, the class
 is "downgraded" to a regular ABC that cannot be used with structural
-subtyping.
+subtyping. The rationale for this rule is that we don't want to accidentally
+have some class act as a protocol just because one of its base classes
+happens to be one. We still slightly prefer nominal subtyping over structural
+subtyping in the static typing world.
 
 A subprotocol can be defined by having *both* one or more protocols as
 immediate base classes and also having ``typing.Protocol`` as an immediate
@@ -447,24 +456,24 @@ base class::
 
   from typing import Sized, Protocol
 
-  class SizedAndCloseable(Sized, Protocol):
+  class SizedAndClosable(Sized, Protocol):
       def close(self) -> None:
           ...
 
-Now the protocol ``SizedAndCloseable`` is a protocol with two methods,
+Now the protocol ``SizedAndClosable`` is a protocol with two methods,
 ``__len__`` and ``close``. If one omits ``Protocol`` in the base class list,
 this would be a regular (non-protocol) class that must implement ``Sized``.
-If ``Protocol`` is included in the base class list, all the other base classes
-must be protocols. A protocol can't extend a regular class.
-
-Alternatively, one can implement ``SizedAndCloseable`` like this, assuming
-the existence of ``SupportsClose`` from the example in `definition`_ section::
+Alternatively, one can implement ``SizedAndClosable`` protocol by merging
+the ``SupportsClose`` protocol from the example in `definition`_ section
+with ``typing.Sized``::
 
   from typing import Sized
 
-  class SupportsClose(...): ...  # Like above
+  class SupportsClose(Protocol):
+      def close(self) -> None:
+          ...
 
-  class SizedAndCloseable(Sized, SupportsClose, Protocol):
+  class SizedAndClosable(Sized, SupportsClose, Protocol):
       pass
 
 The two definitions of ``SizedAndClosable`` are equivalent.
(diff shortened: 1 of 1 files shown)


pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
+Now the protocol ``SizedAndClosable`` is a protocol with two methods,
 ``__len__`` and ``close``. If one omits ``Protocol`` in the base class list,
 this would be a regular (non-protocol) class that must implement ``Sized``.
-If ``Protocol`` is included in the base class list, all the other base classes
-must be protocols. A protocol can't extend a regular class.
-
-Alternatively, one can implement ``SizedAndCloseable`` like this, assuming
-the existence of ``SupportsClose`` from the example in `definition`_ section::
+Alternatively, one can implement ``SizedAndClosable`` protocol by merging
+the ``SupportsClose`` protocol from the example in `definition`_ section
```
gvanrossum: ... in **the** definition section ...

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
-
   class Iterable(Protocol[T]):
       @abstractmethod
       def __iter__(self) -> Iterator[T]:
           ...
 
-Note that ``Protocol[T, S, ...]`` is allowed as a shorthand for
-``Protocol, Generic[T, S, ...]``.
+``Protocol[T, S, ...]`` is allowed as a shorthand for
+``Protocol, Generic[T, S, ...]``. Declaring variance is not necessary for
```
gvanrossum: Start new paragraph at "Declaring variance ..."

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
+protocol classes, since it can be inferred from a protocol definition.
+Example::
+
+  class Box(Protocol[T]):
+      content: T
+
+  x: Box[float]
+  y: Box[int]
+  x = y  # This is OK due to inferred covariance of the protocol 'Box'.
```
gvanrossum: ...due to **the** inferred covariance ...

I'm also not sure how covariance is inferred here. Could you give examples of inferred covariance, contravariance, invariance?

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
 class names can be given as strings as specified by PEP 484. Recursive
-protocols will be useful for representing self-referential data structures
+protocols are useful for representing self-referential data structures
 like trees in an abstract fashion::
 
   class Traversable(Protocol):
       leaves: Iterable['Traversable']
 
+Note that for recursive protocols, a class is considered a subtype of
+the protocol in situations where such decision depends on itself.
```
gvanrossum: ... where the decision depends on the class itself.

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
+Continuing the previous example::
+
+  class SimpleTree:
+      leaves: List['SimpleTree']
+
+  root: Traversable = SimpleTree()  # OK
+
+  class Tree(Generic[T]):
+      def __init__(self, value: T,
+                         leaves: List['Tree[T]']) -> None:
```
gvanrossum: This indent looks funny, it `leaves` should be aligned with `self`.

pep-0544.txt
```diff
@@ -576,7 +610,7 @@ Example::
   cached_func((1, 2, 3)) # OK, tuple is both hashable and sequence
 
 If this will prove to be a widely used scenario, then a special
-intersection type construct may be added in future as specified by PEP 483,
+intersection type construct could be added in future as specified by PEP 483,
```
gvanrossum: I can't help congratulation on this correct use of "could". :-)

pep-0544.txt
```diff
@@ -704,9 +739,10 @@ one can use properties::
       def d(self) -> int:   # ... or it can be abstract
           return 0
 
-In Python 2.7 the function type comments should be used as per PEP 484.
-The ``typing`` module changes proposed in this PEP will be also
-backported to earlier versions via the backport currently available on PyPI.
+Also the function type comments could be used as per PEP 484 (for example
```
gvanrossum: "Also function type comments can be used ..."

pep-0544.txt
```diff
@@ -704,9 +739,10 @@ one can use properties::
...
+to provide compatibility with Python 2). The ``typing`` module changes
+proposed in this PEP will be also backported to earlier versions via the
```
gvanrossum: be also -> also be

pep-0544.txt
```diff
@@ -842,6 +904,43 @@ of simplicity, we propose to not support optional methods or attributes.
...
+----------------------------
+
+There was an idea to make some methods "non-protocol" (i.e. not necessary
+to implement, and inherited in explicit subclassing), but it was rejected,
+since this complicates things. For example, consider this function::
+
+  def fun(m: Mapping):
+      m.keys()
+
+The question is should this be an error? We think most people would expect
```
gvanrossum: Honestly I don't understand why this is even a question. OF COURSE it is valid, keys() is part of the Mapping class/protocol/ABC.

I do think I understand what's going on here, you're considering that keys() is a non-abstract method in Mapping, and the question is whether non-abstract methods are considered part of the protocol by default. I just think the example is a poor one (unless one happens to know that keys() has a default implementation in the Mapping class). So I think you just have to switch to an example that doesn't rely on a standard ABC.

pep-0544.txt
```diff
@@ -842,6 +904,43 @@ of simplicity, we propose to not support optional methods or attributes.
...
+this to be valid. The same applies to most other methods in ``Mapping``,
+people expect that they are provided by ``Mapping``. Therefore, to be on
+the safe side, we need to require these methods to be implemented.
+If one looks at definitions in ``collections.abc``, there are very few methods
+that could be considered "non-protocol". Therefore, it was decided to not
+introduce "non-protocol" methods.
+
+There is only one downside for this: it will require some boilerplate for
```
gvanrossum: for -> to

pep-0544.txt
JukkaL: Since the `content` attribute is mutable, `Box` should be invariant? And for consistency with normal classes, I'd prefer variance to be explicit. Especially if a generic protocol inherits from other protocols, it can be unclear whether a particular type variable should be covariant or invariant, so making this explicit would make code easier to understand. Users are often confused by variance, so it's likely that they often can't predict what the inferred variance would be. And what if I'd like to explicitly make something invariant for some reason, even if inference thinks it can be covariant?

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
 
   class Traversable(Protocol):
       leaves: Iterable['Traversable']
 
+Note that for recursive protocols, a class is considered a subtype of
+the protocol in situations where such decision depends on itself.
+Continuing the previous example::
+
+  class SimpleTree:
+      leaves: List['SimpleTree']
```
JukkaL: Since `leaves` is a mutable attribute, `SimpleTree` doesn't look compatible with `Traversable`. If the attribute in `Traversable` would be a read-only property, this would be fine.

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
 ===============
 
 Subtyping relationships with other types
 ----------------------------------------
 
-Protocols cannot be instantiated, so there are no values with
-protocol types. For variables and parameters with protocol types, subtyping
-relationships are subject to the following rules:
+Protocols cannot be instantiated, so there are no values whose
+*exact* type is a protocol. For variables and parameters with protocol types,
```
JukkaL: Maybe use 'runtime type' instead of 'exact type' as it's less open to interpretation.

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
+subtyping relationships are subject to the following rules:
 
 * A protocol is never a subtype of a concrete type.
-* A concrete type or a protocol ``X`` is a subtype of another protocol ``P``
+* A concrete type ``X`` is a subtype of protocol ``P``
```
JukkaL: Maybe also mention that the implementations must have compatible types.

pep-0544.txt
```diff
@@ -439,98 +448,123 @@ but a static type checker will handle them specially. Subclassing a protocol
...
-                           leaves: 'List[Tree[T]]') -> None:
-            self.value = value
-            self.leafs = leafs
-
-    def walk(graph: Traversable) -> None:
-        ...
-    tree: Tree[float] = Tree(0, [])
-    walk(tree) # OK, 'Tree[float]' is a subtype of 'Traversable'
+* A protocol ``P1`` is a subtype of another protocol ``P2`` if ``P1`` defines
+  all protocol members of ``P2``.
```
JukkaL: Again, maybe mention that the members must have compatible types.

pep-0544.txt
```diff
@@ -842,6 +904,43 @@ of simplicity, we propose to not support optional methods or attributes.
 We can always revisit this later if there is an actual need.
 
 
+Allow only protocol methods and force use of getters and setters
+----------------------------------------------------------------
+
+One could argue that protocols typically only define methods, but not
+variables. However, using getters and setters in cases where only a
+simple variable is needed would be quite unpythonic. Moreover, the widespread
+use of properties (that are often act as validators) in large code bases
```
JukkaL: 'thar are often act' -> 'that often act'. Also, I'm unsure what 'act as validators' means here. Maybe rephrase or give an example?

pep-0544.txt
```diff
@@ -842,6 +904,43 @@ of simplicity, we propose to not support optional methods or attributes.
...
+the safe side, we need to require these methods to be implemented.
+If one looks at definitions in ``collections.abc``, there are very few methods
+that could be considered "non-protocol". Therefore, it was decided to not
+introduce "non-protocol" methods.
+
+There is only one downside for this: it will require some boilerplate for
+implicit subtypes of ``Mapping`` and few other "large" protocols. But, this
+applies to few "built-in" protocols (like ``Mapping`` and ``Sequence``) and
+people are already subclassing them. Also, such style is discouraged for
+user defined protocols. It is recommended to create compact protocols and
```
JukkaL: Maybe write 'user defined' as 'user-defined' (with a hyphen)?

pep-0544.txt
```diff
@@ -949,6 +1048,44 @@ this in type checkers for non-protocol classes could be difficult. Finally, it
 will be very easy to add this later if needed.
 
 
+Prohibit explicit subclassing of protocols by non-protocols
+-----------------------------------------------------------
+
+This was rejected mainly for two reasons:
```
JukkaL: There are additional reasons:

* Explicit subclassing makes it explicit that a class implements a particular protocol, making subtyping relationships easier to see.
* Type checkers can warn about missing protocol members or members with incompatible types more easily, without having to use hacks like dummy assignments.

pep-0544.txt
```diff
@@ -949,6 +1048,44 @@ this in type checkers for non-protocol classes could be difficult. Finally, it
...
+  For example in the case of ``Sequence`` one only needs to implement
+  ``__getitem__`` and ``__len__`` in an explicit subclass, and one gets
+  ``__iter__``, ``__contains__``, ``__reversed__``, ``index``, and ``count``
+  for free.
+
+
+Backwards Compatibility
+=======================
+
+This PEP is fully backwards compatible.
```
JukkaL: If we turn `Sequence` and friends into runtime protocols, the results of some `isinstance` checks are going to change in some edge cases. (Also, the performance of these `isinstance` checks might be affected, though not sure about it.)

pep-0544.txt
ilevkivskyi: Yes, good catch!

pep-0544.txt
ilevkivskyi: @JukkaL 
Yes, ``Box`` should be invariant, completely forgot about "mutability" in my implementation. Fortunately, this is easy to fix, I will just check subclassing both ways for members with ``IS_SETTABLE`` flag (basically treating a settable member as a pair of getter/setter methods).
> And for consistency with normal classes, I'd prefer variance to be explicit.

OK, I agree, I will just remove the text mentioning that variance could be omitted.
> And what if I'd like to explicitly make something invariant for some reason, even if inference thinks it can be covariant?

Good question! I am not sure what to do here. Currently, I do two-step checking:
* First check nominal subtyping using user-declared variance, if it succeeds, then OK, otherwise go to next step.
* Try structural subtyping. Here I check compatibility of every member (types and flags like ``IS_SETTABLE``, ``IS_CLASSVAR``, etc) I think this is quite robust, so that the inferred variance should be reliable (unless I am missing something).

In principle, I probably could always use the user-declared variance, but this will have two downsides:
* 90% of users use invariant variables, so that protocols will be interpret as invariant. For nominal classes, this makes perfect sense, but for protocols, where we anyway do a lot of work checking all members, we could reliably infer variance, and making them all invariant will be quite sad.
* Less important, but this will also make implementation more complex. Currently, when I check ``is_subtype(C, Proto[int])``, I just substitute ``T`` with ``int`` in ``Proto`` and check all members. With user declared variance I will first need something like ``map_instance_to_supertype`` and then use variance.

ilevkivskyi:
@gvanrossum @JukkaL Thank you for your reviews! I agree with all points and will implement them soon. The only open question is about user-defined vs inferred variance for protocols, see above.

JukkaL:
Covariant overriding of attributes should be rejected (https://github.com/python/mypy/issues/3208). Not sure why it's allowed -- may have been an oversight. Also, I'm not sure I actually understood correctly what you meant by inferred variance. It's worth specifying this in more detail, since now it's open to interpretation.

ilevkivskyi:
@JukkaL 
> Covariant overriding of attributes should be rejected (python/mypy#3208).

OK, agreed.

> Also, I'm not sure I actually understood correctly what you meant by inferred variance. It's worth specifying this in more detail, since now it's open to interpretation.

I was referring to this discussion https://github.com/python/peps/pull/243#discussion_r112501384

Here are some examples:
```python
class ProtoCo(Protocol[T]):
    def meth(self) -> T:
        ...
class ProtoCotra(Protocol[T]):
    def meth(self, arg: T) -> None:
        ...
class ProtoInv(Protocol[T]):
    def first(self) -> T:
        ...
    def second(self, arg: T) -> None:
        ...
```
In current implementation, ``ProtoCo[int]`` is a subtype of ``ProtoCo[float]``, and ``ProtoContra[flat]`` is a subtype of ``ProtoContra[int]``. This happens "naturally" in my implementation without additional effort, this is what I call the "inferred variance".

I would like to keep this behavior instead of treating all the above protocols as invariant. It fits a simple mental model (if ``Proto[A]`` can be an explicit subclass of ``Proto[B]``, then it is a subtype). Also there are additional arguments in  https://github.com/python/peps/pull/243#discussion_r112501384

What do you think?

ilevkivskyi:
I just made a quick search on approaches in other languages. Here is a brief summary:
* Go: interfaces are always invariant (and people complain about this).
* TypeScript: interfaces are bivariant, but right now there is a discussion about switching to inferred variance.
* C#: interface variance is user declared.
* Java: question is not applicable, there is only use-site variance.

It looks like there is no wide consensus between different languages about this. In the meantime I have noticed that most people who complain about default invariance of interfaces are not aware about variance, they just feel naturally that their ``AnInterface<Sub>`` is a subtype of ``AnInterface<Super>``, i.e. they intuitively want covariance in this case. Therefore, I think that going with inferred variance will actually reduce the "decision burden" for users.

gvanrossum:
Thank you for the research. I have merged this version, since it is your PEP and it is your proposal to use implicit variance. Then we can have an additional discussion without this PR going stale.

gvanrossum:
FWIW my own opinion on inferred invariance is wishy-washy. Looking at your [three examples](https://github.com/python/peps/pull/243#issuecomment-296169818) I could easily see evolution of a protocol that is currently covariant accidentally make it invariant. Contravariance is also counter-intuitive (whenever it comes up I always have to derive the reasons from a basic example).

I think two situations are useful:

(a) Invariant protocols; sometimes there are good reasons for this (e.g. they represent interfaces to mutable objects, like MutableSequence). We'll have to explain forever that MutableContainer[<subclass>] is no a subtype of MutableContainer[<super>] but so be it, and the reason is a good one -- my assumption is that basically everyone has to bump into this once and then they know.

(b) Covariant protocols; these are easier to use (they are more intuitive somehow, since everyone somehow *expects* that List[<subclass>] is a subtype of List[<superclass>]). But they cannot represent mutable objects.

My worry with default inferred variance is that people won't understand why adding a certain method to their (previously covariant) protocol suddenly makes it invariant, hence breaking some code that's unchanged. Had variance been explicitly declared they would have been forced to think about this sooner and it would be no surprise that they can't add a method that violates the declared variance. (No action-at-a-distance, or in this case error-at-a-distance.)

There's an additional concern. I think it's come up in the attempts to [make Mapping covariant in its keys](https://github.com/python/mypy/issues/1114). This is disallowed because arguments need to be contravariant (this is the most counter-intuitive part of the whole thing). It is annoying because there seems to be no way to create the unsafe scenario for Mapping that justifies the ban on covariant arguments. But why can't the type system tell the difference? Frankly this baffles me.

ilevkivskyi:
Here are some random thoughts:

1. Action-at-a-distance is typical for structural subtyping, most changes in protocol method signatures will break some distant code, whether it is accompanied by a variance change or not. I think most users understand this, everyone who changes e.g. ``Iterable`` expects that a lot of code will break, and one will do such things only for a good reason.

2. There are still warnings at protocol declaration site for obviously incorrect use of type variables:
    ```python
    class Proto(Protocol[T_co, T_contra]):
        def one(self, x: T_co) -> None: ...  # Error!
        def other(self) -> T_contra: ...  # Error!
    ```
    the semantics is practically only changed for plain ``T``, it is invariant for nominal classes, but "auto" for protocols.

3. ``Mapping`` invariant in key is a "guard" against excessively inventive users. For example, this will pass type check with ``Mapping`` _covariant_ in key type, but will fail at runtime:
    ```python
    def fun(m: Mapping[float, int]) -> int:
        return m[0.1]

    class StrangeMap(Mapping[int, int]):
        def __getitem__(self, item: int) -> int:
            return item << 5

    fun(StrangeMap())  # Boom!
    ```
    IIRC, there was a discussion at some point about ignoring users who write such code, but finally we decided not to do this.

4. Currently, bivariant types are not supported at all, but actually the protocols PR correctly detects bivariant types, e.g. ``Container`` as it is currently defined in typeshed, without any additional effort.

PR #257, opened 2017-05-15
bpo-9850: PEP 4: macpath deprecated in Python 3.7

haypo:


pep-0004.txt
```diff
@@ -280,6 +280,12 @@ Deprecated modules
     Date:          2013-08-12
     Documentation: Deprecated as of Python 3.4.
 
+    Module name:   macpath
+    Rationale:     Obsolete macpath module dangerously broken
```
DimitrisJim: To go along with other entries, the rationale could just state "Platform is obsolete/unsupported", otherwise LGTM

haypo: done

PR #261, opened 2017-05-18
PEP 526: Add using ClassAttr instead of ClassVar to rejected ideas

ilevkivskyi:
I also fix some reST formatting.


brettcannon:
Thanks!

PR #264, opened 2017-05-22
PEP 517: Specify UTF-8 for captured output

takluyver:
From a discussion on the mailing list.

This is not yet completely agreed upon, but it's my best take on the solution.

### DIFF
--- a/pep-0517.txt
+++ b/pep-0517.txt
@@ -262,6 +262,15 @@ tree, and MAY print arbitrary informational text on stdout and
 stderr. They MUST NOT read from stdin, and the build frontend MAY
 close stdin before invoking the hooks.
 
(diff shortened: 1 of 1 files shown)


brettcannon:
Since you're an author on the PEP, @takluyver , we will commit this when you want us to (wasn't sure since you said there wasn't agreement in your PR message).

takluyver:
There's still a debate going on over whether to use UTF-8 everywhere, or to use platform dependent locale encodings. I'm very much hoping that we don't have to mess around with locale encodings - especially since they may not be able to represent all characters - but hold off on merging until there's some kind of consensus.

ncoghlan:
The resolution of the discussion on distutils-sig was to accept this change: https://mail.python.org/pipermail/distutils-sig/2017-May/030504.html

PR #271, opened 2017-05-30
PEP 484: Allow deriving NewType from another unique type

ilevkivskyi:
See discussion in https://github.com/python/mypy/pull/3465


gvanrossum:
Done!

PR #276, opened 2017-05-31
reSTify 5 PEPs

csabella:
For #4.

### DIFF
--- a/pep-0342.txt
+++ b/pep-0342.txt
@@ -5,591 +5,596 @@ Last-Modified: $Date$
 Author: Guido van Rossum, Phillip J. Eby
 Status: Final
 Type: Standards Track
-Content-Type: text/plain
+Content-Type: text/x-rst
 Created: 10-May-2005
 Python-Version: 2.5
 Post-History:
 
+
 Introduction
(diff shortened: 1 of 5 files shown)


pep-0342.txt
```diff
@@ -5,591 +5,596 @@ Last-Modified: $Date$
...
+CVS HEAD is available as SourceForge patch #1223381
+(http://python.org/sf/1223381).
 
 
 Specification: Sending Values into Generators
+=============================================
+
+New generator method: ``send(value)``
+
+   A new method for generator-iterators is proposed, called ``send()``.  It
```
brettcannon: I just want to double-check that this indentation ends up looking okay in the final HTML?

csabella: @brettcannon  Yes, it had looked good (like the original) in the html, but I went back and checked PEP-12 and see your point.  I should use 4 spaces to indent if it's a block quote.  But, looking at it fresh, maybe I should make the 'New generator method:' type lines into subheaders.  I'll rework it.

brettcannon:
Thanks!

PR #277, opened 2017-05-31
PEP 517: Add prepare_build_files hook

takluyver:
Alternative to #275 

@dstufft @pfmoore @leorochael

### DIFF
--- a/pep-0517.txt
+++ b/pep-0517.txt
@@ -218,6 +218,8 @@ metadata. The directory passed in by the build frontend MUST be
 identical to the directory created by ``get_wheel_metadata``,
 including any unrecognized files it created.
 
+Mandatory.
+
 ::
 
   def prepare_build_files(build_directory, config_settings):
@@ -235,11 +237,9 @@ Optional. If this hook is not defined, the copying step will be skipped, and the
 ``build_wheel`` hook invoked in the original source directory. In this case, the
 backend is responsible for not cluttering up the source directory.
 
-Mandatory.
-
 .. note:: Editable installs
 
-   This PEP originally specified a fourth hook, ``install_editable``, to do an
+   This PEP originally specified another hook, ``install_editable``, to do an
    editable install (as with ``pip install -e``). It was removed due to the
    complexity of the topic, but may be specified in a later PEP.
 
--- a/pep-0517.txt
+++ b/pep-0517.txt
@@ -225,11 +225,12 @@ Mandatory.
   def build_sdist(sdist_directory, config_settings):
       ...
 
-Must build a ``.tar.gz`` source distribution file, and place it in the specified
+Must build an unpacked source distribution in the specified
 ``sdist_directory``.
 
-A source distribution (sdist) is a tarball containing a single top-level
-directory called ``{name}-{version}``. This directory should contain the
+An unpacked source distribution (sdist) consists of a directory called
+``{name}-{version}`` (e.g. ``foo-1.0``), containing the source files of the
+package. This directory must also contain the
 ``pyproject.toml`` from the build directory, and a PKG-INFO file containing
 metadata in the format described in
 `PEP 345 <https://www.python.org/dev/peps/pep-0345/>`_.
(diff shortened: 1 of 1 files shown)


pep-0517.txt
```diff
@@ -218,6 +218,23 @@ metadata. The directory passed in by the build frontend MUST be
...
+  def prepare_build_files(build_directory, config_settings):
+      ...
+
+Must copy or create any files needed to build this package into
+``build_directory``. For instance, it is expected  that ``pyproject.toml`` will
+exist in the root of this directory after this hook runs. For tools such as
+`setuptools_scm <https://github.com/pypa/setuptools_scm>`_, this may include
+extracting some information from a version control system.
+The ``build_wheel`` hook will subsequently be run from the ``build_directory``
+populated by this hook.
```
pfmoore: It's probably obvious, but should be stated explicitly, that ```build_wheel``` run from the directory created by this step must produce an identical wheel to the one produced by running ```build_wheel``` in the original directory.

Should it also be stated that ```pyproject.toml``` must be copied unchanged (to prohibit the sort of silliness I described on distutils-sig)?

But overall, I like this version. It's clean and straightforward.

pep-0517.txt
```diff
@@ -218,6 +218,23 @@ metadata. The directory passed in by the build frontend MUST be
...
+
+Optional. If this hook is not defined, the copying step will be skipped, and the
+``build_wheel`` hook invoked in the original source directory. In this case, the
+backend is responsible for not cluttering up the source directory.
+
 Mandatory.
```
pfmoore: I suspect "Mandatory" is misplaced or should be removed?

pep-0517.txt
```diff
@@ -218,6 +218,23 @@ metadata. The directory passed in by the build frontend MUST be
...
+``build_directory``. For instance, it is expected  that ``pyproject.toml`` will
+exist in the root of this directory after this hook runs. For tools such as
+`setuptools_scm <https://github.com/pypa/setuptools_scm>`_, this may include
+extracting some information from a version control system.
+The ``build_wheel`` hook will subsequently be run from the ``build_directory``
+populated by this hook.
+
+Optional. If this hook is not defined, the copying step will be skipped, and the
+``build_wheel`` hook invoked in the original source directory. In this case, the
+backend is responsible for not cluttering up the source directory.
```
pfmoore: I'm not sure I like this. I can't see a problem with it, but I've a nagging feeling that I'd prefer the front end to be in charge of the "copy or not" decision. Also, it's not necessarily the backend that clutters the source directory, it could be the user. But to repeat, I can't think of an *actual* issue here, so unless someone else comes up with a specific way that this could break something, I'm not going to object.

takluyver: I don't feel strongly about this, my rationale is just that if the backend knows it can build a wheel without creating clutter in the source directory (as flit does), then we can save copying the files. But the normal operation should be installing from a pre-built wheel anyway, so I don't think performance is that critical for this step.

ncoghlan:
I went ahead and merged this to bring the published PEP in line with current distutils-sig discussion, but will add a couple of comments inline for this I think are still a bit ambiguous.

pep-0517.txt
```diff
@@ -220,9 +220,45 @@ including any unrecognized files it created.
 
 Mandatory.
 
+::
+
+  def build_sdist(sdist_directory, config_settings):
+      ...
+
+Must build an unpacked source distribution in the specified
```
ncoghlan: With the sdist now being unpacked, a better name for the hook is likely to be `export_sdist`.

It's also not entirely clear if the sdist contents should be exported directly into the supplied `sdist_directory`, or if the backend should be creating an appropriately named subdirectory.

My inclination would be to leave the directory naming up to the frontend, and have the backend just dump the sdist contents into whichever path is given (as happens with the wheel metadata export)

pep-0517.txt
ncoghlan: +1 for stating explicitly that `pyproject.toml` must be exported unchanged, and that the build output should be the same regardless of whether or not the frontend actually calls this hook.

PR #278, opened 2017-06-01
Address concerns from python-dev for PEP 546.

Lukasa:


Lukasa:
Done. =)

pep-0546.txt
```diff
@@ -3,6 +3,7 @@ Title: Backport ssl.MemoryBIO and ssl.SSLObject to Python 2.7
 Version: $Revision$
 Last-Modified: $Date$
 Author: Victor Stinner <victor.stinner@gmail.com>,
+        Cory Benfield <cory@lukasa.co.uk>,
```
haypo: just to be explicit: i asked Cory to do this change, so I approve it ;-)

PR #286, opened 2017-06-05
Cross-reference PEP 3131 from PEP 8

ncoghlan:
PEP 3131 (Unicode identifiers) includes a policy section
covering their use within the standard library.

This adds an explicit cross-reference to that section from
the prescriptive naming conventions section of PEP 8


ncoghlan:
Thanks for the reviews, folks!

PR #288, opened 2017-06-09
PEP 517: Misc changes from discussion

takluyver:
Ping @ncoghlan - I think this implements everything you suggested on  #277 except for the sdist directory question (does the backend create a `foo-1.0` directory, or just dump the files directly into `build_directory`?). I have no strong feeling on that question, but you didn't sound certain, so I thought I'd leave it for further discussion before making a change.

### DIFF
--- a/pep-0517.txt
+++ b/pep-0517.txt
@@ -222,10 +222,10 @@ Mandatory.
 
 ::
 
-  def build_sdist(sdist_directory, config_settings):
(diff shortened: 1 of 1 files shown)


ncoghlan:
Thanks. I'll start a new distutils-sig thread covering what I consider to be the open questions (mainly around deciding on a consistent approach to the artifact export step.

PR #289, opened 2017-06-12
Updates to PEP 544

ilevkivskyi:
Here are few (minor) updates based on a recent discussion on python-dev. Most notable change is the change of title.

Attn: @JukkaL @ambv @gvanrossum

### DIFF
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -1,5 +1,5 @@
 PEP: 544
-Title: Protocols
+Title: Protocols: Structural subtyping (static duck typing)
 Version: $Revision$
 Last-Modified: $Date$
 Author: Ivan Levkivskyi <levkivskyi@gmail.com>, Jukka Lehtosalo <jukka.lehtosalo@iki.fi>, Łukasz Langa <lukasz@langa.pl>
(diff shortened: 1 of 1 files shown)


pep-0544.txt
```diff
@@ -1238,6 +1270,32 @@ Since there is a reasonable alternative for such cases with existing tooling,
...
 
+Call structural base types "interfaces"
+---------------------------------------
+
+In addition to the reasons given for the term "protocols" in `rationale`_,
+we don't want to use the term "interfaces" since it is already used
+by many languages with subtly different semantics. In particular, protocols
+are different from Java interfaces in one important aspect:
+protocols don't require explicit declaration of implementation,
+they are mainly oriented on duck-typing.
```
ambv: Well, there's a few differences between Java interfaces. Implicit implementation is one but also the fact that protocols can have default implementations and store state. Also, this argument is a bit flawed since the term "protocols" is also used by a number of languages with subtly different semantics from ours. For example, Objective C and Swift have protocols but **also require explicit conformance** (what you call here the "declaration of implementation").

Instead, I would focus on the other reason you pointed out on the mailing list: "protocol" is a term already widely used in Python to describe duck typing contracts. You can give examples.

ilevkivskyi: Yes, you are right. I have now reformulated this in a new commit.

ilevkivskyi:
Maybe this can be merged now, so that the new version will be displayed (including the new title)?

gvanrossum:
Sorry!

PR #293, opened 2017-06-15
reSTify PEP225, PEP234, PEP255, PEP450

csabella:
For #4.


serhiy-storchaka:
Would be better to avoid reflowing lines. This would make reviewing easier.

csabella:
@serhiy-storchaka   Sorry, I thought one of the steps in converting to reST was to make the lines fit close to 79 characters.  Should I not do that?

brettcannon:
@csabella I think @serhiy-storchaka makes an interesting point of reviewability. I say don't worry about reflowing since almost everyone views the PEPs as rendered HTML anyway.

brettcannon:
Thanks!

PR #295, opened 2017-06-17
Remove spurious whitespace

lelit:
There was a spurious space at the beginning of the line, that transformed the paragraph into a definition list item.


ned-deily:
Thanks for the report!

PR #299, opened 2017-06-26
Mark PEP 508 as replacing PEP 496

brettcannon:


ncoghlan:
@brettcannon Hmm, I guess technically 496 should really be marked as rejected, since it was never accepted - it and the more comprehensive 508 were formally competitors from a process perspective.

brettcannon:
OK, then the opening text in PEP 496 is a bit misleading since it says it was replaced with PEP 508.

PR #300, opened 2017-06-28
Use correct status for PEP 496

ncoghlan:
PEP 496 was never Accepted or Active, so the correct status
is "Rejected in favour of PEP 508" rather than "Superseded
by PEP 508".


ncoghlan:
Alternative to #299 that fixes the status of PEP 496, rather than adding the Supersedes header to PEP 508.

PR #308, opened 2017-07-14
517: Add note about fresh subprocess per build

takluyver:
Also a missing heading.


takluyver:
cc @ncoghlan you mentioned this on my previous PR

PR #309, opened 2017-07-15
PEP 484: Clarify that submodules in a package are exported

JelleZijlstra:
See discussion in https://github.com/python/typeshed/pull/1484#discussion_r127114355.

cc @ilevkivskyi @gvanrossum


gvanrossum:
Great! I couldn't have done it better.

PR #310, opened 2017-07-15
PEP 517: update backend example

ncoghlan:
- update hook names and signatures
- clearly separate sdist building & wheel building
- add support for out-of-tree wheel builds

### DIFF
--- a/pep-0517.txt
+++ b/pep-0517.txt
@@ -198,20 +198,25 @@ If ``build_directory`` is not None, it is a unicode string containing the
 path to a directory other than the source directory (the working directory where
 the hook is called) where intermediate build artifacts may be stored.
 
+These out-of-tree builds should have the same result as first building an sdist
+and then building a wheel from that sdist. If the backend cannot otherwise
+ensure that unexpected files won't end up in the built wheel archive, it should
+copy any essential input files it needs to ``build_directory`` and perform the
+build there.
+
 The provided build directory may be empty, or it may contain artifacts from a
 previous build to be used as a cache. The backend is responsible for determining
-whether any cached artifacts are outdated. When a build directory is provided,
-the backend should not create or modify any files in the source directory. If
-the backend cannot reliably avoid modifying the directory it builds from, it
-should copy any files it needs to ``build_directory`` and perform the build
-there.
+whether any cached artifacts are outdated.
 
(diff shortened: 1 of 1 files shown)


ncoghlan:
@takluyver This updates the minimalist backend example in PEP 517 to match the latest API definition (including full out-of-tree build support by way of creating and unpacking the sdist)

pep-0517.txt
```diff
@@ -636,14 +634,51 @@ build backend::
...
+        sdist.extractall()
+        os.remove(sdist_filename)
+        os.chdir(SDIST_NAME)
+
+    def build_wheel(wheel_directory, build_directory=None,
+                    metadata_directory=None, config_settings=None):
+        """PEP 517 wheel creation hook"""
+        # First check if the frontend has requested an out-of-tree build
+        out_of_tree_build = (build_directory is not None and
+                             not os.path.samefile(build_directory, os.getcwd())
```
takluyver: Do we want to make a note about this case (`build_directory==source_directory`) in the spec? I would have written the code assuming that if build_directory is not None, we're doing an out of tree build. I don't mind whether we say that it can be the same as the source directory or that it should be different, but we should probably make it explicit either way.

ncoghlan: I think we should say that frontends *should* indicate that case as "build_directory is None", but a defensively coded backend may want to check for it with `os.path.samefile()` anyway.

takluyver: Sounds good to me.

ncoghlan:
@takluyver I've pushed an update that attempts to clarify the handling of build_directory.

I also realised that `metadata_directory` was underspecified for the case where the backend *doesn't* define prepare_wheel_metadata, so I added a paragraph that basically says "backends can do whatever they want" (which shouldn't be a problem, since frontends will have already built the wheel as part of getting the metadata in that case).

ncoghlan:
@pfmoore pointed out on distutils-sig that the current wording of the out-of-tree build description in build_wheel doesn't explicitly say "this should give the same result as building via sdist would", so I updated it to just say that.

ncoghlan:
@takluyver I went ahead and merged this version in order to get the updates live on python.org, and also to make way for an upcoming PR to remove the remaining references to `prepare_wheel_metadata` that @dstufft noticed.

PR #312, opened 2017-07-20
PEP 517: Fix metadata preparation example

ncoghlan:
- use plain strings rather than pathlib.Path
- drop config_settings as irrelevant
- return relevant absolute paths from the example
  helper functions


ncoghlan:
Thanks for the review!

PR #313, opened 2017-07-20
517: Move comparison to PEP 516 into an appendix

takluyver:
We have a long section comparing the two PEPs which is now mostly of historical interest, so I've moved it to an appendix.

I'd still be in favour of making it shorter as well, because we keep needing to update its examples as the specification evolves (and I keep forgetting to do so). But I haven't done that yet.

cc @ncoghlan @njsmith


ncoghlan:
Cool - this at least moves it out of the way so it isn't interrupting the main flow of the PEP anymore, and is clearly an optional addon.

PR #340, opened 2017-08-13
Fix typos in PEP 491

jwodder:
Two lines of RFC822-style headers were being formatted as a single-line blockquote instead of a preformatted code block.  In addition, the prose immediately before them called them "files" instead of "fields."  This PR fixes that.


brettcannon:
@dholth care to review this?

brettcannon:
@dholth Thanks!

PR #341, opened 2017-08-13
Fix footnote/reference links in PEP 496

jwodder:
The footnote/reference links in PEP 496 weren't actually hyperlinks.  This fixes that.


brettcannon:
Thanks! I believe we also automatically link anything formatted as `PEP NNNN`, so it might also work in the future to just ditch the hyphen to get hyperlinking to work.

PR #362, opened 2017-08-24
PEP 551: Add "discussions-to" header

Mariatta:
Discussions to: security-sig@python.org

### DIFF
--- a/pep-0551.rst
+++ b/pep-0551.rst
@@ -3,6 +3,7 @@ Title: Security transparency in the Python runtime
 Version: $Revision$
 Last-Modified: $Date$
 Author: Steve Dower <steve.dower@python.org>
+Discussions-To: <security-sig@python.org>
 Status: Draft
 Type: Standards Track
 Content-Type: text/x-rst


zooba:
I'm not sure - this will eventually go to python-dev, it's just at security-sig for initial review.

My understanding of `discussions-to` is that it is mainly intended for distutils-sig, since those PEPs never come past python-dev. Probably all we want is to mark the post history as having had a thread in security-sig.

zooba:
Yeah, [PEP 1](https://github.com/python/peps/blob/master/pep-0001.txt#L229) says it's when discussion _and resolution_ takes place somewhere other than python-dev. That won't be the case here - it'll definitely go to python-dev for resolution.

ncoghlan:
That header is actually used for both purposes (and the "initial design discussion before sending it to python-dev" usage is the older of the two): https://github.com/python/peps/blob/master/pep-0001.txt#L421

So setting it temporarily for the benefit of folks that read the PEP without having seen your initial security-sig post may make sense.

Mariatta:
Cool! Thanks for the clarification @ncoghlan and @zooba 😃

PR #364, opened 2017-08-25
PEP 517: Remove build_directory parameter

takluyver:
We seem to be reaching the conclusion that this is not needed, and removing it allows the spec to be simpler.


ncoghlan:
Even if we make further changes and clarifications, I figure it makes sense to bring this part of the spec in line with the latest round of discussions.

PR #382, opened 2017-08-31
Update pep-0539.txt

embray:
A few minor language updates and clarifications--~~no substantive changes~~ one substantive change removing the use of `bool` types..  @ma8ma

### DIFF
--- a/pep-0539.txt
+++ b/pep-0539.txt
@@ -162,7 +162,7 @@ Default Value      (``-1`` as key creation        ``Py_tss_NEEDS_INIT``
 Requirement        native threads                 native threads
                    (since CPython 3.7 [9]_)
 Restriction        No support for platforms       Unable to statically allocate
-                   wher native TLS key is         keys when ``Py_LIMITED_API``
+                   where native TLS key is        keys when ``Py_LIMITED_API``
                    defined in a way that cannot   is defined.
                    be safely cast to ``int``.
 =================  =============================  =============================
--- a/pep-0539.txt
+++ b/pep-0539.txt
@@ -210,9 +210,8 @@ Platform Support Changes
 A new "Native Thread Implementation" section will be added to PEP 11 that
 states:
 
-* As of CPython 3.7, in the case of enabling threads support in the
-  interpter, all platforms are required to provide at least one of native
-  thread implementation (such as pthreads or Windows) to implement the TSS
+* As of CPython 3.7, all platforms are required to provide a native thread
+  implementation (such as pthreads or Windows) to implement the TSS
   API.  Any TSS API problems that occur in an implementation without native
   threads will be closed as "won't fix".
 
@@ -319,16 +318,15 @@ Rejected Ideas
   wishing to be supported by CPython should follow the requirements of
   PEP-11.  As explained above, while this would be a fair argument if
   CPython were being to asked to make changes to support particular quirks
-  or features of a specific platform, in this case it is quirk of CPython
+  or features of a specific platform, in this case it is a quirk of CPython
   that prevents it from being used to its full potential on otherwise
   POSIX-compliant platforms.  The fact that the current implementation
   happens to work on Linux is a happy accident, and there's no guarantee
   that this will never change.
 
 * Affected platforms should just configure Python ``--without-threads``:
-  This is a possible temporary workaround to the issue, but only that.
-  Python should not be hobbled on affected platforms despite them being
-  otherwise perfectly capable of running multi-threaded Python.
+  this is no longer an option as the ``--without-threads`` option has
+  been removed for Python 3.7 [16]_.
 
 * Affected platforms should use CPython's built-in TLS implementation
   instead of a native TLS implementation: This is a more acceptable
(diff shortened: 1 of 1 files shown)


pep-0539.txt
```diff
@@ -150,20 +151,20 @@ Features           - create key                   - create key
...
                    failure)
-Requirement        native thread                  native thread
+Requirement        native threads                 native threads
                    (since CPython 3.7 [9]_)
-Restriction        Not support platform where     Unable to statically allocate
-                   native TLS key is defined in   key when ``Py_LIMITED_API``
-                   a way that cannot be safely    is defined.
-                   cast to ``int``.
+Restriction        No support for platforms       Unable to statically allocate
+                   wher native TLS key is         keys when ``Py_LIMITED_API``
```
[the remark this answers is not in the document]
embray: Thanks!

brettcannon:
@ma8ma if you're happy with the changes, then just let us know and someone can merge this.

ma8ma:
I'd change PyThread_tss_is_crated declaration's bool that is [pointed out][p] by @zooba.
@embray Would you change the specification as follows?

```diff
diff --git a/pep-0539.txt b/pep-0539.txt
index a9905a7..5a7ab18 100644
--- a/pep-0539.txt
+++ b/pep-0539.txt
@@ -77,7 +77,7 @@ The specification also adds a few new features:
 
       PyAPI_FUNC(Py_tss_t *) PyThread_tss_alloc(void)
       PyAPI_FUNC(void) PyThread_tss_free(Py_tss_t *key)
-      PyAPI_FUNC(bool) PyThread_tss_is_created(Py_tss_t *key)
+      PyAPI_FUNC(int) PyThread_tss_is_created(Py_tss_t *key)
 
   The first two are needed for dynamic (de-)allocation of a ``Py_tss_t``,
   particularly in extension modules built with ``Py_LIMITED_API``, where
@@ -87,8 +87,8 @@ The specification also adds a few new features:
   dynamic allocation failure.  The behavior of ``PyThread_tss_free`` involves
   calling ``PyThread_tss_delete`` preventively, or is no-op if the value
   pointed to by the ``key`` argument is ``NULL``.  ``PyThread_tss_is_created``
-  returns ``true`` if the given ``Py_tss_t`` has been initialized (i.e. by
-  ``PyThread_tss_create``).
+  returns non-zero if the given ``Py_tss_t`` has been initialized (i.e. by
+  ``PyThread_tss_create``), otherwise returns zero.
 
 The new TSS API does not provide functions which correspond to
 ``PyThread_delete_key_value`` and ``PyThread_ReInitTLS``, because these

```

[p]: https://github.com/python/cpython/pull/1362#pullrequestreview-59884901

embray:
Yeah, I agree with that change.  I'll make the update.

ma8ma:
@embray Thank you for updating :+1: 
@brettcannon It's OK to merge!

pep-0539.txt
```diff
@@ -327,14 +330,15 @@ Rejected Ideas
   Python should not be hobbled on affected platforms despite them being
   otherwise perfectly capable of running multi-threaded Python.
```
ma8ma: I'd suggest to add a sentence at the end `This idea cannot be adopted because the ``--without-threads`` option has since been removed [16]_.`. And the footnotes `[16]: https://bugs.python.org/issue31370`.

pep-0539.txt
```diff
@@ -209,10 +210,11 @@ Platform Support Changes
 A new "Native Thread Implementation" section will be added to PEP 11 that
 states:
 
-* As of CPython 3.7, in the case of enabling thread support, all platforms are
-  required to provide at least one of native thread implementation (as of
-  pthreads or Windows) to implement TSS API. Any TSS API problems that occur
-  in the implementation without native thread will be closed as "won't fix".
+* As of CPython 3.7, in the case of enabling threads support in the
+  interpter, all platforms are required to provide at least one of native
```
ma8ma: The note `in the case of enabling threads support in the  interpter,` no longer makes any sense, it could be removed.

ncoghlan:
I incorporated @ma8ma's changes to account for the `--without-threads` removal, so I'm going to go ahead and merge this.

If you submit this version python-dev for pronouncement, I'll happily accept it :)

embray:
Oof--I wasn't happy for a second about the removal of `--without-threads`, but I guess with MicroPython being a thing maybe it's not very important now.  Still seems like a thing that happened with very minimal input from the broader community...  Do we *actually* know how many people are using it or just making up statistics like "99.99%".  Oh well, off-topic for this.  Thanks @ncoghlan for making the change.

bitdancer:
We've actually been discussing it off and on for years, with as far as I remember there have been only theoretical objections to doing it (ie: no actual users have stepped forward to object).  We end up breaking the --without-threads build on a fairly frequent basis, and the complaints about it being broken come from the core team or gentoo (which is probably the only distro that supports python without threads, as an option).  (I found one exception to that pattern in my googling, but only one).  That's fairly strong evidence that it isn't actually used, and we do have time before 3.7 is released to find out we're wrong, if we're wrong :)

embray:
Thanks, for the additional background, that makes sense.  I'm always wary of such changes, especially since I was relying on this to build Python on Cygwin for some time now (and have fixed quite a few of the non-threads cases).  But this was always meant as a temporary workaround and now that we have this PEP it won't be necessary anymore 🙌

PR #396, opened 2017-09-07
pep 550: Mention why complete isolation in generators was rejected.

elprans:


gvanrossum:
@1st1 Should I merge this?

1st1:
I don't like how this PR describes the motivation behind chained lookups.  The actual motivation is make it a guarantee that context managers that use the PEP will affect the code that they wrap.  That includes calling `next(generator)`.

With chained lookups, both snippets below will work in the same way:

```python
with Context():
    for i in gen():
        print(i)
```

and

```python
g = gen()
with Context():
    for i in g:
        print(i)
```

Nathaniel's use case relies on that guarantee and the PEP can use it as an example, but the real motivation is general correctness.

PR #410, opened 2017-09-09
Clarify Python versioning in stub file search path

ethanhs:
The previous version was not entirely clear.

cc @ilevkivskyi

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -1834,11 +1834,11 @@ they're updated often.
 Third-party stub packages can use any location for stub storage.  Type
 checkers should search for them using PYTHONPATH.  A default fallback
 directory that is always checked is ``shared/typehints/pythonX.Y/`` (for
-any PythonX.Y version, not just the installed version).  Since there can
-only be one package installed for a given Python version per environment,
-no additional versioning is performed under that directory (just like bare
-directory installs by ``pip`` in site-packages).  Stub file package authors
-might use the following snippet in ``setup.py``::
+some PythonX.Y as determined by the type checker, not just the installed
+version).  Since there can only be one package installed for a given Python
+version per environment, no additional versioning is performed under that
+directory (just like bare directory installs by ``pip`` in site-packages).
+Stub file package authors might use the following snippet in ``setup.py``::
 
   ...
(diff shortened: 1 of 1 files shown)


pep-0484.txt
```diff
@@ -1833,12 +1833,12 @@ they're updated often.
...
 Third-party stub packages can use any location for stub storage.  Type
 checkers should search for them using PYTHONPATH.  A default fallback
-directory that is always checked is ``shared/typehints/python3.5/`` (or
-3.6, etc.).  Since there can only be one package installed for a given
-Python version per environment, no additional versioning is performed
-under that directory (just like bare directory installs by ``pip`` in
-site-packages).  Stub file package authors might use the following
-snippet in ``setup.py``::
+directory that is always checked is ``shared/typehints/pythonX.Y/`` (for
+any PythonX.Y version, not just the installed version).  Since there can
```
gvanrossum: I think it's still not clear. When I run e.g. `mypy --python-version 3.5` which directories will it check? python3.5 only, or python3.3, python3.4, python3.5?

ethanhs: Ah, this is true. Perhaps "... (for some PythonX.Y as determined by the type checker, not just the installed Python version)."?

gvanrossum: So, just pythonX.Y? Then that proposed sentence sounds right.

ethanhs: Hm, as I think about this more, would it be possible to drop this entirely? I don't think stubs for one Python version should live in the install of another. It would be better to have this be replaced by https://github.com/python/typing/issues/84.

ethanhs: I'll make the change you approved, the removal can be discussed in the PEP discussion.

gvanrossum: Well, it's not implemented in mypy, and I doubt it's implemented by other checkers. And it's somewhat ambiguous where "shared/typeshed" is rooted. But if we just want to delete it, I think we should just wait until you have a PEP with a better proposal and leave PEP 484 unchanged.

ethanhs: My thinking exactly. I will add that to the PEP Im working on.

PR #411, opened 2017-09-09
New PEP: Core support for generic types

ilevkivskyi:
This a mini-PEP that proposes to add some core support for generic types.

cc @gvanrossum @JukkaL


pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 03-Sep-2017
+Post-History: 09-Sep-2017
+
+
+Abstract
+========
+
+Initially PEP 484 was designed in such way that would not introduce
```
ethanhs: "... way that __it__ would not introduce..."

ilevkivskyi: Thanks, fixed.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
+PEP: 9999
```
gvanrossum: Let's assign this PEP number 560.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+Title: Core support for generic types
+Author: Ivan Levkivskyi <levkivskyi@gmail.com>
+Status: Draft
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 03-Sep-2017
+Post-History: 09-Sep-2017
```
gvanrossum: Insert `Python-Version: 3.7` before this line.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+
+Abstract
+========
+
+Initially PEP 484 was designed in such way that it would not introduce
+*any* changes to the core CPython interpreter. Now that type hints and
+the ``typing`` module are extensively used by the community, this restriction
+can be removed. It is proposed to add two special methods
```
gvanrossum: I wonder if PEP 526 (variable annotations) and PEP 557 (dataclasses) should also be mentioned as an indication that type hints are becoming mainstream.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+better support of generic types.
+
+
+Rationale
+=========
+
+The restriction to not modify the core CPython interpreter lead to some
+design decisions that became questionable when the ``typing`` module started
+to be widely used. There are three main points of concerns:
+performance of the ``typing`` module, metaclass conflicts, and large amount
```
gvanrossum: ... and the large number of hacks...

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+------------
+
+The ``typing`` module is one of the heaviest and slowest modules in
+the standard library even with all the optimizations made. Mainly this is
+because subscripted generic types (see PEP 484 for definition of terms
+used in this PEP) are class objects (see also [1]_). The three main ways how
+the performance can be improved with the help of the proposed special methods:
+
+- Creation of generic classes is slow since the ``GenericMeta.__new__`` is
+  very slow, we will not need it anymore.
```
gvanrossum: comma -> semicolon (this is a common thing I notice in your writing).

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+- Very long MROs for generic classes will be twice shorter, they are present
```
gvanrossum: ditto

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+- General complexity of the ``typing`` module, the new proposal will not
+  only allow to remove the above mentioned hacks/bugs, but also simplify
+  the implementation, so that it will be easier to maintain.
+
+
+Specification
+=============
+
+Idea of ``__class_getitem__`` is very simple, it is an exact analog of
```
gvanrossum: "The idea of ``__class_getitem__`` is simple: it is ..."

Also, perhaps it's not that simple. The spec doesn't indicate how it's called and doesn't give an example of how one should implement it. (It is a class method?) Maybe you can give a complete example, e.g.
```py
class MyList:
    def __getitem__(self, index):
        "Called e.g. for `a = MyList(...); x = a[index]`"
        <return the indexed item from the instance>
    @classmethod
    def __class_getitem__(cls, subscript):
        "Called e.g. for `t = MyList[subscript]`"
        <return an object representing a subscripted type>
```

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+``__getitem__`` with an exception that it is called on a class that
+defines it, not on its instances, this allows us to avoid
+``GenericMeta.__getitem__`` for things like ``Iterable[int]``. Note that this
+method is used as a fallback, so if a metaclass defines ``__getitem__``,
+then it will have the priority.
```
gvanrossum: it -> that

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+If an object that is not a class object appears in bases of a class
```
gvanrossum: ...appears in *the* bases of...

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+definition, the ``__subclass_base__`` is searched on it. If found,
+it is given an original tuple of bases as an argument. If the result of call
```
gvanrossum: ... it is *called with the* original tuple of bases as an argument. If the result of *the* call ...

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+is not ``None``, then it is substituted instead of this object.
+Otherwise, the base is just removed. This is necessary to avoid
```
gvanrossum: I assume the "Otherwise" clause only applies if `__subclass_base__` is found?

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+inconsistent MRO errors, that are currently prevented by manipulations in
+``GnericMeta.__new__``. After creating the class, original bases are saved
```
gvanrossum: bases -> *the* original bases

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+in ``__orig_bases__`` (now this is also done by the metaclass).
```
gvanrossum: now -> currently

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+These two method names are reserved for exclusive use by the ``typing`` module
```
gvanrossum: Prefix with "NOTE:".

Also I'm not sure this restriction is worth it. We can't enforce it, and I don't (yet) want to disallow other uses of annotations. (Maybe in Python 4.0. :-)

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+and the generic types machinery, and any other use is strongly discouraged.
+The reference implementation (with tests) can be found in [4]_, the proposal
```
gvanrossum: Where in [4] is the behavior implemented that a metaclass with `__getitem__` still has preference? Or does that just happen to fall out of something else?

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+was originally posted and discussed on the ``typing`` tracker, see [5]_.
+
+
+Backwards compatibility and impact on users who don't use ``typing``:
+=====================================================================
+
+This proposal will break code that currently uses the names
```
gvanrossum: This proposal *may* break ...

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+``__class_getitem__`` and ``__subclass_base__``.
+
+This proposal will allow to have almost full backwards compatibility with
+current public generic types API, moreover the ``typing`` module is still
```
gvanrossum: comma -> semicolon

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+was originally posted and discussed on the ``typing`` tracker, see [5]_.
+
+
+Backwards compatibility and impact on users who don't use ``typing``:
+=====================================================================
+
+This proposal will break code that currently uses the names
+``__class_getitem__`` and ``__subclass_base__``.
+
+This proposal will allow to have almost full backwards compatibility with
```
gvanrossum: This proposal will *support almost complete* backwards compatibility with *the* ...

Also, how will compatibility with the latest 3.6 typing.py not be complete?

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+current public generic types API, moreover the ``typing`` module is still
+provisional.
+
+With the reference implementation I measured negligible performance effects
+(under 1%) for regular (non-generic) classes.
```
gvanrossum: Would like to see how you measured this -- in a microbenchmark or for some app? And have you found perf improvements for generics?

pep-9999.rst
JelleZijlstra: Also, typo in "GnericMeta".

pep-9999.rst
ilevkivskyi: Yes, I will clarify this.

pep-9999.rst
ilevkivskyi: I could imagine that without this note, the PEP may be perceived as something like ``__init_subclass__`` that is intended to be used to customize classes.

pep-9999.rst
ilevkivskyi: This is because I added ``->tp_as_mapping->mp_subscript`` to ``type``, but it is overwritten in user defined metclasses if it defines ``__getitem__``.

Actually, after long thinking it seems to me that it is better to move the code to the end of ``PyObject_GetItem`` this will follow the same logic, maybe will be a bit slower, _but_ will have a benefit that it will not auto-generate descriptor ``__getitem__`` on ``type`` (otherwise preventing this would require some special-casing in ``PyType_Ready()``). What do you think?

pep-9999.rst
ilevkivskyi: Yes, I tested it on a micro-benchmark, here is the code:
```python
class A: ...
class B: ...

for i in range(100000):
    class E(A, B):
        cx = 0
        cy = 0
        def __init__(self, x, y):
            self.x = x
            self.y = y
```
The difference is statistically insignificant:
* master: 3.19 +- 0.04 s
* proposed: 3.15 +- 0.04 s

It is hard to measure speed-up for generics in ``typing`` without writing the full new ``typing`` implementation, but here are some basic observations:
* The first ``List[int]`` will be at least 10x faster.
* All next appearances of ``List[int]`` will take the same time, since type cache look-up is 2x faster than the new ``__class_getitem__``.
* Method calls on instances of generic classes may be up to 10% faster.
* Instantiating generic classes probably will be few percent faster, but it is hard to say for sure now.

Also there will be a decrease in memory consumption. It is hard to say how much, but my expectations are that ``typing`` will be around 2x "lighter". Also annotations in user code will save some memory (although again this is not dramatic due to generic type cache).

To summarize, the speed/memory improvements are rather incremental (although quite good), this is why I have two additional motivation points: avoiding metaclass conflicts and the removal of various hacks.

pep-9999.rst
ilevkivskyi: ``__class_getitem__`` is automatically a class method (like ``__init_subclass__``), so that ``@classmethod`` decorator is not needed here. I will add this example.

ilevkivskyi:
@gvanrossum Thanks for review! I think I now implemented all comments.

ilevkivskyi:
@gvanrossum In the meantime I have found a bug in reference counting in my implementation that caused a refleak, it is now fixed.

pep-9999.rst
gvanrossum: I see how it works now. I'm not sure yet how adding this to `PyObject_GetItem` would be better or worse -- maybe one of the currently active core devs who actually know the code can advise on that. What's the reason for not wanting to auto-generate that descriptor?

ilevkivskyi: > What's the reason for not wanting to auto-generate that descriptor?

I think it may confuse people if they type ``dir(type)`` and/or ``dir(<user defined class>)``. Also now that I moved the code to ``PyObject_GetItem`` it looks simpler and more straightforward.

gvanrossum:
Merged! Ready for python-ideas discussion.

PR #416, opened 2017-09-11
Updates for PEP 561

ethanhs:
These changes are in response to @JelleZijlstra's review of PEP 561.

### DIFF
--- a/pep-0561.rst
+++ b/pep-0561.rst
@@ -23,15 +23,15 @@ architecture.
 Rationale
 =========
 
-Currently, package authors are wishing to distribute code that has
-inline type information, but there is no standard method to distribute
+Currently, package authors wish to distribute code that has
+inline type information. However, there is no standard method to distribute
 packages with inline type annotations or syntax that can simultaneously
 be used at runtime and in type checking. Additionally, if one wished to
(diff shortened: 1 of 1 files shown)


pep-0561.rst
```diff
@@ -94,7 +106,9 @@ a normal package. The normal resolution order of checking ``*.pyi`` before
 ``*.py`` will be maintained. The value of the ``typed`` argument to 
 ``setup()`` is ``"stubs"`` for this type of distribution. The author of the
 package is suggested to use ``package_data`` to assure the stub files are
-installed alongside the runtime Python code.
+installed alongside the runtime Python code. In addition, the package should
```
JelleZijlstra: I think this should go in the next section (which indicates that the title of this section is confusing, since it's actually about packages containing both code and stub files: maybe it should be "Packages with Stubs" or something).

ethanhs: Good catch, that was meant for the next section!

ethanhs:
I believe this is ready to merge unless there is other feedback. Thanks for the review @JelleZijlstra !

ethanhs:
Could this be merged please? I'm hoping to submit version 2 to python-ideas for more feedback. Thanks so much!

PR #421, opened 2017-09-13
Mark PEP 212 as Rejected

Mariatta:
- author unreachable
- enumerate() covers the use case

Closes https://github.com/python/peps/issues/207


Mariatta:
Thanks :)

PR #423, opened 2017-09-13
reSTify PEP 308

mozillazg:
refer #4

### DIFF
--- a/pep-0308.txt
+++ b/pep-0308.txt
@@ -5,317 +5,331 @@ Last-Modified: $Date$
 Author: Guido van Rossum, Raymond Hettinger
 Status: Final
 Type: Standards Track
-Content-Type: text/plain
+Content-Type: text/x-rst
 Created: 7-Feb-2003
 Post-History: 7-Feb-2003, 11-Feb-2003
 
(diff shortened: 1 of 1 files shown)


pep-0308.txt
```diff
@@ -5,318 +5,332 @@ Last-Modified: $Date$
...
+    (if <condition>: <expression1> else: <expression2>)
 
-    Note that the enclosing parentheses are not optional.
+Note that the enclosing parentheses are not optional.
 
-    The resulting expression is evaluated like this:
+The resulting expression is evaluated like this:
 
-    - First, <condition> is evaluated.
+- First, <condition> is evaluated.
```
merwok: Brackets don't need to be escaped?

mozillazg: @merwok Yes, `docutils` will do it.

merwok: I didn’t mean HTML-escaped, but reST-escaped, since angle brackets have a syntactic role inside `ref` markup.  But it seems like this usage is ok!

mozillazg:
ping @brettcannon

PR #428, opened 2017-10-06
V3 revision of PEP 561: Distributing Type Information

ethanhs:
I have made some more edits to my PEP. In summary:

* Move to adding a new metadata specifier so that more tools can participate
* Clarify version matching between third party stubs and runtime packages.
* various other fixes for clarity, readability, and removal of repetition

@ncoghlan I would appreciate a review of this new version if you have time. Thanks!


ncoghlan:
Thanks, that makes the distribution metadata proposal much clearer, and more appropriate for this kind of spec.

PR #436, opened 2017-10-23
PEP 539: Fix wording about Py_tss_NEEDS_INIT

ma8ma:
Wording-only change, replacing the word "default value" with "initializer" making more sense.

Related PR: https://github.com/python/cpython/pull/4096


ma8ma:
@ncoghlan Thank you for merging!

PR #456, opened 2017-11-10
Create pep-9999

JCWDFCS:
Rename 'object` as 'instance' in an address


Mariatta:
The proposed text is not a valid PEP, so I'm closing this.
Are you proposing a change to Python? If so, please first discuss it in [python-ideas](https://mail.python.org/mailman/listinfo/python-ideas) mailing list.
Thanks.

PR #457, opened 2017-11-10
pep-9999.rst

JCWDFCS:
Rename 'object` as 'instance' in an address.


Rosuav:
Please stop opening issue spam about this. Admins, do we need to block this user?

brettcannon:
@JCWDFCS this is an official warning that any more abuse across any Python organization repository of needless issues or pull requests will lead to a banning.

PR #459, opened 2017-11-10
Small edits for PEP 560

gvanrossum:
@ilevkivskyi This started as an attempt to fix a few grammatical nits; it expanded in more punctuation fixes, more grammar fixes, removal of `:` from section headings, addition of two section headings, and a clarification that new dunder names are allowed by the language reference.

### DIFF
--- a/pep-0560.rst
+++ b/pep-0560.rst
@@ -25,19 +25,19 @@ generic types.
 Rationale
 =========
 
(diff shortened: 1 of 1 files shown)


gvanrossum:
It's a service I provide for most PEPs as part of the review process, if I am leaning towards acceptance. :-)

PR #460, opened 2017-11-10
Updates to PEP 560

ilevkivskyi:
This is based on discussion on python-ideas list.
Reference implementation will be updated soon.

cc: @gvanrossum

### DIFF
--- a/pep-0560.rst
+++ b/pep-0560.rst
@@ -146,17 +146,25 @@ Otherwise (if the result is ``None``), the base is just removed. This is
 necessary to avoid inconsistent MRO errors, that are currently prevented by
 manipulations in ``GenericMeta.__new__``. After creating the class,
 the original bases are saved in ``__orig_bases__`` (currently this is also
-done by the metaclass). For example, with this change ``typing.List``
-will behave in the following way::
+done by the metaclass). For example::
 
-  from typing import List
+  class GenericAlias:
+      def __init__(self, origin, item):
+          self.origin = origin
+          self.item = item
+      def __mro_entry__(self, bases):
+          return self.origin
 
-  class Tokens(List[int]):
+  class NewList:
+      def __class_getitem__(cls, item):
+          return GenericAlias(cls, item)
+
+  class Tokens(NewList[int]):
       ...
 
-  assert Tokens.__bases__ == (list,)
-  assert Tokens.__orig_bases__ == (List[int],)
-  assert Tokens.__mro__ == (Tokens, list, object)
+  assert Tokens.__bases__ == (NewList,)
+  assert Tokens.__orig_bases__ == (NewList[int],)
+  assert Tokens.__mro__ == (Tokens, NewList, object)
 
 NOTE: These two method names are reserved for use by the ``typing`` module
 and the generic types machinery, and any other use is discouraged.
--- a/pep-0560.rst
+++ b/pep-0560.rst
@@ -146,13 +146,45 @@ Otherwise (if the result is ``None``), the base is just removed. This is
 necessary to avoid inconsistent MRO errors, that are currently prevented by
 manipulations in ``GenericMeta.__new__``. After creating the class,
 the original bases are saved in ``__orig_bases__`` (currently this is also
-done by the metaclass).
+done by the metaclass). For example, with this change ``typing.List``
+will behave in the following way::
 
-NOTE: These two method names are reserved for exclusive use by
-the ``typing`` module and the generic types machinery, and any other use is
-strongly discouraged. The reference implementation (with tests) can be found
-in [4]_, and the proposal was originally posted and discussed on
-the ``typing`` tracker, see [5]_.
+  from typing import List
+
+  class Tokens(List[int]):
+      ...
+
+  assert Tokens.__bases__ == (list,)
+  assert Tokens.__orig_bases__ == (List[int],)
+  assert Tokens.__mro__ == (Tokens, list, object)
+
+NOTE: These two method names are reserved for use by the ``typing`` module
+and the generic types machinery, and any other use is discouraged.
+The reference implementation (with tests) can be found in [4]_, and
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
@gvanrossum OK, I have some sketch of implementation for refactored ``typing`` using new methods. It looks like the speed-up of import is relatively modest, ``importlib.reload(typing)`` is only up to 3x faster. I have some more ideas about how to improve this, we can discuss it later if necessary.

Now I want to discuss "logistic aspects" of PEP 560. The core part of implementation is ready. The update to ``typing`` is very raw (many tests still fail). How do we organize this? I see three options:
* One mega-PR that contains both core components and the ``typing`` refactoring.
* Two PRs: one small with core components, and one large with ``typing`` updates.
* Many smaller PRs for ``typing`` module.

What do you think? I prefer the second option. It may be hard to review the ``typing`` PR, but on the other hand, splitting it into smaller PRs is not easy, the changes touch many "entangled" parts.

gvanrossum:
I agree the second option sounds best (presumably followed by some small
iterations as people try edge cases). A speedup of 3x without even trying
is nothing to sneeze at!

This will cause the separate typing repo to become relevant only for
backports, right? That is, Python 2.7 and 3.4, essentially. Of those, I
expect that only Python 2.7 will remain important, as a lot of people (not
just Dropbox) still have large amounts of legacy code that they are typing
and where they might want to use certain new features such as Protocols
(and Literal types if they become a thing, etc.). Hm, in some cases
typing_extensions may be the place. Oh well.

ilevkivskyi:
@gvanrossum 
> This will cause the separate typing repo to become relevant only for backports, right?

Yes. Btw this will also allow to get rid of several backward compatibility hacks in the 3.7 version.

In the meantime I tried several more ideas, and it looks like we can get another 3-5x speedup for import (so around 10x total) if we tweak how ``__mro_entry__`` works. Currently, there are two options:
* ``__mro_entry__()`` returns non-``None``: we just replace the base with the result.
* ``__mro_entry__()`` returns ``None``: we remove the base.

Sometimes, however it would be more efficient and convenient to be able to replace a base with several bases (instead of having some "intermediary" classes that inherit from these bases). So I propose the following rules instead:
* ``__mro_entry__()`` returns non-tuple: we just replace the base with the result.
* ``__mro_entry__()`` returns a tuple: we unpack the tuple.

This will allow substitutions with multiple bases, and the removal of a base (previously done by ``None``) can be done by returning an empty tuple ``()``.

Note that the proposed rules are also somehow similar to ``issubclass(cls, C)`` vs. ``issubclass(cls, (A, B))``. What do you think about this? (Just to clarify this will not give any significant speed-up for "post-import" performance, but will make ``import typing`` another 3-5x faster). I already pushed the C part, will try the ``typing`` performance and report later with more precise numbers.

Finally, importing ``re`` takes a lot of time. I think we can hide ``typing.Pattern`` and ``typing.Match`` behind module ``__getattr__`` when PEP 562 is accepted.

---------------
Here are some "problems" that appear with the new implementation:
* User classes will not have the fancy ``repr`` as they do now. This seems to be unfixable, because we don't have ``GenericMeta`` anymore (and I don't want ``__class_repr__``). So that user classes will be repr'd as ``<class 'usermod.UserClass'>``, not ``usermod.UserClass``. Note however that all things it ``typing`` can still be nicely formatted (since they are not classes actually) and subscripted user classes have nice repr: ``usermod.UserClass[int, str]``.
* Before ``issubclass(SomeGeneric[int], SomeGeneric)`` always returned ``True``. Now it is impossible to keep this for user classes (although possible for things in ``typing``). I think for consistency we can just raise ``TypeError`` instead of trying to make some complex rules when it works. (This problem is already mentioned in the PEP).

ilevkivskyi:
@gvanrossum OK, it is less than I thought, a bit more than 2x. But anyway, in total we have 7x speed-up for import with the tweaked ``__mro_entry__`` and between 1.1x and 2x for various "post-import" operations (subscription, instantiation, method lookup, etc).

I think these are all low hanging fruits. There may be some other optimizations, but they will probably be rather incremental. (As a side comment, ``typing`` now has around twice less lines of code, while 99.5% of tests pass, some don't as I mentioned in the previous comment.)

gvanrossum:
Great news!  Why not simplify mro_entry and make it always return a tuple?
And rename to mro_entries.

ilevkivskyi:
> Why not simplify mro_entry and make it always return a tuple? And rename to mro_entries.

The point is that in 95% of cases there will be only one entry (for example in all user classes). Zero or two entries are necessary only for special things in ``typing``. So I am not sure here.

I will finish the ``typing`` part and will make a PR to ``python/peps`` repo with the changed specification for ``__mro_entry__`` then it probably can be posted to python-dev.

gvanrossum:
Ironically we'd have a function introduced to support typing whose
signature has a problematic type. OTOH maybe the cost of creating a
singleton tuple is worth saving in those 95% of cases, so I'm not sure
(though there is the cost on the receiving end to check for multiple
possibilities).

gvanrossum:
FWIW it's hard to time the creation of singleton tuples but it seems fast compared to a Python function call:
```
$ python -m timeit -s 'def f(x): return (x)' 'f(0)'
10000000 loops, best of 3: 0.076 usec per loop
$ python -m timeit -s 'def f(x): return (x,)' 'f(0)'
10000000 loops, best of 3: 0.0984 usec per loop
$ 
```
That makes me think the tuple adds ~22 nsec to a call overhead of 76 nsec. But surely the actual call does a lot more than that!

ilevkivskyi:
Yes, the performance difference is minor. I was a bit unsure about practicality of often writing ``(x,)`` instead of just ``x``. But this anyway will be not very often used beyond ``typing``, so I am OK with always returning a tuple. I will update the implementation and will make a PR to ``python/peps`` (probably do both tomorrow).

PR #461, opened 2017-11-11
Update PEP 561 to use package metadata in place of distribution metadata

ethanhs:
Note that the implementations are not quite updated yet, but they will be by the end of the day.

### DIFF
--- a/pep-0561.rst
+++ b/pep-0561.rst
@@ -6,32 +6,33 @@ Type: Standards Track
 Content-Type: text/x-rst
 Created: 09-Sep-2017
 Python-Version: 3.7
-Post-History: 
+Post-History: 10-Sep-2017, 12-Sep-2017, 06-Oct-2017, 26-Oct-2017
 
 
 Abstract
 ========
 
 PEP 484 introduced type hinting to Python, with goals of making typing
-gradual and easy to adopt. Currently, typing information must be distributed 
-manually. This PEP provides a standardized means to package and distribute
-type information and an ordering for type checkers to resolve modules and 
-collect this information for type checking using existing packaging
-architecture.
+gradual and easy to adopt. Currently, typing information must be distributed
+manually. This PEP provides a standardized means to leverage existing tooling
+to package and distribute type information with minimal work and an ordering
+for type checkers to resolve modules and collect this information for type
+checking.
 
 
 Rationale
 =========
 
-Currently, package authors wish to distribute code that has
-inline type information. However, there is no standard method to distribute
-packages with inline type annotations or syntax that can simultaneously
-be used at runtime and in type checking. Additionally, if one wished to
-ship typing information privately the only method would be via setting
-``MYPYPATH`` or the equivalent to manually point to stubs. If the package
-can be released publicly, it can be added to typeshed [1]_. However, this
-does not scale and becomes a burden on the maintainers of typeshed. 
-Additionally, it ties bugfixes to releases of the tool using typeshed.
+Currently, package authors wish to distribute code that has inline type
+information. Additionally, some maintainers would like to distribute stub files
+to keep Python 2 compatibility while using newer annotation syntax. However,
+there is no standard method to distribute packages with inline type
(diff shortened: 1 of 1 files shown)


pep-0561.rst
```diff
@@ -42,40 +43,73 @@ checkers to fail on packages that are highly dynamic (e.g. sqlalchemy
...
 In order to make packaging and distributing type information as simple and
-easy as possible, the distribution of type information, and typed Python code
-is done through existing packaging frameworks. This PEP adds a new item to the
-``*.distinfo/METADATA`` file to contain metadata about a package's support for
-typing. The new item is optional, but must have a name of ``Typed`` and have a
-value of either ``inline`` or ``stubs``, if present.
+easy as possible, packaging and distribution is done through existing
+frameworks.
+
+To enable the safe distribution of typing information, packaging tools MUST
```
ncoghlan: If I've understood your revised proposal correctly, this isn't quite the right framing, as packaging tools don't need to do anything new, packaging *publishers* are being asked to include a marker file inside their package directories that can be accessed with https://docs.python.org/3/library/pkgutil.html#pkgutil.get_data given the location of the package itself.

For a package published with `setuptools`, that will look like the following [package_data](https://packaging.python.org/tutorials/distributing-packages/#package-data) entry:

    package_data={
        'pkgname': ['py.typeinfo'],
    },

(I know I originally advised *against* mentioning `setuptools` APIs in the PEP, but with the revised proposal, I think it will make it clearer that you're proposing to use existing packaging system features, rather than proposing new ones)

pep-0561.rst
```diff
@@ -42,40 +43,73 @@ checkers to fail on packages that are highly dynamic (e.g. sqlalchemy
...
+install a ``.typeinfo`` file per installed package that indicates it supports
```
ncoghlan: Some Windows tools have problems with dotfiles, so `py.typeinfo` would likely be a better filename for this purpose.

You should also be explicit that for namespace package portions, the `py.typeinfo` file would go inside the namespace submodule, not at the namespace package level.

pep-0561.rst
ethanhs: I was originally thinking the maintainer would indicate through the packaging tool (such as a new keyword to `setup()`) that the package supported typing (through a dict mapping package names to the type of information). Then it would be the responsibility of the tooling to install a generated `py.typeinfo` for listed packages.

However, your proposal has quite a few advantages. It requires no change to tooling, and should work on all Python versions back to 2.7, if I am correct. I am slightly leery of putting the onus of making the `py.typeinfo` on the maintainer however, as in the current spec the format is in the METADATA style, which is meant for tooling to both read and write AIUI. Perhaps the contents of the file can be changed to be either plain `inline` or `stubs`. I'm interested in your thoughts on this.

pep-0561.rst
ethanhs: Ah, if the onus is on the maintainer to create the file, then yes, this would be an annoyance. 👍 the name sounds good.

pep-0561.rst
ncoghlan: Tools may be updated to make this easier (such as with a `type_hints=True` parameter), but it's better for the PEP itself to not *rely* on tooling updates (since it complicates the implementation).

As far as the format of the file itself goes, you may even want to encode the setting in the filename itself (`py.inline.typeinfo` vs `py.stubs.typeinfo`), such that the setting can be checked with os.listdir(), rather than having to actually open and read multiple files in multiple directories (which can be very slow, especially on network drives).

ethanhs: Yes, I was considering putting it in the file name as well. I will update the text now. Thank you for the feedback, it was very helpful!

ncoghlan:
@ethanhs I went ahead and merged the update - if it turns out there are more changes you want to make after updating the implementation, just submit a fresh PR.

ethanhs:
@ncoghlan thanks! I think that is it. I will go ahead and post to python-dev I think.

PR #462, opened 2017-11-11
Rename all .txt PEP files to .rst

mozillazg:
Only rename files, others PR will update pep2html.py and pep0.


Mariatta:
Thanks @mozillazg!

Mariatta:
Sorry @mozillazg. After I merged your PR, it erased the history of the PEPs; each pep-*.rst was treated as a brand new file. So I had to revert your changes to retain the history.

methane:
Please wait for reverting.

> it erased the history of the PEPs; each pep-*.rst was treated as a brand
new file.

Do you mean ’git log’ or github’s file history?

git doesn't save history. It calculate history. So it can't remove history.

I think there is some way to give some hint when making history.

>

Mariatta:
@methane, sorry, 'erased' was the wrong word I shouldn't have used it. What happened was that when a file was renamed, GitHub is only showing the new file history since the rename. So after pep-0008.txt was renamed to pep-0008.rst, on GitHub I could no longer see all the revision ever made to PEP 8.
On the command line, `git log pep-0008.rst` also does not show the complete revision history. Only that it was renamed. We have to do `git log --follow pep-0008.rst` to get the complete history.
I think not being able to easily see the history in GitHub is inconvenient and I'm not comfortable with it.
So I reverted the commit for now. (everything is in . txt again).

PR #463, opened 2017-11-11
reSTify PEP 0

mozillazg:
* [x] reSTify PEP 0
* [x] test via `make -j`
* [x] PR to pythondotorg
* [x] test via pythondotorg(`./manage.py generate_pep_pages`)
* [x] wait for pythondotorg update

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 pep-0000.txt
+pep-0000.rst
 pep-????.html
 __pycache__
 *.pyc
--- a/Makefile
+++ b/Makefile
@@ -17,11 +17,11 @@ PYTHON=python3
 
 TARGETS= $(patsubst %.rst,%.html,$(wildcard pep-????.rst)) $(patsubst %.txt,%.html,$(wildcard pep-????.txt))  pep-0000.html
 
-all: pep-0000.txt $(TARGETS)
+all: pep-0000.rst $(TARGETS)
 
 $(TARGETS): pep2html.py
 
-pep-0000.txt: $(wildcard pep-????.txt) $(wildcard pep-????.rst) $(wildcard pep0/*.py)
+pep-0000.rst: $(wildcard pep-????.txt) $(wildcard pep-????.rst) $(wildcard pep0/*.py)
 	$(PYTHON) genpepindex.py .
 
 rss:
@@ -32,6 +32,7 @@ install:
 
 clean:
 	-rm pep-0000.txt
+	-rm pep-0000.rst
 	-rm *.html
 
 update:
--- a/genpepindex.py
+++ b/genpepindex.py
@@ -36,7 +36,7 @@ def main(argv):
     peps = []
     if os.path.isdir(path):
         for file_path in os.listdir(path):
-            if file_path == 'pep-0000.txt':
+            if file_path.startswith('pep-0000.'):
                 continue
             abs_file_path = os.path.join(path, file_path)
             if not os.path.isfile(abs_file_path):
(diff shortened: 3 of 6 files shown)


Mariatta:
Thanks for the PR!
I tried checking out your PR locally and build the PEP, but getting an error while rendering PEP 0 in pythondotorg.

My steps are:
- checked out your PR
- delete my copy of pep-0000.txt
- `make -j`
- in my local pythondotorg directory: `./manage.py generate_pep_pages`.

Problem is at https://github.com/python/pythondotorg/blob/master/peps/converters.py#L40:

```
body_children = list(soup.body.children)
```

The new PEP 0 does not have the `<body>` tag in the HTML output, so the above line errored out.

We'll need to coordinate and update pythondotorg before merging this PR.

mozillazg:
@Mariatta Thanks for your review. I'll submit a PR to pythondotorg for fixing this.

mozillazg:
@Mariatta PR to pythondotorg is here: https://github.com/python/pythondotorg/pull/1192

berkerpeksag:
FYI, I've just merged and deployed https://github.com/python/pythondotorg/pull/1192.

berkerpeksag:
Is it possible to add a "Type" column to the table in "Index by Category" section?

screenshot from 2018-06-07 20 58 56

mozillazg:
@berkerpeksag Done.

image

mozillazg:
@Mariatta This PR is ready for review.

gvanrossum:
@Mariatta Is anything holding this up?

.gitignore
```diff
@@ -1,4 +1,4 @@
-pep-0000.txt
+pep-0000.rst
```
ncoghlan: Old checkouts may still have the text file around, so best to keep both files here for now.

Makefile
```diff
@@ -31,7 +31,7 @@ install:
 	echo "Installing is not necessary anymore. It will be done in post-commit."
 
 clean:
-	-rm pep-0000.txt
+	-rm pep-0000.rst
```
ncoghlan: Old checkouts may still have the text file around, so best to remove both files here for now.

genpepindex.py
```diff
@@ -36,7 +36,7 @@ def main(argv):
     peps = []
     if os.path.isdir(path):
         for file_path in os.listdir(path):
-            if file_path == 'pep-0000.txt':
+            if file_path == 'pep-0000.rst':
```
ncoghlan: Best to switch this to `file_path.startswith('pep-0000.')` so old text files also get ignored.

pep0/output.py
```diff
@@ -26,15 +26,58 @@
...
-        underline_headers[key] = constants.text_type(len(value) * '-')
-    print(constants.column_format % underline_headers, file=output)
+        markups[key] = constants.text_type(len(value) * '=')
+    markup_line = (column_format % markups).strip()
+    print(markup_line, file=output)
+    print((column_format % column_headers).strip(), file=output)
+    print(markup_line, file=output)
+
+
+def write_table(columns, output, include_header=False):
```
ncoghlan: The parameter here is a sequence of `rows`, rather than a sequence of columns

pep0/output.py
```diff
@@ -26,15 +26,58 @@
...
+    if not columns:
+        return
+
+    lengths = [[] for x in columns[0]]
+    for column in columns:
```
ncoghlan: `for row in rows:` (I won't make any further comments that relate to the name change)

Makefile
pradyunsg: `rm non-existent-file` would result in an exit code of 1, which would make `make` abort the build. Adding a `-` before the command would make `make` ignore these errors:

```
clean:
        -rm -f *.o
```

(example from https://www.gnu.org/software/make/manual/html_node/Errors.html)

berkerpeksag:
FWIW, I find Nick's PR a bit simpler than this one. The downside is that it requires Python 3.6 and I don't think we have Python 3.6 installed on python.org servers (but it might not that hard to create a separate venv for the [peps repository](https://github.com/python/psf-salt/blob/66de808f988466be2a207102c4aea1b00ca2d7dc/salt/pydotorg/init.sls))

mozillazg:
@berkerpeksag I also think Nick's PR is better.

PR #468, opened 2017-11-14
PEP 561: Add note about module only distributions

ethanhs:
As requested by @gvanrossum.


ethanhs:
Your edits look good to me @gvanrossum, thank you!

PR #476, opened 2017-11-17
Add suggestion about multiple comment syntax

ethanhs:
cc @gvanrossum You suggested I make a PR about this a bit over a week ago.


JelleZijlstra:
Not sure if either of you saw, but there was a recent thread on Python-ideas about this ("A proliferation of (un-)Pythonically programmatic pragmas"). There is some talk about coming up with a general standard for this sort of comment.

gvanrossum:
I did see that, but I doubt that much will come of it (this seems to be how I feel about many python-ideas threads these days). In any case the wording I added just describe the status quo, so I think it's fine to see it as a clarification, not as a new requirement.

Issue #388, opened 2017-09-03
Description-Content-Type does not appear to be in accepted PEPs


xoviat:
This field, used in PKG-INFO is now in production use despite not being defined in a PEP. Someone should take this up.

di:
@ncoghlan suggested that this field does not need to be defined in a PEP -- see https://mail.python.org/pipermail/distutils-sig/2016-June/029078.html

xoviat:
Thanks.

xoviat:
@agronholm

ncoghlan:
To be completely clear, there *is* open work to be done in this area:
* https://github.com/pypa/pypa.io/issues/11
* https://github.com/pypa/pypa.io/issues/11

It's just meta-process work to make the process change clearer (including adding a cross-reference to the new home from PEP 345), rather than defining a new iteration of the metadata using the current process.

agronholm:
So I take it that https://packaging.python.org/specifications/ is the canonical source of valid field names? Is somebody going to update the JSON schema for packaging metadata? Wheel's test suite started failing the moment this new field was introduced by a new setuptools release. How can we prevent this from happening in the future?

agronholm:
Specifically this file here: https://github.com/python/peps/blob/master/pep-0426/pydist-schema.json
@ncoghlan this seems to be your file, so would you mind adding the new field there? I will then copy it over to the wheel codebase.

ncoghlan:
PEP 426 isn't released yet - the metadata.json file defined by `wheel` is wholly under @dholth's control.

agronholm:
Speaking of which – what's currently blocking that PEP?

ncoghlan:
A credible belief on my part that it actually solves a real world problem in a way that will prompt people to invest their time and energy in implementing (I should really put it back to Deferred status on those grounds).

Specifically,  I think we're likely to be better off just adding a command to `pip` to emit a JSON-ified version of the current metadata: https://github.com/pypa/pip/issues/4691

agronholm:
If that is the case, should wheel be writing that data to .dist-info in the first place, as it does now?

ncoghlan:
Yes, projects are allowed to write arbitrary files to dist-info (that's by design). wheel calls it metadata.json because we've promised never to use that for any of the formal packaging specifications.

tseaver:
@ncoghlan Without a PEP defining the new fields, how do we figure out what metadata version supports them?

di:
[...]

PR #490, opened 2017-12-05
PEP 566: Metadata for Python Software Packages 1.3

di:
This adds a packaging metadata 1.3 PEP (as discussed in #388).

The goal here is not to provide a full specification for all fields, but to:
* Motivate and describe the addition of the new fields or changes to existing fields;
* Point to the [Core Metadata Specifications](https://packaging.python.org/specifications/core-metadata/) reference document as the canonical source for the specification.

(CC: @ncoghlan)

### DIFF
--- a/pep-0566.txt
+++ b/pep-0566.txt
@@ -1,0 +1,159 @@
+PEP: 566
+Title: Metadata for Python Software Packages 1.3
+Version: $Revision$
+Last-Modified: $Date$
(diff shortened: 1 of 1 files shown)


ncoghlan:
@di Thanks! This will show up at https://www.python.org/dev/peps/pep-0566/ once the relevant cache headers expire in the CDN.

PR #498, opened 2017-12-10
Clean up / clarify grammar in PEP 540.

larryhastings:


vstinner:
Merged, thanks. I pushed a following commit to truncate to 72 columns.

PR #508, opened 2017-12-13
pep-567: Add C API section.

1st1:
Add C API section.  cc @gvanrossum @Rosuav

### DIFF
--- a/pep-0567.rst
+++ b/pep-0567.rst
@@ -263,10 +263,37 @@ as follows::
             ...
 
 
-CPython C API
--------------
+C API
+-----
 
-TBD
+1. ``PyContextVar * PyContextVar_New(char *name)``: create a
+   ``ContextVar`` object.
+
+2. ``PyObject * PyContextVar_Get(PyContextVar *)``:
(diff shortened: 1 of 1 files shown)


pep-0567.rst
```diff
@@ -263,10 +263,37 @@ as follows::
...
+1. ``PyContextVar * PyContextVar_New(char *name)``: create a
+   ``ContextVar`` object.
+
+2. ``PyObject * PyContextVar_Get(PyContextVar *)``:
+   return the value of the variable in the current context.
+
+3. ``PyContextToken * PyContextVar_Set(PyContextVar *, PyObject *)``:
+   set the value of the variable in the current context.
+
+4. ``PyContextVar_Reset(PyContextToken *)``:
```
gvanrossum: I think it's surprising that this doesn't take a ContextVar argument. Even though it's technically redundant, I'd add it -- it makes the API more regular and gives the caller the confidence that it's resetting the correct variable.

1st1: Makes sense.

PR #516, opened 2017-12-16
Fix PEP summary/email line indent in pep-0000.html

jsalter:
The PEP summary/email lines that aren't the first line in their group
have an extra leading space because the previous line is printed with a
trailing space. The special lines read from pep-0000.txt should be
printed with no extra ending to be consistent with what the python2
version did with a print statement that ends with a comma.


ncoghlan:
Good catch! I'd never really looked at the raw HTML output, and the rendered HTML looks the same regardless.

PR #522, opened 2017-12-25
Doc is served by nginx, not apache.

JulienPalard:
Not much.


JulienPalard:
@Mariatta It can be seen here I should have linked this: https://github.com/python/psf-salt/blob/master/salt/docs/init.sls

PR #528, opened 2017-12-31
Fix typo: Set -> set

martinbroadhurst:


the-knights-who-say-ni:

Rosuav:
This is a trivial change (I can't see any reason for this to be a specialized class here), so I'm going to override the bot. I recommend checking the CLA requirements for next time, though, as it'll smooth things out. Thanks for fixing this!

bitdancer:
This was not a typo, and so I don't think this should have been merged.  At the time that pep was written, Set was the correct spelling.  'set' did not get added to the language until python 2.4, the same version in which this PEP landed.  However, now that it is merged it is not also not worth reverting, since both peps did both land in 2.4, and the distinction is now of only historical interest.  (But don't forget that most PEPs are historical documents :)

Rosuav:
Ah, I'm sorry, I was looking through for a reason for that and missed seeing the version. My bad! Sorry.

bitdancer:
Well, those of us who have been doing this long enough that we *remember* when set was Set have a bit of an advantage in spotting such a thing :)

Rosuav:
Haha, yeah. I was thinking "this proves how big a noob I am", given that I didn't seriously start using Python until about 2.6. And then I talk to JavaScript people and realise that I've been using Python for longer than Node.js has been around. Noob? Old hand? It's all relative...

PR #529, opened 2018-01-02
Clarify the prohibition on locals() and globals().

gvanrossum:
It applies to the implementation, not to end users passing these in to
f-string -- the latter is fine (e.g. print(f"Locals are {locals()}")).

CC: @mariatta


Mariatta:
Thanks for the clarification, @gvanrossum :)

PR #536, opened 2018-01-16
pep-567: V3.

1st1:
/cc @gvanrossum

### DIFF
--- a/pep-0567.rst
+++ b/pep-0567.rst
@@ -98,7 +98,7 @@ The ``ContextVar.get()`` method looks for the variable in the current
 It is not possible to get a direct reference to the current ``Context``
 object, but it is possible to obtain a shallow copy of it using the
 ``contextvars.copy_context()`` function.  This ensures that the
-caller of ``Context.run()`` is the sole owner of its ``Context``
+*caller* of ``Context.run()`` is the sole owner of its ``Context``
 object.
 
 
@@ -162,21 +162,23 @@ current context to the value it had before the ``set()`` operation
 that created the ``token`` (or to remove the variable if it was
 not set)::
 
-    assert var.get(None) is None
+    # Assume: var.get(None) is None
 
+    # Set 'var' to 1:
     token = var.set(1)
     try:
-        ...
+        # var.get() == 1
     finally:
         var.reset(token)
 
-    assert var.get(None) is None
+    # After reset: var.get(None) is None,
+    # i.e. 'var' was removed from the current context.
 
 ``ContextVar.reset()`` method is idempotent and can be called
 multiple times on the same Token object: second and later calls
 will be no-ops.  The method raises a ``ValueError`` if:
 
-* called with a token object created by another variable; or
+* it is called with a token object created by another variable; or
 
 * the current ``Context`` object does not match the one where
   the token object was created.
@@ -222,22 +224,28 @@ be contained in the ``ctx`` context::
     var = ContextVar('var')
     var.set('spam')
 
-    def function():
-        assert var.get() == 'spam'
-        assert ctx[var] == 'spam'
+    def main():
+        # 'var' was set to 'spam' before
+        # calling 'copy_context()' and 'ctx.run(main)', so:
+        # var.get() == ctx[var] == 'spam'
 
         var.set('ham')
-        assert var.get() == 'ham'
-        assert ctx[var] == 'ham'
+
+        # Now, after setting 'var' to 'ham':
+        # var.get() == ctx[var] == 'ham'
 
     ctx = copy_context()
 
-    # Any changes that 'function' makes to 'var' will stay
-    # isolated in the 'ctx'.
-    ctx.run(function)
+    # Any changes that the 'main' function makes to 'var'
+    # will be contained in 'ctx'.
+    ctx.run(main)
 
-    assert var.get() == 'spam'
-    assert ctx[var] == 'ham'
+    # The 'main()' function was run in the 'ctx' context,
+    # so changes to 'var' are contained in it:
+    # ctx[var] == 'ham'
+
+    # However, outside of 'ctx', 'var' is still set to 'spam':
+    # var.get() == 'spam'
 
 ``Context.run()`` raises a ``RuntimeError`` when called on the same
 context object from more than one OS thread, or when called
@@ -328,8 +336,17 @@ using a copy-on-write approach and built-in dict type::
         def __init__(self):
             self._mapping = dict()
 
-        def get(self, key):
+        def __getitem__(self, key):
             return self._mapping[key]
+
+        def __contains__(self, key):
+            return key in self._mapping
+
+        def __len__(self):
+            return len(self._mapping)
+
+        def __iter__(self):
+            return iter(self._mapping)
 
         def set(self, key, value):
             copy = _ContextData()
@@ -365,9 +382,6 @@ Every OS thread has a reference to the current ``Context`` object::
                     f'cannot enter context: {self} is already entered')
 
             ts: PyThreadState = PyThreadState_Get()
-            if ts.context is None:
-                ts.context = Context()
-
             self._prev_context = ts.context
             try:
                 ts.context = self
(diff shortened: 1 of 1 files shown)


gvanrossum:
If you have more changes please don't squash commits, I am already reviewing.

1st1:
No that was one and only force-push. I only added this:

```diff
+``Context.run()`` raises a ``RuntimeError`` when called on the same
+context object from more than one OS thread, or when called
+recursively.
```

pep-0567.rst
```diff
@@ -153,14 +174,19 @@ that created the ``token``::
 
 ``ContextVar.reset()`` method is idempotent and can be called
 multiple times on the same Token object: second and later calls
-will be no-ops.
+will be no-ops.  The method raises a ``ValueError`` if:
+
+* called with a token object created by another variable; or
```
gvanrossum: "_it is_ called with ..."

pep-0567.rst
```diff
@@ -203,9 +224,11 @@ be contained in the ``ctx`` context::
 
     def function():
         assert var.get() == 'spam'
+        assert ctx[var] == 'spam'
```
gvanrossum: I'd much prefer it if instead of `assert` statements you'd use comments e.g. `# Now ctx[var] == 'spam'`. (The first `assert` in the PEP is a pre-condition, you can phrase that as `# Assume <blah>`.)

pep-0567.rst
```diff
@@ -348,48 +343,59 @@ using ``dict.copy()``::
...
+                self._prev_context = None
+
+        def copy(self):
+            new = Context()
+            new._data = self._data
+            return new
 
         # Mapping API methods are implemented by delegating
-        # `get()` and other Mapping calls to `self._data`.
+        # `get()` and other Mapping methods to `self._data`.
```
gvanrossum: I'd like to see these written out -- tedious and boring, but it could help getting everyone on the same page, and would avoid later surprises.

pep-0567.rst
```diff
@@ -348,48 +343,59 @@ using ``dict.copy()``::
...
 
         def run(self, callable, *args, **kwargs):
-            ts : PyThreadState = PyThreadState_Get()
-            saved_data : _ContextData = ts.context_data
+            if self._prev_context is not None:
+                raise RuntimeError(
+                    f'cannot enter context: {self} is already entered')
 
+            ts: PyThreadState = PyThreadState_Get()
+            if ts.context is None:
```
gvanrossum: I think the lazy initialization of `ts.context` (also below in `copy_context()`) is a distraction. `context` is not declared `Optional` in the pseudo-declaration of `class PyThreadState` above; we can just assume (or state in the text) that a new `Context` is put in there when a new thread (state) is created.

pep-0567.rst
```diff
@@ -419,35 +424,56 @@ points to a ``_ContextData`` object::
...
+implementation of the cache in great detail.
+
+The ``Token`` class is implemented as follows::
 
     class Token:
 
         MISSING = object()
 
-        def __init__(self, var, old_value):
+        def __init__(self, ctx, var, old_value):
```
gvanrossum: Maybe spell out `context`? (But not `variable`.) My brain keeps stumbling over `ctx` when I speak the code in my head. :-)

pep-0567.rst
```diff
@@ -461,37 +487,64 @@ points to a ``_ContextData`` object::
...
+   reset the value of the context variable.
+
+5. ``PyContext * PyContext_New()``: create a new empty context.
+
+6. ``PyContext * PyContext_Copy()``: get a copy of the current context.
+
+7. ``int PyContext_Enter(PyContext *)`` and
+   ``int PyContext_Exit(PyContext *)`` allow to set and restore
+   the context for the current OS thread.  It is required to always
+   restore the previous context::
```
gvanrossum: This requirement has no teeth -- you cannot enforce it, can you?
Also, someone on the mailing list (Paul?) asked why we need separate Enter and Exit functions -- is this answered anywhere?

pep-0567.rst
```diff
@@ -551,6 +604,53 @@ code unmodified, but will automatically enable support for
 asynchronous code.
 
 
+Examples
+========
+
+Converting code that uses threading.local()
+-------------------------------------------
+
+A typical code that uses ``threading.local()`` usually looks like
```
gvanrossum: code -> code fragment (code is not countable so you can't say "a code").

pep-0567.rst
```diff
@@ -551,6 +604,53 @@ code unmodified, but will automatically enable support for
...
+        # Subclass threading.local to specify a default value.
+        value = 'spam'
+
+    mylocal = mylocal()
+
+    # To set a new value:
+    mylocal.value = 'new value'
+
+    # To read the current value:
+    mylocal.value
```
gvanrossum: Assign it to a variable or use it in a larger expression.

pep-0567.rst
```diff
@@ -551,6 +604,53 @@ code unmodified, but will automatically enable support for
...
+-------------------------------------------
+
+A typical code that uses ``threading.local()`` usually looks like
+the following snippet::
+
+    class mylocal(threading.local):
+        # Subclass threading.local to specify a default value.
+        value = 'spam'
+
+    mylocal = mylocal()
```
gvanrossum: This has a variable of the same name as its type. Please be more creative. :-)

(In general, the example would probably benefit from being specific to some application rather than being generic like this. You don't have to add more code to make the example complete, just pick variable names that suggest a use case -- perhaps "precision" would do?)

pep-0567.rst
```diff
@@ -551,6 +604,53 @@ code unmodified, but will automatically enable support for
...
+-------------------------------------
+
+It is possible to run code in a separate OS thread using a copy
+of the current thread context::
+
+    executor = ThreadPoolExecutor()
+    current_context = contextvars.copy_context()
+
+    executor.submit(
+        lambda: current_context.run(some_function))
```
gvanrossum: Since `submit()` takes both a callable and optional args, this can be simplified to
```py
executor.submit(current_context.run, some_function)
```

pep-0567.rst
```diff
@@ -76,20 +76,30 @@ different asynchronous tasks that exist and execute concurrently
...
+* use ``ContextVar.set()`` to set a new value for the context
+  variable.
+
+The ``ContextVar.get()`` method looks for the variable in the current
+``Context`` object using ``self`` as a key.
+
+It is not possible to get a direct reference to the current ``Context``
+object, but it is possible to obtain a shallow copy of it using the
+``contextvars.copy_context()`` function.  This ensures that the
+caller of ``Context.run()`` is the sole owner of its ``Context``
```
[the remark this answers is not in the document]
1st1: I've emphasized the "caller".

I've tried to mention asyncio.Task here too, but failed to find a way to do it in a concise and a non-distracting way.

pep-0567.rst
1st1: fixed

pep-0567.rst
1st1: Fixed all asserts.

pep-0567.rst
1st1: Done: I've implemented all abstract methods in `collections.abc.Mapping`. `Context` actually subclasses it, so the implementation was indeed incomplete.

pep-0567.rst
1st1: Changed it to 'context'.

pep-0567.rst
1st1: fixed

pep-0567.rst
1st1: fixed

pep-0567.rst
1st1: fixed

pep-0567.rst
1st1: fixed

pep-0567.rst
1st1: > This requirement has no teeth -- you cannot enforce it, can you?

No, it's not enforceable, but I think we should document it that way.  Otherwise `PyContext_Exit` will get out of order, which will in turn result in errors.

> Also, someone on the mailing list (Paul?) asked why we need separate Enter and Exit functions -- is this answered anywhere?

This is just how we usually do this in C.  We could design an API that would accept a pointer to a function and run it within the context, but such API would be painful to use.  Hence enter/exit methods.

1st1:
@gvanrossum Thanks for the quick review!  I've pushed a commit that should address your comments.

1st1:
Thanks!  BTW, are you OK with changing `ContextVar.reset()` to raise an error on 2nd call with the same token?

gvanrossum:
> are you OK with changing `ContextVar.reset()` to raise an error on 2nd call with the same token?

Yes.

PR #540, opened 2018-01-17
PEP 567: avoid _NO_DEFAULT, use [default] syntax

vstinner:
Replace ContextVar(name, *, default=_NO_DEFAULT) with ContextVar(name, *, [default]).

### DIFF
--- a/pep-0567.rst
+++ b/pep-0567.rst
@@ -124,7 +124,7 @@ contextvars.ContextVar
 ----------------------
 
 The ``ContextVar`` class has the following constructor signature:
-``ContextVar(name, *, default=_NO_DEFAULT)``.  The ``name`` parameter
+``ContextVar(name, *, [default])``.  The ``name`` parameter
 is used for introspection and debug purposes, and is exposed
 as a read-only ``ContextVar.name`` attribute.  The ``default``
 parameter is optional.  Example::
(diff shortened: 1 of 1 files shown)


vstinner:
cc @1st1 @gvanrossum

1st1:
I like this change, Guido, what do you think?

vstinner:
Other examples of [optional_parameter] syntax:

* https://docs.python.org/dev/library/os.html#os.sendfile
* https://docs.python.org/dev/library/functions.html#breakpoint (bytearray, bytes)
* https://docs.python.org/dev/library/functions.html#format
* etc.

I'm not sure if we already have examples of keyword-only optional parameters written with [arg] syntax in the doc.

gvanrossum:
I don't like the `ContextVar(name, *, [default])` case -- it suggests you can write `ContextVar('foo', 42)`. Let's leave well enough alone. This is a PEP, not end user docs.

vstinner:
Aha, I found another example of optional keyword-only parameters:

```
max(iterable, *[, key, default])
```

https://docs.python.org/dev/library/functions.html#max

Builtin help():

```
vstinner@apu$ ./python
Python 3.7.0a4+ (heads/master-dirty:44a70e9336, Jan 17 2018, 15:40:30) 
>>> help(max)
Help on built-in function max in module builtins:

max(...)
    max(iterable, *[, default=obj, key=func]) -> value
    max(arg1, arg2, *args, *[, key=func]) -> value
    
    With a single iterable argument, return its biggest item. The
    default keyword-only argument specifies an object to return if
    the provided iterable is empty.
    With two or more arguments, return the largest argument.
```

gvanrossum:
I still don't like it for the PEP. Please let it go.

vstinner:
> I still don't like it for the PEP. Please let it go.

Don't worry, it just was a proposal. I'm fine to keep the current syntax :-)

I just wanted my PR to be complete ;-)

PR #547, opened 2018-01-18
PEP 566: More updates

di:


di:
@ncoghlan Could you merge?

ncoghlan:
Done! (Not sure if we might want a short section explaining the version number jump, but that isn't necessary for PEP approval)

PR #555, opened 2018-01-21
pep-567: Tweak C API

1st1:
Changes:

* Rename `PyContext * PyContext_Copy()` to `PyContext * PyContext_CopyCurrent()`;

* Add a new `PyContext * PyContext_Copy(PyContext *)` function.

This brings parity between C & Python APIs. 

@gvanrossum: this is the last change I can currently think of.


slavikf:
75

PR #563, opened 2018-01-29
Fix rendering of editor's first initial

mikesep:
If R. David Murray's first initial is not escaped, it is interpreted as
the alphabetic ordinal in an enumerated list and rendered as
"18. David Murray".


Rosuav:
Maybe that's his street address.

PR #567, opened 2018-02-01
bpo-32744: Fix double colons typos in code

delimitry:
https://bugs.python.org/issue32744

### DIFF
--- a/pep-0342.txt
+++ b/pep-0342.txt
@@ -420,7 +420,7 @@ Examples
                 destination.send(page)
 
     @consumer
-    def jpeg_writer(dirname)::
+    def jpeg_writer(dirname):
         fileno = 1
         while True:
             filename = os.path.join(dirname,"page%04d.jpg" % fileno)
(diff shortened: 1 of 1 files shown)


Rosuav:
My suspicion is that these aren't typos - they look like reStructuredText formatting. But these ones appear to be superfluous. (Maybe these blocks were copied and pasted from stand-alone preformatted blocks?) Merging.

delimitry:
Thanks @Rosuav, it's fixed now!
https://www.python.org/dev/peps/pep-0342/

berkerpeksag:
> (Maybe these blocks were copied and pasted from stand-alone preformatted blocks?)

https://github.com/python/peps/commit/329ed7e935e420f4433217ac067e969d6d90bdc7 added these superfluous colons (this is why I was against automatic reST conversion at the time; see https://github.com/python/peps/commit/d1d344880b045db8784c7beec081614db2efeaa5 for a similar bug)

PR #568, opened 2018-02-05
Change to -stubs over _stubs

ethanhs:
Also note that a py.typed file is not needed for stub only packages.

cc @gvanrossum I believe you wanted these changes based on last Friday's discussion?


gvanrossum:
Yup!

ethanhs:
Great! I will update the implementation later today.

PR #584, opened 2018-02-28
Readme: Change pep-0999.txt to pep-0999.rst

Mariatta:
`.rst` file extension is now supported. I think new PEPs should be in `.rst`.


Rosuav:
Agreed - this matches the recommendation in PEP 12.

Mariatta:
Thanks!

PR #589, opened 2018-03-10
Cleaning up text - role of the PSF

MarkMangoba:
Adding text for Packaging Workgroup role.

### DIFF
--- a/pep-0541.txt
+++ b/pep-0541.txt
@@ -262,7 +262,7 @@ The Package Index maintainers can escalate issues covered by this
 document for resolution by the Packaging Workgroup if the matter is not clear
 enough.  Some decisions *require* additional judgement by the Board,
 especially in cases of Code of Conduct violations or legal claims.
-Recommendations made by the Board are are sent to the Packaging Workgroup [9]_ for review.
+Recommendations made by the Board are sent to the Packaging Workgroup [9]_ for review.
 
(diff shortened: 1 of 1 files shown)


ncoghlan:
As per the packaging-wg discussion, I'd also suggest we remove footnote 7 (the link to the source forge tracker) and the final paragraph in the Intellectual Property Policy section that references it.

pep-0541.txt
```diff
@@ -259,12 +259,12 @@ The Python Software Foundation [8]_ is the non-profit legal entity that
 provides the Package Index as a community service.
 
 The Package Index maintainers can escalate issues covered by this
-document for resolution by the PSF Board if the matter is not clear
+document for resolution by the Packaging Workgroup if the matter is not clear
 enough.  Some decisions *require* additional judgement by the Board,
 especially in cases of Code of Conduct violations or legal claims.
-Decisions made by the Board are published as Resolutions [9]_.
+Recommendations made by the Board are are sent to the Packaging Workgroup for review.
```
ncoghlan: Rather than deleting the reference to footnote 9 entirely, I'd suggest changing the footnote link to be https://wiki.python.org/psf/PackagingWG, and move the reference up to the first mention of the working group.

ncoghlan:
I'll fix the source forge link in a separate PR.

PR #601, opened 2018-03-26
Module State Access from C Extension Methods

Traceur759:


pep-0573.rst
```diff
@@ -0,0 +1,502 @@
...
+
+.. note::
+
+    The defining class is not ``type(self)``, since ``type(self)`` might
+    be a subclass of ``Foo``.
+
+Implicitly, the last three of those rely on name-based lookup via the function's ``__globals__``:
+either the ``Foo`` attribute to access the defining class and Python function object, or ``__name__`` to find the module object in ``sys.modules``.
+In Python code, this is feasible, as ``__globals__`` is set appropriately when the function definition is executed, and
+even if the namespace has been manipulated to return a different object, at worst an exception will be raised.
```
ncoghlan: Mention `__class__` here, but point out that the same basic analysis applies: just as reference by name relies on `__globals__` being set on the function, the cell reference relies on `__closure__` being set.

pep-0573.rst
```diff
@@ -0,0 +1,502 @@
...
+Two possible solutions have been proposed to this problem:
+
+    * Look up the class through walking the MRO.
+      This is potentially expensive, but will be useful if performance is not
+      a problem (such as when raising a module-level exception).
+    * Storing a pointer to the defining class of each slot in a separate table,
+      ``__typeslots__`` [#typeslots-mail]_.  This is technically feasible and fast,
+      but quite invasive.
+
+Due to the invasiveness of the latter approach, this PEP proposes adding a MRO walking helper for use in slot method implementations, deferring the more complex alternative as a potential future optimisation.
```
ncoghlan: Modules affected by this concern also have the option of using thread local state or context variables, or else defining their own reload-friendly lookup caching scheme.

pep-0573.rst
Traceur759: Mentioning @encukou

encukou: I don't think it's necessary . The Background gives the normal way to get the defining class in Python, mainly to more illustratively explain what the defining class *is*. Mentioning the class closure alternative way isn't of much use, IMO.
Also, isn't the closure a CPython implementation detail? We're explaining "This is how you do it in Python, now we want to do the same safely in C". Only the latter is CPython-specific.

ncoghlan: Zero-arg super() and the `__class__` closure reference are part of the language spec: https://docs.python.org/3/reference/datamodel.html?highlight=__class__#creating-the-class-object

It's passing `__classcell__` to `type.__new__` via the class execution namespace that's a CPython implementation detail. (It could be interesting to try to replace that with a context variable so it doesn't need to be explicitly preserved by metaclass authors)

ncoghlan: Anyway, I agree it isn't essential background for this PEP, but as someone that *does* know about `__class__`, an `# or __class__` comment after the name references would be reassuring :)

encukou: Ha! Until a minute ago I was in that group (). Thanks for the link! Will fix.

PR #605, opened 2018-04-02
PEP 575: Unifying function/method classes

jdemeyer:


Rosuav:
The next PEP number is 573 - go ahead and claim that number.

jdemeyer:
PEP 573 exists (Module State Access from C Extension Methods), so I guess you mean 574.

Rosuav:
That's what I get for pulling changes and not noticing that git popped an error into the console. 575 is correct. Thanks @jdemeyer and sorry for the misinformation!

jdemeyer:
> go ahead and claim that number.

I'm not entirely sure what you meant with this, but I created `pep-0575.rst` in the repo and updated the header.

Rosuav:
Yep! That's exactly what I meant. Your PEP should be live shortly at https://www.python.org/dev/peps/pep-0575/ and you can post to python-ideas with that URL.

PR #611, opened 2018-04-07
Mention how to calculate the ABI flag for CPython

brettcannon:
It's not obvious where to find the suffix for the ABI tag if you don't happen to just know.

### DIFF
--- a/pep-0425.txt
+++ b/pep-0425.txt
@@ -109,7 +109,8 @@ ABI Tag
 The ABI tag indicates which Python ABI is required by any included
 extension modules.  For implementation-specific ABIs, the implementation
 is abbreviated in the same way as the Python Tag, e.g. `cp33d` would be
-the CPython 3.3 ABI with debugging.
+the CPython 3.3 ABI with debugging (calculated by concatenating the Python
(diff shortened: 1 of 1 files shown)


pep-0425.txt
```diff
@@ -109,7 +109,8 @@ ABI Tag
 The ABI tag indicates which Python ABI is required by any included
 extension modules.  For implementation-specific ABIs, the implementation
 is abbreviated in the same way as the Python Tag, e.g. `cp33d` would be
-the CPython 3.3 ABI with debugging.
+the CPython 3.3 ABI with debugging (calculated by concatenating the Python
+tag with `sys.abiflags()`).
```
ncoghlan: *gasp* You mean I don't have to go digging in `sysconfig` for that? 

Also, this is just a string, so the parentheses shouldn't be there.

What does `sys.abiflags` get set to on Windows?

pfmoore: On Windows, "AttributeError: module 'sys' has no attribute 'abiflags'"

brettcannon:
Closing this since this is complicated enough that we should just have a library do this. 😉

PR #614, opened 2018-04-10
 #433 Update pep-0526.txt - Typo

toonarmycaptain:
Fixed format so that both Yes//No bullet points display consistently with bullet points.

### DIFF
--- a/pep-0526.txt
+++ b/pep-0526.txt
@@ -366,7 +366,7 @@ There should be no space before the colon. If an assignment has right hand
 side, then the equality sign should have exactly one space on both sides.
 Examples:
 
--  Yes::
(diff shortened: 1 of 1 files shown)


brettcannon:
Is the extra space in `-  Yes` on purpose? And why the change in bullet point style?

toonarmycaptain:
No, the extra space was inadvertent (and has been rectified). No reason for changing the bullet style, I just copied the `-` style from elsewhere in the document, as I wasn't aware of any convention.

brettcannon:
@toonarmycaptain no convention, just that both are allowed. But if the PEP itself is flipping back and forth then normalizing doesn't hurt.

PR #615, opened 2018-04-11
Delta114514

delta114514:
this proposal will make flat the nested iterators easier.


brettcannon:
As outlined in [PEP 1](https://www.python.org/dev/peps/pep-0001/), we do ask all PEPs first be vetted to the general community such as on python-ideas. Since I have not seen a discussion about this there, I'm closing this PR.

PR #618, opened 2018-04-12
Update pep-573

Traceur759:
Mentioning: @ncoghlan

### DIFF
--- a/pep-0573.rst
+++ b/pep-0573.rst
@@ -46,8 +46,6 @@ Process-Global State
 C-level static variables. Since this is very low-level
 memory storage, it must be managed carefully.
 
-(XXX Use this term everywhere)
-
 Per-module State
 ----------------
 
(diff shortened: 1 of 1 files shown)


pep-0573.rst
```diff
@@ -385,30 +394,51 @@ and return -1.
...
+A new structure, ``PyType_offsets``, will have two members containing the
+offsets of ``__dict__`` and ``__weakref__``::
+
+    typedef struct {
+        Py_ssize_t dict;
+        Py_ssize_t weaklist;
+    } PyType_offsets;
+
+The new slot, ``Py_offsets``, will be used to pass a ``PyType_offsets *``
+structure containing the mentioned data.
```
ncoghlan: Due to the stable ABI restrictions, it would be better if these were two separate slots - that way, if we need a new offset in the future, it would just be another slot definition.

If we use a struct, we can never make it bigger, as that could lead to a newer Python reading past the end of the memory allocated by an older extension module.

PR #625, opened 2018-04-20
PEP 574: rename BYTEARRAY to BYTEARRAY8

ogrisel:
Because the `BYTEARRAY` opcode needs to include the 64-bit length of the data (like `BINBYTES8` does already), I would find it more consistent to name that opcode `BYTEARRAY8` directly.

This is actually what @pitrou ended implementing in his draft branch for pickle protocol 5.

### DIFF
--- a/pep-0574.rst
+++ b/pep-0574.rst
@@ -274,7 +274,7 @@ Protocol changes
 
 Three new opcodes are introduced:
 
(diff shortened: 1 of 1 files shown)


pitrou:
You're right, I forgot to update the PEP for this.

PR #629, opened 2018-04-24
Update PEP 575

jdemeyer:
Also fix a typo in PEP 573.


Rosuav:
Rebased and pushed without squashing - they look like separate commits to me. Effectively merged though.

PR #634, opened 2018-04-27
Fix pep2rss to specify output encoding

### LINKED ISSUE #633: pep2rss can raise UnicodeEncodeError on certain platforms
This is due to the default encoding of [`open()`](https://docs.python.org/3/library/functions.html#open) being platform dependent.

For example, on Windows (CP-1252), it can raise
`UnicodeEncodeError: 'charmap' codec can't encode character '\u0141' in position 1545: character maps to <undefined>`
when handling the character, 'Ł', in the name of the author of [PEP 569](https://www.python.org/dev/peps/pep-0569/).
https://github.com/python/peps/blob/433faca4cf1bb8a0550d01b86b2b4581b3eda460/pep-0569.rst#L5

Harmon758:
Fixes #633

### DIFF
--- a/pep2rss.py
+++ b/pep2rss.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # usage: pep-hook.py $REPOS $REV
 # (standard post-commit args)
--- a/pep2rss.py
+++ b/pep2rss.py
@@ -69,5 +69,5 @@ rss = rssgen.RSS2(
     lastBuildDate = datetime.datetime.now(),
     items = items)
 
-with open(RSS_PATH, 'w') as fp:
+with open(RSS_PATH, 'w', encoding="utf-8") as fp:
     fp.write(rss.to_xml())
--- a/pep2rss.py
+++ b/pep2rss.py
@@ -4,13 +4,12 @@
 # (standard post-commit args)
 
 import os, glob, time, datetime, stat, re, sys
-import codecs
 import PyRSS2Gen as rssgen
 
 RSS_PATH = os.path.join(sys.argv[1], 'peps.rss')
 
 def firstline_startingwith(full_path, text):
-    for line in codecs.open(full_path, encoding="utf-8"):
+    for line in open(full_path, encoding="utf-8"):
         if line.startswith(text):
             return line[len(text):].strip()
     return None
@@ -69,5 +68,5 @@ rss = rssgen.RSS2(
     lastBuildDate = datetime.datetime.now(),
     items = items)
 
-with codecs.open(RSS_PATH, 'w', encoding="utf-8") as fp:
+with open(RSS_PATH, 'w', encoding="utf-8") as fp:
     fp.write(rss.to_xml(encoding="utf-8"))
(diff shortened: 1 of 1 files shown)


pep2rss.py
```diff
@@ -69,5 +69,5 @@ def pep_creation_dt(full_path):
     lastBuildDate = datetime.datetime.now(),
     items = items)
 
-with open(RSS_PATH, 'w') as fp:
+with open(RSS_PATH, 'w', encoding="utf-8") as fp:
```
zware: This looks dangerous; if it's run on Python 2 (which is quite possible; the shebang is `/usr/bin/env python`), this will fail.  We should do `from io import open` at the top.

Harmon758: Ah, I overlooked this.
I've gone ahead and changed it to `codecs.open` for consistency and because `rss.to_xml()` returns `str` and not `unicode` (which would have to be converted to not raise a `TypeError` in Python 2 when using `fp.write` with `io.open`).
This does cause line endings to output as '\n' instead of the platform-specific representation, but since line endings are supposed to be normalized to '\n' when processing XML anyway, I don't think this is of significance.

Harmon758:
I've noticed that the RSS feed generated still specifies "iso-8859-1" as the encoding in the XML declaration.

However, changing
https://github.com/python/peps/blob/499a7c25c6d07a6fc0cb85cac41e6689e59c40c3/pep2rss.py#L75
to
```py
fp.write(rss.to_xml(encoding="utf-8"))
```
results in 
```
  File "pep2rss.py", line 75, in <module>
    fp.write(rss.to_xml(encoding="utf-8"))
  File "C:\Python27\lib\codecs.py", line 706, in write
    return self.writer.write(data)
  File "C:\Python27\lib\codecs.py", line 369, in write
    data, consumed = self.encode(object, self.errors)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc5 in position 1537: ordinal not in range(128)
```
in Python 2.

This can be fixed by manually converting to `unicode` like so:
```py
fp.write(unicode(rss.to_xml(encoding="utf-8"), "utf-8"))
```
but I'm not sure of the best way to make this compatible with Python 3, beyond manually checking `sys.version_info[0]`, or whether this seemingly redundant method is the best way to go about this.

brettcannon:
Does it even need to support Python 2?

zware:
Changing the shebang would be simpler :)

holdenweb:
But wouldn't help on systems lacking Python 3 support, whose users might be
disappointed to see this project end-of-life before the end of next year ...

Steve Holden

brettcannon:
@holdenweb who is being disappointed from not generating the RSS feed for the PEPs using Python 2?

I'm with @zware and the shebang should simply be changed and we should just rely on Python 3.

gvanrossum:
This seems stalled -- who is waiting for whom here?

brettcannon:
@gvanrossum I think @zware and I are waiting for the PR to be updated to update the shebang and use Python 3 instead of trying to support Python 2.

Harmon758:
Ah, ok, since there don't seem to be any objections, I'll go ahead and drop Python 2 support.

zware:
Thanks, @Harmon758!

PR #646, opened 2018-05-10
PEP 544: Specify usage of modules as subtypes of protocols

ilevkivskyi:
See https://github.com/python/mypy/issues/5018 for motivation.

Attn: @gvanrossum @JukkaL @ambv

### DIFF
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -738,6 +738,54 @@ aliases::
   CompatReversible = Union[Reversible[T], SizedIterable[T]]
 
 
+Modules as subtypes of protocols
+--------------------------------
+
+A module object is accepted where a protocol is expected if the public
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
See https://github.com/python/mypy/issues/5322

gvanrossum:
Whoa. If you don't feel like working on this, fine, but there's no need to close the PR.

brettcannon:
I tweaked the title to say "WIP" so people know it's still open on purpose.

ilevkivskyi:
As we discussed recently, there is no need to wait for full implementation before acceptance (this part of the PEP is not implemented in mypy). I looked at this again, and seems to me this can be merged as is, the current proposal matches the discussion in https://github.com/python/mypy/issues/5322.

PR #647, opened 2018-05-10
PEP 544: Specify self-types in protocols

ilevkivskyi:
This is already implemented in mypy in exactly the same way it works for nominal classes (and thus shares most bugs of self-types in nominal classes). The question is: if self-types in protocols just follow the specification of PEP 484 do we need this additional subsection? (Or is it still helpful just as an example?)

Attn: @gvanrossum @JukkaL @ambv

### DIFF
--- a/pep-0544.txt
+++ b/pep-0544.txt
@@ -590,8 +590,8 @@ Continuing the previous example::
 Self-types in protocols
 -----------------------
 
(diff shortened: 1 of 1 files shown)


gvanrossum:
Yes, I think this is useful to remind readers that this works.

PR #649, opened 2018-05-15
Update PEP 575

jdemeyer:


jdemeyer:
The fix to PEP 539 is just to fix a build problem with the peps. It is unrelated to PEP 575 but I wanted to fix that issue anyway.

encukou:
Ah, I see! That threw me off.
PEP pull requests get squashed, so the fix-up should have its own PR. I filed it as https://github.com/python/peps/pull/651.

jdemeyer:
I just didn't want to get the blame for failing the build (which was already failing before) and I saw an easy fix.

jdemeyer:
I removed the commit touching PEP 539.

PR #650, opened 2018-05-15
New PEP 'Rationalize Built-in function classes'. 

markshannon:


jdemeyer:
Allow me to comment on this not-yet-released PEP:

**Performance**

Most importantly, changing bound methods of extension types from `builtin_function_or_method` to `bound_method` will certainly yield a performance loss. It might be possible to mitigate this somewhat by adding specific optimizations for calling `bound_method`. However, that would add extra complexity and it will probably still be slower than the existing code.

And I would also like to know whether it will be possible for custom built-in function subclasses to implement `__get__` to change a function into a method (like Python functions) and whether/how the `LOAD_METHOD` opcode will work in that case.

**Introspection**

When I want "introspection support", that goes beyond the call signature. Also `inspect.getfile` should be supported. Currently, that simply raises an exception for built-in functions.

I think it's important to specify the semantics of `inspect.isfunction`. Given that you don't mention it, I assume that `inspect.isfunction` will continue to return `True` only for Python functions. But that way, these new function classes won't behave like Python functions.

jdemeyer:
Is `Post-History: 10-May-2018` a typo or did you really post it on that date? In the latter case, I missed it...

PR #652, opened 2018-05-16
Add a teachability section to suggested PEP content in PEP-0001

willingc:
Closes #485


encukou:
#485 also has discussion on changing PEP 12. Should I open a new issue for that?

PR #656, opened 2018-05-16
Defer PEP 547

encukou:
Officially defer PEP 547. It is not ready, yet.

@Traceur759, do you agree?


ncoghlan:
Deferring until the PEP can actually be tested against Cythonised modules sounds like a good idea to me - otherwise we'd be at fairly high risk of standardising something Cython couldn't actually use.

PR #657, opened 2018-05-16
Update pep-0282.txt

daxm:
Modify mymodule.py, myapp.py, and the resulting output to work with Python 3.6.

### DIFF
--- a/pep-0282.txt
+++ b/pep-0282.txt
@@ -67,42 +67,42 @@ used to generate simple logging output on stderr.
 
     --------- mymodule.py -------------------------------
     import logging
+
     log = logging.getLogger("MyModule")
 
     def doIt():
-            log.debug("Doin' stuff...")
-            #do stuff...
-            raise TypeError, "Bogus type error for testing"
+        log.debug("Doin' stuff...")
+        # do stuff...
+        raise TypeError("Bogus type error for testing")
     -----------------------------------------------------
 
     --------- myapp.py ----------------------------------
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

When your account is ready, please add a comment in this pull request
and a Python core developer will remove the `CLA not signed` label
to make the bot check again.

daxm:
I've signed the CLA but my account appears broken.  Even though I verified my email address my login only gives "Invalid Login" error message when I attempt to log in.

JelleZijlstra:
Final PEPs are a historical record; I don't think we should update their syntax.

daxm:
Then I'd suggest that we do one of the following:
1.  Document the fact that the example shown only works in Python 2.
2.  Add a new example that is Python 3 compliant.
3.  Delete the example.

Personally I don't like any of these options.  It would be my preference to upkeep the PEP documentation to what is useful for the current needs of Python.  What is the use of an example if it no longer works?  What is the use of having a "historical" example, especially when people are looking to this documentation in an effort to learn how this feature works?

pfmoore:
The Python documentation should have sufficient information - if it doesn't, then that's where changes should be made to ensure that people wanting to use the feature don't *need* to refer to the PEP.

brettcannon:
If someone wants to leave a comment saying that the PEP is now for historical record and the documentation should be referred to how things work then that's fine, but updating the PEP's examples isn't necessary and worth trying to keep up-to-date.

PR #667, opened 2018-06-02
Explain why we didn't use setup.cfg for PEP 518

brettcannon:


ncoghlan:
For reference, this is the thread that ultimately lead to the acceptance of PEPs 517 & 518: https://mail.python.org/pipermail/distutils-sig/2016-May/028726.html

Perhaps it would be worth linking to that from the new subsection, and saying that standardising on setup.cfg was our starting point, but we moved away from that option due to the stated downsides?

pep-0518.txt
```diff
@@ -375,6 +375,17 @@ An example Python literal file for the proposed data would be::
      }
 
 
+Sticking with ``setup.cfg``
+---------------------------
+
+There are two issues with ``setup.cfg`` used by setuptools as a general
+format. One is that they are ``.ini`` files which have issues as mentioned
+in the configparser_ discussion above. The other is that the schema for
+that file has never been rigourlessly defined and thus it's unknown which
```
JelleZijlstra: Do you mean "rigorously" (not sure of the non-U.S. spelling) instead of "rigourlessly"?

ncoghlan:
I can't find the link right now, but I think we should also copy @njs's gist evaluating the formats in depth into the PEPs repo, since that goes into a lot more detail than the PEP does, and was the actual basis for the format selection decision.

ncoghlan:
Ah, that would explain why I couldn't find any public gists under the other account :)

Quick fix PR linking to your current gist https://github.com/python/peps/pull/668 (but it would still be good to archive that directly in the PEPs repo)

pep-0518.txt
brettcannon: Fixed

sametmax: > ``.ini`` files which have issues as mentioned
+in the configparser_ discussion above

I LOVE toml. But if we go the parser route, the current toml parser in Python is atrocious. It's brittle and makes syntax error very hard to debug.

> The other is that the schema for
+that file has never been rigourlessly defined

It's an exaggeration. It's not perfect, but it's quite complete and still way easier than to start with setup.py:
 
http://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

It also had fantastic goodies like including the version of the package in one line.

PR #671, opened 2018-06-13
[pep8] Document empty blocks

ambv:
Original discussion: https://twitter.com/llanga/status/1006974601723723776

### DIFF
--- a/pep-0008.txt
+++ b/pep-0008.txt
@@ -310,6 +310,18 @@ Note, some editors and web-based code viewers may not recognize
 control-L as a form feed and will show another glyph in its place.
 
 
+Empty blocks
+------------
+
+Use ``pass`` to denote a block is intentionally empty.  In empty
+class or function bodies, prefer a meaningful docstring instead of
+``pass``.
+
+Do not use ``...`` or ``None`` for this purpose.  The former suggests
+"details supressed" instead of "there is nothing there".  The latter
+looks like an unfinished thought.
+
+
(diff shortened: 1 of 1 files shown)


pep-0008.txt
```diff
@@ -310,6 +310,18 @@ Note, some editors and web-based code viewers may not recognize
...
 
+Empty blocks
+------------
+
+Use ``pass`` to denote a block is intentionally empty.  In empty
+class or function bodies, prefer a meaningful docstring instead of
+``pass``.
+
+Do not use ``...`` or ``None`` for this purpose.  The former suggests
+"details supressed" instead of "there is nothing there".  The latter
```
willingc: typo: suppressed

pep-0008.txt
```diff
@@ -310,6 +310,18 @@ Note, some editors and web-based code viewers may not recognize
 control-L as a form feed and will show another glyph in its place.
 
 
+Empty blocks
+------------
+
+Use ``pass`` to denote a block is intentionally empty.  In empty
```
willingc: ...block which is intentionally left empty.

gvanrossum:
I still don't think this needs to make it into PEP 8 -- I don't think I've ever seen a debate about this or code doing it wrong until Łukasz brought it up on Twitter. We should exercise restraint when adding to PEP 8 -- it's already getting long, and there are more important things that it's not even mentioning.

PR #674, opened 2018-06-16
Begins splitting PEP 551 into PEP 578

zooba:
At Guido's request, I am splitting PEP 551 into two PEPs. The end result will be:
* PEP 578 is a standards track PEP for Python 3.8 adding runtime audit and open-for-import hooks
* PEP 551 is an informational PEP providing guidance for those integrating Python into secure/audited environments

The `spython` implementation will become a sample associated with PEP 551 and will not be checked in with the PEP 578 changes.

### DIFF
--- a/pep-0551.rst
+++ b/pep-0551.rst
@@ -141,7 +141,8 @@ assist system administrators with deploying a security transparent
 version of Python that can integrate with their existing auditing and
 protection systems.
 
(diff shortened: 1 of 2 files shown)


zooba:
This is a work in progress still, but I wanted to try and claim PEP 578 and also share diffs.

PR #681, opened 2018-06-21
PEP 581: (First draft) Using GitHub Issues

Mariatta:


pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
+- With more than 28 million users, an open source contributor is more likely
+  to already have an account, and familiar with GitHub interface, making it
+  easier to start contributing.
+
+- Email notifications contain metadata [#email-meta]_, integrated with GMail, and
+  allows you to systematically filter emails.
+
+- Email address can remain hidden by choice.
+
```
vstinner: Few more advantages:

* Single namespace for issue numbers and pull request numbers: `#xxx` (instead of `bpo-xxx` vs `GH-xxx`). Many people omits `bpo-` or `GH-` prefix :-(
* Ability to "quote" code using a permalink to lines of code in the Git repository. Example: `https://github.com/python/cpython/blob/bcd3a1a18d841338f57c39f6a7de8cf14d0c3e03/Lib/asyncio/base_events.py#L375-L377` creates:

http://haypo.alwaysdata.net/tmp/quote.png

(The URL to the picture is a temporary link, please copy the picture somewhere!)

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+Why GitHub
+----------
+
+It has a lot of nice features readily available out of the box, that are not
+currently available in Roundup / bpo.
+
+- APIs we can use to build integrations and automations.
+
+- Ability to attach screenshots and files in pull requests and issues.
```
vstinner: Hum, that's a basic feature of Roundup, whereas GitHub only accepts a small set of file formats, no? Last time I tried to attach a file, I got an error because the file format was denied.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
+- Administrators and core developers can edit issues comments and pull requests.
```
vstinner: I'm not sure that it's a big difference with bugs.python.org. The difference here is that the GitHub administrators can more easily give bug triage permission: single platform (GitHub) instead of two (GitHub + bugs.python.org). Single user directory.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+- Ability to reply to issues and pull request conversations via email.
+
+- It supports two factor authentication.
+
+- It supports markdown and emoji.
+
+- Core developers don't have to maintain it, giving us more time to focus on
+  the development of Python.
+
+- We can automatically close issues when a PR has been merged [#close-github-issue]_.
```
vstinner: We had that previously with Subversion, maybe also with Mercurial. It seems like we lost the feature when we migrated to Git and GitHub, likely because nobody "ported" the feature to Git.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+CPython's development moved to GitHub on February 2017. All other projects within
+The PSF's organization are hosted on GitHub and are using GitHub issues.
+CPython is still using Roundup as the issue tracker in bugs.python.org (bpo).
+
+Why GitHub
+----------
+
+It has a lot of nice features readily available out of the box, that are not
+currently available in Roundup / bpo.
+
```
vstinner: I would start by saying that the form is much simpler, so it's easier to report a bug, no?

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+- Ability to attach screenshots and files in pull requests and issues.
+
+- Administrators and core developers can edit issues comments and pull requests.
+
+- Ability to reply to issues and pull request conversations via email.
+
+- It supports two factor authentication.
+
+- It supports markdown and emoji.
```
vstinner: There is also a Preview tab, very useful for non-trivial formatting.

It also supports votes, but I'm not sure that it's an advantage :-D

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
+Issues with bpo / Roundup
+-------------------------
+
+- It is not actively maintained.
+
+- Less than five people maintain bpo. Some of them are core developers.
+
+- It is in mercurial. Without any CI available, it puts heavy burden on existing
+  (few) maintainers in terms of reviewing, testing, and applying patches.
```
vstinner: It would help to count them. I counted 2, but it's better to ask Ezio for a more accurate answer. Ezio recently wrote: https://mail.python.org/pipermail/python-committers/2018-June/005515.html

In the mercurial logs, I see commits from 3 developers: Ezio, Berker, Maciej.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
+- At its current state, it is not equipped to accept lots of contributions from
+  people who aren't already familiar with the code base.
+
+- The upstream roundup is in mercurial. If bpo does move to GitHub, it will become
+  difficult to update patches from upstream.
```
vstinner: I don't understand "If bpo move to GitHub": do you mean migrating the Mercurial repository to Git and move it to GitHub? Or move Python bugs (bugs.python.org) to GH?

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
+- The user interface needs update and redesign. It will require UX/UI research
+  to keep it up to date with current web standards.
+
+- Email address is exposed. There is no choice to mask it.
+
+- There is no REST API available. There is an open issue in Roundup for adding
+  REST API  [#roundup-rest-api]_. Last activity was in 2016.
+
+- I personally don't want to receive email each time someone added themselves as "nosy"
```
vstinner: You might express it without writing "I", maybe: Email notifications are sent each time someone adds themself as "nosy" and it's not possible to opt-out for such notification.

In general, Roundup sends many useless notifications and it's not possible or hard to configure notifications. GitHub has a global "Ignore" option per project, and a similar thing to "nosy": Subscribe/Unbscribe per issue. But no notification to everybody when someone subscribes (good!).

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
+- Creating an account has been a hassle. We've had reports when people have
+  trouble creating accounts or logging in.
+
+Why not GitLab
+--------------
+
+Had we migrated to GitLab instead of GitHub in 2017, this PEP would have been
+titled "Using GitLab Issues for CPython".
+
```
vstinner: You should mention somewhere "commercial" or "vendor lock-in" and explain why you consider that it's not a blocker issue. I'm sure that the question will arise.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+In order to really improve Roundup / bpo, it needs to first migrate to GitHub,
+add CI and bots. As I understand it, there is hesitation because upstream Roundup
+is still im mercurial. Someone more familiar with Roundup / bpo needs
+to champion this effort. (I'm not volunteering, I'm sorry).
+
+I believe the effort of creating and maintaining GitHub integrations and bots
+is much less than the effort needed to get Roundup up to speed and then to continue
+maintaining it.
+
```
vstinner: Since GitHub is popular, there are also a lot of existing tooling around GitHub. I don't think that it's the case for Roundup. GitHub also has builtin webhooks and a wide choice of "services". Enabling a service is just 2 clicks.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+==============
+
+Backup of GitHub data
+---------------------
+
+This effort has been started and being tracked as an issue in core-workflow [#backup-github]_.
+We're using GitHub Migrations API [#gh-migrations-api]_ to download GitHub data
+for CPython on daily basis. The archives will be dropped in an S3 bucket.
+
+Thanks to Ernest W. Durbin III for working on this.
```
vstinner: Download data: ok, but how can we make sure that data are usable? Is it possible to get a read-only review of these backups?

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
+If a core developer is still interested in the issue, they can
+indicate so in the bpo issue, and later use the button to migrate it over to GitHub.
+
+Make bpo readonly
+-----------------
+
+This should be the final step. Once we start using GitHub issues, make bpo
+readonly, not shut it down.
+Do not accept new registrations. Do not allow comments or issues to be created.
```
vstinner: An alternative is to keep bpo: have bpo *and* GH.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+Question: Will it be possible to completely automate the CLA process, so
+it does not require human intervention?
+
+Create labels for issue triage
+------------------------------
+
+In bpo, we currently have the following fields for each issues:
+
+Type: behavior, crash, compile error, resource usage, security, performance, enhancement.
+Components: 2to3, Argument Clinic, asyncio, Build, Cross-build, ctypes, ...
```
vstinner: IMHO components will be useful on pull requests are well. We already have "type-doc" label for example. It helps when someone wants to review all documentation PRs  for example.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+Priority: release blocker, deferred blocker, critical, high, normal, low
```
vstinner: Ditto: when a PR has no associated bug, priority might be useful on PRs as well. But when a PR has an associated bug, the priority would be redundant and may become inconsistent if only updated on one side :-/

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+
```
vstinner: We need a field to specify the Python version that should be fixed, like the Python Versions field on bpo.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+We will create the corresponding labels::
+
+   type-behavior, type-crash, type-compile error, type-resource usage, ...
+
+   components-2to3, components-argument clinic, components-asyncio, ...
+
+   priority-release blocker, priority-deferred blocker, priority-critical, ...
+
+In addition, we'll create ``needs triage`` label.
```
vstinner: Please elaborate. How do you qualify that a bug is triagged or not?

pep-0581.rst
Mariatta: Hm perhaps I should say "embed" screenshots, that I can see images directly in the issue, instead of having to click and open a new tab for the image.

pep-0581.rst
Mariatta: I cannot edit or delete a message in bpo.
If I made a mistake in my message, I had to post a new message to explain my previous mistake.
In GitHub I can edit my own messages, and core devs can edit everyone's messages.

pep-0581.rst
Mariatta: I like these features too. Will add these in my next iteration :) Thanks.

pep-0581.rst
Mariatta: I meant to move the codebase of bugs.python.org from mercurial repository to GitHub. Will clarify in the next iteration. Thanks.

pep-0581.rst
Mariatta: See https://github.com/python/core-workflow/issues/20#issuecomment-396238565 for what kind of data are getting backed up. Perhaps ask Ernest about getting a copy of the backup if you're interested?

pep-0581.rst
Mariatta: See "create issue template" below.

If the issue does not have at least one of `type-`, `component`, `priority-`, `needs backport to-` labels, then bot will apply `needs triage` label.

pep-0581.rst
Mariatta: No. Because then we need to support two different workflows and further complicate our life.

pep-0581.rst
ammaraskar: Github's all or nothing permission model is kind of a hindrance in this case. Right now there are triagers on bpo who aren't core devs, but can help categorize and resolve issues on bpo. Moving to Github means it would be harder to give them access.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+Why not GitLab
+--------------
+
+Had we migrated to GitLab instead of GitHub in 2017, this PEP would have been
+titled "Using GitLab Issues for CPython".
+
+Why not other issue tracker
+---------------------------
+
+Using another issue tracker will require yet another learning curve, of having
```
ammaraskar: Using Github to track issues has its own learning curve, do you mean its less steep than other bug trackers?

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+back with a PEP.
+
+Why not focus on improving Roundup / bpo
+----------------------------------------
+
+GitHub has many features we like that are already available. We still need to
+build out additional integrations and update our bots, but this is something
+we already know how to do.
+
+In order to really improve Roundup / bpo, it needs to first migrate to GitHub,
```
ammaraskar: I think this is really only a concern with the roundup fork, which might not even be necessary anymore. The `python-dev` repo which hosts our custom plugins for roundup should be able to migrate to git without issue.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+At the moment the CLA is hosted within bpo. It needs to be updated such that
+signing the CLA does not require a bpo account, and it should be hosted outside
+of the bpo.
+
+The current CLA process itself is not ideal. Currently contributors to
+devguide, peps, and core-workflow need to sign a CLA,  and it requires a bpo
+account. A bpo account should not be required for those projects.
+
+Currently the CLA process involves real human to manually check the records.
+Question: Will it be possible to completely automate the CLA process, so
```
ammaraskar: This would be a great time to start a thread with PSF legal on this. On the technical side, there are already great projects to handle this like http://clabot.github.io/

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+Create labels for issue triage
+------------------------------
+
+In bpo, we currently have the following fields for each issues:
+
+Type: behavior, crash, compile error, resource usage, security, performance, enhancement.
+Components: 2to3, Argument Clinic, asyncio, Build, Cross-build, ctypes, ...
+Priority: release blocker, deferred blocker, critical, high, normal, low
+
+We will create the corresponding labels::
```
ammaraskar: Hmm, this is a lot of labels. Have you given thought or experimented to see how well the Github interface handles this many labels?

ammaraskar:
As far as downsides of Github, I think one thing to note is that its pretty much impossible to get them to do anything. This should be a major indicator of that: https://github.com/dear-github/dear-github

This means that we shouldn't really expect anything beyond the current feature set they provide.

pep-0581.rst
```diff
@@ -0,0 +1,310 @@
...
+- It supports two factor authentication.
+
+- It supports markdown and emoji.
+
+- Core developers don't have to maintain it, giving us more time to focus on
+  the development of Python.
+
+- We can automatically close issues when a PR has been merged [#close-github-issue]_.
+
+- With more than 28 million users, an open source contributor is more likely
```
ammaraskar: This is a great point and worth highlighting, bpo right now is quite intimidating to people trying to report issues. Unless you've used it for a while the interface is really prehistoric and un-intuitive.

pep-0581.rst
pradyunsg: quote

(the image linked to by @vstinner; attached via drag-drop on GitHub's web UI)

pep-0581.rst
ncoghlan: The repo exports that Ernest (has set up? is setting up?) are definitely worth mentioning.

ncoghlan: Ah, they're already mentioned later on in a slightly different context. Still, that's the key point for addressing any lock-in concerns: there are PSF hosted data backups, and the only thing that gets potentially tricky is tracking identity following a future migration away from GitHub, and the CLA signing process can cover that.

pep-0581.rst
ncoghlan: But the opportunity can potentially be taken to trim the component list

pep-0581.rst
ncoghlan: Honestly, I think "release blocker" and "deferred blocker" are the only actually useful priority levels we currently have, and those will always be associated with issues rather than PRs.

pep-0581.rst
ncoghlan: I think we can use the existing backport labels for that - just set them on the issue to indicate which labels the PR should get.

pep-0581.rst
Mariatta: I meant that since we (existing Python core devs, The PSF, and contributors to Python) have been using GitHub for everything else, there is less learning to do, than if we're to use a different issue tracker.

pep-0581.rst
Mariatta: I was thinking that this itself might require a separate PEP 😅

pep-0581.rst
Mariatta: I'm for trimming it :)  but perhaps we'll need more discussions with other core devs.

pep-0581.rst
Mariatta: Yeah I plan to use the existing "needs backport to:" labels. These can be attached to the issue, and later automatically applied to the PR.

pep-0581.rst
Mariatta: I haven't seen any mention of limit of labels on GitHub anywhere. These are labels created but in really not all of them will be applied to the issue/PR. 
In coala we've been working with [65 labels](https://github.com/coala/coala/labels?page=1&sort=name-asc), and we also use bots to automatically categorize [PRs](https://github.com/coala/coala/pulls).

Mariatta:
Thanks for all the reviews!

> This means that we shouldn't really expect anything beyond the current feature set they provide.

I'm not too concerned. I believe we can work with GitHub as is. This PEP did not mention anything about needing GitHub to provide us additional features.

pep-0581.rst
ammaraskar: Aah perfect, you already have a project that has a lot of labels 👍

ammaraskar:
>This PEP did not mention anything about needing GitHub to provide us additional features.

Right, I just mean it might be a problem in the future if a new feature is required. Though with how hard it is to build new features into bpo it might be a moot point.

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+
+We will create the corresponding labels::
+
+   type-behavior, type-crash, type-compile error, type-resource usage, ...
+
+   components-2to3, components-argument clinic, components-asyncio, ...
+
+   priority-release blocker, priority-deferred blocker, priority-critical, ...
+
+In addition, we'll create ``needs triage`` label.
```
emilyemorehouse: I think it could be helpful to utilize GitHub's multiple issue templates [(more info here)](https://blog.github.com/2018-05-02-issue-template-improvements/). I found this to be really helpful on other OSS projects to keep general usage bugs out of the issue tracker by redirecting them elsewhere. 

IMO, Babel [does this really well](https://github.com/babel/babel/issues/new/choose) (and has great emoji usage too).

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+
+This will require actual update to the bpo. But I believe this effort needed
+is much less than a complete overhaul.
+
+We will create a button in bpo, that will copy over the issue description
+and associated comments into a GitHub issue.
+
+We should not be moving all open issues to GitHub. Issues with little or no
+activity should just be closed. Issues with no decision made for years should
+just be closed.
```
terryjreedy: No, No, No.

Closing bpo issues is a separate topic from and irrelevant to creating a usable GH cpython issue tracker.  It can and has been done without this PEP.  More could be done, but not so blindly as you propose.  Doing so is a separate discussion, which I won't continue here.

Implementing this PEP is unaffected by bpo issue closures.  The main affect of closure is limiting bpo search results. Bpo issue closures do not belong on this PEP and should be removed.  Please don't repeat Chris Angelico's mistake of piggybacking a separate proposal onto his main proposal of assignment expressions.

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+
+If a core developer is still interested in the issue, they can indicate so in
+the bpo issue, and later use the button to migrate it over to GitHub.
+
+Make bpo readonly
+-----------------
+
+This should be the final step. Once we start using GitHub issues, make bpo
+readonly, not shut it down.
```
terryjreedy: It was not practical to use hg and git simultaneously.  In my view, and some others, apparently, github was not fully ready to use when the switch was made.  I would say not until Miss Islington started doing backports.  (They still take more time than on hg, but the difference is tolerable.)  The result was that many people stopped patching python for various periods.

It *would* be practical to have both issue trackers in use for a short time.  So I think disabling bpo should wait until the new system has been used and approved by people other than the creators.  Unlike with the repository, testing can be done with real issues for cpython PRs.

So I think approval of this PEP should be provisional, subject to final acceptance review, which would then signal 'stop using bpo'.  At that point, shut off new registrations and issues, (which could be discourage earlier), and new comments.  'Readonly' should mean that search still works, and that issues can still be closed instead of transferred.

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+----------
+
+GitHub has a lot of nice features readily available out of the box, that are not
+currently available in Roundup / bpo.
+
+- APIs we can use to build integrations and automations. There are various existing
+  integrations and applications available from GitHub Marketplace to help with
+  our workflow. New applications are easily installed and enabled.
+
+- Ability to embed screenshots and files in pull requests and issues.
```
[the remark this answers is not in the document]
ammaraskar: This was discussed above: https://github.com/python/peps/pull/681#discussion_r197403355

While bpo might support uploading images, you have to click away to view them as opposed to allowing them to be embedded in the issue page.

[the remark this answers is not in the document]
terryjreedy: Images are rare, and at least some I have seen are well or better viewed on a separate tab.  And I still prefer them listed with other files.   Multi-screen text is definitely better on a new tab.

pep-0581.rst
terryjreedy: This image is a png image, not a permalink to the code itself.  Clicking on the image gets you to the image on a screen by itself, not to the code.

pep-0581.rst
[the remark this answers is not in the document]
ammaraskar: You can still click them to open them in a new tab on Github, there's not really a concept of a list of attached files to an issue on here though. That goes for images, test scripts etc. You'll have to find the particular comment where it was attached to download it.

pep-0581.rst
ammaraskar: That was just an example from elsewhere, here's one from this repo: 
https://github.com/python/peps/blob/8ace3c5e1fa67967824fcf2bcf24435ad3cfef2f/pep0/pep.py#L14-L19

zware: @terryjreedy The image is a representation of what you get from dropping a link like `https://github.com/python/peps/blob/8ace3c5e1fa67967824fcf2bcf24435ad3cfef2f/genpepindex.py#L17-L27` in a message:

https://github.com/python/peps/blob/8ace3c5e1fa67967824fcf2bcf24435ad3cfef2f/genpepindex.py#L17-L27

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+  to already have an account, and familiar with GitHub interface, making it
+  easier to start contributing.
+
+- Email notifications contain metadata [4]_, integrated with GMail, and
+  allows you to systematically filter emails.
+
+- Email address can remain hidden by choice.
+
+
+Issues with bpo / Roundup
```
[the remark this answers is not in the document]
ammaraskar: >This UI design is FAR, FAR superior to the anti-design of just jumbling all options together under Labels.

I would agree with this, putting the dropdowns etc in the issue text and having a bot pick them is a little bit lackluster compared to a purpose built form. (There's been an open issue about this as far back as 2014 https://github.com/isaacs/github/issues/148)

pep-0581.rst
terryjreedy: On bpo, I copy-paste the link and then the quote, between "". 

`https://github.com/python/peps/blob/8ace3c5e1fa67967824fcf2bcf24435ad3cfef2f/genpepindex.py#L4-L5`

"Generating the PEP index is a multi-step process.  To begin, you must first
parse the PEP files themselves, which in and of itself takes a couple of steps:
"
Given that one wants the complete lines, it is definitely easier to only copy-paste the link to get the same effect.

https://github.com/python/peps/blob/8ace3c5e1fa67967824fcf2bcf24435ad3cfef2f/genpepindex.py#L4-L5

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+
+- It supports two factor authentication.
+
+- It supports markdown and emoji. It has a preview tab, so we can see how our
+  comment will be rendered, before actually posting.
+
+- It supports voting (reactions).
+
+- It supports permalink [2]_, allowing us to quote and copy paste
+  source code easily.
```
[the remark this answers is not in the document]
terryjreedy: After seeing yours and Zach's proper examples, and trying it for myself, I agree that link + quote of *complete lines* (typical for code) is easier.  But it is pretty easy on bpo.  So I would agree with "copy paste source code lines more easily."

Copying partial or complete sentences from docs, with extraneous material before and after the desired quote removed, as it typical for doc issues, requires a separate copy paste on either tracker.

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+Provide explanation in the devguide about new issue workflow and triage labels.
+
+Add a button in bpo to migrate the issue to GitHub
+--------------------------------------------------
+
+This will require actual update to the bpo. But I believe this effort needed
+is much less than a complete overhaul.
+
+We will create a button in bpo, that will copy over the issue description
+and associated comments into a GitHub issue.
```
terryjreedy: What about the existing files, which are typically patch files?  Create a comment to embed them in? Or put a link to the original in?

pep-0581.rst
willingc: To minimize the burden of maintaining two issue reporting/commenting workflows, one for GitHub and one for bpo, the following constraints will be made to bpo when GitHub is used for issues:
- Disable new user registration to bpo. 
- Disable new issue creation on bpo.
- Lock comments on existing bpo issues. Additional comments would be done on GitHub with a link to the archival bpo issue.

pep-0581.rst
Mariatta: I'm assuming your good intention, but I think your last sentence is unnecessary.

pep-0581.rst
matrixise: I think there is an other issue, currently, there are 2 lists, the contributors and the members. do you think you will add a new list for the 'triage' ? on bpo, my bpo account can change the associated component of an issue, could I do the same here?

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+   Needs backport to: 2.7 | 3.6 | 3.7
+   ---
+
+The idea is to allow the issue creator to help us triage the issue.
+The above values are pre-filled in the template. Issue creator will remove texts
+that do not apply to their issue.
+
+Based on the above headers, bedevere-bot can apply the necessary labels to the
+issue. If issue creator did not supply the above headers, the bot will apply
+``needs triage`` label. At that point it will require a core developer to
```
matrixise: a core developer to properly label the issue, and what happens for the bpo accounts with the 'triage' flag? as mine.

pep-0581.rst
```diff
@@ -0,0 +1,362 @@
...
+
+- APIs we can use to build integrations and automations. There are various existing
+  integrations and applications available from GitHub Marketplace to help with
+  our workflow. New applications are easily installed and enabled.
+
+- Ability to embed screenshots and files in pull requests and issues.
+
+- Administrators and core developers can edit issues comments and pull requests.
+
+- Ability to reply to issues and pull request conversations via email.
```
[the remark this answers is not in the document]
freddrake: This is often caused by not realizing it was an email from a tracker (at least in my case).  Having whatever tracker we use make it visually clear that it's an automated tracker notification usually gets me to click through, and not trust the email return path.

PR #692, opened 2018-07-04
PEP 572 accepted by Guido van Rossum

vstinner:


vstinner:
Guido wrote:
> And if you're asking whether it's likely that I'll change my mind, no.

https://mail.python.org/pipermail/python-dev/2018-July/154247.html

The PEP 572 is currently heavily discussed on the Internet, so marking it as accepted would clarify the situation.

For example, when I wrote on python-dev that the PEP is acepted, Ivan Pozdeev replied:
> As of now, https://www.python.org/dev/peps/pep-0572/ is marked as "draft".

https://mail.python.org/pipermail/python-dev/2018-July/154278.html

gvanrossum:
I really don't care what the internet says about this PEP, I want to wait about a week before formally marking it as accepted. There will be a follow-up discussion on python-dev first, another posting for (editorial) feedback, and a better message to link to for acceptance.

vstinner:
Ok.

PR #693, opened 2018-07-04
PEP 505: prepare for 3.8

zooba:
Add myself as PEP 505 coauthor (with Mark's permission), and prepare for the rewrite!


pradyunsg:
@zooba Is this rewrite still underway?

If so, would it make sense to update this back to a "Draft"? If not, could we add back a "PEP Deferral" section to go with the "Deffered" status?

PR #695, opened 2018-07-06
PEP 572: Add examples from the stdlib

vstinner:


pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+site
+^^^^
+
++-----------------------------------------------------+------------------------------------------------------------+
+| Current code                                        | Replaced with                                              |
++-----------------------------------------------------+------------------------------------------------------------+
+|                                                     |                                                            |
+| ::                                                  | ::                                                         |
+|                                                     |                                                            |
+|  env_base = os.environ.get("PYTHONUSERBASE", None)  |   if (env_base := os.environ.get("PYTHONUSERBASE", None)): |
```
gvanrossum: The outermost parentheses on the right here are redundant. Please remove them.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+copy
+^^^^
+
++---------------------------------------------------------------+---------------------------------------------------------------+
+| Current code                                                  | Replaced with                                                 |
++---------------------------------------------------------------+---------------------------------------------------------------+
+|                                                               |                                                               |
+| ::                                                            | ::                                                            |
+|                                                               |                                                               |
+|   reductor = dispatch_table.get(cls)                          |   if (reductor := dispatch_table.get(cls)):                   |
```
gvanrossum: Ditto.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+|   if reductor:                                                |       rv = reductor(x)                                        |
+|       rv = reductor(x)                                        |   elif (reductor := getattr(x, "__reduce_ex__", None)):       |
```
gvanrossum: And here.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+|   else:                                                       |       rv = reductor(4)                                        |
+|       reductor = getattr(x, "__reduce_ex__", None)            |   elif (reductor := getattr(x, "__reduce__", None)):          |
```
gvanrossum: And here.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
++-----------------------------------------------------+-----------------------------------------------------+
+| Current code                                        | Replaced with                                       |
++-----------------------------------------------------+-----------------------------------------------------+
+|                                                     |                                                     |
+| ::                                                  | ::                                                  |
+|                                                     |                                                     |
+|   s = _format_time(self._hour, self._minute,        |   s = _format_time(self._hour, self._minute,        |
+|                    self._second, self._microsecond, |                    self._second, self._microsecond, |
+|                    timespec)                        |                    timespec)                        |
+|   tz = self._tzstr()                                |   if (tz := self._tzstr()):                         |
```
gvanrossum: Again.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+sysconfig
+^^^^^^^^^
+
++------------------------------------+-----------------------------------------+
+| Current code                       | Replaced with                           |
++------------------------------------+-----------------------------------------+
+|                                    |                                         |
+| ::                                 | ::                                      |
+|                                    |                                         |
+|   while True:                      |   while (line := fp.readline()):        |
```
gvanrossum: Again.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+|       line = fp.readline()         |       if (m := define_rx.match(line)):  |
```
gvanrossum: And here.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+|       if not line:                 |           n, v = m.group(1, 2)          |
+|           break                    |           try:                          |
+|       m = define_rx.match(line)    |               v = int(v)                |
+|       if m:                        |           except ValueError:            |
+|           n, v = m.group(1, 2)     |               pass                      |
+|           try:                     |           vars[n] = v                   |
+|               v = int(v)           |       elif (m := undef_rx.match(line)): |
```
gvanrossum: And here.

pep-0572.rst
```diff
@@ -359,6 +359,125 @@ found in assignment statements:
...
+|      if ans:                                 |                                                                       |
+|          return ans                          |                                                                       |
+|                                              |                                                                       |
++----------------------------------------------+-----------------------------------------------------------------------+
+
+Avoid nested if and remove one indentation level.
+
+copy
+^^^^
+
```
gvanrossum: Maybe add a note here that this example is duplicated from Appendix A -- I don't mind the duplication but I don't want people to wonder if there perhaps is a subtle difference. (Subtlety in use of assignment expressions is an anti-pattern.)

tim-one:
If the redundant parens are removed, I'm also +1.   And also don't mind the duplicated example - if this is merged, it will look like I wrote the Appendix independent of the body of the PEP and didn't notice the duplication, which is fine by me.   I'd like to keep the example in the Appendix too so that the credit I gave to the person who first pointed it out survives.  They deserve credit for looking at real code before nearly else could even be beaten into trying it.

gvanrossum:
Thanks for digging these up Victor! They really help. I committed them in #698.

PR #698, opened 2018-07-07
Add PEP 572 examples from stdlib

gvanrossum:
(Forked from #695, mostly to remove the redundant parentheses that Victor added.)

### DIFF
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -400,7 +400,7 @@ Avoid nested if and remove one indentation level.
 copy.py
 ^^^^^^^
 
-(See Appending A for the origin of this example.)
+(See Appendix A for the origin of this example.)
 
 +---------------------------------------------------------------+---------------------------------------------------------------+
 | Current code                                                  | Replaced with                                                 |
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -359,6 +359,125 @@ found in assignment statements:
 Examples
 ========
 
+Examples from the Python standard library
+-----------------------------------------
+
+site
+^^^^
+
++-----------------------------------------------------+------------------------------------------------------------+
+| Current code                                        | Replaced with                                              |
++-----------------------------------------------------+------------------------------------------------------------+
+|                                                     |                                                            |
+| ::                                                  | ::                                                         |
+|                                                     |                                                            |
+|  env_base = os.environ.get("PYTHONUSERBASE", None)  |   if (env_base := os.environ.get("PYTHONUSERBASE", None)): |
+|  if env_base:                                       |       return env_base                                      |
+|      return env_base                                |                                                            |
+|                                                     |                                                            |
++-----------------------------------------------------+------------------------------------------------------------+
+
+*env_base* is only used on these lines, putting its assignment on the if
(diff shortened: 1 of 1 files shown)


pep-0572.rst
```diff
@@ -359,6 +359,127 @@ found in assignment statements:
...
+|          return ans                          |                                                                       |
+|                                              |                                                                       |
++----------------------------------------------+-----------------------------------------------------------------------+
+
+Avoid nested if and remove one indentation level.
+
+copy.py
+^^^^^^^
+
+(See Appending A for the origin of this example.)
```
tim-one: s/Appending/Appendix/

pep-0572.rst
```diff
@@ -359,6 +359,127 @@ found in assignment statements:
...
+|                    timespec)                        |                    timespec)                        |
+|   tz = self._tzstr()                                |   if tz := self._tzstr():                           |
+|   if tz:                                            |       s += tz                                       |
+|       s += tz                                       |   return s                                          |
+|   return s                                          |                                                     |
+|                                                     |                                                     |
++-----------------------------------------------------+-----------------------------------------------------+
+
+*tz* is only used for ``s += tz``, moving its assignment inside the if
+helps to show its scope.
```
tim-one: I wonder whether it's worthwhile to add more-general text with the observation that it's _often_ the case that targets in `if` and `while` tests are conceptually local to the block construct.  Victor seemed really keen on that, and Chris, after all, tried to enforce it in earlier versions of the PEP.  But "implied intended scoping" isn't my personal itch, so I'll leave that to them 😉.

pep-0572.rst
```diff
@@ -359,6 +359,127 @@ found in assignment statements:
...
+|           m = undef_rx.match(line) |                                         |
+|           if m:                    |                                         |
+|               vars[m.group(1)] = 0 |                                         |
+|                                    |                                         |
++------------------------------------+-----------------------------------------+
+
+Calling ``fp.readline()`` in the ``while`` condition and calling
+``.match()`` on the if lines make the code more compact without making
+it harder to understand.
+
```
tim-one: Just leaving a note of appreciation for this example - I find the right-side version significantly easier to follow!

pep-0572.rst
gvanrossum: Hm, when I read Victor's post on that topic I was pretty confused. We'd have to introduce a term like "intuitive scope" to refer to the notion that the author intends a variable's scope to end after a given block. And honestly that concept applies to classic code (without assignment expressions) too -- e.g. when people write
```py
m = re.match(pat, line)
if m:
    return m
```
there is no use for `m` after the end of that block.

So I don't think we should add a discussion of this idea to the PEP.

tim-one: Yes, in that case Victor's idea is that using an assignment expression instead would _announce_ the intent that `m`'s value will never be used beyond the end of the block, even you replaced `return m` with - say - `print(m)`.  The intent is announced by the form of the initial binding, not by studying the content of the dependent block(s).

It's not something I ever really have a problem with, though, so it's not my itch.   So if they want to pursue it, it's up to them to push it.

PR #710, opened 2018-07-08
PEP 572: Move the proposed evaluation order change to "Syntax and semantics"

gvanrossum:
Remove "Open questions and TODOs" since they're now all decided.

### DIFF
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -307,6 +307,21 @@ This PEP recommends always putting spaces around ``:=``, similar to
 PEP 8's recommendation for ``=`` when used for assignment, whereas the
 latter disallows spaces around ``=`` used for keyword arguments.)
 
+Change to evaluation order
+--------------------------
+
+In order to have precisely defined semantics, the proposal requires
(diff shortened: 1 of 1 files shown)


pep-0572.rst
```diff
@@ -307,6 +307,21 @@ This PEP recommends always putting spaces around ``:=``, similar to
...
+requirement, as function calls may already have side effects.  Python
+already has a rule that subexpressions are generally evaluated from
+left to right.  However, assignment expressions make these side
+effects more visible, and we propose a single change to the current
+evaluation order:
+
+- In a dict comprehension ``{X: Y for ...}``, ``Y`` is currently
+  evaluated before ``X``.  We propose to change this so that ``X`` is
+  evaluated before ``Y``.  (In a dict display like ``{X: Y}}`` this is
+  already the case.)
```
tim-one: Perhaps add ", and also in \`\``dict((X, Y) for ..`\`\` which should clearly be equivalent to the dict comprehension`.

PR #711, opened 2018-07-09
PEP 572: Explain why 'as TARGET' is inferior

gvanrossum:
Also explain what's wrong with `->`.

### DIFF
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -620,6 +620,10 @@ Broadly the same semantics as the current proposal, but spelled differently.
    would create unnecessary confusion or require special-casing
    (e.g. to forbid assignment within the headers of these statements).
 
+   (Note that ``with EXPR as VAR`` does *not* simply assing the value
+   of ``EXPR`` to ``VAR`` -- it calls ``EXPR.__enter__()`` and assigns
+   the result of *that* to ``VAR``.)
+
    Additional reasons to prefer ``:=`` over this spelling include:
 
    - In ``if f(x) as y`` the assignment target doesn't jump out at you
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -615,12 +615,36 @@ Broadly the same semantics as the current proposal, but spelled differently.
 
        stuff = [[f(x) as y, x/y] for x in range(5)]
 
-   Since ``EXPR as NAME`` already has meaning in ``except`` and ``with``
-   statements (with different semantics), this would create unnecessary
-   confusion or require special-casing (eg to forbid assignment within the
-   headers of these statements).
+   Since ``EXPR as NAME`` already has meaning in ``import``,
+   ``except`` and ``with`` statements (with different semantics), this
+   would create unnecessary confusion or require special-casing
+   (e.g. to forbid assignment within the headers of these statements).
 
-   TODO: Explain more why this is bad, since this keeps coming up (it's also readability)
+   Additional reasons to prefer ``:=`` over this spelling include:
+
+   - In ``if f(x) as y`` the assignment target doesn't jump out at you
+     -- it just reads like ``if f x blah blah`` and it is too similar
+     visually to ``if f(x) and y``.
+
+   - In all other situations where an ``as`` clause is allowed, even
+     readers with intermediary skills are led to anticipate that
+     clause (however optional) by the keyword that starts the line,
+     and the grammar ties that keyword closely to the as clause:
+
+     - ``import foo as bar``
+     - ``except Exc as var``
+     - ``with ctxmgr() as var``
+
+     To the contrary, the assignment expression does not belong to the
+     ``if`` or ``while`` that starts the line, and we intentionally
+     allow assignment expressions in other contexts as well.
+
+   - The parallel cadence between
+
+     - ``NAME = EXPR``
+     - ``if NAME := EXPR``
+
+     reinforces the visual recognition of assignment expressions.
 
 2. ``EXPR -> NAME``::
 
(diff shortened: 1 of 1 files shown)


Rosuav:
Yep. May also be worth mentioning that 'with... as' is NOT simply binding the expression to its left (nor is 'except.. as', but less subtly), so that syntax is dangerous. Otherwise, LGTM.

gvanrossum:
@Rosuav 
> May also be worth mentioning that 'with... as' is NOT simply binding the expression to its left

It already says "with different semantics" (I think you wrote that originally). I don't feel the need to reiterate the well-known semantics of `with ... as` here. If people want to argue on the basis of their own misunderstanding about `with ... as` then that's their problem.

tim-one:
I agree more words about `with` aren't needed, but, to be fair to Chris, it was obvious during many threads that almost nobody knew that `with f() as name` didn't simply bind `name` to the result of `f()`.  Which kinda makes it our problem too 😉

gvanrossum:
Okay, I'll consider adding a brief explanation (tomorrow).

FWIW we're nearing the end of the tunnel. Apart from this I'd like to review the "Steve Dower thread" and then I'm ready to do another post to python-dev. I also realize that I never posted the major update from May 22 (#654) there. In retrospect that was a mistake.

And @Rosuav -- can you explain why you'd rather not see any of the exceptional cases?  Apart from the first one I am not wedded to them, as they could be ruled out as style guides. I do feel strongly about not allowing `a := 1` as an alternative to `a = 1` at the top level -- we don't need two ways to do that.

pep-0572.rst
```diff
@@ -615,12 +615,36 @@ Broadly the same semantics as the current proposal, but spelled differently.
...
+     To the contrary, the assignment expression does not belong to the
+     ``if`` or ``while`` that starts the line, and we intentionally
+     allow assignment expressions in other contexts as well.
+
+   - The parallel cadence between
+
+     - ``NAME = EXPR``
+     - ``if NAME := EXPR``
+
+     reinforces the visual recognition of assignment expressions.
```
tim-one: And there's a rhyming cadence in that `for` loop headers also put the targets to the left and the values to the right.

pep-0572.rst
```diff
@@ -630,8 +654,10 @@ Broadly the same semantics as the current proposal, but spelled differently.
    programmable calculators. (Note that a left-facing arrow ``y <- f(x)`` is
    not possible in Python, as it would be interpreted as less-than and unary
    minus.) This syntax has a slight advantage over 'as' in that it does not
-   conflict with ``with`` and ``except`` statements, but otherwise is
-   equivalent.
+   conflict with ``with``, ``except`` and ``import``, but otherwise is
+   equivalent.  But it is entirely unrelated to Python's other use of
+   ``->`` (function return type annotations), and compared to ``:=``
+   (which dates back to Algol-58) it has a much weaker tradition.
```
tim-one: I'm not sure `->` had enough advocates to be worth the bother of "refuting" - but since you already did, keep it 😄

pep-0572.rst
```diff
@@ -615,12 +615,36 @@ Broadly the same semantics as the current proposal, but spelled differently.
...
+   ``except`` and ``with`` statements (with different semantics), this
+   would create unnecessary confusion or require special-casing
+   (e.g. to forbid assignment within the headers of these statements).
 
-   TODO: Explain more why this is bad, since this keeps coming up (it's also readability)
+   Additional reasons to prefer ``:=`` over this spelling include:
+
+   - In ``if f(x) as y`` the assignment target doesn't jump out at you
+     -- it just reads like ``if f x blah blah`` and it is too similar
+     visually to ``if f(x) and y``.
```
tim-one: Huh!  I wrote a longish comment here which appears to have wholly vanished ☹️ 

Didn't ask for any changes.  Short course:  many complained that the PEP was destroying that Python "reads like English", as if punctuation didn't play a vital role in written English too.  Your point here is, to me, very strong.  `:=` prevents the text from looking like "uniform identifier salad", driving the eye to that something "important" is going on.

Just FYI, while Nick's `given` reads better to me in _some_ cases, one of its advocates  (not Nick) really soured me on it by trying to sell this as an "excellent" use:

    if some_condition and stuff.lower().strip().replace('-', '') == existing_value given stuff = get_user_input():

"Uniform identifier salad"!  I had to scan the line 3 times before I even noticed the crucial `given`, which, because of the underscore, "looks like" just the third of four identifiers in a row.  It's a Good Thing that `:=` doesn't look like an identifier!

Rosuav:
I mainly dislike special cases. "It could be ugly" feels like a slim reason to outright forbid something; feels like that should be the job of style guides, not grammar. I understand that there are going to need to be SOME forbiddances, but IMO the fewer the better.

gvanrossum:
> I mainly dislike special cases. [...]

It so happens that all these special cases come down to forbidding `X = Y := Z`, constructs that have both `=` and `:=` at the same level.

This is also a corner of the syntax where the *actual* grammar used by our much abused LL(1) parser generator comes down to `EXPR ['=' EXPR]` even though what we want to accept is of the form `[VAR =] EXPR`, since the LL(1) parser wants to know what path through the grammar to take from looking at a single token:

When the parsing engine sees the token `a` at the start of the line, it doesn't know whether to launch into the grammar alternative for `EXPR` or the one for `VAR = EXPR`, and yet (due to the restrictions of LL(1) parsers) it needs to know. This is an ambiguity in the grammar, and Python's pgen has no way to resolve it. So we write the rule as `EXPR ['=' EXPR]`, and in a post-processing pass we check that if there's an `'='` token, the parse tree to the left is a valid `VAR` (e.g. it can be an identifier or an attribute reference but it can't be a function call or an operator).

For keyword args we do similar things but the thing on the left of the `'='` token is only allowed to be a simple `NAME`.

Now, in general for assignment expressions we have to use a similar trick -- again, accept `EXPR [':=' EXPR]` and then later check that if there is an `:=` the `EXPR` to its left is actually a simple (unparenthesized) `NAME`. While I'm sure we *could* come up with a grammar and a separate check for valid subtrees based on the operators found, I *really* would rather be limited to a single special operator per tree. (The worst case is actually function call arguments, where we want to allow `NAME = EXPR` as well as `NAME := EXPR`, and, of course `EXPR`.)

PS. I didn't use the names from the [actual grammar](https://github.com/python/cpython/blob/master/Grammar/Grammar) since those are kind of historic (and differ from what we publish in the reference manual).

Rosuav:
Cool. So we're down to one special case that is actually a real point of confusion. :+1:

Issue #706, opened 2018-07-08
Please always mention the PEP number in PR/commit subject lines


gvanrossum:
I'd like to introduce a convention that PRs (and direct commits) for this repo should always have the PEP number in their subject line, with the exception of changes that affect the infrastructure (e.g. pep2html.py). I'd recommend the format "PEP NNN: <description of change>" but it's also okay to use e.g. "Fix typos in PEP NNN".

What's a good place to put such a convention?

Mariatta:
It can be added to a .github/contributing.md file that will become a [contributing guideline](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) for the peps repo.

PR #712, opened 2018-07-09
Add PEPs Contributing Guideline, CoC, and PR template

Mariatta:
Closes https://github.com/python/peps/issues/706


.github/contributing.rst
```diff
@@ -0,0 +1,13 @@
+Contributing Guideline
+======================
+
+To learn more about the purpose of PEPs and how to go about writing a PEP, please
+start reading at PEP 1 (pep-0001.txt in this repo). Note that PEP 0, the index PEP,
+is now automatically generated, and not committed to the repo.
```
gvanrossum: I'd like to see a (strong) suggestion added here to first discuss an idea on python-ideas before attempting to write a PEP.

Mariatta:
No prob!
I've kept COC and PR template in .github directory, but moved CONTRIBUTING.rst to the root level.
I also linked it from README.rst.

Mariatta:
And if we find people continuing to forgot about the PEP number, I can write a bot to check for it 😄

gvanrossum:
What's the reason for keeping the other two in the .github subdir? In mypy we have ISSUE_TEMPLATE.md at the top level and it works fine. (I guess I am just in disagreement with GitHub's recommendation to hide these files from view.)

Mariatta:
The templates will work in both root level or inside .github.
I just wanted to keep the root level clean of unnecessary things.

In my mind, the templates are really not relevant unless you're on GitHub web UI, opening an issue or PR, so I think these should remain in .github folder.

But I can move the CoC to the root.

gvanrossum:
Sorry. I see (for the first time now :-) that the cpython repo has a .github directory, so I withdraw my objection. I do think that the COC (here as well as there) belongs at the top level, since I would hope that people using the repo without using GitHub still ought to be reminded of the COC.

Mariatta:
Moved Code_of_conduct.rst to root level.
I saw that CPython's [contributing guideline](https://github.com/python/cpython/blob/master/.github/CONTRIBUTING.rst#code-of-conduct ) also mentions CoC, so I copied the text here too.

PR #717, opened 2018-07-10
PEP 572: Delete objection about ugly code

gvanrossum:
See https://twitter.com/asmeurer/status/1016522016160022528

@Rosuav This was your paragraph IIRC -- do you mind if we delete it?


tim-one:
I'm betting it felt really good to write at the time 😉  But, ya, I agree the PEP would read better without it now.

PR #718, opened 2018-07-10
PEP 572: Tweaks suggested by Glenn Linterman

gvanrossum:


pep-0572.rst
```diff
@@ -225,11 +225,11 @@ comprehension, for example::
 
 An exception to this special case applies when the target name is the
 same as a loop control variable for a comprehension containing it.
-This is invalid.  This exception exists to rule out edge cases of the
+This exception exists to rule out edge cases of the
 above scope rules as illustrated by ``[i := i+1 for i in range(5)]``
 or ``[[(j := j) for i in range(5)] for j in range(5)]``.  Note that
 this exception also applies to ``[i := 0 for i, j in stuff]``, as well
-as to cases like ``[i+1 for i in i := stuff]``.
+as to cases like ``[i+1 for i in i := stuff]``.  Such code is invalid.
```
tim-one: Eh?  I'm having a real problem with adding comments that just seem to vanish ☹️ 

So, again, pasting in a comment from my recent email:

Rather than:

"""
An exception to this special case applies when the target name is the
same as a loop control variable for a comprehension containing it. ...
"""

I think I'd prefer:

"""
However, an assignment expression target name cannot be the same as a ``for``-target name appearing in any comprehension containing the assignment expression.  The latter names are local to the comprehension in which they appear, so it would be contradictory for a contained use of the same name to refer to the scope containing the outermost comprehension instead.

For example, ``[i := i+1 for i in range(5)]`` is invalid:  the ``for i`` part establishes that ``i`` is local to the comprehension, but the ``i :=`` part insists that ``i`` is not local to the comprehension.  The same reason makes these examples invalid too:

``[[(j := j) for i in range(5)] for j in range(5)]``
``[i := 0 for i, j in stuff]``
``[i+1 for i in i := stuff]``
"""

PR #719, opened 2018-07-10
PEP 572: Assorted fixes suggested by Rob Cliffe

gvanrossum:


pep-0572.rst
```diff
@@ -126,11 +130,14 @@ that value::
...
     if (match := pattern.search(data)) is not None:
-        ...
+        # Do something with match
 
-    # A more explicit alternative to the 2-arg form of iter() invocation
-    while (value := read_next_item()) is not None:
-        ...
+    # A loop that can't be trivially rewritten using 2-arg iter()
+    while (value := read_next_item(file)) is not None:
+        # Do something with value
```
tim-one: Ick - this example seems wholly contrived.  How about keeping the original `# A more explicit ...` comment but changing the example to match it?

    Before:
        for line in iter(somefile.readline, ""):
            process(line)
    After:
        while line := somefile.readline():
            process(line)

Maybe that just especially appeals to me because I can never remember what 2-arg `iter()` does ☹️

pep-0572.rst
```diff
@@ -255,7 +262,8 @@ Relative precedence of ``:=``
 
 The ``:=`` operator groups more tightly than a comma in all syntactic
 positions where it is legal, but less tightly than all operators,
```
tim-one: s/all operators/all other operators/ ?  That change would seem to be in the spirit of the OP's complaints.

Rosuav: According to the table at https://docs.python.org/3/reference/expressions.html#operator-precedence the comma is not an operator in Python. It's always a syntactic construct. I suppose you could argue that it's an operator taking two or more operands, and returns a tuple containing them, but that's pushing the definition a bit.

tim-one: Right, comma isn't "really" an operator, but pretending otherwise is common in explanatory text.  The idea is that the RHS of `:=` goes on until you hit a comma or unmatched right paren/bracket/brace.  The fiction is useful because specifying the actual grammar is full of details that - while necessary for the implementation - aren't really illuminating.

Rosuav: Okay. Then let's go with "other"; it'll be technically correct in that "all other operators" means "everything except `:=`", and colloquially correct in that it means "everything except the comma".

pep-0572.rst
ncoghlan: The problem with file based examples is that they typically collapse down to `for line in somefile: ...` in real code. It really is only in cases where native iteration support isn't available for some reason that you need to resort to the loop-and-a-half pattern instead.

giampaolo: Another real-world example involving files is:

‘’’
while chunk := file.read(8192):
     process(chunk)
‘’’

tim-one: @giampaolo, I think that would be an excellent example to replace the one here!  The clumsy 2-arg iter way of spelling that is

    for chunk in iter((lambda: somefile.read(8192)), ""):

pep-0572.rst
gvanrossum: The "all other operators" change is already part of https://github.com/python/peps/pull/718.

pep-0572.rst
gvanrossum: Adding this now.

PR #720, opened 2018-07-10
PEP 572: disallow ungrouped assign-expr in annotations

birkenfeld:
A case equally confusing to default values, I think.

### DIFF
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -182,6 +182,18 @@ in order to avoid ambiguities or user confusion:
   style recommendation against mutable default values), and also to
   echo the similar prohibition in calls (the previous bullet).
 
+- Unparenthesized assignment expressions are prohibited as annotations
+  for arguments, return values and assignments.  Example::
+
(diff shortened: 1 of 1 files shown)


Rosuav:
Is this actually ambiguous, or are you calling fr it to be disallowed simply because it's ugly?

ncoghlan:
It's genuinely ambiguous, because it's hard to tell the difference between:

    answer: p := expr1 = expr2 # Assigns `expr1` to `p`, assigns `expr2` to `answer`
    answer: p := expr1 == expr2 # Assigns `expr1 == expr2` to both `p` and `answer`

In any case where `LHS = RHS` is permitted by the language grammar (which I believe is only in annotations), then a top-level assignment expression should be prohibited in `LHS` to avoid the `=`/`==` ambiguity.

gvanrossum:
Good catch. This totally should be prohibited -- also for the same (unmentioned) reason that the the poor LL(1) parser would have to bend over backwards to allow it. Will merge now.

PR #726, opened 2018-07-10
Linkify the PEP numbers in "Superseded-By" header.

Mariatta:
We're already linking PEP numbers in "Replaces:", and "Requires:".

Adjusted the code so it can link "Superseded-By:".

Add the "Replaces:" header where they're missing.

Closes https://github.com/python/peps/issues/477

### DIFF
--- a/pep2html.py
+++ b/pep2html.py
@@ -235,7 +235,7 @@ def fixfile(inpath, input_lines, outfile):
                 else:
                     mailtos.append(part)
             v = COMMASPACE.join(mailtos)
-        elif k.lower() in ('replaces', 'replaced-by', 'requires'):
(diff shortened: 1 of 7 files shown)


pep2html.py
```diff
@@ -235,7 +235,7 @@ def fixfile(inpath, input_lines, outfile):
                 else:
                     mailtos.append(part)
             v = COMMASPACE.join(mailtos)
-        elif k.lower() in ('replaces', 'replaced-by', 'requires'):
+        elif k.lower() in ('replaces', 'superseded-by', 'requires'):
```
Mariatta: There's no "replaced-by" header anywhere.

PR #727, opened 2018-07-10
Update .travis.yml, build using Python 3.7-dev instead of 3.5

Mariatta:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
 python:
-    - 3.5
+    - "3.7-dev"
```
JelleZijlstra: Why dev and not just 3.7?

Mariatta: Living on the edge! 😄 
I just copy 🍝 from [core-workflow](https://github.com/python/core-workflow/blob/master/.travis.yml), [bedevere](https://github.com/python/bedevere/blob/master/.travis.yml), and [the-knights-who-say-ni](https://github.com/python/the-knights-who-say-ni/blob/master/.travis.yml), all of them using "-dev".

JelleZijlstra: I suppose those were written before 3.7.0 was released. I don't think it matters a lot but it seems safer to use the released version. :)

Mariatta: 3.7 on travis is only available using `dist: xenial`, and it is not yet officially documented [here](https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions).

JelleZijlstra: Makes sense, thanks for looking into this!

pradyunsg: FWIW, it seems the 3.7-dev version is actually a beta or something similar and not the CPython's current master branch.

None the less, IIRC, we didn't have any major changes made in the rc period anyway. :)

PR #729, opened 2018-07-10
PEP 572: Add another UAEAP bullet point

Rosuav:
(This section would gzip very efficiently. "Unparenthesized assignment
expressions are prohibited...")

### DIFF
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -201,6 +201,19 @@ in order to avoid ambiguities or user confusion:
   ungrouped assortment of symbols and operators composed of ``:`` and
   ``=`` is hard to read correctly.
 
+- Unparenthesized assignment expressions are prohibited in lambda functions.
+  Example::
+
+    (lambda: x := 1) # INVALID
+    lambda: (x := 1) # Valid, but unlikely to be useful
+    (x := lambda: 1) # Valid
+
+  This allows ``lambda`` to always bind less tightly than ``:=``; creating
+  a name binding inside the lambda function is unlikely to be of value, as
+  there is no way to make use of it. In cases where the name will be used
+  more than once, the expression is likely to need parenthesizing anyway,
+  so this prohibition will rarely affect code.
+
 Scope of the target
 -------------------
 
--- a/pep-0572.rst
+++ b/pep-0572.rst
@@ -207,11 +207,12 @@ in order to avoid ambiguities or user confusion:
     (lambda: x := 1) # INVALID
     lambda: (x := 1) # Valid, but unlikely to be useful
     (x := lambda: 1) # Valid
+    lambda line: (m := re.match(pattern, line)) and m.group(1) # Valid
 
-  This allows ``lambda`` to always bind less tightly than ``:=``; creating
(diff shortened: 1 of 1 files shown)


pep-0572.rst
```diff
@@ -201,6 +201,19 @@ in order to avoid ambiguities or user confusion:
   ungrouped assortment of symbols and operators composed of ``:`` and
   ``=`` is hard to read correctly.
 
+- Unparenthesized assignment expressions are prohibited in lambda functions.
+  Example::
+
+    (lambda: x := 1) # INVALID
+    lambda: (x := 1) # Valid, but unlikely to be useful
```
gvanrossum: Hm. I'd much rather show an example here that is actually useful (and not obviously terrible style). E.g.
```
lambda line: (m := re.match(pattern, line)) and m.group(1)
```
Such a lambda might be useful (or be the starting point for something useful) with `sorted(lines, key=...)`.

pep-0572.rst
```diff
@@ -201,6 +201,19 @@ in order to avoid ambiguities or user confusion:
...
+    (x := lambda: 1) # Valid
+
+  This allows ``lambda`` to always bind less tightly than ``:=``; creating
+  a name binding inside the lambda function is unlikely to be of value, as
+  there is no way to make use of it. In cases where the name will be used
```
gvanrossum: "creating a name binding inside the lambda function" ->
"having a name binding at the top level inside a lambda function"

Rosuav: Sure

pep-0572.rst
JelleZijlstra: That one wouldn't work in Python 3 because None and `str` are not comparable (`None < str` throws an error). Something like `lambda line: m.group(1) if m := re.match(pattern, line) else ''` would work I think.

Rosuav: I'll keep the super-simple one, but also add the more useful one.

It won't really work as a sort key function, since None isn't comparable. Unfortunately, the `if/else` variant reads somewhat badly:

    lambda line: m.group(1) if m := re.match(pattern, line) else ''

(it's confusingly similar to a comparison), and I'm not even sure how I'd write a bool-tuple version as a clean one-liner.

Rosuav: Added.

gvanrossum:
AFUAEAPBP? :-)

pep-0572.rst
tim-one: I didn't ignore this.  In the back of my head, I've been trying all day to dream up a realistic example of using an assignment expression in a lambda that didn't plain suck.  The problem:  if a lambda is fancy enough to benefit from one, it's already so fancy that I'd almost certainly write it as a `def` instead.  This is the closest I got to a legitimate use, and it's still on the "use a `def`" side to me:

```py
lowest_terms = lambda n, d: (n // (g := gcd(n, d)), d // g)
```

gvanrossum: Agreed. I think here we're just looking to define the edge cases of the grammar precisely, and future generations may find a use for what the grammar allows. (As with so many other grammar edge cases.)

PR #732, opened 2018-07-10
PEP 572:  clarify Appendix C example

tim-one:
PEP 572: Make very clear in the Appendix C example that the function local remains unbound until the caller executes the genexp.


tim-one:
Ya, I looked at the log for the failing build and 572 built fine - the failure was "out of memory" long after it was done.

PR #733, opened 2018-07-10
PEP 572: pass mandatory bufsize sock.recv() argument

giampaolo:


pep-0572.rst
```diff
@@ -571,8 +571,8 @@ an ``if`` or ``while`` statement::
         print("Fallback found:", match.group(0))
 
     # Reading socket data until an empty string is returned
-    while data := sock.recv():
-        print("Received data:", data)
+    while data := sock.recv(8192):
+        print("Received data: ", data)
```
gvanrossum: The space you added is not needed -- print() already separates its arguments with a space.

gvanrossum:
Thanks again!

PR #740, opened 2018-07-13
PEP 541: document issue tracker

ssbarnea:
Adds information about the location of the issue tracker used for processing ownership change requests.

### DIFF
--- a/pep-0541.txt
+++ b/pep-0541.txt
@@ -109,6 +109,7 @@ following means of contact:
 
 The maintainers stop trying to reach the user after six weeks.
 
+Ownership change requests are to be raised at [9]_
 
 Abandoned projects
 ------------------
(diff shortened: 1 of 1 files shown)


ssbarnea:
@zxiiro

pep-0541.txt
```diff
@@ -109,6 +109,7 @@ following means of contact:
 
 The maintainers stop trying to reach the user after six weeks.
 
+Ownership change requests are to be raised at [9]_
```
ambv: This should be moved under "Removal of an abandoned project" instead. Please keep two newlines at the end of the section.

pradyunsg:
I don't think there's a decision on this yet. As per my understanding, the support process is still being discussed: https://github.com/pypa/warehouse/issues/1506#issuecomment-375736028.

PR #742, opened 2018-07-16
PEP 574: change buffer callback semantics

pitrou:
`buffer_callback` now takes a single buffer argument and can return a true-ish value if the buffer is to be considered in-band.


pitrou:
@ncoghlan

PR #746, opened 2018-07-21
Multiple PEPs: Fix dead links for issues

emilyemorehouse:
`http://www.python.org/sf/<issue-num>` or `http://python.org/sf/<issue-num>` redirects to `https://legacy.python.org/sf/<issue-num>`. This returns a 403 as the legacy site has been partially phased out.

Any reference to `http://www.python.org/sf/<issue-num>` or `http://python.org/sf/<issue-num>` has been changed to `https://bugs.python.org/issue<issue-num>`


Mariatta:
Thanks!

PR #769, opened 2018-09-11
PEP-0581: use auto-numbering for footnotes.

ezio-melotti:
Use auto-numbering for footnotes to make it easier to add/remove notes.


Mariatta:
Thanks!

PR #771, opened 2018-09-11
PEP 8002: Fix markup

berkerpeksag:
This broke python.org's PEP builder.


berkerpeksag:
Thanks for the reviews!

PR #775, opened 2018-09-13
PEP 580: Reword section on cc_parent

encukou:
Here is a proposed update about ``cc_parent``.
@jdemeyer, do you agree?

### DIFF
--- a/pep-0580.rst
+++ b/pep-0580.rst
@@ -122,38 +122,35 @@ Parent
 ------
 
 The ``cc_parent`` field (accessed for example by a ``__parent__``
-or ``__objclass__`` descriptor from Python code) can be any Python object.
-For methods of extension types, this is set to the class.
-For functions of modules, this is set to the module.
-However, custom classes are free to set ``cc_parent`` to whatever they want,
-it can also be ``NULL``.
-It is only used by the C call protocol if the ``CCALL_OBJCLASS`` flag is set.
+or ``__objclass__`` descriptor from Python code) can be any Python
+object, or NULL.
+Custom classes are free to set ``cc_parent`` to whatever they want.
+It is only used by the C call protocol if the
+``CCALL_OBJCLASS`` flag is set.
 
-The parent serves multiple purposes: for methods of extension types
-(more precisely, when the flag ``CCALL_OBJCLASS`` is set),
-it is used for type checks like the following::
+For methods of extension types, ``cc_parent`` points to the class
+that defines the method (which may be a superclass of ``type(self)``).
+This is currently is non-trivial to retreive from a method's code.
+(In the future, this can be used to access to the module state via
+the defining class. See the rationale of PEP 573 for details.)
+
+When the flag ``CCALL_OBJCLASS`` is set (as it will be methods of
+extension types), ``cc_parent`` is used for type checks like the following::
 
     >>> list.append({}, "x")
     Traceback (most recent call last):
       File "<stdin>", line 1, in <module>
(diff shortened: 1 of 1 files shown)


pep-0580.rst
```diff
@@ -122,39 +122,36 @@ Parent
...
+It is only used by the C call protocol if the
+``CCALL_OBJCLASS`` flag is set.
 
-The parent serves multiple purposes: for methods of extension types
-(more precisely, when the flag ``CCALL_OBJCLASS`` is set),
-it is used for type checks like the following::
+For methods of extension types, ``cc_parent`` points to the class
+that defines the method (which may be a superclass of ``type(self)``).
+This is currently is non-trivial to retreive from a method's code.
+(In the future, this can be used to access to the module state via
```
jdemeyer: "used to access to the module state" looks like a typo

pep-0580.rst
```diff
@@ -122,39 +122,36 @@ Parent
...
+the defining class. See the rationale of PEP 573 for details.)
+
+When the flag ``CCALL_OBJCLASS`` is set (as it will be methods of
+extension types), ``cc_parent`` is used for type checks like the following::
```
jdemeyer: "as it will be methods of extension types" looks like a typo

pep-0580.rst
```diff
@@ -122,39 +122,36 @@ Parent
...
-which works directly for functions of a module.
-For methods, this works indirectly through the class,
-assuming that the class has a pointer to the module.
-
-The parent would also typically be used to implement ``__qualname__``.
-The new C API function ``PyCCall_GenericGetQualname()`` does exactly that.
-
-**NOTE**: for functions of modules,
-the parent is exactly the same as ``__self__``.
+For functions of modules, `cc_parent` is set to the module.
```
jdemeyer: Minor style issue: I use double {{{``}}} for code since the meaning for that in ReST syntax is unambiguous.

pep-0580.rst
```diff
@@ -122,39 +122,36 @@ Parent
...
 
-The parent serves multiple purposes: for methods of extension types
-(more precisely, when the flag ``CCALL_OBJCLASS`` is set),
-it is used for type checks like the following::
+For methods of extension types, ``cc_parent`` points to the class
+that defines the method (which may be a superclass of ``type(self)``).
+This is currently is non-trivial to retreive from a method's code.
+(In the future, this can be used to access to the module state via
+the defining class. See the rationale of PEP 573 for details.)
+
```
jdemeyer: It would be good to mention here the value of `cc_parent` for functions of a module. It looks strange to me to wait with that until the end of this section.

encukou: I think it's good to complete the "For methods of extension types" discussion with the type check, before diving into "For functions of modules".

jdemeyer: > I think it's good to complete the "For methods of extension types" discussion with the type check

On the other hand, the parts about "methods of extension types" and "functions of a module" both describe the mapping of current functionality onto the new protocol. The part about `CCALL_OBJCLASS` is describing the protocol itself. It feels strange to me to mix that.

Maybe a good solution would be to put the last part about functions of module first, so we get the order:
- functions of modules
- methods of extension types
- `CCALL_OBJCLASS`

jdemeyer:
I don't want to bikeshed this, so feel free to merge it as is or to apply my last comment about the ordering.

encukou:
OK. If you go through this again, feel free to reorder to your liking.

PR #778, opened 2018-09-16
PEP 8013: Completed first draft

zooba:


jdemeyer:
> GUIDO has responsibility for reviewing controversial decisions in the form of PEPs written by members of the core development team.

> GUIDO only pronounces on PEPs submitted to python-committers. There is no expectation that GUIDO follows or participates on any other mailing lists.

Really? So non-core-developers would no longer be allowed to submit PEPs? That would be a *major* change over the current process, where everybody can submit a PEP. In my opinion, such a big change should be made more explicit, preferably with a rationale.

zware:
Agreed that a brief rationale for this change should be added, as it was articulated at the meeting.

For @jdemeyer and others' benefit, my recollection of the rationale is basically that anyone can still author and submit a PEP to all but the final stage of the process, but as part of the vetting process you need buy-in from at least one active committer before it can reach the stage of possibly being accepted -- if you can't convince a single committer to work with you on the idea, it likely isn't ready to be accepted anyway.  This is mostly just a formalization of something that already happened with most PEPs under the former system.

jdemeyer:
> basically that anyone can still author and submit a PEP to all but the final stage of the process, but as part of the vetting process you need buy-in from at least one active committer before it can reach the stage of possibly being accepted

So all discussions about the PEP happen on python-dev or whatever, but the actual final submission happens by a core dev on python-committers? That makes sense, but it isn't how I read the PEP. So yes, this should be clarified.

zooba:
PR #782 has additional wording.

PR #783, opened 2018-09-18
PEP 566: Support for direct references and permit public index servers to add restrictions

pradyunsg:
- Permit direct references as an alternative to version specifiers
- Permit public index servers to prohibit strict matching specifiers
  and direct references

Closes #763 

/cc @di @pfmoore


brettcannon:
Thanks!

PR #786, opened 2018-09-20
[DO-NOT-MERGE][WIP] Remove "Legacy C Locale" section

vstinner:
FreeBSD 11 does not provide a C locale with UTF-8 encoding, but it's
still supported. The UTF-8 Mode is now enabled if the LC_CTYPE locale
is "C" or "POSIX". Moreover, the UTF-8 Mode can also be used on
Windows.

### DIFF
--- a/pep-0011.txt
+++ b/pep-0011.txt
@@ -144,18 +144,6 @@ patches will be accepted. Such build files will be removed from the
 source tree 3 years after the extended support for the compiler has
 ended (but continue to remain available in revision control).
 
-Legacy C Locale
----------------
-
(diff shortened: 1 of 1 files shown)


ncoghlan:
No, the qualification on FreeBSD support described in the PEP holds:

> Any Unicode-related integration problems that occur only in the legacy ``C``
locale and cannot be reproduced in an appropriately configured non-ASCII
locale will be closed as "won't fix".

This is not a supportable configuration, and it WILL be broken in various ways. You can patch as many things in CPython as you like, but you won't fix readline, you won't fix other C/C++ extensions, all of which will still be using ASCII, even when CPython itself switches to UTF-8 mode.

PR #787, opened 2018-09-21
Fix PEP 567 to reflect changes in 3.7.1 (see bpo-34762)

1st1:


1st1:
@elprans please take a look.

PR #797, opened 2018-10-03
PEP 580: Petr Viktorin as BDFL-Delegate

jdemeyer:
See discussion at https://mail.python.org/pipermail/python-dev/2018-October/155403.html

### DIFF
--- a/pep-0580.rst
+++ b/pep-0580.rst
@@ -1,6 +1,7 @@
 PEP: 580
 Title: The C call protocol
 Author: Jeroen Demeyer <J.Demeyer@UGent.be>
+BDFL-Delegate: Petr Viktorin
 Status: Draft
 Type: Standards Track
 Content-Type: text/x-rst


gvanrossum:
Given the lack of a BDFL, was this discussed somewhere? A reference would be appropriate (in the PR, not necessarily in the PEP). (Writing this with my PEP-editor hat on.)

jdemeyer:
> Given the lack of a BDFL, was this discussed somewhere? A reference would be appropriate (in the PR, not necessarily in the PEP). (Writing this with my PEP-editor hat on.)

There is ongoing discussion, I added a link to python-dev. If you think of "BDFL Delegate" as "PEP Reviewer" (which is what it is in practice), it shouldn't be a problem that there is no BDFL. I should also note that I did ask Antoine Pitrou, INADA Naoki and Nick Coghlan (in private emails) as "experts" and they all approved of Petr becoming BDFL-Delegate.

gvanrossum:
But the role of BDFL-Delegate is different from a reviewer. Anyone can review a PEP. But a BDFL-Delegate can *accept* a PEP. I think you're jumping the gun here. Petr should do a thorough review without the BDFL-Delegate appointment. Implementation (in a fork) can start when he says it's ready.

jdemeyer:
@gvanrossum: I'm not entirely following what you're trying to say here. Are you saying that it's too early to appoint a BDFL-Delegate because PEP 580 hasn't been sufficiently looked at? Or are you saying that it's impossible to follow the BDFL-Delegate process from PEP 1 because of the governance crisis? Or both?

encukou:
- @ambv, the release manager for 3.8 (closest we have to an authority), considers this premature. 
- `BDFL-Delegate` is probably the wrong term anyway.

I'll close this for now.

PR #799, opened 2018-10-06
PEP 544: Make Guido the BDFL-Delegate (Protocols)

gvanrossum:
See recent thread "PEP 544 status (forked off "Petr Viktorin as BDFL-Delegate for PEP 580")" in python-dev.


Rosuav:
LGTM. Merging per discussion.

PR #807, opened 2018-10-12
PEP 561: Clarify ordering of module resolution

ethanhs:
As brought up in https://github.com/python/mypy/issues/5520.

cc @gvanrossum


sproshev:
@ethanhs If package is distributed with stubs (e.g. `attr`: `__init__.py` and `__init__.pyi`), what category should be used from `Module Resolution Order` for it?

gvanrossum:
For any `.py` file, including `__init__.py`, if there is a corresponding `.pyi` file in the same directory, that `.pyi` is used exclusively.

PR #812, opened 2018-10-23
PEP 8015: spelling conventions and errors

danishprakash:


gvanrossum:
Thanks!

pep-8015.rst
```diff
@@ -314,15 +314,15 @@ Special Case: Ban a core developer
 
 As any other member of the Python community, the PSF Code of Conduct
 Workgroup can ban a core developer for a limited amount of time. In this
-case, the core developer immediately looses their core developer status.
+case, the core developer immediately loses their core developer status.
 
 In general, a ban is only the last resort action when all other options
-have been exausted.
+have been exhausted.
```
vstinner: That's strange. I already fixed this typo (a friend already reported it to me), but it seems like I pushed to the wrong Git repository...

pep-8015.rst
```diff
@@ -314,15 +314,15 @@ Special Case: Ban a core developer
...
 
 At the end of the ban, the developer is allowed to contribute again as a
 regular contributor.
 
-If the developer changes their behaviour, another core developer can
+If the developer changes their behavior, another core developer can
```
vstinner: Oh, I always hesitated between the Britain and American English spelling... It seems like the Americain English spelling won :-)

danishprakash:
@gvanrossum your welcome!


> Thanks! I'm not a native English speaker,

@vstinner Neither am I but these stood out while reading. Tbh, it's one of the most interesting and well-thought out PEP I've read recently, great job. Has the voting already started for this?

pep-8015.rst
danishprakash: Honestly, I'm not even sure why I did that, maybe because Python is more closely associated with the American subcontinent but then again, this is debatable.

PR #827, opened 2018-11-02
PEP 8016 - initial draft

njsmith:
Incomplete – please don't merge yet. For discussion, see:

  https://discuss.python.org/t/working-discussion-for-pep-8016-the-boringest-possible-steering-council-model/


njsmith:
Okay, I think this is a pretty solid first draft. @dstufft, want to do the honors and hit merge? :-)

PR #830, opened 2018-11-03
PEP 8001: Use CVIS instead of a github repository

dstufft:
Suggested change to use CVIS instead of a github repository to enable private ballots.

### DIFF
--- a/pep-8001.rst
+++ b/pep-8001.rst
@@ -77,14 +77,43 @@ to November 30 (Anywhere-on-Earth).
 Where is the vote?
 ------------------
 
-The vote will happen through a private GitHub repository named
-"vote-governance" within the Python organization.  Every committer
-will push a single file named after their GitHub username.  Inside the
-file, preferences should be stated one PEP per line.
+The vote will happen using a "private" poll on the
+`Condorcet Internet Voting Service <https://civs.cs.cornell.edu/>`_. Every committer
+will recieve an email with a link allowing them to rank the PEPs in their order of
+preference.
 
-Committers are allowed to change their vote while voting is open.
+The election wil be supervised by Ernest Durbin.
 
-The repository will be made read-only and public on December 1st.
+The results of the election, including anonymized ballots, will be made public on
+December 1st, after the election has closed.
+
+The following settings will be used for the vote in the CIVS system::
+
+    [x] Private
+    [ ] Make this a test poll: read all votes from a file.
+    [ ] Do not release results to all voters.
+    [x] Enable detailed ballot reporting.
+        [ ] In detailed ballot report, also reveal the identity of the voter with each ballot.
+    [ ] Allow voters to write in new choices.
+    [ ] Present choices on voting page in exactly the given order.
+    [ ] Allow voters to select “no opinion” for some choices.
+    [ ] Enforce proportional representation
+
+Which will have the effect of:
+
+* Making the election "private", or in other words, invite only.
+* The results of the election will be released to all voters.
+* The contents of every ballot will be released to the public, along
+  with a detailed report going over how the winner was elected.
+* The detailed ballots will *not* include any identifying information
+  and the email addresses of the voters will be thrown away by the CIVS
+  system as soon as the email with their voting link has been sent.
+* Voters will *not* be able to write in new choices, and will be limited
+  only to the options specified in the election.
+* The default ordering for each ballot will be randomized to remove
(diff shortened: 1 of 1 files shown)


ambv:
I understand why you'd like the vote to be private *while* it happens but why insist on anonymizing identity of voters after the vote?

It's open source, we are supposed to work in the open. Being able to see who voted enables seeing what the voter population looks like. In other words, what pockets of votes are there and who did not vote at all. It also makes it obvious if somebody who shouldn't vote, does. Finally, it allows me to confirm that my own vote was recorded right.

dstufft:
@ambv You may want to ask this question on either python-committers (the long thread) or [Discourse](https://discuss.python.org/t/pep-8001-public-or-private-ballots/374). On python-committers three people (myself, Barry Warsaw, and Stefan Krah) expressed concerns about having public ballots, but said that we would still be voting, while Tim Peters said he planned on abstaining from the vote completely due to it being public. In the thread in discourse Mariatta has said that she would much prefer it if they vote was private, but would still be voting if it was public. Three additional people favorited my post proposing a private ballot, but it's hard to determine if they just liked the extensiveness of the argument or some other part of the post, or if they felt similarly about a private ballot.

I can't answer for Barry, Stefan, or Tim but my own personal reason is that I genuinely feel uncomfortable with a public ballot. A secret ballot allows every individual to vote how they want, without worrying about any potential fallout from the way they voted. Now I don't specifically think that say, rank PEP 8012 low that you're going to go out of your way to make my life miserable. But let's say I rank 8010 low but it ultimately wins. Will that hurt my own chances of being able to become "GUIDO" if I so desire it? Effectively, a secret ballot removes all extrinsic motivations for voting one way or another, and allows people to vote their true preferences. One can argue that we don't think this election is likely going to result in any sort of extrinsic pressure, but I don't think that is even possible (the mildest form of extrinsic pressure is just worrying that a friend of yours might be upset or hurt if you rank their proposal lower).

I don't believe that a secret ballot somehow runs counter to the idea of working in the open. The discussion was held in the open, the entire process was as well, including the full contents of every ballot. The only thing that has been kept secret is who specifically voted for what. In a discussion if someone is worried about speaking out, they can try to launder their thoughts through a trusted friend, but you can't do anything like that with voting.

I'm also not at all worried about knowing who voted and who didn't (honestly, if someone chooses not to vote, are they going to get blowback? Who knows!). I also don't feel like people are going to vote if they are not allowed to. While extrinsic pressures can be implicit and happen subconsciously, voting when you weren't eligible to vote is a very overt act of malevolence that I don't believe anyone involved is going to behave in that way. 

It's true that the specific implementation of private ballot that this PR implements does not allow you to confirm that your own vote was recorded correctly while a novel encryption schemes generally would. We unfortunately don't have a novel encryption scheme handy. It is a downside to a public ballot, but I think it is a pretty small one, since I don't believe that CIVS is going to try to secretly undermine our election.

It feels to me like there is a fair amount of support amongst the core developers for using a private ballot and I don't see much of a desire for public ballots. That could simply mean that people who want a public ballot have checked out of the discussion or are generally being quiet because the PEP is currently written the way they prefer, I dunno. I think it's a least important to consider the fact that we have 4 people uncomfortable but still voting, 1 person completely abstaining, and possibly 3 other people uncomfortable all because of the use of a public ballot and whether that means that public ballot is not the right way to go here, and whether the use of a public ballot is going to meaningfully effect the outcome of the election (remembering that if someone doesn't feel comfortable posting their ballot public, they may not feel comfortable  arguing against the PEPs public ballot at all (or feeling like it won't have any effect), and may simply choose to not vote).

tim-one:
I haven't said I won't vote if the ballot is public, just that I probably won't.  That's how I'm _leaning_.

Voting for a government is very different from voting on a technical proposal, and that's very broadly accepted.  That's why, e.g., I don't believe anyone has ever suggested that votes for the PSF Board be wholly public - or in just about any other organization.  Contrarily, nobody ever asks that +1/0/-1 polls on technical issues be made private.

Voting for a governance _structure_ isn't wholly in either camp, which is why I'm only "leaning" so far.  It's close _enough_ to "pure politics" that it makes me queasy.  If, e.g., the PSF membership decided to vote for its Official Religion, I'd resist making that a public vote too :wink:.

dstufft:
Ah sorry, that discussion happened when I was tired and I misremembered how strongly you felt. My bad! Thanks for clarifying.

PR #849, opened 2018-12-07
Check in the memory model PEP.

benjaminp:
This old draft PEP came up in an in-person conversation. It turns out to be almost vanished from the internet; the only source of it I could find is on Launchpad: https://code.launchpad.net/~jyasskin/python/memory-model-pep So, I'm assigning it a number, marking it Withdrawn, and checking it in so as not to lose a historical record.


jyasskin:
Wow, hi old memories. :)

PR #850, opened 2018-12-07
PEP 3157: an old PEP: Concurrency Memory Model by Jeffrey Yasskin

gvanrossum:
@benjamin, @msullivan and I were waxing nostalgically about this and realized that
it was never submitted as a PEP draft.  Even if it's just meant to be
informational and withdrawn, I think it would be nice to have as a
historical document, and possibly as a starting point for something
more modern.

I propose to use a number in the 3000 range because it was written
during the heady Py3k days (last-modified says 2007-06-28).

Retrieved from:
https://bazaar.launchpad.net/~jyasskin/python/memory-model-pep/view/head:/pep-memory-model.txt

@jyasskin Are you okay with this? Did you ever sign the CLA?

### DIFF
--- a/pep-3157.txt
+++ b/pep-3157.txt
@@ -2,7 +2,7 @@ PEP: 3157
 Title: A Concurrency Memory Model for Python
 Version: $Revision$
 Last-Modified: $Date$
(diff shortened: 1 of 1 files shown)


gvanrossum:
D'oh.

PR #443, opened 2017-10-27
Pep 508 parser fixes

michaelplaing:
Fix 1 bug. Enhance handling of trailing whitespace for URLs. Correct spelling.

Add tests.

PR #851, opened 2018-12-10
PEP 508: Grammar fixes

doismellburning:
Taken from https://github.com/pypa/interoperability-peps/pull/66 and
https://github.com/python/peps/pull/443

### DIFF
--- a/pep-0508.txt
+++ b/pep-0508.txt
@@ -355,48 +355,48 @@ Complete Grammar
 
 The complete parsley grammar::
 
-    wsp           = ' ' | '\t'
-    version_cmp   = wsp* <'<=' | '<' | '!=' | '==' | '>=' | '>' | '~=' | '==='>
-    version       = wsp* <( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+>
-    version_one   = version_cmp:op version:v wsp* -> (op, v)
-    version_many  = version_one:v1 (wsp* ',' version_one)*:v2 -> [v1] + v2
-    versionspec   = ('(' version_many:v ')' ->v) | version_many
-    urlspec       = '@' wsp* <URI_reference>
-    marker_op     = version_cmp | (wsp* 'in') | (wsp* 'not' wsp+ 'in')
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Thanks again for your contribution, we look forward to reviewing it!

doismellburning:
Is there any precedent/plan for automated testing of "PEP code"? Should there be?

I've taken the liberty of throwing together https://github.com/doismellburning/pep508 because I kept picking up slightly buggy grammar variants but I can't imagine this is the first time something subtle has slipped through...

doismellburning:
(I've signed the CLA, just waiting for it to be received and processed)

doismellburning:
@mindw Would you be able and willing to sign the CLA please?

pep-0508.txt
```diff
@@ -355,48 +355,48 @@ Complete Grammar
...
+    marker_var     = wsp* (env_var | python_str)
+    marker_expr    = marker_var:l marker_op:o marker_var:r -> (o, l, r)
+                   | wsp* '(' marker:m wsp* ')' -> m
+    marker_and     = marker_expr:l wsp* 'and' marker_expr:r -> ('and', l, r)
+                   | marker_expr:m -> m
+    marker_or      = marker_and:l wsp* 'or' marker_and:r -> ('or', l, r)
+                       | marker_and:m -> m
+    marker         = marker_or
+    quoted_marker  = ';' wsp* marker
+    identifier_end = letterOrDigit | (('-' | '_' | '.' )* letterOrDigit)
```
merwok: Could you make this fix without changing spacing in all lines? (for a better history, annotate, etc)

doismellburning: I'm no fan of vertical alignment for similar reasons, but I was aiming to remain consistent with the author's style

I feel like if I were to PR a fix *just* to this line, that would be the worst-of-both-worlds, and that either this PR (retain alignment) or a two-commit approach (the first to break alignment, the second to fix the bug) are both better options - which would you prefer?

merwok: Doing a second commit has the same issues (misleading history etc).

It also wouldn’t work on Python repos as PRs use squash and merge.

PR #854, opened 2018-12-18
Talk about line length instead of window size

andresdelfino:
Avoid unnecessary mixing of line length and window width, as it can be confusing.


andresdelfino:
@warsaw could you take a look at this?

PR #858, opened 2018-12-25
PEP 484: Add @type_check_only

srittau:
Per discussion in python/typing#597.

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -2047,6 +2047,10 @@ Convenience definitions:
 * @no_type_check_decorator, a decorator to create your own decorators
   with the same meaning as ``@no_type_check`` (see below)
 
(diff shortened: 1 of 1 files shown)


srittau:
@gvanrossum @jukkal @ambv as authors.

pep-0484.txt
```diff
@@ -2047,6 +2047,10 @@ Convenience definitions:
 * @no_type_check_decorator, a decorator to create your own decorators
   with the same meaning as ``@no_type_check`` (see below)
 
+* @type_check_only, a decorator only available during type checking
```
brettcannon:
```suggestion
* ``@type_check_only``, a decorator only available during type checking
```

srittau: While this makes sense to me, the other definitions in this list are also not marked for some reason.

PR #860, opened 2018-12-28
PEP 008: modified two typos in the documentation

meysam81:


Rosuav:
The current text is grammatically correct in both cases. The second one could be valid either way, but the first edit definitely should be a singular "is" to go with a singular "someone".

PR #864, opened 2019-01-04
PEP 13: Add info about the limited access to members list

sblondon:
The core team members list is only available for core members. Otherwise a 404 error is displayed so people will try to fix what seems to be a broken link (like me).

This PR is done to close #863

I already signed the CLA to contribute to cpython.

Of course, feel free to rewrite it; english is not my native tongue.


brettcannon:
Thanks!

PR #866, opened 2019-01-07
PEP 8100: Election logistics

ewdurbin:
Adds information on the Voter Roll, which was updated in https://github.com/python/voters/pull/3

Adds information on the Election Implementation.

### DIFF
--- a/pep-8100.rst
+++ b/pep-8100.rst
@@ -50,6 +50,65 @@ Once the nomination period opens, candidates will be listed here:
 1. *Candidate (nominated by)*
 
 
+Voter Roll
+==========
+
+All CPython core developers are eligible to vote. Ballots will be distributed
+based on the `The Python Voter Roll for this election
+<https://github.com/python/voters/blob/master/2019-01-21-2019-python-steering-committee-election-election.csv>`__.
+
+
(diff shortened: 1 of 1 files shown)


pep-8100.rst
```diff
@@ -50,6 +50,65 @@ Once the nomination period opens, candidates will be listed here:
 1. *Candidate (nominated by)*
 
 
+Voter Roll
+==========
+
+All CPython core developers are eligible to vote. Ballots will be distributed
```
njsmith: This is perfectly fine, but I wonder if we want to be even more pedantic and use the language from PEP 8016/13? The technical rules are "All active Python core team members are eligible to vote". Since we haven't added any new members this month, "active core team member" means everyone on the github team (i.e., exactly the same people who could vote in the last election), plus any past core devs who were inactive but decide to mark themselves "active" again.

(Everything else looks great to me.)

ewdurbin: 👍

ewdurbin:
@njsmith, @gpshead brings up a good point [here](https://discuss.python.org/t/2019-python-steering-committee-election-information/619/2), should the template for nominations be updated to include employer?

PR #898, opened 2019-02-23
create package target

ewdurbin:
the peps are built... may as well publish a tarball directly from travis.

python.org hosting is being updated and i'm trying to move it away from local
disk storage, which the current pep integration relies on.

this will allow python.org to pull the tarball out of S3 and do it's work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,12 @@
 pep-0000.txt
 pep-0000.rst
 pep-????.html
+peps.rss
 __pycache__
 *.pyc
 *.pyo
 *~
 *env
 .vscode
+*.swp
+/build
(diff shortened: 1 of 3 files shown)


ewdurbin:
Notes:
  - A user `peps-release-key` has been created in the PSF AWS account with only permissions to perform these uploads.
  - Credential for that user was added to the travis settings as encrypted environment variables.

IAM policy:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "s3:PutObject",
                "s3:PutObjectAcl"
            ],
            "Resource": [
                "arn:aws:s3:::pythondotorg-assets-staging/peps.tar.gz",
                "arn:aws:s3:::pythondotorg-assets/peps.tar.gz"
            ]
        }
    ]
}
```

PR #900, opened 2019-02-24
PEP 219: Fix grammatical mistakes

### LINKED ISSUE #899: spelling error in PEP219 introduction
Am moving an issue from the bug tracker here. This was reported by 景云鹏 here : https://bugs.python.org/issue36090

In the introduction for PEP 219. In the second paragraph here : https://www.python.org/dev/peps/pep-0219/#introduction

It is written "in more that a year" instead of "in more that a year"

Harmon758:
Fixes #899


Rosuav:
LGTM. The third one still feels a bit awkward, but this is an improvement. The other two look fine.

PR #903, opened 2019-03-01
Introduce the concept of sponsors

brettcannon:


pep-0001.txt
```diff
@@ -166,10 +156,22 @@ initial concerns about the proposal.
...
 
-Following a discussion on python-ideas, the proposal should be submitted as a
-draft PEP via a `GitHub pull request`_.  The draft must be written in PEP
-style as described below, else it will fail review immediately (although minor
-errors may be corrected by the editors).
+Following a discussion on python-ideas, the workflow varies based on whether
+the PEP author is a core developer. If the PEP author is **not** a
+core developer then the PEP author will need to find a core developer
+*sponsor* for the PEP. The sponsor's job is to provide guidance to the PEP
+author to help them through the PEP process (somewhat acting like mentor).
```
willingc: Perhaps: ...through the logistics of the PEP process...

jdemeyer:
If "The sponsor's job is to provide guidance to the PEP
author to help them through the PEP process (somewhat acting like mentor)", then wouldn't it better be called "mentor" instead of "sponsor"? To me, the word "sponsor" conveys a somewhat different meaning, namely that the sponsor should also strongly supports the actual PEP.

pep-0001.txt
willingc: Taking into account @jdemeyer's feedback, I would also strike ~~somewhat acting like mentor~~.

It may also be useful to clarify that the sponsor's primary responsibility is walking a contributor through the process of the PEP. The sponsor should feel that the change would be an asset to the language/its development process. Although the sponsor may, it is not required that the sponsor strongly advocate for the PEP.

jdemeyer:
We should just be careful not to make it more difficult for non-core developers to submit a PEP. This should help people to get a PEP drafted instead of being an additional hurdle.

willingc:
> We should just be careful not to make it more difficult for non-core developers to submit a PEP. This should help people to get a PEP drafted instead of being an additional hurdle.

Agreed. The intent is to help non-core developers to navigate the PEP process.

pep-0001.txt
gvanrossum: Hmm... I guess there are different definitions of "sponsor", but the one I've heard is actually one where the sponsor *does* advocate for someone. E.g. https://inclusion.slac.stanford.edu/sites/inclusion.slac.stanford.edu/files/The_Key_Role_of_a_Sponsorship_for_Diverse_Talent.pdf

And honestly I think it a bit weird to expect a core dev to sponsor a PEP that they *do* not care about. It's like reviewing a PR -- if you don't think the PR should be merged, why not just say so rather than start nitpicking about code style or INCREF/DECREF patterns?

There is (or should be) a long process on python-ideas that should happen before a PEP draft should be written, and if there are no core devs interested at that point, a PEP has no chance. And, Jeroen's feedback notwithstanding, I don't think it would be a good use of anybody's time to make it easier for outsiders to propose PEPs -- there just are too many outsiders who lack language design experience but are eager to change Python to be more like their favorite language (be it Haskell or JavaScript :-). *Good* proposals by qualified outsiders (like Jeroen!) will not have problems.

pep-0001.txt
```diff
@@ -166,10 +156,22 @@ initial concerns about the proposal.
...
+For the core developer sponsoring, being a sponsor does **not** disqualify
+them from becoming a co-author or BDFL-Delegate later on. The core developer
```
gvanrossum: (but not both).

pep-0001.txt
```diff
@@ -488,6 +490,7 @@ optional and are described below.  All other headers are required. ::
     PEP: <pep number>
     Title: <pep title>
     Author: <list of authors' real names and optionally, email addrs>
+  * Sponsor: <real name of core developer sponsoring>
```
gvanrossum: Beware that this requires a change to the pep2html.py script.

brettcannon: Yep, that's the next step before merging. 😄 I just wanted to make sure the idea had traction and support before I did the coding work.

pep-0001.txt
brettcannon: So my feeling matches with Guido's: if not a single core developer is willing to help advocate for a PEP then it has no chance of being accepted anyway, so I would expect a sponsor to strongly advocate.

pep-0001.txt
brettcannon: https://github.com/python/peps/pull/911

brettcannon:
Thanks everyone for the reviews! I'll email python-dev shortly about the change.

PR #909, opened 2019-03-04
Re-order and flesh out what belongs in a PEP

brettcannon:


pep-0001.txt
```diff
@@ -411,23 +411,13 @@ Each PEP should have the following parts:
...
-
-4. Specification -- The technical specification should describe the
-   syntax and semantics of any new language feature.  The
-   specification should be detailed enough to allow competing,
-   interoperable implementations for at least the current major Python
-   platforms (CPython, Jython, IronPython, PyPy).
-
-5. Motivation -- The motivation is critical for PEPs that want to
+3. Motivation -- The motivation is critical for PEPs that want to
    change the Python language.  It should clearly explain why the
```
gvanrossum: Maybe "language or library"? Or even "language, library or ecosystem"?

pep-0001.txt
```diff
@@ -411,23 +411,13 @@ Each PEP should have the following parts:
...
    existing language specification is inadequate to address the
    problem that the PEP solves.  PEP submissions without sufficient
    motivation may be rejected outright.
 
-6. Rationale -- The rationale fleshes out the specification by
+4. Rationale -- The rationale fleshes out the specification by
    describing what motivated the design and why particular design
```
gvanrossum: Maybe drop "what motivated the design and"? Since that should already be in the Motivation section.

pep-0001.txt
```diff
@@ -455,15 +460,26 @@ Each PEP should have the following parts:
...
-9. How to Teach This -- For a PEP that adds new functionality or changes
-   language behavior, it is helpful to include a section on how to
-   teach users, new and experienced, how to apply the PEP to their
-   work.
-
-   This section may include key points and recommended documentation
-   changes that would help users adopt a new feature or migrate their
-   code to use a language change.
+   
+9. Rejected Ideas -- Throughout the discussion of a PEP, various ideas
```
gvanrossum: (Aside: this is in a sense part of the rationale -- it contains the rationale for rejecting various alternatives that were brought up. I do agree it is better done as a separate section.)

pep-0001.txt
```diff
@@ -455,15 +460,26 @@ Each PEP should have the following parts:
...
+    so people know that they are being thought about but do not have a
+    concrete resolution. This helps make sure all issues required for the
+    PEP to be ready for consideration are complete as well as prevent people
+    from bringing them up independently.
+    
+11. References -- A collection of URLs used as references through the PEP. 
+   
+12. Copyright/public domain -- Each PEP must either be explicitly
+    labeled as placed in the public domain (see this PEP as an
+    example) or licensed under the `Open Publication License`_.
```
gvanrossum: Should we perhaps make a security section mandatory too? Or at least mention that if there *are* any security concerns they should be collected in a separate section? (Because if most PEPs don't have security issues, requiring everyone to have a section "Security concerns" that says "There are none" is a bit silly.)

pep-0001.txt
```diff
@@ -436,14 +426,29 @@ Each PEP should have the following parts:
...
 
-7. Backwards Compatibility -- All PEPs that introduce backwards
+6. Backwards Compatibility -- All PEPs that introduce backwards
    incompatibilities must include a section describing these
    incompatibilities and their severity.  The PEP must explain how the
    author proposes to deal with these incompatibilities.  PEP
    submissions without a sufficient backwards compatibility treatise
    may be rejected outright.
 
+7. How to Teach This -- For a PEP that adds new functionality or changes
```
willingc: Yes!!!

pep-0001.txt
```diff
@@ -455,15 +460,26 @@ Each PEP should have the following parts:
...
-   work.
-
-   This section may include key points and recommended documentation
-   changes that would help users adopt a new feature or migrate their
-   code to use a language change.
+   
+9. Rejected Ideas -- Throughout the discussion of a PEP, various ideas
+   will be proposed which are not accepted. Those rejected ideas should
+   be recorded along with the reasoning as to why they were rejected.
+   This boths helps record the thought process behind the final version
```
willingc: both

pep-0001.txt
```diff
@@ -455,15 +460,26 @@ Each PEP should have the following parts:
...
+   of the PEP as well as preventing people from bringing up the same
+   rejected idea again in subsequent discussions.
+
+10. Open Issues -- While a PEP is in draft, ideas can come up with warrant
```
willingc: with -> which

pep-0001.txt
```diff
@@ -455,15 +460,26 @@ Each PEP should have the following parts:
...
+    further discussion. Those to-be-discussed ideas should be recorded
```
willingc: ~~to-be-discussed~~

pep-0001.txt
```diff
@@ -455,15 +460,26 @@ Each PEP should have the following parts:
...
+    so people know that they are being thought about but do not have a
+    concrete resolution. This helps make sure all issues required for the
+    PEP to be ready for consideration are complete as well as prevent people
```
willingc: ...complete and reduces people duplicating prior discussion.

pep-0001.txt
willingc: Similar to Section 7: How to Teach, there could be a `Section x: Deployment, Configuration, and Security` that highlights any specific deployment, configuration, or security processes or best practices when used.

pep-0001.txt
brettcannon: That was actually already there, but people have not been doing it. 😉

brettcannon:
@gvanrossum I think since this is basically a touch-up on pre-existing procedures I don't think we really need to worry about getting the whole SC involved (especially since you and @willingc  + me make majority 😉 ).

@willingc I added a security section, but for the deployment and configuration I don't know if that comes up often enough to warrant a specific section.

And I think I addressed everyone's comment, PTAL.

brettcannon:
And FYI in case you aren't aware, GitHub now has a feature for [suggestions](https://help.github.com/en/articles/commenting-on-a-pull-request) so that for typo fixes and such you can directly suggest the change and then I can just click a button to have the PR updated.

PR #913, opened 2019-03-06
PEP 12: provide templates for the header and sections

brettcannon:
To help people not to forget something.

### DIFF
--- a/pep-0012.rst
+++ b/pep-0012.rst
@@ -134,6 +134,7 @@ directions below.
   replacing all this gobbledygook with your own text. Be sure to
   adhere to the format guidelines below, specifically on the
   prohibition of tab characters and the indentation requirements.
+  See "Suggested Sections" below for a template of sections to include.
 
 - Update your References and Copyright section.  Usually you'll place
   your PEP into the public domain, in which case just leave the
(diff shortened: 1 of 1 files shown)


pep-0012.rst
```diff
@@ -148,6 +149,25 @@ directions below.
 
 - Send your PEP submission to the PEP editors at peps@python.org.
 
+For reference, here are all of the possible header fields::
```
gvanrossum: Most of these have a specific format. Maybe (in a follow-up PR?) we should describe those? Or is there reference documentation for pep2html.py already?

pep-0012.rst
```diff
@@ -580,6 +600,47 @@ should be avoided.  For ordinary text, use ordinary 'single-quotes' or
...
 
     ``literal text: in here, anything goes!``
+    
+    
+Suggested Sections
+==================
+Various sections are found to be common across PEPs and are outlined in
+PEP 1 [1]_. Those sections are provided here for convenience (sans the
+"Copyright" section and the metadata footer).
+::
```
gvanrossum: Again I wonder if we could put a sentence in each explaining what goes where?

brettcannon: So the only reason I didn't put it in was to minimize coupling with PEP 1, but I'm fine with adding one-line guidance in each section.

pep-0012.rst
brettcannon: We can put examples in there.

brettcannon:
@gvanrossum PTAL

brettcannon:
@warsaw since your name is on the PEP, any feedback, or can I merge this?

PR #915, opened 2019-03-07
PEP 8100: Publish full vote counts from election

ewdurbin:
We have recieved consent from all candidates except one. No objection was heard
from the remaining candidate, and 4 attempts to solicit consent or objection
were made.


pradyunsg:
Thanks for publishing these @ewdurbin! :D

PR #932, opened 2019-03-15
Defer PEP 480

brettcannon:


brettcannon:
Got private approval from @dstufft .

PR #937, opened 2019-03-17
PEP 584: Fix examples from likely copy-paste bug

EmilStenstrom:


the-knights-who-say-ni:

Thanks again for your contribution, we look forward to reviewing it!

EmilStenstrom:
CLA has been signed, just waiting for the document to be accepted on your side.

stevendaprano:
Sorry guys, I'm still having technology difficulties that make it hard 
for me to interact with Github's website.

It does look like a copy-paste error to me, but the suggested correction 
isn't quite right since d has already been modified in place. The 
previous call to print(d) gives

{'spam': 1, 'eggs': 2}

so the 'cheese' key has been removed. After ``d -= {'spam', 'parrot'}`` 
we should have

{'eggs': 2}

and the second example ``d -= [('spam', 999)]`` needs to be deleted in 
full. To be clear, here's the relevant section in full:

# --- cut here %< ---

Dict difference ``-`` will return a new dict containing the items from 
the left operand which are not in the right operand.

    >>> d = {'spam': 1, 'eggs': 2, 'cheese': 3}
    >>> e = {'cheese': 'cheddar', 'aardvark': 'Ethel'}
    >>> d - e
    {'spam': 1, 'eggs': 2}
    >>> e - d
    {'aardvark': 'Ethel'}

Augmented assignment will operate in place.

    >>> d -= e
    >>> print(d)
    {'spam': 1, 'eggs': 2}


Like the merge operator and list concatenation, the difference operator 
requires both operands to be dicts, while the augmented version allows 
any iterable of keys.

    >>> d - {'spam', 'parrot'}
      ...
    TypeError: cannot take the difference of dict and set

    >>> d -= {'spam', 'parrot'}
    >>> print(d)
    {'eggs': 2}


# --- cut here %< ---



-- 
Steven

gvanrossum:
OK, I'll send a new PR and you can approve that. Thanks @EmilStenstrom for noticing something was wrong here!

EmilStenstrom:
Oh, I read the three examples to be independent, so that you could compare two different ways of doing the same thing. Happy to see it fixed!

PR #942, opened 2019-03-22
PEP 556: Mark deferred

pitrou:
... and explain why


pitrou:
@gvanrossum @pablogsal

PR #943, opened 2019-03-22
Make Nick Coghlan the BDFL Delegate for PEP 574

brettcannon:


brettcannon:
FYI @pitrou

pitrou:
LGTM. Please be aware that PEP 574 is not finalized yet.

ncoghlan:
Note that BDFL-Delegate needs to go after Author, but before Status: https://www.python.org/dev/peps/pep-0001/#pep-header-preamble

PR #945, opened 2019-03-24
PEPs 576, 579, 580: Appoint Petr Viktorin as BDFL-Delegate

ncoghlan:
Petr had been mentioned as a potential delegate prior to the governance
changes, and with Petr's agreement, the Steering Council is appointing
him to review the competing proposals in these PEPs, and make the
decision on which approach to pursue to resolve the issues with the
status quo.


ncoghlan:
@encukou FYI :)

PR #948, opened 2019-03-26
PEP 570: Add references to PEP 399, 3rd party modules using the notation and minor additions

pablogsal:


pablogsal:
CC: @mariocj89

pablogsal:
CC: @mariocj89 @vstinner

pep-0570.rst
```diff
@@ -176,6 +204,54 @@ This is a well discussed, recurring topic on the Python mailing lists:
 * May 2006: `Benji York: [Python-Dev] Positional-only Arguments
   <https://mail.python.org/pipermail/python-dev/2006-May/064790.html>`_
 
+Positional-only parameters have also the (minor) advantage of enforcing
+some logical order when calling interfaces that make use of them. For
+example, the ``range`` function takes all its parameters posionally and
```
willingc: sp. positionally

pep-0570.rst
```diff
@@ -176,6 +204,54 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+
+Another important aspect that motivates positional only arguments is
+[#PEP399]_: Pure Python/C Accelerator Module Compatibility Requirements.
+This PEP states that:
+
+    This PEP requires that in these instances that the C code must pass
+    the test suite used for the pure Python code so as to act as much as
+    a drop-in replacement as reasonably possible
+
+Is clear that if the C code is implemented using the existing capabilities
```
willingc: It's clear...

pep-0570.rst
```diff
@@ -176,6 +204,54 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+to implement positional only parameters using the argument clinic and
+relating machinery, is not possible for the pure Python counterpart to match
```
willingc: ...machinery; then, it is not...

tonybaloney:
2 things I would add: 
1. `METH_FASTCALL` only works for positional-only arguments, so this could be a way to improve performance on method calls by insisting that they be called as positional
2. The construction and use of a tuple during frame construction in `_PyEval_EvalCodeWithName()` (or just say `PyEval_EvalCodeEx()` instead of a dictionary would give better performance on evaluating code objects.

pep-0570.rst
```diff
@@ -76,7 +78,8 @@ the developer to maintain its name forever, as users might pass ``x`` as a
 keyword.
 
 Another good example is an API that wants to transmit the feeling
-of ownership through positional arguments, see::
+that one of its parameters is the "main" parameter through positional
+arguments. For example, see ::
 
     class MyDecorator:
         def __init__(self, original_function):
```
vstinner: note for myself, I didn't know that self can be passed as a keyword argument :-)

```
class A:
    def meth(self, arg):
        print(self, arg)

A.meth(self="hello", arg="arg")
```

Output:

```
hello arg
```

I'm not sure that it's a good API than self=value is allowed :-)

pep-0570.rst
```diff
@@ -132,10 +146,25 @@ Python function signature and what the function actually accepts,
...
 and the lack of clarity in the resulting signature.
 
+As mentioned before, this syntax is already being used outside the
+CPython code base for similar use cases [#numpy-ufuncs]_ [#scipy-gammaln]_,
+remarking that these scenarios are not exclusive of CPython and the
+standard library.
+
+The main reason this notation is received with surprise currently
+by users is because it has not been properly documented until very
+recently[#document-positional-only]_ and is not possible to use it
```
vstinner: "recently[#document-positional-only]_" missing space before the link?

pep-0570.rst
pablogsal: Good point!

vstinner: More fun with self.

```
>>> import io
>>> help(io.FileIO.write)
Help on method_descriptor:

write(self, b, /)
    Write buffer b to file, return number of bytes written.

>>> f=io.FileIO("/dev/null", "wb")
>>> io.FileIO.write(f, b"data")  # ok
4

>>> io.FileIO.write(self=f, b=b"data") # wtf?
TypeError: descriptor 'write' of '_io.FileIO' object needs an argument
```

It seems like 'method_descriptor' handles self as a positional-only argument since (at least) Python 2.7, but the error message isn't helpful. I created: https://bugs.python.org/issue36437

vstinner:
Rendered text: " but this goes against the spirit of [10]". Would it be possible to have a more readable links to get something like "the spirit of PEP 399" with a link on "PEP 399"?

Maybe:

```
against the spirit of `PEP 399 "Pure Python/C Accelerator Module Compatibility Requirements" <https://www.python.org/dev/peps/pep-0399>`__
```
It seems like you want to get a list of references. I'm not sure what is the best way to combine that.

pep-0570.rst
```diff
@@ -75,19 +77,30 @@ The name of the parameter provides no value whatsoever, and forces
...
+By being able to define positional-only arguments, we can change the
+name of those at will or even change them to ``*args``. There are multiple
+interfaces in the standard library that fall into this category, for
+example the "main" argument of ``collections.defaultdict`` (called
+*default_factory* in its documentation) can only be passed positionally.
+
+Another essential aspect to consider is [#PEP399]_, that mandates that pure
+Python versions of modules in the standard library *must* have the same
+interface and semantics that the accelerator modules implemented in C ().
+For example, if ``collections.defaultdict`` would have a pure Python
```
pganssle: Instead of "would have" use "had" or possibly "were to have".

pep-0570.rst
```diff
@@ -59,14 +60,28 @@ builtin functions that can specify positional-only
...
+surprise and confusion by users encountering functions with positional-only
+arguments. Notably, major third-party packages are already using the "/"
+notation in their interfaces [#numpy-ufuncs]_ [#scipy-gammaln]_.
+
+Positional-only arguments may be useful in several situations. One of the more
+extreme situations is in a function that can take any keyword parameter but
+also can take a positional one. Well-known examples for this situation are
+``Formatter.format`` and ``dict.update``. For instance, ``dict.update``
+accepts a dictionary (positionally) and/or any set of keyword parameters to use
+as key/value pairs. In this case, if the dictionary parameter were not
```
mariocj89: Note: We need to change the implementation to make this work, and it might make things quite more complex.

pablogsal: What do you mean? This is working with the current implementation

pablogsal: Hummmmm, I see. Indeed, not sure if this is even possible. Maybe we should modify this paragraph....

pep-0570.rst
```diff
@@ -149,13 +203,12 @@ Providing positional-only arguments will allow for maintaining the
 interface when creating pure Python implementation of C modules, which
 provides not only the API benefits outlined in this document but it is
 also faster. See this thread about converting keyword arguments to positional:
```
pganssle: This sentence has a lot packed into it and I think it ends up being a bit of a conceptual jumble, to untangle, I think that there are three semi-unrelated points being made here:

1. Having some arguments be positional-only is valuable, because it has the API benefits described in the earlier section.
2. Functions in extension modules use positional-only modules, and it is difficult or impossible to properly maintain a pure Python implementation of these functions at the moment.
3. Positional-only arguments are faster in extension modules, and when we have proper syntax for them, they'll *definitely* be faster in pure Python than a corresponding `*args`-based implementation would be.

I think point 2 explains why, *given that we have positional-only arguments*, we should have a syntax to specify them in function definitions, while points 1 and 3 explain why it's a good thing that we have positional-only arguments. A potential rewording for this whole paragraph:

"In addition to the API benefits outlined earlier in this document, positional-only arguments are also faster, as demonstrated in this thread about converting keyword arguments to positional: [#thread-keyword-to-positional]_. Providing a way to specify positional-only arguments in Python will make it easier to maintain pure Python implementations of C modules and will allow users to take advantage of these benefits even in code written only in Python."

pep-0570.rst
```diff
@@ -149,13 +203,12 @@ Providing positional-only arguments will allow for maintaining the
...
-[#thread-keyword-to-positional]_ and PEP-399 [#PEP399]_, which requires the
-same API for C accelerators as the Python implementation.
+[#thread-keyword-to-positional]_.
 
 There have been multiple changes in builtin functions that moved away
 from keyword arguments, like ``bool``, ``float``, ``list``, ``int``, ``tuple``
-which is a non-backward compatible change. By having proper support for
-positional-only arguments, these kind of APIs, where it is clear that
+which is a non-backwards compatible change. By having proper support for
```
pganssle:
```suggestion
which is not a backwards-compatible change. By having proper support for
```

Alternatively:

```
which is a backwards-incompatible change. By having proper support for
```

pep-0570.rst
```diff
@@ -149,13 +203,12 @@ Providing positional-only arguments will allow for maintaining the
...
+positional-only arguments, these kinds of APIs, where it is clear that
 passing a keyword argument provides no clarity, it would be possible to
 follow a similar approach as these builtins, without breaking users.
```
pganssle: This part is a bit unclear to me. The earlier part of the paragraph mentions making backwards-incompatible changes, but this is talking about non-breaking changes.

I also think this is another example where you are mixing arguments in favor of using positional-only arguments *at all*, and arguments in favor of adding Python syntax for them. Perhaps this paragraph and the previous one can be reworked along those lines, e.g.:

>In addition to the API benefits outlined earlier in this document, positional-only arguments are also faster, as demonstrated in this thread about converting keyword arguments to positional: [#thread-keyword-to-positional]_. In fact, because of these benefits there has even been a recent trend towards moving builtins *away* from keyword arguments: recently, backwards-incompatible changes were made to disallow keyword arguments to ``bool``, ``float``, ``list``, ``int``, ``tuple``.

> Providing a way to specify positional-only arguments in Python will make it easier to maintain pure Python implementations of C modules and will allow users to take advantage of these benefits even in code written only in Python. It will also encourage users to start with positional-only arguments when they believe that passing a keyword argument provides no clarity; unlike making a keyword argument positional-only, allowing a positional argument to be passed positionally is not a breaking change.

Not sure how strong that second paragraph is, but I think you can see the way I'm trying to arrange the argument.

pep-0570.rst
```diff
@@ -176,6 +229,59 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+some logical order when calling interfaces that make use of them. For
+example, the ``range`` function takes all its parameters positionally and
+this disallows forms like:
+
+    range(stop=5, start=0, step=2)
+    range(stop=5, step=2, start=0)
+    range(step=2, start=0, stop=5)
+    range(step=2, stop=5, start=0)
+
+at the price of disallowing using keyword arguments for the (unique)
```
pganssle:
```suggestion
at the price of disallowing the use of keyword arguments for the (unique)
```

pep-0570.rst
```diff
@@ -176,6 +229,59 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+PEP 399 [#PEP399]_: Pure Python/C Accelerator Module Compatibility
+Requirements.  This PEP states that:
+
+    This PEP requires that in these instances that the C code must pass
+    the test suite used for the pure Python code to act as much as
+    a drop-in replacement as reasonably possible
+
+It is clear that if the C code is implemented using the existing capabilities
+to implement positional-only parameters using the argument clinic and
+relating machinery; then, it is not possible for the pure Python counterpart
```
pganssle:
```suggestion
related machinery, it is not possible for the pure Python counterpart
```

pep-0570.rst
```diff
@@ -176,6 +229,59 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+    >>> import binascii; binascii.crc32(data=b'data')
+    TypeError: crc32() takes no keyword arguments
+
+    $ pypy3 # PyPy 6.0.0
+    >>>> import binascii; binascii.crc32(data=b'data')
+    2918445923
+
+Other Python implementations can, of course, reproduce the CPython APIs
+manually, but this goes against the spirit of PEP 399 [#PEP399]_ that
+intends to avoid duplication effort by mandating that all modules added
```
pganssle:
```suggestion
intends to avoid duplication of effort by mandating that all modules added
```

pep-0570.rst
```diff
@@ -176,6 +229,59 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+to Python's standard library **must** have a pure Python implementation
+with the same interface and semantics.
+
+A final argument in favor of positional-only arguments is that they allow
+some new optimisations like the ones already present in the argument clinic
```
pganssle: "optimisations" is the British spelling. Not sure if there's a style guide or if it recommends BrE or AmE. Probably best to go with style guide if it exists, otherwise just consistency throughout (which, again, I haven't checked for yet :P)

pep-0570.rst
```diff
@@ -176,6 +229,59 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+since said parameters must be passed in strict order. For instance, CPython's
+internal *METH_FASTCALL* calling convention has been recently speciallized
+for functions with positional-only parameters to eliminate the cost for
+handling empty keywords. Similar kind of performance improvements can be
```
pganssle:
```suggestion
handling empty keywords. Similar performance improvements can be
```

pep-0570.rst
```diff
@@ -176,6 +229,59 @@ This is a well discussed, recurring topic on the Python mailing lists:
...
+applied when creatin the evaluation frame of Python functions thanks to
```
pganssle:
```suggestion
applied when creating the evaluation frame of Python functions thanks to
```

pep-0570.rst
```diff
@@ -113,30 +153,44 @@ arguments by name::
 via the ``/`` marker, but this at the moment is only documentational,
```
pganssle:
```suggestion
via the ``/`` marker, but this at the moment is only a documentation convention,
```

pep-0570.rst
```diff
@@ -113,30 +153,44 @@ arguments by name::
...
+remarking that these scenarios are not exclusive to CPython and the
+standard library.
+
+The main reason this notation is received with surprise currently
+by users is because it has not been adequately documented until very
+recently [#document-positional-only]_ and is not possible to use it
+in Python code. For these reasons, this notation is currently an
+oddity that appears only in CPython's APIs developed in C. Documenting
+the notation and making it possible to be used in Python code will
+certainly, eliminate this problem.
```
pganssle:
```suggestion
certainly eliminate this problem.
```

I think "certainly" is still a bit strong here, I'd probably go with "likely" or something softer, but I don't feel strongly about it.

pep-0570.rst
```diff
@@ -221,7 +327,7 @@ remains the same as the one for positional-or-keyword. Once
 a positional-only argument is provided with a default,
```
pganssle: In the folded portion of the diff (line 219/325), "The logic around optional values for positional-only argument" should either be:

"The logic around optional values for positional-only arguments"

or

"The logic around optional values for a positional-only argument"

I prefer the first one.

Also "remains the same as the one for positional-or-keyword."

should be

"remains the same as for positional-or-keyword arguments"

or

"remains the same as for a positional-or-keyword argument"

pep-0570.rst
```diff
@@ -221,7 +327,7 @@ remains the same as the one for positional-or-keyword. Once
...
 the following positional-only and positional-or-keyword argument
```
pganssle: Should be "the following positional-only and positional-or-keyword arguments need to have defaults as well" (plural, since it applies to *all* arguments, not just the one immediately following).

pep-0570.rst
```diff
@@ -221,7 +327,7 @@ remains the same as the one for positional-or-keyword. Once
...
 needs to have a default as well. Positional-only parameters that
-don’t have a default value are "required" positional-only parameters.
+do not have a default value are "required" positional-only parameters.
```
pganssle:
```suggestion
do not have a default values are *required* positional-only parameters.
```

Quotes here makes sense but italics has the same meaning and disambiguates between quoting it as a specific term and quoting it as scare quotes (which would indicate that it's a soft or often ignored requirement).

pep-0570.rst
```diff
@@ -237,6 +343,16 @@ Whilst the followings are not::
     def name(p1=None, p2, /, p_or_kw=None, *, kw):
```
pganssle: I would probably use "While" instead of "Whilst" on folded line 234/340. As far as I know that's the correct way to use "whilst", but whilst is archaic and doesn't improve the clarity of the statement.

pablogsal:
Thank you very much to everyone involved in the review! :D

PR #963, opened 2019-03-29
PEP 570: Add a 'How to teach this' section and improve rejected ideas section with feedback from discourse

pablogsal:
CC: @pganssle @mariocj89


pep-0570.rst
```diff
@@ -385,6 +385,202 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
...
+        elif isinstance(template, SlowTemplate):
+            return template.format_slow(**kwargs)
+        else:
+            return template.generic_format(**kwargs)
+
+Now users can call your function with *all* possible keyword parameters ::
+
+    >>> format_template(slow_template, name="Hilbert", template="fractal")
+    "After some slow computations, *Hilbert* has finished the *fractal* template"
+
```
mariocj89: I would remove the alternative section. If we have the syntax users win nothing by knowing how to hack it around. I think it is just noise.

pep-0570.rst
```diff
@@ -385,6 +385,202 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
...
+    format_template(template, /, **kwargs)
+
+The second version is more explicit and clear and can benefit from systems
+that can generate the documentation automatically by reading the function
+signatures. Additionally, if you plan to use type annotations for your
+functions, using a single parameter instead of ``*args`` can be easier to
+deal with.
+
+Some other scenarios where positional-only parameters can appear are when
```
mariocj89: If the syntax is valid in pure python then I think this does not add much value.
The only benefit of explaining something to users here would be they “it is faster”.

pep-0570.rst
```diff
@@ -385,6 +385,202 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
...
+    function.integrate(0)
+    function.integrate(0, method="euler")
+    function.integrate(0, 100)
+    function.integrate(0, 100, method="euler")
+
+As you can see, positional-only arguments are specially convenient when you
+want to make sure that some of the interfaces you provide can be easily
+modified in the future without making backwards incompatible changes in your
+library. It allows you to have clear and explicit function signatures that
+express how arguments should be provided without the need of implementing
```
mariocj89: Expresses? Some native speaker might help rewording this.

pep-0570.rst
pablogsal: What I wanted is to remark that you need to use it if you are proving a fallback for a function that behaves like this and is implemented in C. Notice there are more ways of having this behaviour without the argument clinic.

pep-0570.rst
```diff
@@ -385,6 +385,202 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
...
+            ....
+
+When calling this method, all of these forms are correct ::
+
+    function.integrate(0)
+    function.integrate(0, method="euler")
+    function.integrate(0, 100)
+    function.integrate(0, 100, method="euler")
+
+As you can see, positional-only arguments are specially convenient when you
```
brettcannon:
```suggestion
As you can see, positional-only arguments are especially convenient when you
```

pep-0570.rst
brettcannon: Pablo's actually correct.

pep-0570.rst
pablogsal: I think it helps to realize what are the advantages. Users are already used to `*args` and the hacks with that, so this will help to show how those constructs can be substituted with positional-only.

gvanrossum:
So I just went back to the Tutorial and found this section: [More on Defining Functions](https://docs.python.org/3.8/tutorial/controlflow.html#more-on-defining-functions).

This section is actually pretty thorough -- it goes into defaults, keywords (including `**kwds`), `*args` (definition and call site). I'm sure it can be updated to also mention `/`, but I'm not sure yet what's the best place to do that. (Honestly I'm also unclear why `*args` and `**kwds` are not treated more symmetrically there -- perhaps because different authors updated the tutorial at different times.)

One thing that I'm realizing is that the main motivation for PEP 570 is to enable *good API design*. In particular it looks like an important use case is to avoid painting yourself into a corner when designing a function or method signature that might evolve in the future. Subclassing is a rather important special case of this -- designing for subclassing means anticipating code that hasn't been written yet and over which you don't have control.

Another important reason to practice good API design is to prevent common mistakes for a function's users. (This actually overlaps with design for subclassing.) PEP 570 can help here too by making calls more uniform -- giving the caller less freedom enforces TOOWTDI, which means readers of those call sites will be less distracted by variations that lead to the same call (or wasting time to verify that this is the case).

All these motivations are somewhat different to the motivation for defaults, keywords, and "varargs" (both `*args` and `**kwds`). Those actually enable more variation, by opening up different usages: calling a function without specifying values for all arguments (defaults), specifying arguments by name instead of position (keywords), and allowing unlimited arguments (`*args` and `**kwds`). This is all good, but it is up to the caller to choose the right pattern, and there definitely are traps for unwary users.

I have lost track of how many times I have had to explain (often to experienced users!) that there is a difference between keyword arguments and parameters with default values -- the presence of a default doesn't mean the caller has to use a keyword, and the absence of a default doesn't mean the caller cannot use a keyword. The introduction of plain `*` to force the following arguments to be passed as keywords helped here. PEP 570 is the logical final step here.

pablogsal:
> Hm... At the very least this requires a native speaker to go over the text carefully. But something else about it feels off to me -- it somehow sounds more like a thorough motivation than like an effective tutorial. Don't worry too much though, this is not a mandatory part of the PEP.

For now, I have added in commit 6424bcc a separation between "how to use them" on "when to use them". As this won't be the final document that will be in the documentation, I have not documented **all** possible cases when combining them with the rest of the options but I have added a diagram for the general case.

pablogsal:
> Another important reason to practice good API design is to prevent common mistakes for a function's users. (This actually overlaps with design for subclassing.) PEP 570 can help here too by making calls more uniform -- giving the caller less freedom enforces TOOWTDI, which means readers of those call sites will be less distracted by variations that lead to the same call (or wasting time to verify that this is the case).

Should I add subclassing to the PEP or the "How to teach this section"? Is more or less a consequence of the exposition but there is some benefit of making it explicit. 

Also, feel free to push directly to my branch to avoid iteration if you want to add something directly to the text.

gvanrossum:
I'd add subclassing to the main motivation of the PEP.

I'm going to try and take a somewhat more hands-off attitude for about a
week, I'm going to be very busy (Carol can guess why).

ncoghlan:
Something I'm missing from the motivation in the current PEP draft is the fact that the users can't currently learn what the documentation notation for positional-only arguments means by experimenting at the interactive prompt:

```
>>> def pos_only_arg(arg, /):
  File "<stdin>", line 1
    def pos_only_arg(arg, /):
                          ^
SyntaxError: invalid syntax
>>> def combined_example(pos_only_arg, /, standard_arg, *, kwd_only_arg):
  File "<stdin>", line 1
    def combined_example(pos_only_arg, /, standard_arg, *, kwd_only_arg):
                                       ^
SyntaxError: invalid syntax
```

My working theory is that this is the root cause of the baffled incomprehension encountered by many learners for this notation, as opposed to being able to run through this straightforward sequence to learn about keyword-only arguments:

```
>>> def standard_arg(arg):
...     print(arg)
... 
>>> def kwd_only_arg(*, arg):
...     print(arg)
... 
>>> standard_arg(1)
1
>>> standard_arg(arg=1)
1
>>> kwd_only_arg(1)
TypeError: kwd_only_arg() takes 0 positional arguments but 1 was given
>>> kwd_only_arg(arg=1)
1
```

So I think a "How to teach this" section in the PEP is a good idea, but it should be a relatively simple REPL session, going through the behaviour of 4 example functions that just print the supplied arguments:

```
>>> def pos_only_arg(arg, /):
...     print(arg)
... 
>>> def standard_arg(arg):
...     print(arg)
... 
>>> def kwd_only_arg(*, arg):
...     print(arg)
... 
>>> def combined_example(pos_only, /, standard, *,kwd_only):
...     print(pos_only, standard, kwd_only)
... 
>>> pos_only_arg(1)
1
>>> pos_only_arg(arg=1)
TypeError: pos_only_arg() got an unexpected keyword argument 'arg'
>>> standard_arg(2)
2
>>> standard_arg(arg=2)
2
>>> kwd_only_arg(3)
TypeError: kwd_only_arg() takes 0 positional arguments but 1 was given
>>> kwd_only_arg(arg=3)
3
>>> combined_example(1, 2, 3)
TypeError: combined_example() takes 2 positional arguments but 3 were given
>>> combined_example(1, 2, kwd_only=3)
1 2 3
>>> combined_example(1, standard=2, kwd_only=3)
1 2 3
>>> combined_example(pos_only=1, standard=2, kwd_only=3)
TypeError: combined_example() got an unexpected keyword argument 'pos_only'
```

pep-0570.rst
ncoghlan: This is a draft tutorial aimed at new users - it needs to teach them what the syntax does, it doesn't need to teach them all the API design trade-offs that might lead them to choosing whether or not to use it. Instead, it can just give a couple of specific examples of standard library APIs that work that way, like `dict.update`.

ncoghlan:
As another comment: I don't think it's helpful to propose a direct patch to the CPython tutorial in the PEP (I think the short example REPL session above is better for that purpose), but a tutorial change would be a good thing to have in the reference implementation (but again, that doesn't need to be an extensive discourse on API design, it just needs to introduce the notation, and explain the significance of the 3 forms).

pep-0570.rst
```diff
@@ -385,6 +385,305 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
 the existing typedargslist and varargslist (in the same way keyword-only
 arguments were introduced).
 
+=================
+How to teach this
+=================
+
+This PEP recommends adding a new subsection to the Python documentation,
+in the section "More on Defining Functions", where the rest of the
```
pganssle: Can we add a link to this section?

Looks like it refers to: https://docs.python.org/3/tutorial/controlflow.html#more-on-defining-functions

pep-0570.rst
```diff
@@ -385,6 +385,305 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
...
+compatible change as the language is making sure that users were not using the name
+of the parameters when calling the function. So now, the old way of calling and the
+new way of calling are still possible ::
+
+    >>> hilbert_multiply(elem1, elem2) # Old way of calling the function
+
+    >>> hilbert_multiply(elem1, elem2, elem3, elem 4) # New way of calling the function
+
+If the function was not declared previously using positional-only parameters, then
+users will receive an error when calling the function as they used to do ::
```
pganssle: I'm going to change this to "as they used to", because "as they used to do" sounds wrong to me, but I think in BrE there are certain situations where it is idiomatic to say things like, "I would do" where Americans would say "I would" - I am not sure if this is one of them.

pganssle:
I made some grammar and style checks to the proposed section as written mainly because I think it is valuable content that could be used in a blog post or some other tutorial, but I think it might be a bit too opinionated for the "how to teach this" documentation.

I think it's worth considering that even if this PEP is accepted, there will probably be a decent chunk of people who don't like this pattern, and I think there is a lot more support out there for "make it possible to do this" than there is for "make this the Pythonic right way to design APIs".

I looked around in the existing documentation and I don't see "keyword-only" arguments documented anywhere either. In some ways, I think the best way to teach positional-only arguments is together with keyword-only arguments. Nick's proposed REPL example with one or two paragraphs on either side, and [this diagram](https://github.com/python/peps/pull/963/files#diff-427f4cd921e7fe1d6b2dac27ea990983R494) seems to be the perfect sweet spot, to me.

pep-0570.rst
```diff
@@ -385,6 +385,305 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
...
+How to use positional-only parameters
+-------------------------------------
+
+You can mark a set of parameters in a function to be positional-only by placing
+a "/" marker after them ::
+
+    def process_spams(spam1, spam2, /, complementary_spam):
+        ....
+
+In this example, ``spam1`` and ``spam2`` are positional-only while
```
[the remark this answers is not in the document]
willingc: Agreed. A bit too much spam which adds to the cognitive load. More distinct names would be better as this will likely become end users docs at some point. I'll give some thought to a real world example.

pep-0570.rst
```diff
@@ -385,6 +385,304 @@ as a rule is not LL(1). This is the reason the rule needs to be included in
...
+
+This happens because the name ``spam`` cannot be bound to the first argument, as it
+can only be passed positionally, so it ends up in the ``**kwargs`` dictionary that is
+collecting all the keywords that are not bound to any other name. The next section
+provides some scenarios where this can be very useful.
+
+For completeness, this is the more general declaration for a function
+(ignoring default arguments) ::
+
+    def foo(pos1, pos2, /, posorkeyword1, posorkeyword2, *args, kwonly1, kwonly2, **kwargs):
```
willingc:
```suggestion
    def foo(pos1, pos2, /, pos_or_kw3, pos_or_kw4, *args, kw5, kw6, **kwargs):
```

pep-0570.rst
willingc: It would be good to split out the alternative/advanced usage into its own section. It's useful to document, similar to when we rewrote the asyncio docs, what concepts most users would need from those that advanced users or library authors might need. This seems like a similar case here. The downside of not documenting advanced alternatives (even if we discourage them) is that it leaves it open for interpretation what the intent is and when to use (which the community will likely fill in with multiple different ways).

pep-0570.rst
willingc: Again, I think this could be advanced use cases which could be split out as such.

pganssle:
I created [a PR against this branch](https://github.com/pablogsal/peps/pull/1/files) with a new proposed wording for the "How to teach this" section. Sorry to split the discussion, but I didn't want to directly commit such radical changes without sign off from at least Pablo.

pablogsal:
I have merged Paul's PR simplifying the "How to teach this" section taking into account the feedback and I have incorporated the changes requested on discourse (https://discuss.python.org/t/pep-570-python-positional-only-parameters/1078/138) as well as the previously discussed subclassing argument.

gvanrossum:
Thanks! I'll merge this now, since the tests pass (this is not yet approval of the PEP), so we can all see the improved version on the web. Carol just told me she's going to try and improve the teaching section some more. After that I'll read it through and probably just make a bunch of small edits.

PR #968, opened 2019-04-01
PEP 578: Rename open_for_import to io.open_code

zooba:
Haven't implemented these changes yet, and I want to do that before committing this, but at least getting the new proposed names out there early.


ncoghlan:
I like the new name. The PEP should discuss whether or not the pickle module should be switched over to using the "open code" API rather than the regular API (ditto the XML APIs that support the full XML spec), or whether those should just be their own distinct audit events.

PR #972, opened 2019-04-03
PEP 570: Restructure sections as recommended by PEP 1

ericvw:
Organize PEP 570 according to the recommended ordering of sections
defined in PEP 1.  This presents the proposal in a more organized and
compelling way.

Notable changes are:

- The content more easily reads in from top to bottom by moving core
  content into Abstract, Motivation, Rationale, Specification.
- Making a clear distinction of the impact to library authors and
  callers of APIs.
- Fixed-up usage of "parameters" and "arguments" around the context of the
  what is being discussed.
- Grammatical edits and simplified wording, while maintaining the core
  content and intent of the text.

While the diff looks drastic, the core content of the originally written
PEP remains intact.

Note that I have incorporated grammatical edits and fixes from #971 manually because of the restructuring of the entire PEP (@pganssle).

PR #975, opened 2019-04-04
PEP 570: Edit Sections and How to Teach

willingc:
- Edit PEP sections for clarity and readability


pep-0570.rst
```diff
@@ -19,27 +19,33 @@ Abstract
 ========
 
 This PEP proposes to introduce a new syntax, ``/``, for specifying
-positional-only parameters in Python.
+positional-only parameters in Python function signatures.
```
ericvw: @willingc in #972, I changed all references to "function signature" to "function definition" to be congruent with the text in https://docs.python.org/3/tutorial/controlflow.html#defining-functions

> The keyword `def` introduces a function definition. [...]

For consistency with the rest of the document, does it make sense to keep with "definition" over "signature"?  There are two other references to "signature" further down…

pep-0570.rst
```diff
@@ -541,15 +631,33 @@ definition::
...
 
+-----
+Recap
+-----
+
+The use case will determine which parameters to use in the function definition::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+
+As guidance:
```
ericvw: This is an excellent summary! 👍

pep-0570.rst
```diff
@@ -267,10 +281,14 @@ can be called. It would restrict arguments to be called as positional-only,
 while not allowing them to be called as keyword arguments.
 
 Previously, PEP 457 proposed to define the syntax, but with a much broader
-scope. This PEP takes the original proposal a step further that by justifying
+scope. This PEP takes the original proposal a step further by justifying
 the syntax and providing an implementation for the ``/`` syntax in function
 definitions.
 
+-----------
```
ericvw: Wow — the subsections give more clarity within the rationale. 

Indicating these subsections explicitly was something I was struggling to define when I was working on #972.

👍

pep-0570.rst
```diff
@@ -239,14 +252,15 @@ Improving Language Consistency
 The Python language, itself, would be more consistent with positional-only
 parameters. If the concept is a normal feature of Python rather than a feature
 exclusive to extension modules, it would reduce confusion for users
-encountering functions with positional-only arguments. Again, major
+encountering functions with positional-only arguments. Some major
```
ericvw: Since this is talking about the definition of the function, I think "parameters" should be used instead of "arguments".

pep-0570.rst
```diff
@@ -345,6 +375,10 @@ goes against the spirit of PEP 399 [#PEP399]_ to avoid duplication of effort by
 mandating that all modules added to Python's standard library **must** have a
 pure Python implementation with the same interface and semantics.
 
+-------------------------
+Consistency in Subclasses
+-------------------------
+
 Another scenario where positional-only parameters provide benefit occurs when a
 subclass overrides a method of the base class and changes the name of arguments
```
ericvw: Not part of this diff, but I think I missed this in #972 to change "arguments" to "parameters" here.

pep-0570.rst
```diff
@@ -403,15 +441,20 @@ like::
...
-following positional-only and positional-or-keyword arguments need to have
-defaults as well. Positional-only parameters which do not have a default
-values are *required* positional-only parameters. Therefore the following are
-valid signatures::
+The following would apply:
+
+* All parameters left of the ``/`` are demarcated as positional-only.
+* If ``/`` is not specified in the function definition, that function does not
+  accept any positional-only arguments.
+* The logic around optional values for positional-only arguments remains the
```
ericvw: I think this should be "parameters" as well for this bullet point.  Maybe…

pep-0570.rst
```diff
@@ -403,15 +441,20 @@ like::
...
+  same as for positional-or-keyword arguments.
+* Once a positional-only parameter is specified with a default, the
+  following positional-only and positional-or-keyword arguments need to have
```
ericvw: I think this should also be "parameters" too.

pep-0570.rst
```diff
@@ -460,32 +508,30 @@ in 2012 [#GUIDO]_ :
 How to teach this
 =================
 
-Since this concept is closely analogous to keyword-only arguments, introducing
-a dedicated syntax to mark positional-only arguments may in fact make it
-*easier* to teach the possible function definitions a user may encounter or
-design, by teaching the two concepts together.
+Introducing a dedicated syntax to mark positional-only arguments is closely
```
ericvw: Sorry for a lot of these, but I think this may be "parameters" too.

pep-0570.rst
willingc: Thanks for the thorough review @ericvw :snake: Yeah, we should be consistent throughout. I'll take a look on my flight and make some edits. FYI - keep reviewing PRs, you are doing good work.

pep-0570.rst
willingc: Thanks

pep-0570.rst
```diff
@@ -403,15 +442,20 @@ like::
...
+  accept any positional-only arguments.
+* The logic around optional values for positional-only parameters remains the
+  same as for positional-or-keyword parameters.
+* Once a positional-only parameter is specified with a default, the
+  following positional-only and positional-or-keyword parameters need to have
+  defaults as well.
+* Positional-only parameters which do not have default
+  values are *required* positional-only parameters.
+
+Therefore the following are valid signatures::
```
ericvw: @willingc a few more 😄 ; changing "signatures" to " function definitions"

pep-0570.rst
```diff
@@ -420,7 +464,12 @@ valid signatures::
     def name(p1, p2, /, p_or_kw):
     def name(p1, p2, /):
 
-While the followings are not::
+As today, the following are valid signatures::
```
ericvw: Change "signatures" to "function definitions".

pep-0570.rst
```diff
@@ -541,15 +632,33 @@ definition::
...
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+
+As guidance:
+
+* Use positional-only if names do not matter or have no meaning, and there are
+  only a few arguments which will always be passed in the same order.
+* Use keyword-only when names have meaning and the function definition is
+  more understandable by being explicit with names.
+* Consider positional-only if performance is critical and would be optimized.
```
pablogsal: Question: should we include here the case for a function that takes *any* keyword argument via `**kwargs` and some positional argument?

```
    def log(fmt, **kwargs):
        print(fmt.format_map(kwargs))
```

pep-0570.rst
```diff
@@ -441,7 +490,7 @@ A draft of the proposed grammar specification is::
 It would be added to the actual ``typedargslist`` and ``varargslist``, but for
 more relaxed discussion it is presented as ``new_typedargslist`` and
 ``new_varargslist``. Note that using a construction with two new rules
-(``new_varargslist`` and ``new_varargslist``) is not possible with the current
+(``new_typedargslist`` and ``new_varargslist``) is not possible with the current
 parser as a rule is not LL(1). This is the reason the rule needs to be
 included in the existing ``typedargslist`` and ``varargslist`` (in the same way
 keyword-only arguments were introduced).
```
gvanrossum: That's a lot of words. IMO just showing the entire updated definitions for these two rules would be better.

pep-0570.rst
```diff
@@ -460,32 +509,30 @@ in 2012 [#GUIDO]_ :
...
+may *simplify* how to teach the possible function definitions a user may encounter or
+design.
 
 This PEP recommends adding a new subsection to the Python documentation, in the
 section `"More on Defining Functions"`_, where the rest of the argument types
-are discussed. The following paragraphs serve as a draft for these additions
-that will serve to introduce the notation for both positional-only and
+are discussed. The following paragraphs serve as a draft for these additions.
+They will introduce the notation for both positional-only and
 keyword-only parameters. It does not intend to be exhaustive, nor should it be
```
gvanrossum: "does not intend" -> "is not intended"

pep-0570.rst
```diff
@@ -494,7 +541,44 @@ A function definition which makes use of both of these features may look like::
...
+
+--------------------------
+Positional-only Parameters
+--------------------------
+
+Looking at this in a bit more detail, it is possible to mark certain parameters
+as *positional-only*. If *positional-only*, the parameters' order matters, and
+the parameters cannot be passed by keyword. By placing a ``/`` (forward-slash) in
+the parameters list after the last positional-only parameter, a developer is
+logically separating positional-only parameters from the rest of the parameters.
```
gvanrossum: IMO the key point here is that positional-only parameters come *before* the `/` (and if there's no `/`, there are no pos-only params). IOW I'd emphasize that the pos-only params come *before* the `/`, rather than stating that the `/` comes after the pos-only params.

pep-0570.rst
gvanrossum: I'm not sure I would even mention perf here. Surely this only matters for C functions and even so only in rare cases.

pep-0570.rst
willingc: I removed the wording here, updated a simplified view of the change, and added the full rules. Hoping that clarifies.

pep-0570.rst
willingc: Hmm...I wasn't sure this would be easily summarized here. I think this is something that could be part of the actual documentation. Does that work for you @pablogsal?

willingc:
Other than Pablo's question which I responded to in the review comments, I think that I got all  of the requested changes. Folks may wish to do a quick scan of the Grammar section. Thanks.

pablogsal:
I am ok with not including the **kwargs case in the "how to teach this" section as currently we do not talk about `**kwargs` at all there.

pep-0570.rst
pablogsal: Yep :)

pablogsal:
I am landing this! 

Thank you very much to everyone involved. This has greatly improved the clarity and readability of the PEP!

PR #979, opened 2019-04-07
PEP 336: Change semicolons to colons at function definitions

BTaskaya:
Change semicolons to colons at function definitions


gvanrossum:
Thanks for your attention to detail. I hope you come back with more contributions.

PR #984, opened 2019-04-10
PEP 590: Add "Vectorcall" and "fastcall" keywords

encukou:
- Add "Vectorcall" to the PEP title to make it easier to reference
- Mention "fastcall" in the abstract, for people familiar with the term

@markshannon, are you OK with renaming the PEP like this?


markshannon:
Yes, this does make it a bit clearer.

PR #985, opened 2019-04-10
PEP 578: Add clarifications around open_code() for non-code files

zooba:
Clarifications as requested by @ncoghlan (not sure he agrees with the clarifications, but he asked for them ;) )


ncoghlan:
They're not the result I was expecting, but they're a result I like :)

PR #989, opened 2019-04-12
PEP 394: Allow for more flexibility in handling /usr/bin/python

hroncok:
Co-Authored-By: @encukou

### DIFF
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -103,18 +103,9 @@ For Python script publishers
 * Encourage your end users to use a virtual environment.
   This makes the user's environment more predictable (possibly resulting
   in fewer issues), and helps avoid disrupting their system.
-* In shebang lines, the preferred spelling is ``/usr/bin/env python``,
+* For scripts that are only expected to be run in an activated virtual
+  environment, shebang lines can be written as ``/usr/bin/env python``,
   as this instructs the script to respect the active virtual environment.
-* For Python 3 only scripts that do not support being executed on Python
-  2 at all, it is recommended to instead use ``python3`` and
-  ``/usr/bin/env python3``, as these will never invoke Python 2, and are
-  expected to work for both Python 3 virtual environments and Python 3
-  system installations
-* For Python 2 only scripts that do not support being executed on Python
-  3 at all, it is recommended to instead use ``python2`` and
-  ``/usr/bin/env python2``, as these will never invoke Python 3, and are
-  expected to work for Python 2 virtual environments and at least some
-  Python 2 system installations.
 * In cases where the script is expected to be executed outside virtual
   environments, developers will need to be aware of the following
   discrepancies across platforms and installation methods:
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -51,7 +51,7 @@ For Python runtime distributors
 -------------------------------
 
 * We expect Unix-like software distributions (including systems like macOS and
-  Cygwin) should install the ``python2`` command into the default path
+  Cygwin) to install the ``python2`` command into the default path
   whenever a version of the Python 2 interpreter is installed, and the same
   for ``python3`` and the Python 3 interpreter.
 * When  invoked, ``python2`` should run some version of the Python 2
@@ -104,7 +104,7 @@ For Python script publishers
   This makes the user's environment more predictable (possibly resulting
   in fewer issues), and helps avoid disrupting their system.
 * For scripts that are only expected to be run in an activated virtual
-  environment, shebang lines can be written as ``/usr/bin/env python``,
+  environment, shebang lines can be written as ``#!/usr/bin/env python``,
   as this instructs the script to respect the active virtual environment.
 * In cases where the script is expected to be executed outside virtual
   environments, developers will need to be aware of the following
@@ -114,9 +114,9 @@ For Python script publishers
     refers to Python 2. Most of these distros do *not* provide a
     ``python2`` command.
   * Some newer Linux distributions will provide a ``python`` command that
-    refers to Python 3
+    refers to Python 3.
   * Some Linux distributions will not provide a ``python`` command at
-    all by default, but will provide a ``python3`` command by default
+    all by default, but will provide a ``python3`` command by default.
 
 * When potentially targeting these environments, developers may either
   use a Python package installation tool that rewrites shebang lines for
@@ -148,7 +148,7 @@ These recommendations are the outcome of the relevant python-dev discussions
 in March and July 2011 ([1]_, [2]_), February 2012 ([4]_),
 September 2014 ([6]_), discussion on GitHub in April 2018 ([7]_),
 on python-dev in February 2019 ([8]_), and during the PEP update review
-in May 2019 ([10]_).
+in May/June 2019 ([10]_).
 
 
 History of this PEP
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -13,7 +13,7 @@ Type: Informational
 Content-Type: text/x-rst
 Created: 02-Mar-2011
 Post-History: 04-Mar-2011, 20-Jul-2011, 16-Feb-2012, 30-Sep-2014, 28-Apr-2018,
-              12-Apr-2019
+              26-Jun-2019
 Resolution: https://mail.python.org/pipermail/python-dev/2012-February/116594.html
 
 
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -111,8 +111,7 @@ For Python script publishers
   discrepancies across platforms and installation methods:
 
   * Older Linux distributions will provide a ``python`` command that
-    refers to Python 2. Most of these distros do *not* provide a
-    ``python2`` command.
+    refers to Python 2, and will likely not provide a ``python2`` command.
   * Some newer Linux distributions will provide a ``python`` command that
     refers to Python 3.
   * Some Linux distributions will not provide a ``python`` command at
@@ -267,12 +266,7 @@ making such a change.
   previously installed binary.
 * Even as the Python 2 interpreter becomes less common, it remains reasonable
   for scripts to continue to use the ``python3`` convention, rather than just
-  ``python``. In the event of a future Python 4.0 release, distributors may
-  then choose to simply alias ``python3`` to the same binary as ``python4``
-  (if the degree of change will be the same as in any other Python feature
-  release), or else they may choose to maintain parallel ``python3`` and
-  ``python4`` stacks for a time (depending on the needs of their specific
-  audience).
+  ``python``.
 * If these conventions are adhered to, it will become the case that the
   ``python`` command is only executed in an interactive manner as a user
   convenience, or else when using a virtual environment or similar mechanism.
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -34,7 +34,7 @@ The recommendations in the next section of this PEP will outline the behavior
 when:
 
 * using virtual environments
-* cross-platform scripts with shebangs for either ``python2`` or ``python3``
+* writing cross-platform scripts with shebangs for either ``python2`` or ``python3``
 
 The PEP's goal is to clarify the behavior for script end users, distribution
 providers, and script maintainers / authors.
@@ -79,7 +79,7 @@ For distributors
   This ensures software is used with the latest version of Python available,
   and it can remove a dependency on Python 2.
   The details on what specifics to set are left to the distributors;
-  though, example specifics could include:
+  though. Example specifics could include:
 
   * Changing ``python`` shebangs to ``python3`` when Python 3.x is supported.
   * Changing ``python`` shebangs to ``python2`` when Python 3.x is not yet
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -47,8 +47,8 @@ Recommendation
 Our recommendations are detailed below.
 We call out any expectations that these recommendations are based upon.
 
-For distributors
-----------------
+For Python runtime distributors
+-------------------------------
 
 * We expect Unix-like software distributions (including systems like macOS and
   Cygwin) should install the ``python2`` command into the default path
@@ -94,16 +94,15 @@ For distributors
   The ``python3`` or ``python2`` command (according to the environment's
   interpreter version) should also be available.
 
-For developers
---------------
+For Python script publishers
+----------------------------
 
 * When reinvoking the interpreter from a Python script, querying
   ``sys.executable`` to avoid hardcoded assumptions regarding the
   interpreter location remains the preferred approach.
-* While far from being universally available, ``python`` remains the
-  preferred spelling for explicitly invoking Python, as this is the
-  spelling that virtual environments make consistently available
-  across different platforms and Python installations.
+* Encourage your end users to use a virtual environment.
+  This makes the user's environment more predictable (possibly resulting
+  in fewer issues), and helps avoid disrupting their system.
 * In shebang lines, the preferred spelling is ``/usr/bin/env python``,
   as this instructs the script to respect the active virtual environment.
 * For Python 3 only scripts that do not support being executed on Python
@@ -141,6 +140,19 @@ For developers
   a container or virtual environment) may continue to use the ``python``
   command name.
 
+For end users of Python
+-----------------------
+
+* While far from being universally available, ``python`` remains the
+  preferred spelling for explicitly invoking Python, as this is the
+  spelling that virtual environments make consistently available
+  across different platforms and Python installations.
+* For software that is not distributed with (or developed for) your system,
+  we recommend using a virtual environment, possibly with an environment
+  manager like ``conda`` or ``pipenv``, to help avoid disrupting your system
+  Python installation.
+
+
 These recommendations are the outcome of the relevant python-dev discussions
 in March and July 2011 ([1]_, [2]_), February 2012 ([4]_),
 September 2014 ([6]_), discussion on GitHub in April 2018 ([7]_),
@@ -176,9 +188,9 @@ available. As Python 2 is nearing its end of life in 2020 (PEP 373, PEP 404),
 distributions are making Python 2 optional or removing it entirely.
 This means either removing the ``python`` command or switching it to invoke
 Python 3. Some distributors also decided that their users were better served by
-ignoring the PEP's recommendations, and provided system administrators with the
-freedom to configure their systems based on the needs of their particular
-environment.
+ignoring the PEP's original recommendations, and provided system
+administrators with the freedom to configure their systems based on
+the needs of their particular environment.
 
 
 .. _rationale:
@@ -262,8 +274,8 @@ making such a change.
   versa. That way, if a sysadmin does decide to replace the installed
   ``python`` file, they can do so without inadvertently deleting the
   previously installed binary.
-* Even as the Python 2 interpreter becomes less common, scripts should
-  nevertheless continue to use the ``python3`` convention rather than just
+* Even as the Python 2 interpreter becomes less common, it remains reasonable
+  for scripts to continue to use the ``python3`` convention, rather than just
   ``python``. In the event of a future Python 4.0 release, distributors may
   then choose to simply alias ``python3`` to the same binary as ``python4``
   (if the degree of change will be the same as in any other Python feature
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -193,11 +193,6 @@ cross-platform and cross-distribution experience.
 Accordingly, publishers can expect users of the software to provide a suitable
 execution environment.
 
-As part of that however, it is also appropriate to give Python
-distributors the flexibility they need in order to make the
-behaviour of their systems as similar as possible to the behaviour
-of an activated virtual environment.
-
 
 Future Changes to this Recommendation
 =====================================
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -40,7 +40,8 @@ Recommendation
 ==============
 
 
-Our recommendations are detailed below. We call out any expectations that these recommendations are based upon.
+Our recommendations are detailed below.
+We call out any expectations that these recommendations are based upon.
 
 For distributors
 ----------------
@@ -52,29 +53,36 @@ For distributors
 * When  invoked, ``python2`` should run some version of the Python 2
   interpreter, and ``python3`` should run some version of the Python 3
   interpreter.
-* If the ``python`` command is installed, it is expected to invoke either the same
-  version of Python as the ``python3`` command or as the ``python2``
+* If the ``python`` command is installed, it is expected to invoke either
+  the same version of Python as the ``python3`` command or as the ``python2``
   command.
-* Distributors may choose to set the behavior of the ``python`` command as follows:
-    - ``python2``,
-    - ``python3``,
-    - not provide ``python`` command
-    - allow ``python`` to be configurable by an end user or a system administrator.
-  not provide it at all, or make it configurable by the user or
-  system administrator.
+* Distributors may choose to set the behavior of the ``python`` command
+  as follows:
+
+   * ``python2``,
+   * ``python3``,
+   * not provide ``python`` command,
+   * allow ``python`` to be configurable by an end user or
+     a system administrator.
+
 * The Python 3.x ``idle``, ``pydoc``, and ``python-config`` commands should
   likewise be available as ``idle3``, ``pydoc3``, and ``python3-config``;
   Python 2.x versions as ``idle2``, ``pydoc2``, and ``python2-config``.
-  The commands with no version number should either invoke the same version of Python
-  as the ``python`` command, or not be available at all.
-* When packaging third party Python scripts, distributors are encouraged to change less specific
-  shebangs to more specific ones. This ensures software is used with the latest
-  version of Python available, and it can remove a dependency on Python 2.
-  The details on what specifics to set are left to the distributors; though, example specifics could include: 
-  Examples of this include changing ``python`` shebangs to ``python3``
-  when Python 3.x is supported, changing ``python`` shebangs to ``python2``
-  when Python 3.x is not yet supported, or changing ``python3`` shebangs to
-  ``python3.8`` if the software is built with Python 3.8.
+  The commands with no version number should either invoke the same version
+  of Python as the ``python`` command, or not be available at all.
+* When packaging third party Python scripts, distributors are encouraged to
+  change less specific shebangs to more specific ones.
+  This ensures software is used with the latest version of Python available,
+  and it can remove a dependency on Python 2.
+  The details on what specifics to set are left to the distributors;
+  though, example specifics could include:
+
+  * Changing ``python`` shebangs to ``python3`` when Python 3.x is supported.
+  * Changing ``python`` shebangs to ``python2`` when Python 3.x is not yet
+    supported.
+  * Changing ``python3`` shebangs to ``python3.8`` if the software is built
+    with Python 3.8.
+
 * When a virtual environment (created by the PEP 405 ``venv`` package or a
   similar tool such as ``virtualenv`` or ``conda``) is active, the ``python``
   command should refer to the virtual environment's interpreter and should
@@ -175,13 +183,11 @@ Current Rationale
 =================
 
 As of 2019, activating a Python virtual environment (or its functional
-equivalent) prior to script execution is one way to obtain a consistent cross-platform and
-cross-distribution experience.
+equivalent) prior to script execution is one way to obtain a consistent
+cross-platform and cross-distribution experience.
 
-Accordingly, publishers can expect users of the software to provide a suitable execution environment.
-availability of such an environment to get their software working
-correctly, and push responsibility for handling other environments
-(such as system Python installations) onto consumers of the software.
+Accordingly, publishers can expect users of the software to provide a suitable
+execution environment.
 
 As part of that however, it is also appropriate to give Python
 distributors the flexibility they need in order to make the
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -78,23 +78,42 @@ For distributors
 For developers
 --------------
 
-* In order to tolerate differences across platforms, new code that needs
-  to invoke the Python interpreter should not specify ``python``, but rather
-  should specify either ``python3`` or ``python2`` (or the more specific
-  ``python3.x`` and ``python2.x`` versions; see the `Migration Notes`_).
-  This distinction should be made in shebangs, when invoking from a shell
-  script, when invoking via the system() call, or when invoking in any other
-  context.
 * When reinvoking the interpreter from a Python script, querying
   ``sys.executable`` to avoid hardcoded assumptions regarding the
   interpreter location remains the preferred approach.
-* Scripts that are deliberately written to be source compatible with both
-  Python 3.x and 2.x should use ``python3`` on their shebang line
-  (see `Rationale`_ for details).
-* One exception to this is scripts that are deliberately written to be used
-  on an “*old system*” where ``python3`` (or even ``python2``) is not available
-  (such as the default Python installed on macOS or RHEL 6).
-  Such scripts may continue to use ``python`` on their shebang line.
+* While far from being universally available, ``python`` remains the
+  preferred spelling for explicitly invoking Python, as this is the
+  spelling that virtual environments make consistently available
+  across different platforms and Python installations.
+* In shebang lines, the preferred spelling is ``/usr/bin/env python``,
+  as this instructs the script to respect the active virtual environment.
+* For Python 3 only scripts that do not support being executed on Python
+  2 at all, it is recommended to instead use ``python3`` and
+  ``/usr/bin/env python3``, as these will never invoke Python 2, and are
+  expected to work for both Python 3 virtual environments and Python 3
+  system installations
+* For Python 2 only scripts that do not support being executed on Python
+  3 at all, it is recommended to instead use ``python2`` and
+  ``/usr/bin/env python2``, as these will never invoke Python 3, and are
+  expected to work for Python 2 virtual environments and at least some
+  Python 2 system installations.
+* In cases where the script is expected to be executed outside virtual
+  environments, developers will need to be aware of the following
+  discrepancies across platforms and installation methods:
+
+  * Older Linux distributions will provide a ``python`` command that
+    refers to Python 2. Most of these distros do *not* provide a
+    ``python2`` command.
+  * Some newer Linux distributions will provide a ``python`` command that
+    refers to Python 3
+  * Some Linux distributions will not provide a ``python`` command at
+    all by default, but will provide a ``python3`` command by default
+
+* When potentially targeting these environments, developers may either
+  use a Python package installation tool that rewrites shebang lines for
+  the installed environment, provide instructions on updating shebang lines
+  interactively, or else use more specific shebang lines that are
+  tailored to the target environment.
 * Scripts targeting both “*old systems*” and systems without the default
   ``python`` command need to make a compromise and document this situation.
   Avoiding shebangs (via the console_scripts Entry Points ([9]_) or similar
@@ -105,9 +124,9 @@ For developers
 
 These recommendations are the outcome of the relevant python-dev discussions
 in March and July 2011 ([1]_, [2]_), February 2012 ([4]_),
-September 2014 ([6]_), discussion on GitHub in April 2018 ([7]_)
-and on python-dev in February 2019 ([8]_).
-
+September 2014 ([6]_), discussion on GitHub in April 2018 ([7]_),
+on python-dev in February 2019 ([8]_), and during the PEP update review
+in May 2019 ([10]_).
 
 
 History of this PEP
@@ -137,11 +156,10 @@ However, these recommendations implicitly assumed that Python 2 would always be
 available. As Python 2 is nearing its end of life in 2020 (PEP 373, PEP 404),
 distributions are making Python 2 optional or removing it entirely.
 This means either removing the ``python`` command or switching it to invoke
-Python 3, invalidating respectively the first or second recommendation.
-Also, some distributors decided that their users are better served by
-ignoring the PEP's recommendations, making the PEP's nominally
-cross-platform recommendations on ``python`` and ``python2`` in shebangs
-increasingly unreliable.
+Python 3. Some distributors also decided that their users were better served by
+ignoring the PEP's recommendations, and provided system administrators with the
+freedom to configure their systems based on the needs of their particular
+environment.
 
 
 .. _rationale:
@@ -149,23 +167,19 @@ increasingly unreliable.
 Current Rationale
 =================
 
-As of 2019, nearly all new systems include Python 3 and the ``python3``
-command. This makes the ``python3`` command the best general choice for
-code that can run on either Python 3.x or 2.x, even though it is not
-available everywhere.
+As of 2019, activating a Python virtual environment (or its functional
+equivalent) is the best way to obtain a consistent cross-platform and
+cross-distribution experience.
 
-The recommendation is skewed toward current and future systems, leaving
-behind “*old systems*” (like RHEL 6 or default Python installed on macOS).
-On these systems, Python software is rarely updated and any recommendations
-this PEP makes would likely be ignored.
+Accordingly, it is entirely reasonable for publishers to expect the
+availability of such an environment to get their software working
+correctly, and push responsibility for handling other environments
+(such as system Python installations) onto consumers of the software.
 
-Also, since distributors often ignored recommendations the PEP gave
-regarding the ``python`` command (for what they saw as legitimate special
-needs), this PEP now gives them broad control over the command.
-Correspondingly, users are advised to not use the ``python`` command
-in cross-platform code.
-Instead, this PEP specifies the expected behavior of the ``python3`` and
-``python2`` commands, which is not controversial.
+As part of that however, it is also appropriate to give Python
+distributors the flexibility they need in order to make the
+behaviour of their systems as similar as possible to the behaviour
+of an activated virtual environment.
 
 
 Future Changes to this Recommendation
@@ -349,6 +363,9 @@ References
 
 .. [9] The console_scripts Entry Point
    (https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point)
+
+.. [10] May 2019 PEP update review
+   (https://github.com/python/peps/pull/989)
 
 
 Copyright
--- a/pep-0394.txt
+++ b/pep-0394.txt
@@ -39,31 +39,38 @@ interpreter invoked by the ``python`` command.
 Recommendation
 ==============
 
+
+Our recommendations are detailed below. We call out any expectations that these recommendations are based upon.
+
 For distributors
 ----------------
 
-* Unix-like software distributions (including systems like macOS and
+* We expect Unix-like software distributions (including systems like macOS and
   Cygwin) should install the ``python2`` command into the default path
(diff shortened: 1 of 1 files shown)


pep-0394.txt
```diff
@@ -5,104 +5,167 @@ Last-Modified: $Date$
...
+* If the ``python`` command is installed, it should either invoke the same
+  version of Python as the ``python3`` command or as the ``python2``
+  command.
+* Distributors may choose to set the behavior of the ``python`` command,
+  not provide it at all, or make it configurable by the user or
   system administrator.
-* In order to tolerate differences across platforms, all new code that needs
+* The Python 3.x ``idle``, ``pydoc``, and ``python-config`` commands should
+  likewise be available as ``idle3``, ``pydoc3``, and ``python3-config``
+  (and similarly for Python 2.x). The original commands should either invoke
```
warsaw: "and similarly for Python 2.x" is a bit ambiguous.  Are you suggesting that if distributors include these tools for Python 2, that they be named `idle2`, `pydoc2`, and `python2-config`?  Perhaps it's better to be explicit here.

hroncok: abf3bf3

pep-0394.txt
```diff
@@ -175,14 +236,13 @@ making such a change.
...
-* If the Python 2 interpreter becomes uncommon, scripts should nevertheless
+* When the Python 2 interpreter becomes uncommon, scripts should nevertheless
   continue to use the ``python3`` convention rather that just ``python``. This
   will ease transition in the event that yet another major version of Python
   is released.
 * If these conventions are adhered to, it will become the case that the
   ``python`` command is only executed in an interactive manner as a user
-  convenience, or to run scripts that are source compatible with both Python
-  2 and Python 3.
+  convenience.
```
ncoghlan: extra clause: "... convenience, or else when using a virtual environment or similar mechanism."

pep-0394.txt
```diff
@@ -175,14 +236,13 @@ making such a change.
   versa. That way, if a sysadmin does decide to replace the installed
   ``python`` file, they can do so without inadvertently deleting the
   previously installed binary.
-* If the Python 2 interpreter becomes uncommon, scripts should nevertheless
+* When the Python 2 interpreter becomes uncommon, scripts should nevertheless
```
ncoghlan: We should reword this bullet point, since we now expect Python 3 -> Python 4 to be a "enough changes have accumulated since Python 3.0 that they're effectively different languages" kind of event, rather than a Python 3.0 style compatibility break. Something like:

> Even as the Python 2 interpreter becomes less common, scripts should nevertheless continue to use the `python3` convention rather than just `python`. In the event of a future Python 4.0 release, distributors may then choose to simply alias `python3` to the same binary as `python4` (since the degree of change is now expected to be the same as in any other Python feature release), or else they may choose to maintain parallel `python3` and `python4` stacks for a time (depending on the needs of their specific audience).

hroncok: I'd happy to change this, however is there some authoritative source I can link to? I don't want this PEP to be the place that promises/defines/guesses the plans for Python 4. If not, we could also reword this to work either way. Something like: "When Python 2 is forgotten, still use "python3". If Python 4 will be a disaster, it will help you, OTOH if Python 4 will be peace of cake, you can either alias "python3" to "python4" or decide to keep them parallel installable for a time."

pep-0394.txt
```diff
@@ -5,104 +5,167 @@ Last-Modified: $Date$
...
-  Python 2 only scripts should either be updated to be source compatible
-  with Python 3 or else to use ``python2`` in the shebang line.
-
+* Distributions can choose whether ``python`` is installed and, if it is,
+  whether it refers to the same target as ``python2`` or ``python3``.
+* End users should be aware that the ``python`` command is inconsistent
+  across Unix-like systems.
+* In activated virtual environments, ``python`` always refers to the
+  environment's Python version.
+* Upstream projects should prefer the ``python3`` shebang even if they support
```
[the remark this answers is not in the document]
hroncok: 4a4cdf7c

pep-0394.txt
```diff
@@ -5,104 +5,167 @@ Last-Modified: $Date$
...
+   Python 2 and 3 (since it was available on all systems, even those that
+   already aliased it to Python 3).
+2. The ``python`` command should always invoke Python 2 (to prevent
+   hard-to-diagnose errors when Python 2 code is run on Python 3).
+3. The ``python2`` and ``python3`` commands should be available to specify
+   the version explicitly.
+
+However, these recommendations implicitly assumed that Python 2 would always be
+available. As Python 2 is nearing its end of life in 2020 (PEP 373, PEP 404),
+distributions are making Python 2 optional or removing entirely.
```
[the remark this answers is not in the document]
hroncok: 4a4cdf7c

pep-0394.txt
```diff
@@ -5,104 +5,167 @@ Last-Modified: $Date$
...
+This means either removing the ``python`` command or switching it to invoke
+Python 3, invalidating respectively the first or second recommendation.
+Also, some distributors decided that their users are better served by
+ignoring the PEP's recommendations, making the PEP's supposedly
```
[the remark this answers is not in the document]
hroncok: 4a4cdf7c

pep-0394.txt
hroncok: 4a4cdf7c

pep-0394.txt
brettcannon: There is no authoritative statement beyond some email threads where that seems to be the prevailing feeling towards Python 4.

hroncok: That's my understanding as well. I don't want this PEP to be used as a reference about Python 4. I'll try to go with a statement that is more universal.

hroncok: 36e61fab

gvanrossum:
You should probably get the Steering Council to appoint a BDFL-Delegate who can approve it. Last time we had MvL as PEP Czar but he's not around, and BDFL-Delegate is the closest thing. I'm not touching this myself. :-)

encukou:
Thanks for the advice! just before I read your comment, I asked for, essentially, just that :)
https://github.com/python/steering-council/issues/8

warsaw:
Since Nick, Petr, and myself are all authors, who would we get to be a BD?

gvanrossum:
The SC could deliberate it for a bit during our next meeting.

ncoghlan:
Definitely an SC level discussion (and it probably makes sense for us to just review the PR changes directly).

Some potentially useful context for that discussion:

* https://fedoraproject.org/wiki/Changes/Mass_Python_2_Package_Removal (Fedora 30 is feature complete and will be released in the next few weeks)
* https://fedoraproject.org/wiki/Changes/F31_Mass_Python_2_Package_Removal (Fedora 31 is scheduled to be released in October
* https://fedoraproject.org/wiki/Changes/RetirePython2 (Fedora 32 doesn't have a final schedule yet, but late April/early May is the usual time frame - the key point for our purposes is that it will be the first Fedora release after Python 2 goes completely end of life upstream)

hroncok:
To complete the list of Fedora changes posted by Nick:

My idea is to also include a Fedora 31 change that moves `/usr/bin/python` (and others, such as `nose`, `pytest`, `pip`) to Python 3 (with possible deferral to F32), together with package provides (`dnf install python` to install Python 3, `dnf install python-requests` to install `python3-requests`). I just haven't drafted it yet, waiting for the resolution of this PEP update.

https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3#Phase_3:_Switch_python_to_refer_to_python3

pep-0394.txt
```diff
@@ -5,104 +5,167 @@ Last-Modified: $Date$
...
-  ``python`` command even if ``python2`` is available.
-  (All software in such a controlled environment must use ``python3`` or
-  ``python2`` rather than ``python``, which means scripts that deliberately
-  use ``python`` need to be modified for such environments.)
-* When a virtual environment (created by the PEP 405 ``venv`` package or a
-  similar tool) is active, the ``python`` command should refer to the
-  virtual environment's interpreter. In other words, activating a virtual
-  environment counts as deliberate user action to change the default
-  ``python`` interpreter.
+* Scripts that are deliberately written to be source compatible with both
```
ncoghlan: I think we need to present additional options here, as our actual advice to developers is "don't rely on hardcoded shebang lines, and instead let tools rewrite them for you".

That means the preferred option is "Use a Python package installer that supports shebang line rewriting, like `pip install --user` or `pipx install`, rather than copying around plain text files" (perhaps with a reference to https://packaging.python.org/guides/installing-stand-alone-command-line-tools/ ).

Option 2 is "Use the environment-aware `/usr/bin/env python` shebang line, and require that the script be executed inside an activated Python virtual environment"

Option 3 is "Use the generic `/usr/bin/env python` shebang line, but ensure that will resolve to a suitable interpreter on the systems that you're targeting if the script needs to run outside a virtual environment". Adding in `ln -s ~/.local/bin/python /usr/bin/python3` or `sudo ln -s /usr/local/bin/python /usr/bin/python3` for the target systems that you care about is a much lower impact change than rewriting all your shebang lines to add a version qualifier (especially since making that change will break the scripts on systems that only have Python 2).

The only time that using the `python3` shebang genuinely makes sense is when the script wouldn't run on Python 2 anyway - in those cases, using the more specific shebang only increases your target platform compatibility, it never reduces it.

ncoghlan: If we don't want that level of complexity in the PEP, than I think we should make the guidance explicitly:

- `/usr/bin/env python` for source compatible scripts, or Python 3 scripts that are unlikely to ever be exposed to target environments where that invocation won't work (i.e. Python 3 only systems when outside a virtual environment)
- `/usr/bin/env python3` for scripts that require Python 3 or later
- `/usr/bin/env python2` for scripts that don't support Python 3 at all
- where possible, use a Python package or script installer (e.g. `pip` or `pipx`) instead of copying scripts directly to a target system, as they are able to rewrite shebang lines to suit the target environment

Hardcoding specific paths to Python binaries should be restricted to non-portable scripts that only target particular platforms and don't need to respect virtual environments.

ncoghlan:
@hroncok Ah, nice! I think that would tie in better with my proposed guidance then, as once you've implemented that change, then `/usr/bin/env python` shebangs will start working on Python 3 only Fedora systems, just as they'd work on RHEL 8 systems where an admin has run `alternatives --set python /usr/bin/python3` (as per https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/)

encukou:
About setting `/usr/bin/python`, that article says:
> Note, We do not recommend this approach. We recommend you explicitly refer to python3 or python2. That way, your scripts and commands will work on any machine that has the right version of Python installed.

I still think that is good advice. Setting `/usr/bin/python` is a decision of the sysadmin, not the user, and it's machine-global – if anything on the machine grows to depend on the decision, it practically can't be changed.
Generic scripts are better off using `python3`.

Keep in mind this PEP's advice will be applied to *new* code. New code generally won't be deployed on systems without a `python3` command. Not having `python3` signalizes an old system with a high cost of upgrading (not only the system, but also software on it).

ncoghlan:
I don't think always using "/usr/bin/python3" is generically good advice for new code, since I think requiring the `python3` spelling is ultimately a transient thing - I expect most distros to follow Fedora's approach, and make `python` refer to `python3` by default once they drop their full Python 2 stacks, and I expect we'll eventually change the Windows launcher to at least fall back to Python 3 for unqualified shebang lines, and potentially even prefer it over Python 2.7 once the latter goes end-of-life.

So I think it's entirely reasonable for publishers to continue to use "/usr/bin/env python" as their default approach, and tell anyone with a system where that doesn't work to either:

1. Use an appropriately configured virtual environment; or
2. Set up a shell alias or user level symlink so it works for their personal account; or
3. Convince the system administrators to make it work by default across the entire machine; or
4. Edit the shebang line to be correct for their system before running the script (if editing a shebang line is problematic, folks should be using installers, not downloading scripts directly)

While specifying `/usr/bin/env python3` instead is an entirely reasonable alternative, that offers some potential benefits and is hence well worth mentioning, that's a far cry from telling all current publishers "Your script shebang lines are now wrong, and you should change them" (or, even more irritatingly, encouraging folks to go around submitting PRs to open source projects along the lines of "PEP 394 told me that `/usr/bin/env python` is wrong, so now I am going around and submitting 'Change `python` to `python3` in shebang lines' PRs to every open source Python project I can find").

encukou:
> I think requiring the `python3` spelling is ultimately a transient thing

Is that your personal view, or can the SC agree on that? Last time I suggested this, it was explicitly shot down (https://github.com/python/peps/pull/630#issuecomment-384431259). This PR tries to reflect consensus from the mailing list (as much as possible while giving sound suggestions).

This is really frustrating – it feels like no one has the authoriity to decide, but too many people have a right to veto. And meanwhile, the status quo is inadequate; the PEP needs to be changed now if it's to keep any hope of relevance. This issue is good burnout material.

The explicit `python3` will work *now*. It's good guidance for script authors that want their software to just work on users' machines. 

> Your script shebang lines are now wrong, and you should change them

Would it be OK to just add a disclaimer that while while the `python3` suggestion works, different projects have different needs and other shebangs aren't necessarily wrong?

ncoghlan:
(I'm currently commenting from a personal point of view, but I expect the merge decision on this PR will be made by SC consensus and/or vote)

The key outcome I want to avoid is anyone claiming "PEP 394 says that using `/usr/bin/env python` for source compatible scripts is always wrong", so yeah, I'd be OK with wording that made the points:

* `/usr/bin/env python3` is compatible with virtual environments, Python 3 only systems, and dual stack systems today, so choose this option if compatibility with newer Python 3 only systems is higher priority than compatibility with older Python 2 only systems (even for source-compatible scripts). This is often going to be the case for open source projects.
* `/usr/bin/env python` is compatible with virtual environments, Python 2 only systems and dual stack systems today, and appears likely to become increasingly compatible with Python 3 only systems over time (assuming distros continue on the design trajectories that they're currently on), so choose this option if either compatibility with legacy Python 2 only systems is a high priority, or else if you have existing shebang lines that you'd prefer to leave unmodified for now.

pep-0394.txt
ncoghlan: After discussing this with @encukou in person, I created a PR against the PR branch to better convey what I'm proposing here: https://github.com/hroncok/peps/pull/1/files

ncoghlan:
@gvanrossum @brettcannon @gvanrossum @willingc I think this is ready for broader steering council review now.

General scope of change:
* we grant Python distributors more freedom to either make the ``python`` name refer to the Python 3 runtime (the same way it will in a Python 3 virtual environment), or else make it configurable at the system level (again, more like virtual environments)
* we clarify for publishers some of the trade-offs between the different options for targeting system installations directly

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
-* When a virtual environment (created by the PEP 405 ``venv`` package or a
-  similar tool) is active, the ``python`` command should refer to the
-  virtual environment's interpreter. In other words, activating a virtual
-  environment counts as deliberate user action to change the default
-  ``python`` interpreter.
+* While far from being universally available, ``python`` remains the
+  preferred spelling for explicitly invoking Python, as this is the
+  spelling that virtual environments make consistently available
+  across different platforms and Python installations.
+* In shebang lines, the preferred spelling is ``/usr/bin/env python``,
```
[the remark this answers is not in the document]
hroncok: Note that we proposed this PR without this recommendation so I leave the decision to you and @ncoghlan. I'd happily keep it or revert it back as you see fit.

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
 Resolution: https://mail.python.org/pipermail/python-dev/2012-February/116594.html
 
 
 Abstract
 ========
 
 This PEP provides a convention to ensure that Python scripts can continue to
-be portable across ``*nix`` systems, regardless of the default version of the
-Python interpreter (i.e. the version invoked by the ``python`` command).
+be portable across ``*nix`` systems, regardless of the version of the Python
```
willingc: As a point of clarification, are we considering MacOS as a ``*nix`` system?

willingc: Perhaps:

This PEP outlines the behavior of Python scripts when the ``python`` command is invoked. Depending on a distribution or system configuration, ``python`` may or may not be installed. If ``python`` is installed its target interpreter may refer to ``python2`` or ``python3``. End users may be unaware of this inconsistency across Unix-like systems. This PEP's goal is to reduce user confusion about what ``python`` references and what will be the script's behavior.

The recommendations in the next section of this PEP will outline the behavior when:
- using virtual environments
- cross-platform scripts with shebangs for either ``python2`` or ``python3`` 

The PEP's goal is to clarify the behavior for script end users, distribution providers, and script maintainers / authors.

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+  when Python 3.x is not yet supported, or changing ``python3`` shebangs to
+  ``python3.8`` if the software is built with Python 3.8.
+* When a virtual environment (created by the PEP 405 ``venv`` package or a
+  similar tool such as ``virtualenv`` or ``conda``) is active, the ``python``
+  command should refer to the virtual environment's interpreter and should
+  always be available.
+  The ``python3`` or ``python2`` command (according to the environment's
+  interpreter) should also be available.
+
+For developers
```
willingc: Do we want to split this into script developers/maintainers and script users (who may not be the script's author)?

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+  shebang.
 
 Recommendation
 ==============
 
-* Unix-like software distributions (including systems like Mac OS X and
+For distributors
+----------------
+
+* Unix-like software distributions (including systems like macOS and
```
willingc:
```suggestion
* We expect Unix-like software distributions (including systems like macOS and
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+* Cross-platform scripts should prefer the ``python3`` shebang even if they
+  support both Python 3 and Python 2.
+* Cross-platform scripts only supporting Python 2 should use the ``python2``
+  shebang.
 
 Recommendation
 ==============
 
-* Unix-like software distributions (including systems like Mac OS X and
+For distributors
```
willingc:
```suggestion

Our recommendations are detailed below. We call out any expectations that these recommendations are based upon.

For distributors
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
   interpreter.
-* If the ``python`` command is installed, it should invoke the same version of
-  Python as the ``python2`` command (however, note that some distributions
-  have already chosen to have ``python`` implement the ``python3``
-  command; see the `Rationale`_ and `Migration Notes`_ below).
-* The Python 2.x ``idle``, ``pydoc``, and ``python-config`` commands should
-  likewise be available as ``idle2``, ``pydoc2``, and ``python2-config``,
-  with the original commands invoking these versions by default, but possibly
-  invoking the Python 3.x versions instead if configured to do so by the
+* If the ``python`` command is installed, it should either invoke the same
```
willingc:
```suggestion
* If the ``python`` command is installed, it is expected to invoke either the same
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+  version of Python as the ``python3`` command or as the ``python2``
+  command.
+* Distributors may choose to set the behavior of the ``python`` command,
```
willingc:
```suggestion
* Distributors may choose to set the behavior of the ``python`` command as follows:
    - ``python2``,
    - ``python3``,
    - not provide ``python`` command
    - allow ``python`` to be configurable by an end user or a system administrator.
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
-  compatible with both Python 2.x and 3.x. Such scripts may continue to use
-  ``python`` on their shebang line.
-* When packaging software that is source compatible with both versions,
-  distributions may change such ``python`` shebangs to ``python3``.
-  This ensures software is used with the latest version of
-  Python available, and it can remove a dependency on Python 2.
+* The Python 3.x ``idle``, ``pydoc``, and ``python-config`` commands should
+  likewise be available as ``idle3``, ``pydoc3``, and ``python3-config``;
+  Python 2.x versions as ``idle2``, ``pydoc2``, and ``python2-config``.
+  The original commands should either invoke the same version of Python
```
willingc: The unspecified commands with no version number should ...

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+  as the ``python`` command, or not be available at all.
+* When packaging software, distributors are encouraged to change less specific
```
willingc:
```suggestion
* When packaging third party Python scripts, distributors are encouraged to change less specific
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+  shebangs to more specific ones. This ensures software is used with the latest
```
willingc:
```suggestion
  shebangs to more explicit and specific ones. This encourages scripts to use the latest
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+  version of Python available, and it can remove a dependency on Python 2.
+  The specifics are left on the distributors.
```
willingc:
```suggestion
  The details on what specifics to set are left to the distributors; though, example specifics could include: 
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+  Examples of this include changing ``python`` shebangs to ``python3``
+  when Python 3.x is supported, changing ``python`` shebangs to ``python2``
+  when Python 3.x is not yet supported, or changing ``python3`` shebangs to
+  ``python3.8`` if the software is built with Python 3.8.
+* When a virtual environment (created by the PEP 405 ``venv`` package or a
+  similar tool such as ``virtualenv`` or ``conda``) is active, the ``python``
+  command should refer to the virtual environment's interpreter and should
+  always be available.
+  The ``python3`` or ``python2`` command (according to the environment's
+  interpreter) should also be available.
```
willingc:
```suggestion
  interpreter version) should also be available.
```

pep-0394.txt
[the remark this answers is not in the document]
willingc: This case should be combined with the case on Line 104.

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+
+  * Older Linux distributions will provide a ``python`` command that
+    refers to Python 2. Most of these distros do *not* provide a
+    ``python2`` command.
+  * Some newer Linux distributions will provide a ``python`` command that
+    refers to Python 3
+  * Some Linux distributions will not provide a ``python`` command at
+    all by default, but will provide a ``python3`` command by default
+
+* When potentially targeting these environments, developers may either
```
willingc: These remaining 3 bullets should be moved under the inconclusive ``python`` case as possible resolutions.

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+environment.
+
+
+.. _rationale:
+
+Current Rationale
+=================
+
+As of 2019, activating a Python virtual environment (or its functional
+equivalent) is the best way to obtain a consistent cross-platform and
```
willingc:
```suggestion
equivalent) prior to script execution is one way to obtain a consistent cross-platform and
```

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+cross-distribution experience.
+
+Accordingly, it is entirely reasonable for publishers to expect the
```
willingc:
```suggestion
Accordingly, publishers can expect users of the software to provide a suitable execution environment.

File: pep-0394.txt [hunk 23]
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+availability of such an environment to get their software working
+correctly, and push responsibility for handling other environments
+(such as system Python installations) onto consumers of the software.
+
+As part of that however, it is also appropriate to give Python
```
willingc: I would remove this paragraph.

pep-0394.txt
warsaw: Pedantically, it's "macOS" but I hope the answer is "yes".

pep-0394.txt
hroncok: (Note for self: several lines bellow need to be removed if this goes in).

pep-0394.txt
hroncok: What about simple: The commands with no version number should...

pep-0394.txt
hroncok: *encourages* actually makes me thing it somehow makes it more likely. In fact, it just makes it so. I still consider *ensures* or *makes sure* to fit better here.

pep-0394.txt
hroncok: Note to self: Following lines need to be adjusted if this goes in.

hroncok:
@willingc Thank you very much for you thoughtful suggestions! I'll leave them open for comments (or thumbs ups) by others and incorporate them to the proposal if agreed upon.

pep-0394.txt
```diff
@@ -5,104 +5,181 @@ Last-Modified: $Date$
...
+
+
+.. _rationale:
+
+Current Rationale
+=================
+
+As of 2019, activating a Python virtual environment (or its functional
+equivalent) is the best way to obtain a consistent cross-platform and
+cross-distribution experience.
```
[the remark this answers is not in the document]
ncoghlan: If you're not writing Linux distro packages, you should absolutely be using a venv for everything.

You do need to use `pipx`, and you do need to reinstall them if you change Python versions.

pep-0394.txt
ncoghlan: There aren't any recommendations for script users (unless you count "You may need to edit shebang lines to get scripts to work on your system" as a recommendation)

pep-0394.txt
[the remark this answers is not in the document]
ncoghlan: If you *don't* assume the use of a venv, then everything is awful:

* `python` probably doesn't work on Windows (or runs something random)
* `py` *only* works on Windows
* `python` is only reliably available inside a venv
* `python2` isn't reliably available anywhere
* `python3` doesn't work on Windows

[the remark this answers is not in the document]
ncoghlan: Or, you run them using an explicit interpreter invocation, and forget about relying on shebang lines.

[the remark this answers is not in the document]
warsaw: This all feels like it's outside the scope of this PEP.  With this view, we have to endorse a third party tool (`pipx`) and a deployment workflow, and teach people how to use virtual environments (and do they use `virtualenv` or `$python -m venv`?).  The reinstallation issue is problematic too.  I don't know what any of this has to do with our recommendations to *nix distributions.  If we really want to make these recommendations, we should do that in a separate PEP, IMHO.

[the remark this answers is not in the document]
warsaw: One other thought: If you're using zip applications (pex, shiv), you should absolutely *not* be using venvs or `/usr/bin/env`.

pep-0394.txt
willingc: I think that it's worth, even though the recommendation is short, to split out script end users from script developers.

pep-0394.txt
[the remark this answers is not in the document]
ncoghlan: It's a response to the original draft of this PR, which advised *all* script publishers to unconditionally use `/usr/bin/python3` in their scripts, which I consider to be a bad default recommendation, since it breaks virtual environments.

`/usr/bin/env python` is better generic advice, since it respects virtual environments, and leaves the choice of specifying something else in their shebang line to folks that specifically want to support running outside a venv, and should have a better idea of which kinds of environments they want to target, and hence which compatibility trade-offs they can make in choosing an exact spelling.

Tools like `pipx`, `zipapp`, `shiv` etc all run an installer at some point, so they allow publishers to rely on console script definitions which write the shebang line at install time, rather than needing to rely on a hardcoded shebang line in the source file(s).

The relevance of all this to PEP 394 is that platform providers need to account for publisher behaviour in designing their Python user experience, and that's easier for them to do if our advice to publishers is "assume platforms will either rely on a Python installer, so you can use generated console scripts, or else that they will make themselves look somewhat like a virtual environment".

brettcannon:
I'm too lazy to 👍 everything @willingc said, so consider this a blanket 👍  😄

brettcannon:
And I will wait to read and comment until after this latest round of edits land.

pep-0394.txt
hroncok:
```suggestion
  The commands with no version number should either invoke the same version of Python
```

pep-0394.txt
hroncok: Let me know if you want e to explicitly put this information in.

pep-0394.txt
hroncok: Sorry. I cannot read any more, could you please point em to where would you except these 3 bullets points?

hroncok:
I've added the last round of edits suggested by @willingc. At least most of them.

I'd appreciate if somebody could read the result as a whole and report back if some half unfinished sentences etc. remain, as I'm now blind for such mistakes.

If nobody would do that, I shall return to it tomorrow.

pep-0394.txt
```diff
@@ -5,104 +5,193 @@ Last-Modified: $Date$
...
-  PEP), so ``python`` should be used in the shebang line only for scripts
-  that are source compatible with both Python 2 and 3.
-* in preparation for an eventual change in the default version of Python,
-  Python 2 only scripts should either be updated to be source compatible
-  with Python 3 or else to use ``python2`` in the shebang line.
+The recommendations in the next section of this PEP will outline the behavior
+when:
+
+* using virtual environments
+* cross-platform scripts with shebangs for either ``python2`` or ``python3``
```
encukou:
```suggestion
* writing cross-platform scripts with shebangs for either ``python2`` or ``python3``
```

pep-0394.txt
```diff
@@ -5,104 +5,193 @@ Last-Modified: $Date$
...
+  likewise be available as ``idle3``, ``pydoc3``, and ``python3-config``;
+  Python 2.x versions as ``idle2``, ``pydoc2``, and ``python2-config``.
+  The commands with no version number should either invoke the same version
+  of Python as the ``python`` command, or not be available at all.
+* When packaging third party Python scripts, distributors are encouraged to
+  change less specific shebangs to more specific ones.
+  This ensures software is used with the latest version of Python available,
+  and it can remove a dependency on Python 2.
+  The details on what specifics to set are left to the distributors;
+  though, example specifics could include:
```
encukou:
```suggestion
  though. Example specifics could include:
```

encukou:
It would be nice to hear some guidance from the SC.
The original proposal was distilled from discussions on the mailing list, as far as we could do that. I hope that Nick's changes made the proposal closer to acceptance, but I'm not sure.

Is there anything I can do to move this forward?

pep-0394.txt
```diff
@@ -175,14 +262,17 @@ making such a change.
   versa. That way, if a sysadmin does decide to replace the installed
   ``python`` file, they can do so without inadvertently deleting the
   previously installed binary.
-* If the Python 2 interpreter becomes uncommon, scripts should nevertheless
-  continue to use the ``python3`` convention rather that just ``python``. This
-  will ease transition in the event that yet another major version of Python
-  is released.
+* Even as the Python 2 interpreter becomes less common, scripts should
```
ncoghlan: The "should" wording here doesn't match the revised weaker wording above. I'd suggest:

> Even as the Python 2 interpreter becomes less common, it remains reasonable for scripts to continue to use the `python` convention, rather than just `python`. In the event ...

hroncok: use the python convention, rather than just python? I don't quite understand that bit.

ncoghlan: Sorry, typo - I left out a critical "3" in the original version of the comment :)

ncoghlan:
@encukou Ah, sorry, we never summarised the last SC discussion here. We're basically happy with it now, and Barry has the delegation to push the "Ship it" button once he's happy with the handling of the virtual environment recommendations.

I know that's scope creep from the point of view of your original concerns (i.e. keeping Python upstream, Fedora, and RHEL all heading in a mostly consistent direction), but leaving environment managers out of the discussion leaves an overly large discrepancy between PEP 394 and the guidance in https://packaging.python.org/tutorials/installing-packages/

willingc:
Thanks @ncoghlan for the few tweaks in wording. This looks good to me. Thanks all.

encukou:
> I know that's scope creep from the point of view of your original concerns

That's expected – it's a PEP, not a distro-specific document! It's important to address other use cases, even if I don't have strong opinions on *how* they're addressed.

My concern was that it looked like the discussion is personal comments, and I was unsure if they contribute to making the change acceptable. (We're past brainstorming now and need to consolidate the ideas.)

@hroncok, will you address the latest comments, or should I?

hroncok:
> @hroncok, will you address the latest comments, or should I?

I don't have a strong opinion. I you can do it now, please do.
I've invited you to my fork.

encukou:
Hopefully this addresses the comments well. Please let me know if there's something I missed.

pep-0394.txt
```diff
@@ -141,6 +140,19 @@ For developers
...
+For end users of Python
+-----------------------
+
+* While far from being universally available, ``python`` remains the
+  preferred spelling for explicitly invoking Python, as this is the
+  spelling that virtual environments make consistently available
+  across different platforms and Python installations.
+* For software that is not distributed with (or developed for) your system,
+  we recommend using a virtual environment, possibly with an environment
+  manager like ``conda`` or ``pipenv``, to help avoid disrupting your system
```
willingc: Should we mention the built-in `venv`?

ncoghlan: Perhaps in parentheses in the first bullet point? "... that virtual environments (as created by the standard library's `venv` module or the PyPA's `virtualenv` project) consistently ..."

brettcannon:
@encukou the shebang line question doesn't seem to be resolved yet.

encukou:
@warsaw, what should the preferred shebang line be?

ncoghlan:
Ping @warsaw. This is just waiting on the question of whether or not you're happy for the stance the PEP takes to be "Unless they specifically know otherwise, script publishers only need to worry about end users that are either going to be using a virtual environment, or else are going to be running a Python package installer".

I think that's the only reasonable stance we can take - at this point, shebang lines are too inconsistent across different platforms for us to do anything else. If you can narrow your focus to a specific subset of target platforms, then there are alternatives available, but that's guidance that documentation for those specific environments should be providing, not our generic upstream suggestions.

encukou:
ping @warsaw. Care to reply? Anything I can do to make your job easier?

brettcannon:
I'm going to put this on the steering council meeting agenda until Barry attends and notices this PR. 😄

encukou:
We have a deadline coming up this Wednesday for Fedora 31 (scheduled to live 2019-10 to 2020-11). I'd like to let users have a working `python` command there by default.

Since this PR's recommendations for distros are, as far as I know, agreed on, I'm tempted to switch `python` to `python3` even if the PEP is not updated in time.

ncoghlan:
@encukou Aye, that makes sense, as the only thing holding this up are the remaining concerns regarding `/usr/bin/env python` as a default shebang recommendation for script publishers, even though it's the only venv-compatible option.

We also have a steering council meeting tomorrow morning, so I'll try to get it resolved then - it's the classic problem of "Most software publishers are *not* system integrators, so what's right for system integrators shouldn't affect the default recommendations" :)

pep-0394.txt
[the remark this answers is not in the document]
warsaw: I want to say that a general `/usr/bin/env python` recommendation outside of venvs is very dangerous.  When I was working on Debian/Ubuntu, such things lead to mysterious problems with serious ramifications.  So we went about actively discouraging such shebangs.

I think we need help from tooling here.  Ideally, end users wouldn't have to worry about it at all, and either the platform builds would modify the shebang (as I *think* but don't really remember any more whether Debuntu does this), or installing into a venv would rewrite the shebang.

So I'm quite concerned that a general recommendation will cause users serious problems.  If we want to make progress on this update and can't agree on shebang recommendations, we should remove it from this update and discuss further in a separate PR, bug, or forum.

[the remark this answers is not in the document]
Rosuav: @warsaw Is there a difference here between system scripts and those distributed by third parties? If Debian ships a tool that happens to be written in Python (eg /usr/bin/calibre), it makes sense for its shebang to avoid /usr/bin/env, but if I'm going to give people a script that depends on Python, it makes more sense to let their environment select it.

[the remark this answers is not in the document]
warsaw: @Rosuav That's exactly the situation that broke people on Ubuntu.  They'd end up using some random Python they forgot they installed, with an incompatible library pip installed and it broke their script.  Then they blamed the wrong people, when really they should have pointed their script to the right Python executable, or fixed their `$PATH` correctly.  Okay, sure this is user error, but I know with a PEP 394 recommendation to use `/usr/bin/env`, it'll happen more often.  It might even be that they *think* they're running in a venv, but really they aren't.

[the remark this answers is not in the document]
hroncok: Note that the original PEP nor our PR have ever gone into `/usr/bin/python...` vs `/usr/bin/env python...` shebangs. What if we keep it that way and than we can argue about that in another PR?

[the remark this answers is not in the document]
warsaw: +1 - This is the only controversial bit still left IIRC.  I'm happy to defer this and push out an update with the bits we do agree with.  @encukou @ncoghlan Would that be okay with you?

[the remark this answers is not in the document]
encukou: Yes!

pep-0394.txt
ncoghlan: I'm going to leave this for now, as these do get mentioned further up.

ncoghlan:
Huh, since I deleted/edited the offending line I can't find a way to get back to the "single comment" thread that @warsaw started on the shebang line recommendation.

Anyway, https://github.com/python/peps/pull/989/commits/4b20f8c130df2c6b9b1a9900ea41eca51999ee05 changes the mention of `/usr/bin/env python` to be specific to activated virtual environments (based on the 2019-06-26 SC discussion).

pep-0394.txt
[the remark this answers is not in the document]
ncoghlan: Resolved by removing the "preferred" wording, and instead qualifying this bullet point as being specific to scripts that are only expected to be run in virtual environments: https://github.com/python/peps/commit/4b20f8c130df2c6b9b1a9900ea41eca51999ee05

I then removed the two paragraphs about Py2 vs Py3, since they weren't needed any more, and the text flows better without them.

pep-0394.txt
```diff
@@ -5,104 +5,196 @@ Last-Modified: $Date$
...
-* When a virtual environment (created by the PEP 405 ``venv`` package or a
-  similar tool) is active, the ``python`` command should refer to the
-  virtual environment's interpreter. In other words, activating a virtual
-  environment counts as deliberate user action to change the default
-  ``python`` interpreter.
+* Encourage your end users to use a virtual environment.
+  This makes the user's environment more predictable (possibly resulting
+  in fewer issues), and helps avoid disrupting their system.
+* For scripts that are only expected to be run in an activated virtual
+  environment, shebang lines can be written as ``/usr/bin/env python``,
```
hroncok:
```suggestion
  environment, shebang lines can be written as ``#!/usr/bin/env python``,
```

pep-0394.txt
```diff
@@ -5,104 +5,196 @@ Last-Modified: $Date$
...
+  as this instructs the script to respect the active virtual environment.
+* In cases where the script is expected to be executed outside virtual
+  environments, developers will need to be aware of the following
+  discrepancies across platforms and installation methods:
+
+  * Older Linux distributions will provide a ``python`` command that
+    refers to Python 2. Most of these distros do *not* provide a
+    ``python2`` command.
+  * Some newer Linux distributions will provide a ``python`` command that
+    refers to Python 3
```
hroncok:
```suggestion
    refers to Python 3.
```

pep-0394.txt
```diff
@@ -5,104 +5,196 @@ Last-Modified: $Date$
...
+  * Some Linux distributions will not provide a ``python`` command at
+    all by default, but will provide a ``python3`` command by default
```
hroncok:
```suggestion
    all by default, but will provide a ``python3`` command by default.
```

pep-0394.txt
```diff
@@ -5,104 +5,196 @@ Last-Modified: $Date$
...
+  manager like ``conda`` or ``pipenv``, to help avoid disrupting your system
+  Python installation.
+
 
 These recommendations are the outcome of the relevant python-dev discussions
 in March and July 2011 ([1]_, [2]_), February 2012 ([4]_),
-September 2014 ([6]_), and discussion on GitHub in April 2018 ([7]_).
+September 2014 ([6]_), discussion on GitHub in April 2018 ([7]_),
+on python-dev in February 2019 ([8]_), and during the PEP update review
+in May 2019 ([10]_).
```
hroncok:
```suggestion
in May/June 2019 ([10]_).
```

hroncok:
The GitHub interface had some whitespace problem and I've broke the RST syntax, Sorry about that, should be good now.

pep-0394.txt
```diff
@@ -5,104 +5,196 @@ Last-Modified: $Date$
...
+  in fewer issues), and helps avoid disrupting their system.
+* For scripts that are only expected to be run in an activated virtual
+  environment, shebang lines can be written as ``#!/usr/bin/env python``,
+  as this instructs the script to respect the active virtual environment.
+* In cases where the script is expected to be executed outside virtual
+  environments, developers will need to be aware of the following
+  discrepancies across platforms and installation methods:
+
+  * Older Linux distributions will provide a ``python`` command that
+    refers to Python 2. Most of these distros do *not* provide a
```
warsaw: The second sentence feels a little too imprecise.  What exactly is an "older Linux distribution" and how many of them count as "most"?  Perhaps:

> Older Linux distributions will provide a ``python`` command that refers to Python 2, and will likely not provide a ``python2`` command

pep-0394.txt
```diff
@@ -175,14 +265,17 @@ making such a change.
   versa. That way, if a sysadmin does decide to replace the installed
   ``python`` file, they can do so without inadvertently deleting the
   previously installed binary.
-* If the Python 2 interpreter becomes uncommon, scripts should nevertheless
-  continue to use the ``python3`` convention rather that just ``python``. This
-  will ease transition in the event that yet another major version of Python
-  is released.
+* Even as the Python 2 interpreter becomes less common, it remains reasonable
+  for scripts to continue to use the ``python3`` convention, rather than just
+  ``python``. In the event of a future Python 4.0 release, distributors may
```
warsaw: I'd suggest leaving out any mention of Python 4.  1) it can create FUD; 2) it's entirely vaporware; 3) we don't even know whether there *will* be a Python 4 or what it will look like.  We can always add a Python 4 recommendation when the time comes.

willingc: Agreed. Good catch @warsaw

encukou:
@ncoghlan I pushed a change addressing the issues above. Do you have any comments on them?

ncoghlan:
LGTM! @warsaw, over to you for the final review & merge :)

hroncok:
Thank you!

PR #990, opened 2019-04-13
PEP 591: Adding a final qualifier to typing

msullivan:
Initial draft of proposal to add a final qualifier to the typing module


pep-0591.rst
```diff
@@ -0,0 +1,293 @@
...
+will not be assigned to. This is a useful feature in several
+situations:
+
+* Preventing unintended modification of module and class level
+  constants and documenting them as constants in a checkable way.
+* Creating a read-only attribute that may not be overridden by
+  subclasses. (``@property`` can make an attribute read-only but
+  does not prevent overriding)
+* Allowing a name to be used in situations where ordinarily a literal
+  is expected (for example as a file name for ``NamedTuple``, a tuple
```
JelleZijlstra: Sorry I'm late to the party, but I'm guessing "file name" is a typo for "field name".

pep-0591.rst
```diff
@@ -0,0 +1,293 @@
...
+
+   class Base:
+       @overload
+       def method(self) -> None: ...
+       @overload
+       def method(self, arg: int) -> int: ...
+       @final
+       def method(self, x=None):
+           ...
+
```
JelleZijlstra: Perhaps say that use of final on a non-method function is an error.

pep-0591.rst
```diff
@@ -0,0 +1,293 @@
...
+   class Base:
+       DEFAULT_ID: Final = 0
+
+   RATE = 300  # Error: can't assign to final attribute
+   Base.DEFAULT_ID = 1  # Error: can't override a final attribute
+
+   for x in [1, 2, 3]:
+      FOO: Final = x  # Error: Cannot use Final inside a loop
+
+error: Cannot use Final inside a loop
```
JelleZijlstra: Stray line

sproshev:
Is it possible to use `Final` in type comments?

ilevkivskyi:
Yes.

gvanrossum:
Maybe there should be an example of that somewhere? (At least in the PEP? And even Python 3 code might want to use type comments if it's straddling with 2.7.)

msullivan:
I'll make a PR soon

sproshev:
In the following sample mypy does not raise an error for `y`:

```
from typing_extensions import Final

x: Final[int] = 1
x: Final[str] = "str"

def foo():
    y: Final[int] = 1
    y: Final[str] = "str"
```

Is it a mypy bug or this case is allowed by PEP?

ilevkivskyi:
@sproshev How about adding an annotation to `foo()` or using `--check-untyped-defs`?

Also please don't spam in this PR. If you have questions, ask them on typing Gitter.

PR #991, opened 2019-04-13
PEP 589: Minor typo/grammar/wording/markup fixes

gvanrossum:
@JukkaL please review

### DIFF
--- a/pep-0589.rst
+++ b/pep-0589.rst
@@ -15,7 +15,7 @@ Abstract
 ========
 
 PEP 484 [#PEP-484]_ defines the type ``Dict[K, V]`` for uniform
-dictionaries, where each value has the same type and arbitrary key
+dictionaries, where each value has the same type, and arbitrary key
 values are supported.  It doesn't properly support the common pattern
 where the type of a dictionary value depends on the string value of
 the key.  This PEP proposes a type constructor ``typing.TypedDict`` to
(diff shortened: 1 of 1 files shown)


gvanrossum:
@JukkaL: I also have the following (somewhat) more substantial editorial questions/suggestions:

> PEP: 589

I wonder if this PEP should be updated to mention that whenever
"literal" values are required (esp. keys and totality), Final values
of the right type are also acceptable?  Note that Final is now
described in PEP 591.  (Not sure if there are applications of Literal,
PEP 586, that matter -- though if the type of a variable is
Literal['x'] maybe it could be accepted for a key named x'.)

> * A TypedDict type with required keys is not consistent with a
>   TypedDict type with non-required keys, since the latter allows keys
>   to be deleted.  Example where this is relevant::

Does this only apply for matching keys?  (For the first bullet that's
the case, and there it's obvious from the meaning of invariance.  But
for this bullet it's not so clear.)

> * Operations with arbitrary ``str`` keys (instead of string literals
>   or other expressions with known string values) should be rejected.
>   This involves both destructive operations such as setting an item
>   and read-only operations such as subscription expressions.

Is td.get(var) disallowed (where var: str)?  Explain why (not).

> The mypy [#mypy]_ type checker supports TypedDict types. A reference
> implementation of the runtime component is provided in the
> ``mypy_extensions`` [#mypy_extensions]_ module.

Maybe mention that we intend to add it to `typing_extensions` too?  (Do we?)
(It's implied that it'll be added to `typing`.)

> * There is no way to individually specify whether each key is required
>   or not.  No proposed syntax was clear enough.

Perhaps mention that the need for this is presumed to be not that great?

JukkaL:
> I wonder if this PEP should be updated to mention that whenever
"literal" values are required (esp. keys and totality), Final values
of the right type are also acceptable?

This would be reasonable, and I think that mypy already supports this, at least to some extent.

> Does this only apply for matching keys? 

Yes. The wording is unclear.

> Is td.get(var) disallowed (where var: str)? Explain why (not).

This is allowed, and the inferred type is `object`. The reason is that this can't cause a type safety violation, and this can be useful when doing introspection-style operations. Similarly `var in td` is allowed.

> Maybe mention that we intend to add it to typing_extensions too? (Do we?)

Yes, a good point.

> Perhaps mention that the need for this is presumed to be not that great?

A good idea.

I'll create a PR that addresses the above issues.

PR #995, opened 2019-04-16
PEP 586: Make backwards compatibility best-effort

Michael0x2a:
This pull request relaxes PEP 586 so that type checkers are now expected to maintain backwards compatibility on a best-effort basis, rather than a mandatory basis (as per the earlier discussion on typing-sig).

It also rearranges the order in which information is presented: the section now opens with and focuses on an example of what a too-disruptive inference strategy would look like.


mrkmndz:
Looks good to me!

PR #996, opened 2019-04-16
PEP 586: Add section about Final

Michael0x2a:
This commit adds a section discussion how Literal types interact with Final from PEP 591, if both PEPs are accepted. In short, it states that type checkers are expected to understand that assignment statements of the form `var: Final = value` are a way of declaring `var` to effectively be of type `Literal[value]` in certain cases.

### DIFF
--- a/pep-0586.rst
+++ b/pep-0586.rst
@@ -624,11 +624,11 @@ example, whether or not the following program type checks is left unspecified::
 
     # Note: The assignment does not exactly match the form 'var: Final = value'.
     bar1: Final[int] = 3
-    expects_three(bar1)    # ??
+    expects_three(bar1)  # May or may not be accepted by type checkers
 
     # Note: "Literal[1 + 2]" is not a legal type.
     bar2: Final = 1 + 2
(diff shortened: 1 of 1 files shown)


pep-0586.rst
```diff
@@ -589,6 +592,43 @@ involving Literal bools. For example, we can combine ``Literal[True]``,
...
+of the form ``var: Final = value`` where ``value`` is a valid parameter for
+``Literal[...]``, type checkers should understand that ``var`` may be used in
+any context that expects a ``Literal[value]``.
+
+Type checkers are not obligated to understand any other uses of Final. For
+example, whether or not the following program type checks is left unspecified::
+
+    # Note: The assignment does not exactly match the form 'var: Final = value'.
+    bar1: Final[int] = 3
+    expects_three(bar1)    # ??
```
ilevkivskyi: Two suggestions here (and same few lines below):
* Use two spaces before comment
* The comment should say something concrete like "May or may not be accepted by type checkers" instead of just "??"

PR #1006, opened 2019-04-19
PEP 573: Defer to Python 3.9

encukou:
I'd like to target the module state access PEP to Python 3.9, and focus on C calling convention PEPs (580 & 590) for Python 3.8.
cc @ericsnowcurrently @Dormouse759 @ncoghlan


ericsnowcurrently:
I have no objections.  PEP 554 is likely moving to 3.9 as well.

PR #1017, opened 2019-04-29
PEP 574: Final updates

pitrou:
Reconcile the PEP contents with the current implementation.


pitrou:
@ncoghlan @pierreglaser This makes the PEP ready to review / pronounce on as far as I'm concerned. The updated backport is also available at https://pypi.org/project/pickle5/

ncoghlan:
@pitrou Nice! The web copy should update soon, so if you post a review request thread to python-dev, I'll accept the PEP :)

PR #1026, opened 2019-05-06
PEP 590: Swap tp_vectorcall_offset and tp_vectorcall slots.

markshannon:
Will allow Cython to use vectorcall in versions before 3.8.


encukou:
Let's also copy the rationale over from PEP 580: https://www.python.org/dev/peps/pep-0580/#replacing-tp-print

jdemeyer:
Please CC me in all future changes regarding PEP 590.

PR #1040, opened 2019-05-09
PEP 457: Remove implicit proposals to support "optional groups".

larryhastings:


larryhastings:
There was a bunch of nonsense in PEP 457 where I wandered into talking about "optional groups".  I want to remove that for posterity.  Doing so left some unfortunate gaps in the concepts I was expressing, so I actually added a little.  Given that we live in a post-PEP-570 world, it's a little? strange to go editing this PEP to _add_ stuff.  But I haven't read 570 in months, and I've forgotten it, so conceptually maybe I still live in a pre-PEP-570 world.  Anyway I just want to get 457 to be internally consistent, and not talk about optional groups, and I'll be happy--we can slap some tags on it like "Final" and "Informational" and move on with our lives.

PR #1045, opened 2019-05-11
PEP 427: specify presence of directory entries.

jaraco:
Reference pypa/packaging-problems#212 and pypa/wheel#294.

cc: @pfmoore @agronholm

### DIFF
--- a/pep-0427.txt
+++ b/pep-0427.txt
@@ -153,6 +153,17 @@ The last three components of the filename before the extension are
 called "compatibility tags."  The compatibility tags express the
 package's basic interpreter requirements and are detailed in PEP 425.
 
+Directory Entries
+'''''''''''''''''
+
+The wheel file should include directory entries for all directories
+in the archive, including empty directories and directories containing
+other entries. Directory entries should follow the convention of
+a trailing slash following the name with a zero-byte payload.
+
+Wheels without directory entries are still considered valid but those
+with directory entries are preferred.
+
 Escaping and Unicode
 ''''''''''''''''''''
 


jwodder:
A line should probably be added about whether directories entries should/may/may not be listed in `RECORD`.  The PEP currently says that `RECORD` is a list of the files in the wheel, but does that mean "regular files" or "files in the *nix sense" (in which all filesystem nodes are files)?

agronholm:
I'm not sure if this could or should be shoehorned into this PR, but it would be nice to clarify what sorts of attributes are allowed to be included in the wheels, in light of issues like https://github.com/pypa/wheel/issues/291.

pfmoore:
@jaraco can you also point this change out on distutils-sig and/or Discourse, so that interested parties are aware of it. If you can add a description of the reason for wanting this change there as well, that would be good.

I'll give my comments on the actual change in response to your posting rather than here, to save having to repeat myself :-)

jaraco:
[Posted here](https://discuss.python.org/t/refine-wheel-spec-to-designate-directory-entries/1667).

/spend 45 min

ncoghlan:
I'm going to close this as an in-place modification to the 1.0 version of the wheel spec due to the fact that:

* consumers still need to deal with the entries being missing (as Paul noted in the Discourse thread)
* making their inclusion official means dealing with their RECORD entries as jwodder noted above and Daniel Holth noted in the Discourse thread

A wheel 1.1 spec could answer those questions and *permit* them in wheel archives, while a more [fleshed out specification](https://github.com/pypa/packaging-problems/issues/244#issuecomment-487628373) for "pyz" files (that allowed them to be used as regular sys.path entries by omitting their top-level `__main__.py` file) could *require* them.

PR #1056, opened 2019-05-17
PEP 587 version 4

vstinner:
Work in progress.

### DIFF
--- a/pep-0587.rst
+++ b/pep-0587.rst
@@ -15,20 +15,27 @@ Abstract
 Add a new C API to configure the Python Initialization providing finer
 control on the whole configuration and better error reporting.
 
-It becomes possible to read the configuration and modify it before it is
-applied. It also becomes possible to completely override how Python
-computes the module search paths (``sys.path``).
+It becomes possible to read the configuration and then override some
+computed parameters before it is applied. It also becomes possible to
+completely override how Python computes the module search paths
+(``sys.path``).
 
-Building a customized Python which behaves as regular Python becomes
-easier using the new ``Py_RunMain()`` function. Moreover, command line
-arguments passed to ``PyConfig.argv`` are now parsed the same way the
-regular Python parses command line options, and ``PyConfig.xoptions`` are
-handled as ``-X opt`` command line options.
+The new `Isolated Configuration`_ provides sane default values to
+isolate Python from the system. For example, to embed Python into an
+application. Using the environment are now opt-in options, rather than
+an opt-out options. For example, environment variables, command line
+arguments and global configuration variables are ignored by default.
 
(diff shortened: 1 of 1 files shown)


vstinner:
@zooba @Yhg1s: I'm sharing this PR with you to show you what I'm currently working on, but this PR is not ready for a review.

The main change is that there are now 2 well separated default configuration:

* `PyConfig_InitPythonConfig()` behaves as the regular Python: configure the locale and C streams, parse the command line arguments, read environment variables, etc.
* `PyConfig_InitIsolatedConfig()` isolates Python from the system: ignore environment variable, ignore command line arguments, ignore global configuration variables, etc.

I think that I will remove `PyConfig_INIT` from the PEP, but only use use it **internally** for backward compatibility in Python 3.7 API: `Py_Initialize()` and `Py_Main()`.

I think that `PyConfig_InitPythonConfig()` should ignore global configuration variables as well. If you want to use them, use the Python 3.7 API: `Py_Initialize()`. The new API would ignore them.

vstinner:
Follow-up of discussion started at https://bugs.python.org/issue36945#msg342732 with Steve.
 
Victor:
> * customized Python interpreter: behaves as the "regular Python"

Steve:
> For this, why wouldn't we say "start by copying all the code in Programs/python.c"? Is there any reason why that file needs to only be one single call into Py_Main? Maybe there's some reason I'm not aware of...

I'm not sure which code do you to copy into python.c? Are you talking about PyConfig_InitPythonConfig()?

The point is that I would like to be able to build a customized Python: read command line arguments & everything, *but* change some parameters. 

" Example of customized Python always running in isolated mode::" is a typical example:

https://github.com/python/peps/pull/1056/files#diff-9369b8a927d9489ef327cab986440d88R709

It's exactly like Python, except that -I command flag is always enabled: isolated=1.

This example is simplified. For more advanced customized, you might want to continue the customization between Py_Initialized() and Py_Main(), that's why I added Py_RunMain(). Pseudo-code:

```
PyConfig_InitPythonConfig(config);
Py_InitializeFromConfig(config);
/* ... more configuration code here! ... */
Py_RunMain()
```

brettcannon:
Just an FYI, but if WIP PRs can be kept in a personal repo so as to help keep this main repo as noise-free as possible that would be appreciated.

vstinner:
> Just an FYI, but if WIP PRs can be kept in a personal repo so as to help keep this main repo as noise-free as possible that would be appreciated.

Ah? I recall that many PEPs are in "draft" module and discussed on pull requets for weeks if not months. I just wanted to make it clear that my PEP update was still in-progress.

Anyway, it's now merged :-)

PR #1060, opened 2019-05-20
PEP 589: Small updates based on feedback from @gvanrossum

JukkaL:
These are the updates I promised in https://github.com/python/peps/pull/991.


JukkaL:
@gvanrossum Please review.

gvanrossum:
Thanks, looks good!

PR #1064, opened 2019-05-21
PEP 590: Mark the main API as private to allow changes in Python 3.9

encukou:
cc @markshannon @jdemeyer

### DIFF
--- a/pep-0590.rst
+++ b/pep-0590.rst
@@ -89,7 +89,7 @@ Descriptor behavior
 
 One additional type flag is specified: ``Py_TPFLAGS_METHOD_DESCRIPTOR``.
 
-``Py_TPFLAGS_METHOD_DESCRIPTOR`` should be set if the the callable uses the descriptor protocol to create a bound method-like object.
+``Py_TPFLAGS_METHOD_DESCRIPTOR`` should be set if the callable uses the descriptor protocol to create a bound method-like object.
 This is used by the interpreter to avoid creating temporary objects when calling methods
 (see ``_PyObject_GetMethod`` and the ``LOAD_METHOD``/``CALL_METHOD`` opcodes).
 
--- a/pep-0590.rst
+++ b/pep-0590.rst
@@ -16,6 +16,9 @@ It introduces a new "vectorcall" protocol and calling convention.
 This is based on the "fastcall" convention, which is already used internally by CPython.
 The new features can be used by any user-defined extension class.
 
+Most new API is private in CPython 3.8, with plans to finalized and
+made public in Python 3.9.
+
 **NOTE**: This PEP deals only with the Python/C API,
 it does not affect the Python language or standard library.
 
@@ -43,6 +46,13 @@ This is inefficient for calls to classes as several intermediate objects need to
 For a class ``cls``, at least one intermediate object is created for each call in the sequence
 ``type.__call__``, ``cls.__new__``, ``cls.__init__``.
 
+This PEP proposes an interface for use by extension modules.
+Such interfaces cannot effectively be tested, or designed, without having the
+consumers in the loop.
+For that reason, we provide private (underscore-prefixed) names.
+The API may change (based on onsumer feedback) in Python 3.9, where we expect
+it to be finalized, and the underscores removed.
+
 
 Specification
 =============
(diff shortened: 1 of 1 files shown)


pep-0590.rst
```diff
@@ -121,13 +133,29 @@ For getting the actual number of arguments from the parameter ``n``,
 the macro ``PyVectorcall_NARGS(n)`` must be used.
 This allows for future changes or extensions.
 
+Finalizing the API
```
jdemeyer: This section should logically come *after* the section where you introduce `_PyObject_Vectorcall`.

jdemeyer:
I have a question related to this, but which doesn't affect the PEP: what about documentation? Should we document those underscore-prefixed names? I think yes since we intend to make them public.

jdemeyer:
`_vectorcallfunc` really should be private too.

encukou:
`vectorcallfunc` is only usable with `_Py_TPFLAGS_HAVE_VECTORCALL`, so users will get the warning that they're using unfinalized API. I don't think it needs to be underscore-prefixed.

pep-0590.rst
```diff
@@ -121,13 +133,29 @@ For getting the actual number of arguments from the parameter ``n``,
...
+------------------
+
+The underscore in the names ``_PyObject_Vectorcall`` and
+``_Py_TPFLAGS_HAVE_VECTORCALL`` indicates that this API may change in minor
+Python versions.
+When finalized (which is planned for Python 3.9), they will be renamed to
+``PyObject_Vectorcall`` and ``Py_TPFLAGS_HAVE_VECTORCALL``, with the former
+names provided using macros.
```
jdemeyer: I would make `with the former names provided using macros` more clear and write something like `the old underscore-prefixed names will remain available as aliases` or something.

jdemeyer:
> `vectorcallfunc` is only usable with `_Py_TPFLAGS_HAVE_VECTORCALL`, so users will get the warning that they're using unfinalized API. I don't think it needs to be underscore-prefixed.

OK, I see what you mean.

pep-0590.rst
```diff
@@ -16,6 +16,9 @@ It introduces a new "vectorcall" protocol and calling convention.
 This is based on the "fastcall" convention, which is already used internally by CPython.
 The new features can be used by any user-defined extension class.
 
+Most new API is private in CPython 3.8, with plans to finalized and
+made public in Python 3.9.
```
jdemeyer: This is not grammatically correct. I would write
```
Most new API is private in CPython 3.8. The plan is to make it public in Python 3.9.
```

pep-0590.rst
```diff
@@ -43,6 +46,13 @@ This is inefficient for calls to classes as several intermediate objects need to
 For a class ``cls``, at least one intermediate object is created for each call in the sequence
 ``type.__call__``, ``cls.__new__``, ``cls.__init__``.
 
+This PEP proposes an interface for use by extension modules.
+Such interfaces cannot effectively be tested, or designed, without having the
+consumers in the loop.
+For that reason, we provide private (underscore-prefixed) names.
+The API may change (based on onsumer feedback) in Python 3.9, where we expect
```
jdemeyer: typo: `onsumer`

pep-0590.rst
pfmoore: "Most of the new API is private..." would be better.

pep-0590.rst
```diff
@@ -64,12 +74,14 @@ Changes to the ``PyTypeObject`` struct
...
+A new ``tp_flags`` flag is added, ``_Py_TPFLAGS_HAVE_VECTORCALL``,
 which must be set for any class that uses the vectorcall protocol.
 
-If ``Py_TPFLAGS_HAVE_VECTORCALL`` is set, then ``tp_vectorcall_offset`` must be a positive integer.
+If ``_Py_TPFLAGS_HAVE_VECTORCALL`` is set, then ``tp_vectorcall_offset`` must be a positive integer.
 It is the offset into the object of the vectorcall function pointer of type ``vectorcallfunc``.
-This pointer may be ``NULL``, in which case the behavior is the same as if ``Py_TPFLAGS_HAVE_VECTORCALL`` was not set.
+This pointer may be ``NULL``, in which case the behavior is the same as if ``_Py_TPFLAGS_HAVE_VECTORCALL`` was not set.
+
+If ``_Py_TPFLAGS_HAVE_VECTORCALL`` is not set, then ``tp_vectorcall_offset`` must be zero.
```
jdemeyer: This is actually new content. We can't make it a hard requirement because of backwards compatibility: it's perfectly legal for an extension type to have a non-NULL value for `tp_print`, which will become a non-zero value for `tp_vectorcall_offset`. Can we please not add this sentence here and defer to the documentation when we implement PEP 590?

pep-0590.rst
```diff
@@ -160,6 +172,23 @@ This restriction may be lifted in the future, but that would require
...
 
+Finalizing the API
+==================
+
+The underscore in the names ``_PyObject_Vectorcall`` and
+``_Py_TPFLAGS_HAVE_VECTORCALL`` indicates that this API may change in minor
+Python versions.
+When finalized (which is planned for Python 3.9), they will be renamed to
+``PyObject_Vectorcall`` and ``Py_TPFLAGS_HAVE_VECTORCALL``.
+The the old underscore-prefixed names will remain available as aliases.
```
jdemeyer: Typo: `The the`

jdemeyer:
I have no further comments.

PR #1072, opened 2019-05-22
Change specification for `# type: ignore`

JelleZijlstra:
As discussed on the typing-sig mailing list, the currently specified behavior for `# type: ignore` on a line by itself has remained mostly unimplemented and doesn't seem useful. Instead, the PEP now mandates a syntax for ignoring all errors in a specific file, which is more obviously useful.

I considered adding a sentence like "Individual type checkers may add additional syntax for silencing errors", but it didn't seem particularly useful since type checkers in general are free to add extensions to what PEP 484 specifies.

While I was at it I added a reference to PEP 526 to an out-of-date sentence adjacent to the modified paragraph.

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -1516,7 +1516,9 @@ error refers to::
   }
 
 A ``# type: ignore`` comment on a line by itself at the top of a file,
-before any docstrings or imports, silences all errors in the file.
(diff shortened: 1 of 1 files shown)


gvanrossum:
Thanks!

PR #1079, opened 2019-05-22
PEP 590: Record BDFL delegate

encukou:
In the Steering Council discussions leading up to the [2019-04-26 update], I was appointed BDFL delegate for PEPs 576, 579, 580, and 590.
PEP 590 was still just a draft on the mailing list when the SC updated the PEPs themselves, so it wasn't updated then.

[2019-04-26 update]: https://github.com/python/steering-council/blob/master/updates/2019-04-26_steering-council-update.md#peps


encukou:
At PyCon US, @ncoghlan explained the story to me and told me to go ahead and add myself. So, I'll do that.

Also: cc @jdemeyer since it's a PEP 590 change :)

PR #1099, opened 2019-06-04
PEP 597: Use UTF-8 for default text file encoding

methane:


methane:
This is discussed in [discourse](https://discuss.python.org/t/use-utf-8-as-default-text-file-encoding/1785).  And @pfmoore suggested me to write a PEP for this idea.

How can I assign PEP number for this?

gvanrossum:
As a core dev you can pick your own PEP number. It seems the next number in the series is 596.

methane:
Thank you, I take 596.

gvanrossum:
I have been bitten by this and am in favor of this.

brettcannon:
Not a huge issue, but you're missing some [suggested sections](https://www.python.org/dev/peps/pep-0012/#suggested-sections).

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
...
+
+When there is device encoding (``os.device_encoding(buffer.fileno())``),
+itstill precedes than this default encoding.
+
+
+Not affected areas
+------------------
+
+Unlike UTF-8 mode, ``locale.getpreferredencoding(False)`` returns
+locale encoding.  Locale encoding is still respected when it makes sense.
```
vstinner: Can't we "fix" the UTF-8 Mode instead? I decided to modify locale.getpreferredencoding(False) because I was too lazy to fix all locations relying directly or implicitly on locale.getpreferredencoding(False). I don't recall the details, but they are way more locations than what I expected.

At the end, making locale.getpreferredencoding(False) to lie looks like a simpler solution to my problem. It makes Python more "consistent" in term of encodings. Even if I dislike making functions lie :-(

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
+PEP: 596
```
vstinner: You loose. Lukasz picked this number!
https://www.python.org/dev/peps/pep-0596/

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
...
+Python-Version: 3.9
+
+Abstract
+========
+
+Currently, ``TextIOWrapper`` uses ``locale.getpreferredencoding(False)``
+when ``encoding`` is not specified.
+
+This PEP proposes changing the default encoding to "UTF-8" regardless of
+platform or locale.
```
vstinner: I made a similar proposition in 2011, but it was too early :-)
"[Python-Dev] open(): set the default encoding to 'utf-8' in Python 3.3?"
https://mail.python.org/pipermail/python-dev/2011-June/112086.html

ambv:
Sorry, I missed your pull request, @methane, and snatched PEP 596 from under you for the Python 3.9 release schedule. I renamed this one as PEP 597 so we can merge it.

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
...
+People assume UTF-8 always
+--------------------------
+
+Package authors using macOS or Linux may forget that the default encoding
+is not always UTF-8.
+
+For example, ``long_description = open("README.md").read()`` in
+``setup.py`` is common mistake.  If there are at least one emoji or any
+other non-ASCII characters in the ``README.md`` file, many Windows users
+cannot install the package by ``UnicodeDecodeError``.
```
vstinner: IMHO we need a way to opt-in for the "locale encoding": current Python 3.8 default encoding when open() if called without specifying an encoding.

I made an attempt to add an "locale" encoding which is the locale encoding.

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
...
+======================
+
+There may be scripts relying locale or code page which is not UTF-8.
+
+They must be rewritten to use ``encoding=locale.getpreferredencoding(False)``,
+or ``encoding="mbcs"`` on Windows when porting them to Python 3.9.
+
+
+Alternative ideas
+=================
```
vstinner: Can you document somewhere that we don't want UTF-8 BOM? Unicode advices to *not* use BOM if I recall properly.

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
...
+
+
+Backward compatibility
+======================
+
+There may be scripts relying locale or code page which is not UTF-8.
+
+They must be rewritten to use ``encoding=locale.getpreferredencoding(False)``,
+or ``encoding="mbcs"`` on Windows when porting them to Python 3.9.
+
```
vstinner: Would it make sense to add an env var / cmdline option to opt-out for this new behavior?

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
...
+
+Not affected areas
+------------------
+
+Unlike UTF-8 mode, ``locale.getpreferredencoding(False)`` returns
+locale encoding.  Locale encoding is still respected when it makes sense.
+
+``stdin``, ``stdout``, and ``stderr`` keep respecting locale too.  For example,
+these commands don't cause mojibake regardless code page::
```
vstinner: That sounds like a high risk of mojibake :-(

Let's say that a file contains a filename. Like a "Makefile" file.

`cat Makefile` and `python3.9 -c 'print(open("Makefile").read(), end="")'` will now output different bytes. cat reads bytes and writes unmodified bytes. python3.9 will decodes from UTF-8 and encodes to the locale encoding. Say hello to mojibake!

pep-0596.rst
```diff
@@ -0,0 +1,191 @@
...
+
+
+Warning
+-------
+
+From 3.8, ``DeprecationWarning`` is shown when encoding is omitted and
+``locale.getpreferredencoding(False)`` is not UTF-8.  This helps not only
+writing forward compatible code, but also investigating unexpected
+``UnicodeDecodeError`` caused by the default encoding.  (See Motivation section
+above.)
```
vstinner: DeprecationWarning means that the feature (open without encoding) will be removed soon? What about my proposed open(encoding="locale") idea?

vstinner:
For the mojibake issues, I suggest to look at the old verison of my PEP 540, before Guido requested make to make it way shorter :-) See the "Use Cases" section:
https://github.com/python/peps/blob/f92b5fbdc2bcd9b182c1541da5a0f4ce32195fb6/pep-0540.txt#L490

I would be interested by the result of each use case with this PEP.

vstinner:
> For the mojibake issues, I suggest to look at the old verison of my PEP 540, before Guido requested make to make it way shorter :-) See the "Use Cases" section:

Hum, this raw reStructured Text view on GitHub is now easy to read. Here is a HTML export (link to Use Cases):
http://haypo.alwaysdata.net/tmp/pep-0540.html#use-cases

vstinner:
Somehow related to the mojibake issues, I wrote a serie of 6 articles about the history of Unicode in Python 3, starting at:
https://vstinner.github.io/python30-listdir-undecodable-filenames.html

The Python 3.2 article mentions removed PYTHONFSENCODING and sys.setfilesystemencoding() which caused many troubles in the past:
https://vstinner.github.io/painful-history-python-filesystem-encoding.html

vstinner:
**If** we decide to switch to UTF-8 by default, it would help to have a mechanism to attempt to read a file from UTF-8, or fallback to the locale encoding on the first decode error. At this point, I cannot say if it would be a good idea to make it the default behavior for open().read() or not.

... It might be tricky to implement it, especially if the decoding error happens on the second line, but the file is read line by line using "for line in text_file: ...". I guess that the only reliable solution is to **first** fully decode the file to ensure that it's encoded to UTF-8 (can be decoded from UTF-8). It might be slow if the text file is larger than 1 GiB... So maybe it should be an helper function properly documented ;-)

pep-0596.rst
methane: I think users of UTF-8 mode will love this lie.  But I'm not sure this lie is good when UTF-8 mode become default.

Anyway, after this PEP is accepted and implemented, ``locale.getpreferredencoding(False)`` is not used in TextIOWrapper.  Checking all callers will be easier than before.

pep-0596.rst
uranusjr: Currently the “locale” encoding (platform encoding?) is available via `open(..., encoding=None)`. Would it be a good idea to keep it available? I.e. instead of changing the behaviour of `encoding=None` to use UTF-8, change the *signature* of `open` to `encoding='utf8'`.

methane: > IMHO we need a way to opt-in for the "locale encoding": current Python 3.8 default encoding when open() if called without specifying an encoding.
> 
> I made an attempt to add an "locale" encoding which is the locale encoding.

On Windows, we have "mbcs" already.
How many scripts which works on other than Windows need shortcut for `locale.getpreferredencoding`?


> Currently the “locale” encoding (platform encoding?) is available via `open(..., encoding=None)`. Would it be a good idea to keep it available? I.e. instead of changing the behaviour of `encoding=None` to use UTF-8, change the _signature_ of `open` to `encoding='utf8'`.

I don't think so.  There are many functions which takes `encoding=None` parameter
and pass it to `open(..., encoding=encoding)`.  We can change all caller in stdlib, but
we can not change all callers in other libs.

pep-0596.rst
methane: It doesn't make sense.
One motivation of this PEP is allowing people to assume default encoding of text file is always UTF-8, and they can omit encoding option safely.
Once we have such option, this goal can not be achieved.

pep-0596.rst
methane: It is impossible to avoid 100% of mojibake.
Even `cat Makefile` cause mojibake when Makefile contains UTF-8 but console uses legacy encoding.

In your example, it doesn't cause mojibake on Windows legacy console, thanks to PEP 528.

pep-0596.rst
pfmoore: Also, do any Windows tools still add BOM? I know Notepad did at one point, and if that's common, how would it affect interoperability?

Also, can we move these discussions to python-dev or Discourse? It's nearly impossible to follow the points being made and the implications when they are against the PR like this :-(

pep-0596.rst
methane: No.  It is not removed, it just changes behavior.

See https://docs.python.org/3/library/warnings.html#warning-categories

> Previously DeprecationWarning and FutureWarning were distinguished based on whether a feature was being removed entirely or changing its behaviour. They are now distinguished based on their intended audience and the way they’re handled by the default warnings filters.

DeprecationWarning is Warning for Python developers.  It can be used to notice
behavior changes.

pep-0596.rst
methane: Of course, I will post the article to Discourse when I merge this PR.

pep-0597.rst
```diff
@@ -162,17 +173,75 @@ And these scripts will become portable when default encoding is changed
...
+Backward compatibility
+======================
+
+There may be scripts relying locale or code page which is not UTF-8.
+They must be rewritten to specify ``encoding`` explicitly.
+
+* If the script assumed ``latin1`` or ``cp932``, use ``encoding="latin1"``
+  or ``encoding="cp932"`` should be used.
+
+* If the script is designed to respece locale encoding,
```
vstinner: typo: respect

methane: nice catch.

PR #1100, opened 2019-06-05
PEP 597: Copy editing

Harmon758:


pep-0597.rst
```diff
@@ -141,69 +143,70 @@ If we enable UTF-8 mode by default, even people using Windows will forget
...
+So changing the default encoding of text files to UTF-8 would be better
+even if UTF-8 mode is enabled by default at some point.
 
 
 Why not change std(in|out|err) encoding too?
 --------------------------------------------
 
-Even when locale encoding is not UTF-8, there will be many UTF-8
-text files.  These files are downloaded from the internet, or
-written by modern text editor same to editing Python source.
```
methane: I didn't mean "Python source files" are "modern text editors".
I meant "modern text editors which are used to edit Python source files".

But I think we can remove just ", e.g. Python source files".

Harmon758: Ah, I did mean for "e.g. Python source files" to be referring to "These files" rather than "modern text editors", but I agree that the phrasing was awkward.
I've gone ahead and moved it to fix its subject and improved the phrasing in the rest of the paragraph.
Let me know if you think "such as Python source files" should still be removed, and I'll go ahead and do so if so.

methane: When I wrote this paragraph, I want to describe why most text files are encoded in UTF-8, especially in environment of new Python programmers.

* They may download text files and open it from Python.  Using locale encoding to open text file from the Internet doesn't make sense at all.  UTF-8 dominated the Internet.

* They may write text file using text editor and open it from Python.  And they will use same text editor to write Python source.  Since Python source is UTF-8 by default, the text editor will use UTF-8 by default.

"Text files such as Python source" seems bit redundant to me.  Text files may and may not Python source.  It doesn't describe well why most text files are encoded in UTF-8 instead of locale encoding.

Harmon758: Right, I understand and don't disagree with you.
I've gone ahead and removed the clause.

methane:
Thank you!

PR #1108, opened 2019-06-15
PEP 598: Introducing minor feature releases

ncoghlan:
Alternative to PEP 596 that aims to reduce feature delivery latency
by giving ourselves access to a new, less disruptive, feature
delivery vehicle, rather than by increasing the frequency with
which we deploy our existing disruptive feature delivery vehicle.

### DIFF
--- a/pep-0598.rst
+++ b/pep-0598.rst
@@ -1,0 +1,527 @@
+PEP: 598
+Title: Introducing minor feature releases
+Version: $Revision$
+Last-Modified: $Date$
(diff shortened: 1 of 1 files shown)


ncoghlan:
@zooba I'm going to go ahead and merge this in order to claim the PEP number, but I think this initial draft is a bit rambly and repetitive, so I'd appreciate a review from someone that's already sympathetic to the idea, and may be able to help explain it clearly :)

PR #1134, opened 2019-07-28
PEP 598: Fix terminology, reference support policy NEP

ncoghlan:
- several Scientific Python projects are working on a
  common version support policy for their projects,
  which we need to take into account when considering
  changes to our release cadence
- reviewing that draft NEP highlighted several problems
  with the way this PEP was using the "major" and "minor"
  terms, so this takes a pass through that, replacing
  inappropriate uses of "major" with "feature", and
  and inapproriate uses of "minor" with "micro"
- I also went through and collected statistics on how many
  changes we've been making in micro releases that
  warranted version added/changed notes in the docs


ncoghlan:
Removing the WIP marker, as I'd like to eliminate the terminology issues with major vs minor vs micro (since the SciPy NEP consistently uses the SemVer major/minor/patch wording).

PR #1138, opened 2019-08-02
PEP 569: Python 3.8.0b3 released

Harmon758:
https://mail.python.org/archives/list/python-announce-list@python.org/thread/BCIHRB5N6AKPCBVO4TQYSYSPOBMV3GXL/


brettcannon:
Thanks!

/cc @ambv in case he has a PR floating on his machine to do this update.

PR #1151, opened 2019-08-30
PEP 601: Forbid return/break/continue breaking out of finally

dpgeorge:
This is a draft PEP to propose forbidding return/break/continue breaking out of a finally suite.

### DIFF
--- a/pep-0601.txt
+++ b/pep-0601.txt
@@ -1,4 +1,4 @@
-PEP: 9999
+PEP: 601
 Title: Forbid return/break/continue breaking out of finally
 Author: Damien George, Batuhan Taskaya
 Sponsor: Nick Coghlan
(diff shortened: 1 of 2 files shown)


ncoghlan:
Thanks - PEP number assigned!

dpgeorge:
Thank you!

ncoghlan:
And wearing my PEP Sponsor hat (rather than a PEP Editor one), my suggested next step would be send an email to python-dev and start the discussion regarding the PEP.

Alternatively, if you'd prefer to use discuss.python.org as the main discussion venue:

* start a thread in the Users area of discuss.python.org (see https://discuss.python.org/t/pep-594-removing-dead-batteries-from-the-standard-library/1704 for example, although I'd recommend at least describing what the PEP covers directly in the first post)
* submit a PR adding the Discussions-To header to the PEP (as per https://www.python.org/dev/peps/pep-0594/)
* send a post to python-dev pointing to the discuss.python.org thread

isidentical:
Thanks for suggestions @ncoghlan , by the way can you look #1153 ?

ncoghlan:
@isidentical Done!

PR #1159, opened 2019-09-10
PEP-573: Rewrite section on passing the "defining class" into C methods.

scoder:
The main change is the signature, but I took the liberty to reword the paragraphs while I was matching them with PEP-590.


encukou:
I took the liberty of adding some more changes. See the individual commits.

encukou:
cc @Dormouse759

PR #1161, opened 2019-09-11
PEP 594: Fix reference to prior PEP

### LINKED ISSUE #1160: Missing reference in PEP 594 to PEP 3818
[PEP 594 - Removing dead batteries from the standard library](https://www.python.org/dev/peps/pep-0594/#abstract) references [PEP 3818](https://www.python.org/dev/peps/pep-3818), but that doesn't exist, or no longer exists. The context in PEP 594 refers to [PEP 3108 - Standard Library Reorganization](https://www.python.org/dev/peps/pep-3108/), so perhaps that is the correct link.

EpicWink:
The context refers to 'Standard Library Reorganization', which is PEP
3108. This change updates the nearby link

Fixes #1160

### DIFF
--- a/pep-0594.rst
+++ b/pep-0594.rst
@@ -18,7 +18,7 @@ and SUN file formats), APIs and operating systems that have been superseded a
 long time ago (e.g. Mac OS 9), or modules that have security implications and
 better alternatives (e.g. password and login).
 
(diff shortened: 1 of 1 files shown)


tiran:
Thanks!

PR #1174, opened 2019-09-23
PEP 8: Discourage use of return/break/continue in finally

dpgeorge:
As per discussion on PEP 601: https://discuss.python.org/t/pep-601-forbid-return-break-continue-breaking-out-of-finally/2239/32


gvanrossum:
Thanks! Just what I would have written. :-)

dpgeorge:
You're welcome!

PR #1181, opened 2019-09-27
PEP 599: Updates

di:
* Per https://github.com/pypa/warehouse/pull/6684#issuecomment-534266929, correctly specifies the reported ARMv7 architecture string as `armv7l`, not `armhfp`.
* Also fixes some poor formatting in the footer.

(cc @brettcannon @ncoghlan)

### DIFF
--- a/pep-0599.rst
+++ b/pep-0599.rst
@@ -166,7 +166,7 @@ the ``manylinux2014`` tag:
        GLIBC_2.17
        CXXABI_1.3.7
        GLIBCXX_3.4.19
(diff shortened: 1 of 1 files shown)


ncoghlan:
For the proposed PEP 571 32-bit build fix, I requested that a Change Note be added to the PEP text: https://github.com/python/peps/pull/1180#issuecomment-536508082

I haven't requested one here, given that the PR prompting it is literally the one allowing manylinux2014 uploads to PyPI.

PR #1185, opened 2019-09-30
PEP 587: Remove struct_size

vstinner:


vstinner:
See https://mail.python.org/archives/list/python-dev@python.org/thread/C7Z2NA2DTM3DLOZCFQAK5A2WFYO3PHHX/ and https://bugs.python.org/issue38326 for the rationale and discussion.

PR #1193, opened 2019-10-04
PEP 487: Remove extra underscore.

brandtbucher:
Just something small I noticed while reading up on `__init_subclass__`.


Rosuav:
LGTM, thanks for spotting that.

PR #1195, opened 2019-10-04
PEP 440: Normalize date version examples

jaraco:
Avoid suggesting versions that will raise warnings in build tools such as pypa/setuptools#302


gvanrossum:
To whoever merges this, please remember to change the Commit subject to start with "PEP 440:"

ncoghlan:
Thanks for the reminder Guido - I quite possibly would have forgotten that editing the PR title doesn't automatically fix the commit message.

PR #1197, opened 2019-10-05
PEP 13: Change the Council Election to use Approval voting.

Yhg1s:
Change PEP 13 to make the Council Election use regular (multiwinner) approval voting, the same thing used for the PSF board elections, instead of restricting the number of votes to the number of seats in the Council.


gvanrossum:
Yeah, +1 from me too.

PR #1200, opened 2019-10-07
PEP 605: Add example user-facing docs

ncoghlan:
* example What's New subsection
* example alpha, beta, and rc announcements
* fix Python-Requires recommendation
* add rationale for allowing mid-stream alpha releases


ncoghlan:
@zooba I went ahead and merged this update, as I'm switching focus to the shared informational PEP.

PR #1201, opened 2019-10-08
PEP 599: Allow CXXABI_TM_1 symbol version

mayeut:
More information in [auditwheel discussion](https://github.com/pypa/auditwheel/pull/192/files#r332459037) of [manylinux2014 Pull Request](https://github.com/pypa/auditwheel/pull/192)

### DIFF
--- a/pep-0599.rst
+++ b/pep-0599.rst
@@ -164,7 +164,7 @@ the ``manylinux2014`` tag:
    they may only depend on the following maximum versions::
 
        GLIBC_2.17
-       CXXABI_1.3.7
+       CXXABI_1.3.7, CXXABI_TM_1 is also allowed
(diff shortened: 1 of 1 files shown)


pep-0599.rst
```diff
@@ -164,7 +164,7 @@ the ``manylinux2014`` tag:
    they may only depend on the following maximum versions::
 
        GLIBC_2.17
-       CXXABI_1.3.7
+       CXXABI_1.3.7, CXXABI_TM_1 is also allowed
```
di:
```suggestion
       CXXABI_1.3.7 (CXXABI_TM_1 is also allowed)
```
Might be good to reference https://gcc.gnu.org/ml/libstdc++/2019-10/msg00023.html here as well.

PR #1206, opened 2019-10-14
PEP 584: Remove subtraction and add more background.

brandtbucher:
The most significant changes here are that Steven and I have withdrawn the subtraction part of the proposal, and updated the pure-Python equivalent code to use `self.copy()`.

The rest of the changes are mostly background info, formatting, and summaries of the latest round of discussions. There will likely be at least one more draft before the PEP is finalized.

CC @Rosuav @stevendaprano


brandtbucher:
Thanks @Rosuav!

PR #1214, opened 2019-10-22
PEP 584: Fix formatting and update copyright notice.

brandtbucher:
CC @stevendaprano @Rosuav


Rosuav:
I have Steven's approval via email. If you have an opportunity to say something on this PR (for the sake of transparency) that'd be great, but I'm not waiting on it.

PR #1222, opened 2019-11-06
PEP 3108: Fixed small grammar typo, removed "has"

samsetegne:
I noticed the sentence "Python 3.0 has presents a chance to...", so I removed "has" so that the sentence reads "Python 3.0 presents a chance to..."


brettcannon:
Thanks!

PR #1224, opened 2019-11-06
PEP 558: Move target version to 3.9

gvanrossum:
@ncoghlan @njsmith This PR is also a ping: what needs to happen for this PEP to actually land in 3.9?


gvanrossum:
OK, I'm going to move the target version. But @njsmith / @ncoghlan please respond. Do you have plans for this or is it better to postpone, withdraw or outright reject?

PR #1233, opened 2019-11-12
PEP 579: Accept (with a reminder on informational PEPs)

encukou:
I approve PEP 579, with an explicit note that acceptance of an Informational PEP doesn't really mean that much. PEP 590 is the important one.

cc @jdemeyer


pep-0579.rst
```diff
@@ -2,13 +2,29 @@ PEP: 579
 Title: Refactoring C functions and methods
 Author: Jeroen Demeyer <J.Demeyer@UGent.be>
 BDFL-Delegate: Petr Viktorin
-Status: Draft
+Status: Accepted
```
brettcannon: Informational PEPs should either be "active" because they are a living document or "final" because you're done making changes.

PR #1234, opened 2019-11-12
PEP 593: Set Python-Version: 3.9

gvanrossum:
@ilevkivskyi Maybe we can even mark this Final, since it's been implemented in the 3.7 stdlib?


ilevkivskyi:
> Maybe we can even mark this Final, since it's been implemented in the 3.7 stdlib?

I don't think `Annotated` is in `typing` yet. We can mark it Final when we move it there.

PR #1235, opened 2019-11-14
Revert "Recording the Direct URL Origin of installed distributions"

brettcannon:
Reverts python/peps#1145


brettcannon:
Decided to just directly edit the PEP to fix my goof of not editing the number ahead of time.

PR #1236, opened 2019-11-16
PEP 8101: Initial draft of pep for election of 2020 term steering council

ewdurbin:


pradyunsg:
A minor process request for the future elections -- would it be feasible to add these details as a PEP, before the nomination period begins (minus the candidates and results obviously)?

ewdurbin:
> A minor process request for the future elections -- would it be feasible to add these details as a PEP, before the nomination period begins (minus the candidates and results obviously)?

Yes, PEP 13 should be updated to specify the process/format for these PEPs as well.

PR #1244, opened 2019-12-02
PEP 594: Mention uu_codec

gvanrossum:
If we're deprecating and removing `uu` we should do the same to the codec of the same name.


ncoghlan:
@tiran Are you happy for us to merge this tweak to PEP 594? (removing the uu codec as part of removing the uu module)

PR #1264, opened 2019-12-30
PEP 573: SystemError->TypeError and misc copy-edits

ncoghlan:
* report TypeError rather than SystemError when bad types are passed
  to new APIs
* add PyType_FromModuleAndSpec to the list of new functions
* wrap text at 80 columns
* add C API docs links for thread-local storage and context variables
* note in abstract that only the initial state lookup from slots is
  inherently slow, with subsequent lookups being amenable to caching
* other minor wording tweaks

### DIFF
--- a/pep-0573.rst
+++ b/pep-0573.rst
@@ -19,19 +19,24 @@ Post-History:
 Abstract
 ========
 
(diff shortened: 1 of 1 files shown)


encukou:
Thanks!

PR #1278, opened 2020-01-22
PEP 613: Explicit Type Aliases

shannonzhu:
Note: Python CLA has been signed but is pending the check at https://check-python-cla.herokuapp.com/

### DIFF
--- a/pep-0613.rst
+++ b/pep-0613.rst
@@ -13,19 +13,33 @@ Post-History: 21-Jan-2020
 Abstract
 ========
 
-Type aliases are user-specified types which may be as complex as any type hint, and are specified with a simple variable assignment on a module top level.
+Type aliases are user-specified types which may be as complex as any type hint,
+and are specified with a simple variable assignment on a module top level.
 
 This PEP formalizes a way to explicitly declare an assignment as a type alias.
 
 Motivation
 ==========
 
-Type aliases are declared as top level variable assignments. In `PEP 484 <https://www.python.org/dev/peps/pep-0484/#type-aliases/>`_, the distinction between a valid type alias and a global variable was implicitly determined: if a top level assignment is unannotated, and the assigned value is a valid type, then the name being assigned to is a valid type alias. Otherwise, that name is simply a global value that cannot be used as a type hint.
+Type aliases are declared as top level variable assignments. In `PEP 484 <https://www.python.org/dev/peps/pep-0484/#type-aliases/>`_,
+the distinction between a valid type alias and a global variable was implicitly
+determined: if a top level assignment is unannotated, and the assigned value is
+a valid type, then the name being assigned to is a valid type alias. Otherwise,
+that name is simply a global value that cannot be used as a type hint.
 
-These implicit type alias declaration rules create confusion when type aliases involve forward aliases, invalid types, or violate other restrictions enforced on type alias declaration. Because the distinction between an unannotated value and a type alias is implicit, ambiguous or incorrect type alias declarations implicitly default to a valid value assignment. This creates expressions that are impossible to express as type aliases and punts error diagnosis of malformed type aliases downstream.
+These implicit type alias declaration rules create confusion when type aliases
+involve forward references, invalid types, or violate other restrictions
+enforced on type alias declaration. Because the distinction between an
+unannotated value and a type alias is implicit, ambiguous or incorrect type
+alias declarations implicitly default to a valid value assignment. This creates
+expressions that are impossible to express as type aliases and punts error
+diagnosis of malformed type aliases downstream.
 
-The following examples each include an illustration of some of the suboptimal or confusing behaviors resulting from existing implicit alias declarations.
-We also introduce explicit aliases of the format :code:`TypeName: TypeAlias = Expression` here for the sake of comparison, but the syntax is discussed in further detail in later sections.
+The following examples each include an illustration of some of the suboptimal
+or confusing behaviors resulting from existing implicit alias declarations.
+We also introduce explicit aliases of the format :code:`TypeName: TypeAlias = Expression`
+here for the sake of comparison, but the syntax is discussed in further detail
+in later sections.
 
 Forward References:
 *******************
@@ -35,14 +49,19 @@ Forward References:
   MyType = "ClassName"
   def foo() -> MyType: ...
 
-This code snippet should not error so long as :code:`ClassName` is defined later on. However, a type checker is forced to throw spurious errors that (1) :code:`MyType` is an unannotated global string, and (2) :code:`MyType` is not a valid type.
+This code snippet should not error so long as :code:`ClassName` is defined
+later on. However, a type checker is forced to throw spurious errors that
+(1) :code:`MyType` is an unannotated global string, and (2) :code:`MyType`
+is not a valid type.
 
 ::
 
   MyType: TypeAlias = “ClassName”
   def foo() -> MyType: ...
 
-Explicit aliases remove ambiguity so neither of the above errors will be thrown. Additionally, if something is wrong with :code:`ClassName` (i.e., it’s not actually defined later), the type checker can throw an error.
+Explicit aliases remove ambiguity so neither of the above errors will be
+thrown. Additionally, if something is wrong with :code:`ClassName`
+(i.e., it’s not actually defined later), the type checker can throw an error.
 
 
 Error Messaging:
@@ -50,17 +69,26 @@ Error Messaging:
 
 ::
 
-  MyType = InvalidType
-  MyType = MyGeneric(int)  # i.e., intention was MyGeneric[int]
+  MyType1 = InvalidType
+  MyType2 = MyGeneric(int)  # i.e., intention was MyGeneric[int]
 
-A type checker should warn on this code snippet that :code:`InvalidType` is not a valid type, and therefore cannot be used to annotate an expression or to construct a type alias. Instead, type checkers are forced to throw spurious errors that (1) :code:`MyType` is a global expression missing an annotation, and (2) :code:`MyType` is not a valid type in all usages of :code:`MyType` across the codebase.
+A type checker should warn on this code snippet that :code:`InvalidType` is not
+a valid type, and therefore cannot be used to annotate an expression or to
+construct a type alias. Instead, type checkers are forced to throw spurious
+errors that (1) :code:`MyType` is a global expression missing an annotation,
+and (2) :code:`MyType` is not a valid type in all usages of :code:`MyType`
+across the codebase.
 
 ::
 
-  MyType: TypeAlias = InvalidType
-  MyType: TypeAlias = MyGeneric(int)
+  MyType1: TypeAlias = InvalidType
+  MyType2: TypeAlias = MyGeneric(int)
 
-With explicit aliases, the type checker has enough information to error on the actual definition of the bad type alias, and explain why: that :code:`MyGeneric(int)` and `InvalidType` are not valid types. When the value expression is no longer evaluated as a global value, unactionable type errors on all usages of :code:`MyType` across the codebase can be suppressed.
+With explicit aliases, the type checker has enough information to error on the
+actual definition of the bad type alias, and explain why: that :code:`MyGeneric(int)`
+and `InvalidType` are not valid types. When the value expression is no longer
+evaluated as a global value, unactionable type errors on all usages of :code:`MyType`
+across the codebase can be suppressed.
 
 Scope Restrictions:
 *******************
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

I'm a bot set up to make sure that the project can legally accept this contribution by verifying everyone involved has signed the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA).



## CLA Missing

Our records indicate the following people have not signed the CLA:

@shannonzhu

For legal reasons we need all the people listed to sign the CLA before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.

If you have recently signed the CLA, please wait at least one business day
before our records are updated.


Thanks again for the contribution, we look forward to reviewing it!

Rosuav:
What's the reference to PEP 1? Is that a leftover?

Is there an official policy on the wrapping of text? I can't remember if that requirement was relaxed. Most PEPs have manually-wrapped lines rather than going way off the right margin.

Do you have a potential sponsor lined up?

shannonzhu:
re: PEP 1 - thanks for pointing that out, I included it by mistake and will remove!

I didn't see any mention of line wrapping in contributing guide or PEP 1, but I'm happy to manually wrap if needed.

I don't have a sponsor, though this topic/syntax has been discussed on typing-sig and in person at a typing meetup including core developers. Would appreciate advice on how to proceed here.

Rosuav:
Most PEPs are wrapped. Unless someone has a strong opinion or a citation to an actual requirement, I'll leave it up to you whether to wrap or not.

Re sponsor: Mention on typing-sig that you're in need of a sponsor, and someone will likely step up. All it requires is for a core dev to say "yep, this is worth moving forward with".

pep-0613.rst
```diff
@@ -0,0 +1,173 @@
...
+
+  MyType = InvalidType
+  MyType = MyGeneric(int)  # i.e., intention was MyGeneric[int]
+
+A type checker should warn on this code snippet that :code:`InvalidType` is not a valid type, and therefore cannot be used to annotate an expression or to construct a type alias. Instead, type checkers are forced to throw spurious errors that (1) :code:`MyType` is a global expression missing an annotation, and (2) :code:`MyType` is not a valid type in all usages of :code:`MyType` across the codebase.
+
+::
+
+  MyType: TypeAlias = InvalidType
+  MyType: TypeAlias = MyGeneric(int)
```
gvanrossum: I'm confused. These look like two different examples somehow merged into one -- I assume you can't declare the same name as a `TypeAlias` twice?

Would this section work if the type aliases were named `MyType1` and `MyType2`?

(It's especially confusing since in the previous example, the two lines of the example had to occur together to illustrate the problem, while here there seem to be two independent lines that each illustrate a different variant of the problem.)

pep-0613.rst
```diff
@@ -0,0 +1,173 @@
...
+    x = ClassName
+
+The outer :code:`x` is a valid type alias, but type checkers must error if the inner :code:`x` is ever used as a type. This is confusing because the alias declaration rule is not explicit, and the error is not thrown on the location of the inner type alias declaration but rather on every one of its subsequent use cases.
+
+::
+
+  x: TypeAlias = ClassName
+  def foo() -> None:
+    x = ClassName
+    x: TypeAlias = ClassName
```
gvanrossum: Again, it's not super clear here that this is really two examples: inside the function you can write either of the two lines shown and their (different) meanings will be clear.

pep-0613.rst
```diff
@@ -0,0 +1,173 @@
...
+
+With explicit aliases, the outer assignment is still a valid type variable, and the inner assignment can either be a valid local variable or a clear error, communicating to the author that type aliases cannot be defined inside a nested scope.
```
gvanrossum: Oh wait, the line inside the function `x: TypeAlias = ClassName` is an error? That makes it doubly confusing that you write them both in one code snippet.

pep-0613.rst
```diff
@@ -0,0 +1,173 @@
...
+
+
+Specification
+=============
+
+The explicit alias declaration syntax attempts to clearly differentiate between the three possible kinds of assignments: typed global expressions, untyped global expressions, and type aliases. This avoids the existence of assignments that break type checking when an annotation is added, and avoids classifying the nature of the assignment based on the type of the value.
```
gvanrossum: Does it really only *attempt* to differentiate? I think you can just write "clearly differentiates". :-)

pep-0613.rst
```diff
@@ -0,0 +1,173 @@
...
+  x: int = 1  # typed global expression
+
+  x = int  # untyped global expression (see note below)
+  x: Type[int] = int  # typed global expression
+
+  x: TypeAlias = int  # type alias
+  x: TypeAlias = “MyClass”  # type alias
+
+
+Note: The examples above illustrate implicit and explicit alias declarations in isolation. For the sake of backwards compatibility, type checkers may choose to support both simultaneously, meaning an untyped global expression :code:`x = int` will still be considered a valid type alias.
```
gvanrossum: I'm not happy with this. There must by now exist millions of lines of code using implicit aliases unambiguously. I would rather not declare those invalid at the whim of the type checker -- it would not promote interoperability. My preference is for `x = int` to be a valid type alias except in some corner cases (like `x = "MyClass"`), basically as things are today.

pep-0613.rst
```diff
@@ -0,0 +1,173 @@
...
+
+
+Backwards Compatibility
+=======================
+
+Explicit aliases provide an alternative way to declare type aliases, but all pre-existing code and old alias declarations will work as before.
```
gvanrossum: I agree with this, but the Note above seems to contradict it.

pep-0613.rst
```diff
@@ -0,0 +1,173 @@
...
+
+
+In comparison, the chosen syntax option :code:`MyType: TypeAlias = int` is appealing because it still sticks with the :code:`MyType = int` assignment syntax, and adds some information for the type checker purely as an annotation.
+
+
+Open Issues
+===========
+
+
+The current implementation utilizes the already-defined class :code:`typing.TypeAlias`, but for conciseness there are suggestions to standardize this to a new class, :code:`typing.Alias`.
```
gvanrossum: I don't see `typing.TypeAlias` in any of the versions of `typing.py` that I'm aware of. And I'm guess the open issue is whether to use `TypeAlias` or just `Alias`? I vote for `TypeAlias`, because "alias" can mean many things in many contexts.

shannonzhu: Hm, I meant it's defined in typing.pyi, and so it's possible already to write `from typing import TypeAlias` and use it in your annotations. I suppose this shouldn't matter in versions where annotations aren't evaluated, but let me try to clarify this wording and note that it should probably be added to typing.py. I'd also vote for TypeAlias, but remember that I got more than one suggestion to shorten from others on the list/at the meetup.

pep-0613.rst
shannonzhu: Ah yes, I'll rename the variables so it's clear they're two separate examples in one code snippet.

pep-0613.rst
shannonzhu: I wrote this with the thought in mind that having an explicit-only system might be more straightforward than a mixed one and the conversion could be an easy codemod, so wanted to keep that option open. However, valuing interoperability makes sense - I'll adjust the wording here.

brettcannon:
The [column width is to be no longer than 80](https://www.python.org/dev/peps/pep-0012/#general).

Rosuav:
Ah! Thanks Brett, that's what I was looking for. Was searching for the word "wrap", but it actually uses "fill".

pep-0613.rst
```diff
@@ -13,19 +13,33 @@ Post-History: 21-Jan-2020
...
-
-These implicit type alias declaration rules create confusion when type aliases involve forward aliases, invalid types, or violate other restrictions enforced on type alias declaration. Because the distinction between an unannotated value and a type alias is implicit, ambiguous or incorrect type alias declarations implicitly default to a valid value assignment. This creates expressions that are impossible to express as type aliases and punts error diagnosis of malformed type aliases downstream.
-
-The following examples each include an illustration of some of the suboptimal or confusing behaviors resulting from existing implicit alias declarations.
-We also introduce explicit aliases of the format :code:`TypeName: TypeAlias = Expression` here for the sake of comparison, but the syntax is discussed in further detail in later sections.
+Type aliases are declared as top level variable assignments. In `PEP 484 <https://www.python.org/dev/peps/pep-0484/#type-aliases/>`_,
```
gvanrossum:
```suggestion
Type aliases are declared as top level variable assignments.
In `PEP 484 <https://www.python.org/dev/peps/pep-0484/#type-aliases/>`_,
```

shannonzhu:
Done, thank you for the feedback!

pep-0613.rst
```diff
@@ -0,0 +1,223 @@
...
+thrown. Additionally, if something is wrong with :code:`ClassName`
+(i.e., it’s not actually defined later), the type checker can throw an error.
+
+
+Error Messaging:
+****************
+
+::
+
+  MyType1 = InvalidType
```
rchen152: It doesn't look like `MyType1` and `MyType2` illustrate two different things, so would it be enough to keep just one of them?

pep-0613.rst
```diff
@@ -0,0 +1,223 @@
...
+Scope Restrictions:
+*******************
+
+::
+
+  x = ClassName
+  def foo() -> None:
+    x = ClassName
+
+The outer :code:`x` is a valid type alias, but type checkers must error if the
```
rchen152: It wasn't obvious to me why type checkers must error here until I got to the explanation at the very end of the section, so I'd suggest moving that explanation here: "type checkers must error if the inner `x` is ever used as a type <b>because type aliases cannot be defined inside a nested scope</b>."

Rosuav:
This pull request has already been merged. I recommend posting future comments to the mailing list.

rchen152:
Ah, I wasn't aware that was the convention - sorry about that. I wanted to add comments on specific lines and thought they would be easier to follow if made on the PR. I'll post to the mailing list in the future.

Rosuav:
Yeah, no probs. Just didn't want you to sit here wondering why nobody was reviewing your comments :)

pep-0613.rst
```diff
@@ -0,0 +1,223 @@
...
+Forward References:
+*******************
+
+::
+
+  MyType = "ClassName"
+  def foo() -> MyType: ...
+
+This code snippet should not error so long as :code:`ClassName` is defined
+later on. However, a type checker is forced to throw spurious errors that
```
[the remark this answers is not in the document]
shannonzhu: Thanks for the feedback!

For this one, I'll incorporate " a type checker is forced to assume that MyType is a value assignment "; will leave out the "not yet a valid type" since "yet" might be confusing in the context of a forward alias.

pep-0613.rst
```diff
@@ -0,0 +1,223 @@
...
+(1) :code:`MyType` is an unannotated global string, and (2) :code:`MyType`
```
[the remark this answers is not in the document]
shannonzhu: reworded to 'may' rather than specifying any behavior typecheckers must take here - thought about this one for a while and want to keep both statements about potential type checker behavior in for clarity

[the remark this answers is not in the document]
rchen152: Makes sense!

pep-0613.rst
nipunn1313: Hi! I'm curious about the rationale behind this limitation.
Here's a sample snippit and how it works currently.

```python
from typing import NewType
from typing_extensions import TypeAlias

class C:
    X = NewType('X', int)
    Y1 = X  # mypy - error: Variable "test.C.Y1" is not valid as a type  pyright - seems to work
    Y2: TypeAlias = X  # mypy - seems to work  pyright - error: A TypeAlias can be defined only within a module scope

x: C.X
y1: C.Y1
y2: C.Y2
```

I have a use case like this where there is a type `C.Y` that I'm trying to rename to `C.X` in a backward compatible way. Type alias would be perfect, but PEP613 disallows it (and behavior between existing typecheckers vary). Other recommendations on how to do this welcome.

The PEP mentions how non-explicit alias declarations are confusing in nested scope. From first principles, something like `Y1 = X` seems ok to allow as a type alias. I may be overlooking something.

nipunn1313: From https://mail.python.org/archives/list/typing-sig@python.org/thread/MWRJOBEEEMFVXE7CAKO7B4P46IPM4AN3/ (@ilevkivskyi)

> ```
> class C: ...
> 
> x = C  # this is a type alias from mypy point of view
> class D:
>     x = C  # this is a variable with type Type[C]
> ```
> This default behaviour is kind of a heuristics that matches most typical
> use in each case, but I think it would be nice to have a simple way to be
> explicit either way:


If provided explicitly (like `Y2: TypeAlias`) - it seems reasonable to be allowable. The use case of backwards-compatible rename of an inner type is compelling to me.

If this makes sense, I'm happy to put some legwork in to update/clarify the PEP

gvanrossum: Probably better to bring this up on typing-sig. I don't have a strong opinion on this either way, your suggestion seems reasonable.

PR #1279, opened 2020-01-22
PEP 611: relax hard limits and make 1 million a soft limit. Include recursion depth in limits.

markshannon:
Also remove comment about Jython, as it is incorrect.

### DIFF
--- a/pep-0611.rst
+++ b/pep-0611.rst
@@ -10,7 +10,8 @@ Post-History:
 
 Abstract
 ========
-This PR proposes a limit of one million (1 000 000) for various aspects of Python code and its implementation.
+This PR proposes a soft limit of one million (1 000 000), and a larger hard limit
+for various aspects of Python code and its implementation.
 
 The Python language does not specify limits for many of its features.
 Not having any limit to these values seems to enhance programmer freedom,
@@ -19,6 +20,8 @@ machines have implicit limits or are forced to assume that the limits are
 astronomical, which is expensive.
 
 This PR lists a number of features which are to have a limit of one million.
+
+For CPython the hard limit will be eight million (8 000 000).
 
 Motivation
 ==========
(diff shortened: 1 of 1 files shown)


pep-0611.rst
```diff
@@ -190,32 +206,49 @@ It is anticipated that CPython will enforce the limits as follows:
...
+ 3.9            16 million
+ 3.10 onward    8 million
 =============  ===============
 
 Given the rarity of code generators that would exceed the one million limits,
 and the environments in which they are typically used, it seems reasonable
 to start issuing warnings in 3.9 if any limited quantity exceeds one million.
 
+Historically the recursion limit has been set at 1000. To avoid breaking code that implicitly relies on the value being small,
+the soft recursion limit will be increased gradually, as follows:
```
gvanrossum: How do you plan to implement this?

markshannon: The Python call stack resides on the heap. Changing the interpreter to handle Python calls without consuming C stack is a reasonably straightforward transformation.

As an aside, separating the protection from runaway recursion and C-stack overflow protection would be a good idea regardless.

gvanrossum: I would be happy to support this *after* a working implementation for CPython has been provided as a PR. Until then I am skeptical. In particular, the C API concerns me.

markshannon: Skeptical of what, exactly?
It is hard for me to address your concerns if I don't know what they are.

gvanrossum:
@markshannon 
> Skeptical of what, exactly?
> It is hard for me to address your concerns if I don't know what they are.

Sorry, I should have been clearer. The current CPython implementation takes up a C stack entry for every function call. It's easy enough to modify the interpreter loop to change CALL opcodes to not consume C stack space. It's a little trickier however in the case where the called function is written in C.

Suppose we write `d[k]` where `d` is a dict. This calls `PyDict_GetItem()` which is implemented in C (I don't suppose you want to change that). Now suppose `d` contains key values that have `__hash__` and `__eq__` implementations written in Python. So `PyDict_GetItem()` is likely to call back into Python code, but now there's a C stack frame interleaved with the Python stack. So it's quite possible that such calls will consume C stack frames. Are you going to implement a different limit for this situation? Trying to get rid of it somehow? (I have no idea how to do this, or I'd done it 30 years ago.)

Even if you manage to fix this for internal C API calls that may call back into Python, the same situation can occur with 3rd party extensions, and we'll have to keep a limit on the C stack to avoid segfaults due to C stack overflows. IIUC it's not easy to increase C stack size sufficiently (especially not for threads).

I would be happy to hear that you've found solutions or answers for all these cases, but so far I haven't seen them.

markshannon:
Yes, tracking the size of the C stack would be necessary as well. 
There are APIs for tracking stack sizes and layout, both for posix and windows. Plus there is the portable (if less robust) variant of just taking the address of a local variable and comparing it to a limit pointer. It shouldn't be a problem to implement this in a fashion that is at least as robust as what we currently have.

My thinking is that a call to `PyDict_GetItem()`  would not count towards the recursion limit, but rely on the C stack overflow check to prevent any crashes.

I'll update the PR to state that.

PR #1285, opened 2020-01-28
Updates to PEP 613 Explicit Type Aliases

shannonzhu:
Incorporating some feedback: https://github.com/python/peps/pull/1278

as well as closing the open issue listed previously.

### DIFF
--- a/pep-0613.rst
+++ b/pep-0613.rst
@@ -51,9 +51,10 @@ Forward References:
   def foo() -> MyType: ...
 
 This code snippet should not error so long as :code:`ClassName` is defined
-later on. However, a type checker is forced to throw spurious errors that
-(1) :code:`MyType` is an unannotated global string, and (2) :code:`MyType`
-is not a valid type.
+later on. However, a type checker is forced to assume that MyType is a value
+assignment rather than a type alias, and therefore may throw spurious errors
+that (1) :code:`MyType` is an unannotated global string, and (2) :code:`MyType`
+cannot be used as a return annotation because it is not a valid type.
 
 ::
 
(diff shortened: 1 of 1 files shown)


pep-0613.rst
```diff
@@ -101,10 +102,11 @@ Scope Restrictions:
...
 The outer :code:`x` is a valid type alias, but type checkers must error if the
-inner :code:`x` is ever used as a type. This is confusing because the alias
-declaration rule is not explicit, and the error is not thrown on the location
-of the inner type alias declaration but rather on every one of its subsequent
-use cases.
+inner :code:`x` is ever used as a type because type aliases cannot be defined
+inside a nested scope.
+This is confusing because the alias declaration rule is not explicit, and because
+a type error will not be thrown on the location of the inner type alias declaration
+but rather on every one of its subsequent use cases.
```
gvanrossum: I just realize something (about the example below, where I can't attach a comment): why don't we allow `x: TypeAlias = ClassName` inside a function? It's entirely unambiguous. (The reason may be that it's complicated for type checkers to handle type aliases with limited scope, although it would seem that nested class declarations work just fine.)

gvanrossum:
(I realize I am being ambiguous. If @shannonzhu comes back with a quick response to my comment we can keep iterating -- if not let's land this first.)

pep-0613.rst
shannonzhu: At the moment MyPy and Pyre both disallow nested aliases, I assume b/c getting them right is equivalent to getting nested classes as types right within the scope of the parent class, but is a little more complicated otherwise.

Due to type variable scoping, it's not always sound to be able to reference a nested type alias outside of its scope using qualification, ie:
```
class Foo(Generic[_T]):
    MyType: TypeAlias = List[_T]

def foo() -> Foo.MyType: ...
```

Pyre currently doesn't have a system for scoping types (ie., disallowing the bad usage above), and I think MyPy and Pyre actually have identical behavior at the moment, where nested aliases are not supported at all, and nested classes can be referred to in-scope as well as with qualification outside of the parent class. Are nested aliases something we want to support?

gvanrossum: Mypy doesn't understand the `: TypeAlias` notation yet; without that, it gives an accurate error on your example, I think:
```
class Foo(Generic[_T]):
    MyType = List[_T]  # E: Can't use bound type variable "_T" to define generic alias
```
(So it seems to understand that the intention is a type alias, and it realizes there is a problem with it.)

But I'm happy not to open this can of worms, so I agree with your current wording. I'll land the PR and then I'll let the SC know.

PR #1293, opened 2020-02-03
PEP 3132: Fix example description

birkenfeld:
With two elements, the starred middle expression just is an empty list.


Rosuav:
Was this an error as it originally was published, or is it something that changed later on? Does anyone recall?

birkenfeld:
I don't think it ever worked (or was supposed to) differently; the equivalent expression with slicing shows my intent.

Rosuav:
Gotcha, thanks for the clarification.

PR #1294, opened 2020-02-03
PEP 458: add note about compressing metadata

trishankatdatadog:
Add note about compressing metadata per @ewdurbin feedback


ewdurbin:
In discussion I referenced this sentence from the very last part of the "Metadata Scalability" section, which also discusses JSON metadata. I'm not sure if it should be duplicated in both places.

trishankatdatadog:
Oh, doh, I didn't realize! Let me just close this then?

ewdurbin:
I think so, I'm not sure which section it fits best in though.

trishankatdatadog:
I think where it is works best

PR #1295, opened 2020-02-04
PEP 458: update list of authors

trishankatdatadog:
@mnm678 Any objections?


Rosuav:
It looks like the build system wants consistent email addresses for people, and PEP 480 has a different address @trishankatdatadog . Was your intention to change your address across the board? If so, just add a commit changing PEP 480 as well. If you wanted to have different addresses on different PEPs, though, I'm not sure how to do that.

trishankatdatadog:
@Rosuav Done, thanks :)

Rosuav:
LGTM!

PR #1298, opened 2020-02-05
PEP 584: Assorted cleanup.

brandtbucher:
Just a few things I noticed on the published build. This also includes a link to the newest discussion on python-dev.


brandtbucher:
Wow that was fast.

PR #1306, opened 2020-02-15
PEP 458: Mark as Accepted

ncoghlan:


ncoghlan:
@dstufft is away from convenient GitHub access at the moment, so I'm updating this as per his Discourse post.

PR #1308, opened 2020-02-16
PEP 458: Use 'signed repository metadata' terminology

ncoghlan:
James Bennett pointed out that "package signing" is widely misunderstood
as referring specifically to end-to-end signing with individual
publisher keys, rather than to metdata signing in general.

This updates the title and abstract to instead use the term "signed
repository metadata", and also updates a few other sections that
still gave the impression that implementing PEP 458 would be
enough to give the full end-to-end signing support that is actually
covered in PEP 480.

### DIFF
--- a/pep-0458.txt
+++ b/pep-0458.txt
@@ -1,5 +1,5 @@
 PEP: 458
-Title: Secure PyPI downloads with package signing
+Title: Secure PyPI downloads with signed repository metadata
 Version: $Revision$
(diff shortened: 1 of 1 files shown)


ncoghlan:
@trishankatdatadog @mnm678 A PEP 458 wording tweak as per the Discourse thread (no functional changes, just making it clearer that PEP 458 does *not* cover full end-to-end package signing from the publisher to the consumer).

PR #1309, opened 2020-02-17
PEP 584: Apply feedback from the Python-Dev discussion.

brandtbucher:
These are just a collection of clarifications that were requested on- and off-list.

### DIFF
--- a/pep-0584.rst
+++ b/pep-0584.rst
@@ -96,14 +96,6 @@ entirely string-keyed::
       ...
     TypeError: keywords must be strings
 
-Even then, it should not be promoted as the "one obvious way".  Guido
-`said it best
-<https://mail.python.org/archives/list/python-dev@python.org/message/FM66WWN5YFCMRGMRWGHACOXLQYTWNXE2/>`_:
-
-    Apparently dict(x, \*\*y) is going around as "cool hack" for "call
-    x.update(y) and return x". Personally I find it more despicable
-    than cool.
-
 
 =========
 Rationale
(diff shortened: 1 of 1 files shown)


pep-0584.rst
```diff
@@ -166,21 +191,24 @@ An *approximate* pure-Python implementation is::
     def __or__(self, other):
         if not isinstance(other, dict):
             return NotImplemented
-        new = self.copy()
+        new = dict.copy(self)
```
gvanrossum: I typically spell this `dict(self)`.

pep-0584.rst
```diff
@@ -82,6 +82,29 @@ ChainMap will modify the original dict::
...
+      ...
+    TypeError: keywords must be strings
+
+Even then, it should not be promoted as the "one obvious way".  Guido
+`said it best
+<https://mail.python.org/archives/list/python-dev@python.org/message/FM66WWN5YFCMRGMRWGHACOXLQYTWNXE2/>`_:
+
+    Apparently dict(x, \*\*y) is going around as "cool hack" for "call
+    x.update(y) and return x". Personally I find it more despicable
+    than cool.
```
gvanrossum: I don't think my emotional response adds much insight to the PEP (I also misstated the trick). I think the fact that it only works for string keys is enough to sink it.

brandtbucher:
Thanks @gvanrossum, I agree on both points. Ready for another review.

gvanrossum:
Brandt,

Your next step would be to email steering-council@python.org with a request
to review and hopefully approve this PEP.

--Guido

gvanrossum:
> Your next step would be to email steering-council@python.org with a request
> to review and hopefully approve this PEP.

Whoops, I forgot I am already BDFL-Delegate myself. So I'll just announce that I'm accepting it.

PR #1310, opened 2020-02-18
PEP 596: Update note now that PEP 602 is accepted

hugovk:
Since PEP 602 ("Annual Release Cycle for Python") has been accepted, update the note.


ambv:
Thanks! ✨ 🍰 ✨

PR #1318, opened 2020-02-25
PEP 483, 484, 585: Use parameterized instead of parametrized

ethanhs:
As requested by @gvanrossum

### DIFF
--- a/pep-0585.rst
+++ b/pep-0585.rst
@@ -34,13 +34,13 @@ and easier for teachers to teach Python.
 Terminology
 ===========
 
-Generic (n.) -- a type that can be parametrized, typically a container.
+Generic (n.) -- a type that can be parameterized, typically a container.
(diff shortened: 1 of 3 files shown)


pep-0585.rst
```diff
@@ -283,38 +283,38 @@ With ``__class_getitem__`` as an identity function::
 The indexing being successful here would likely end up raising an
 exception at a distance, confusing the user.
 
-Disallowing instantiation of parametrized types
+Disallowing instantiation of parameterized types
 -----------------------------------------------
```
gvanrossum: Whoops, need to add an extra `-` to this line.

ethanhs: Saw that, done!

gvanrossum:
(Heh, I found some other uses by searching the PEP for "parametriz". I push fixes without going through the PR process. :-)

PR #1319, opened 2020-02-25
PEP 585: Explain the choice of GenericAlias over GenericType

gvanrossum:
A brief summary of @ncoghlan's explanation on python-dev.


ambv:
Thanks! ✨ 🍰 ✨

PR #1322, opened 2020-02-28
PEP 13: update for the latest steering council

brettcannon:


pep-0013.rst
```diff
@@ -334,6 +331,7 @@ History of council elections
 ----------------------------
 
 * January 2019: PEP 8100
+* December 2020: PEP 8101
```
JelleZijlstra: Surely this was December 2019?

brettcannon: Fixed

PR #1327, opened 2020-03-07
PEP 534: Mark as Deferred

ncoghlan:
None of us are currently working on this PEP, so remove the Python-Version target, mark it as Deferred, and explain the key work that remains to be done.


encukou:
Yes: sadly, we can't find the time for this :(
Thanks for documenting it.

PR #1339, opened 2020-03-24
PEP 616: Python-implementation clarity

jheiv:
In attempt to get out ahead of the `str.__getitem__(self, slice(str.__len__(option), None))` version that was posted ~4 hours ago, this:

- Removes all of the unnecessary copying, casting.  This does two things, (1) it allows for non-builtin string subclasses to work provided they support `__getitem__`, (2) it obviates the need to talk about potentially returning `self` later in the doc.
- Tweaks handling of empty string (without affecting the overall result).  And also removes the (now) unnecessary discussion about it.
- Adds TypeError messages (based on those of `.startswith`, but they may just be unnecessary clutter)

### DIFF
--- a/pep-0616.rst
+++ b/pep-0616.rst
@@ -46,64 +46,58 @@ Specification
 The builtin ``str`` class will gain two new methods with roughly the
 following behavior::
 
-    def cutprefix(self, prefix, /):
-        if not isinstance(self, str):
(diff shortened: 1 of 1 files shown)


jheiv:
This should be a PR against the source, not Python -- self-closing.

PR #1342, opened 2020-03-25
PEP 8: Extract Yes/No flags from code blocks

HanifBirgani:
Unifying "Yes/No" flags format in PEP 8 which described in #1341

### DIFF
--- a/pep-0008.txt
+++ b/pep-0008.txt
@@ -81,10 +81,9 @@ using Python's implicit line joining inside parentheses, brackets and
 braces, or using a *hanging indent* [#fn-hi]_.  When using a hanging
 indent the following should be considered; there should be no
 arguments on the first line and further indentation should be used to
-clearly distinguish itself as a continuation line.
+clearly distinguish itself as a continuation line::
 
-Yes::
-
+    # Correct:
     # Aligned with opening delimiter.
     foo = long_function_name(var_one, var_two,
                              var_three, var_four)
@@ -100,8 +99,9 @@ Yes::
         var_one, var_two,
         var_three, var_four)
 
-No::
+::
 
+    # Wrong:
     # Arguments on first line forbidden when not using vertical alignment.
     foo = long_function_name(var_one, var_two,
         var_three, var_four)
@@ -254,10 +254,9 @@ But this can hurt readability in two ways: the operators tend to get
 scattered across different columns on the screen, and each operator is
 moved away from its operand and onto the previous line.  Here, the eye
 has to do extra work to tell which items are added and which are
-subtracted.
+subtracted::
 
-No::
-
+    # Wrong:
     # operators sit far away from their operands
     income = (gross_wages +
               taxable_interest +
@@ -272,10 +271,9 @@ within a paragraph always break after binary operations and relations,
 displayed formulas always break before binary operations" [3]_.
 
 Following the tradition from mathematics usually results in more
-readable code.
+readable code::
 
-Yes::
-
+    # Correct:
     # easy to match operators with operands
     income = (gross_wages
               + taxable_interest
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

I'm a bot set up to make sure that the project can legally accept this contribution by verifying everyone involved has signed the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA).



## CLA Missing

Our records indicate the following people have not signed the CLA:

@HanifBirgani

For legal reasons we need all the people listed to sign the CLA before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.

If you have recently signed the CLA, please wait at least one business day
before our records are updated.


Thanks again for the contribution, we look forward to reviewing it!

gvanrossum:
Do you have a date by which you're expecting to have the updated version ready? (With Correct/Wrong comments in the code blocks.)

HanifBirgani:
It would be ready tomorrow, but I have a question here #1341

pep-0008.txt
```diff
@@ -639,10 +674,9 @@ When to Use Trailing Commas
...
 the ``print`` statement).  For clarity, it is recommended to surround
-the latter in (technically redundant) parentheses.
-
-Yes::
+the latter in (technically redundant) parentheses::
 
+    # Correct:
     FILES = ('setup.cfg',)
 
 OK, but confusing::
```
gvanrossum, line 682: I think we should add a `# Wrong:` comment to the block below.

pep-0008.txt
```diff
@@ -81,10 +81,9 @@ using Python's implicit line joining inside parentheses, brackets and
 braces, or using a *hanging indent* [#fn-hi]_.  When using a hanging
 indent the following should be considered; there should be no
 arguments on the first line and further indentation should be used to
-clearly distinguish itself as a continuation line.
-
-Yes::
+clearly distinguish itself as a continuation line::
 
+    # Correct:
```
gvanrossum, line 86: For this one (only), can you insert a blank line after Correct, and ditto after Wrong? This is because otherwise the scope of the Correct/Wrong is unclear, since there are several blocks with comments.

There are 1-2 others like this that I'll point out explicitly.

pep-0008.txt
```diff
@@ -1259,10 +1293,9 @@ Programming Recommendations
   any return statement returns an expression, any return statements
   where no value is returned should explicitly state this as ``return
   None``, and an explicit return statement should be present at the
-  end of the function (if reachable).
-
-  Yes::
+  end of the function (if reachable)::
 
+      # Correct:
```
gvanrossum, line 1298: Here I also think we could use an extra blank line after the Correct/Wrong comments.

pep-0008.txt
```diff
@@ -1409,18 +1464,18 @@ similar to those on function annotations described above:
 - There should be no space before the colon.
 
 - If an assignment has a right hand side, then the equality sign should have
-  exactly one space on both sides.
-
-- Yes::
+  exactly one space on both sides::
 
+      # Correct:
```
gvanrossum, line 1469: Again, let's try a blank line after the Correct/Wrong comment here.

pep-0008.txt
HanifBirgani: Should the "OK, but confusing" line be removed?

pep-0008.txt
HanifBirgani: Ok, I'll do it.

pep-0008.txt
HanifBirgani: Ok

pep-0008.txt
HanifBirgani: I'll add it.

pep-0008.txt
gvanrossum: Yeah, I think so. The "OK" part just refers to "this works", which is implicit (a style guide doesn't have to point out things that are syntax errors ;-).

HanifBirgani: Ok

gvanrossum:
Thanks, Hanif!

PR #1345, opened 2020-03-28
PEP 483: Remove reference to implicitly Optional parameters

takluyver:
This shorthand is no longer recommended - see #689 which changed it in PEP 484.


Rosuav:
LGTM. Thanks for flagging with approval @ilevkivskyi .

takluyver:
Thanks, impressively speedy turnaround!

PR #1347, opened 2020-03-29
PEP 615: Resolve "using datetime" and add rationale to updates section

pganssle:
No one has really commented on the section about using the datetime module, so I think we can go ahead and move that to "Rejected ideas".

I've also clarified the rationale for not allowing the `ZoneInfo` object's behavior to change.

### DIFF
--- a/pep-0615.rst
+++ b/pep-0615.rst
@@ -170,10 +170,19 @@ added to the cache.
 Behavior during data updates
 ############################
 
-If a source of time zone data is updated during a run of the interpreter, it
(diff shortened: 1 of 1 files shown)


pganssle:
I have recently re-advertised this and I'm planning to submit it to the SC, so I'm going to go ahead and merge this. If anyone has additional comments or edits, feel free to review and I will address them along with whatever the next set of changes is (responses, changing the status to "accepted", adding acknowledgements, etc).

PR #1351, opened 2020-03-31
PEP 617: New PEG parser for CPython

pablogsal:


pablogsal:
CC: @lysnikolaou @gvanrossum

gvanrossum:
> Comparing the last two bullet points below, this doesn't sound right. Am I missing something?

I'm not sure what seems wrong to you? Are we talking about these two?

> - Compiling to bytecode took 3.290 seconds. That's 227,861 lines/sec,
>   or 8,396,942 bytes/sec. Max RSS 77 MiB.
> 
> - Compiling to bytecode using the current parser took 3.367 seconds.
>   That's 222,620 lines/sec, or 8,203,780 bytes/sec. Max RSS 70 MiB.

The second refers to the old parser (which I call "current" in this section). It's slower but uses less memory.

gvanrossum:
Once @lysnikolaou's comment is resolved let's just land this and claim the PEP number.

pablogsal:
We can certainly change this latter, but now that we are at it I added:

> Discussions-To: Python-Dev <python-dev@python.org>

would you prefer `https://discuss.python.org/` instead?

gvanrossum:
I prefer python-dev, so let's keep that.

lysnikolaou:
> > Comparing the last two bullet points below, this doesn't sound right. Am I missing something?
> 
> I'm not sure what seems wrong to you? Are we talking about these two?
> 
> > * Compiling to bytecode took 3.290 seconds. That's 227,861 lines/sec,
> >   or 8,396,942 bytes/sec. Max RSS 77 MiB.
> > * Compiling to bytecode using the current parser took 3.367 seconds.
> >   That's 222,620 lines/sec, or 8,203,780 bytes/sec. Max RSS 70 MiB.
> 
> The second refers to the old parser (which I call "current" in this section). It's slower but uses less memory.

There is the sentence "Here our parser came out a little slower", but the numbers show that it takes less time to compile to bytecode with our parser than with the current one, right?

gvanrossum:
> There is the sentence "Here our parser came out a little slower",

Oh! You're right. I'm deleting that sentence then, since the conclusion below the bullets is correct. (And the first set of bullets didn't have a sentence above it announcing the conclusion.)

lysnikolaou:
> > There is the sentence "Here our parser came out a little slower",
> 
> Oh! You're right. I'm deleting that sentence then, since the conclusion below the bullets is correct. (And the first set of bullets didn't have a sentence above it announcing the conclusion.)

Nice. Let's land this!!

lysnikolaou:
Is this intentional? 
Screenshot 2020-04-01 at 00 51 32

pablogsal:
> Nit: Is this intentional?

Nop

lysnikolaou:
Are we changing this or can we leave it as is for now?

pablogsal:
> Are we changing this or can we leave it as is for now?

Will do a PR to fix it

PR #1354, opened 2020-04-01
PEP 384: Update PyType_Spec struct per bpo-15140

furkanonder:
https://bugs.python.org/issue15140

In PEP 384, the PyType_Spec struct has  "const char *doc" and "unsigned int flags" these is not present in the implementation (Include/object.h).


furkanonder:
You're welcome!

PR #1364, opened 2020-04-10
PEP 558: Postpone to 3.10

ncoghlan:
Leaving as Draft as I'm still intending to work on it in the coming
months, it just isn't going to be ready in time for the 3.9 beta 1
deadline.

### DIFF
--- a/pep-0558.rst
+++ b/pep-0558.rst
@@ -34,21 +34,6 @@ Python C API/ABI::
 It also proposes the addition of several supporting functions and type
 definitions to the CPython C API.
 
-
-Postponed to Python 3.10
-========================
-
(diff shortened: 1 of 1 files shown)


pep-0558.rst
```diff
@@ -35,6 +35,20 @@ It also proposes the addition of several supporting functions and type
 definitions to the CPython C API.
 
 
+Postponed to Python 3.10
+========================
```
vstinner, line 39: Not sure that this section is useful. Just update "Python-Version:" header, no?

This PEP missed Python 3.9 feature freeze.

ncoghlan: Fair point - I've taken it out.

PR #1368, opened 2020-04-14
PEP 597: 3rd edition.

methane:


brettcannon:
@methane any reason you haven't merged this yet?

PR #1371, opened 2020-04-17
Make `from __future__ import annotations` the default in Python 3.10

ambv:
The Language Summit discussion suggests this should be flipped either in Python 3.9 or in Python 3.10. An informal poll ran at the event with 33 votes has shown that 65% would like to see the change in 3.9, 29% in 3.10, 3% in 4.0, and 3% "never" :-)

This PR is open to gather feedback from those concerned.

Side note: The previously documented policy of raising deprecation warnings if there's any annotation was unrealistic. It would raise huge numbers of warnings, a lot of them in third-party code that the receiver of the warning can do nothing about. That part we need to remove from the document anyway.

### DIFF
--- a/pep-0563.rst
+++ b/pep-0563.rst
@@ -280,13 +280,7 @@ Deprecation policy
 Starting with Python 3.7, a ``__future__`` import is required to use the
 described functionality.  No warnings are raised.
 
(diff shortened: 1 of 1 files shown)


vstinner:
https://bugs.python.org/issue38605

PR #1380, opened 2020-04-23
PEP 571: Change status to Final

brainwane:
Per https://github.com/pypa/manylinux/issues/179 , manylinux2010 is
now implemented in its entirety. Per PEP 1, "When the reference
implementation is complete and incorporated into the main source code
repository, the status will be changed to 'Final'."

Signed-off-by: Sumana Harihareswara <sh@changeset.nyc>


ncoghlan:
The informational PEPs are a little different in this regard - they move to Active & stay there until superseded rather than moving through Accepted & Final.

We've been migrating more of the packaging PEPs to the standards track lifecycle, but manylinux is a special case, as https://packaging.python.org/specifications/platform-compatibility-tags/#platform-tags-for-common-linux-distributions relies on the PEPs to define the details of each of the tags.

PR #1391, opened 2020-05-04
Reject PEP 349: Outdated by Python 3

vstinner:


vstinner:
cc @nascheme @gvanrossum

PR #1406, opened 2020-05-19
PEP 563: Make `from __future__ import annotations` the default in Python 3.10

hugovk:
Follow on from https://github.com/python/peps/pull/1371.

Ping @ambv.


ambv:
Thanks! ✨ 🍰 ✨

mankjass (commented):
n

PR #1407, opened 2020-05-19
PEP 596: 3.9.0b1 was released on 2020-05-19

hugovk:
Python 3.9 beta 1 was released today:

* https://pythoninsider.blogspot.com/2020/05/python-390b1-is-now-available-for.html


ambv:
The release commit and tag is from May 18 and the files were available to download on May 18 on our FTP. I only announced the day later because we were waiting for the Windows installer.

PR #1432, opened 2020-06-15
Update PEP 387 based on discussions

brettcannon:


pep-0387.txt
```diff
@@ -48,31 +49,42 @@ This policy applies to all public APIs.  These include:
...
 
-- Function, class, module, attribute, method, and C-API names and types that
-  are prefixed by "_" (except special names).  The contents of these
-  are also not subject to the policy.
+- Function, class, module, attribute, method, and C-API names and
+  types that are prefixed by "_" (except special names).
+
+- Anything documented publicly as being private.
+
+- Imported modules (unless explicitly documented as part of the public
```
benjaminp, line 64: Does this mean, for example, `import lzma; lzma.io.open()` isn't covered?

pep-0387.txt
```diff
@@ -114,7 +126,8 @@ releases:
    reconsider.
 
 5. The behavior change or feature removal may now be made default or
-   permanent in the N+3 release.  Remove the old version and warning.
+   permanent having reached the declared default/removal version.
```
benjaminp, line 129: I would just deleted "default/removal".

pep-0387.txt
brettcannon: Correct. Unless it's documented that `lzma.io` is a thing, it isn't backwards incompatible to remove it.

benjaminp: Okay. It might be clearer to say "modules imported as a dependencies" or something.

PR #1435, opened 2020-06-17
PEP 618: Accepted!

brandtbucher:
CC @pitrou


pitrou:
I'll let @gvanrossum review as I haven't really followed the latest adventures on python-dev ;-)

PR #1438, opened 2020-06-18
PEP 101: Simplified due to centralisation of switchers in the doc.

JulienPalard:
Do not merge before https://github.com/python/cpython/pull/20969.


JulienPalard:
@ambv @pablogsal @ned-deily you're conserned by this one, don't hesitate to tell me if the wording is not clear.

TL;DR: Changes to `indexsidebar.html` and `switchers.js` in all branches are no longer needed, (a change in docsbuild-script is still needed).

PR #1442, opened 2020-06-19
PEP 619: Add PEP 618 to Python 3.10 release schedule

cool-RR:


pablogsal:
Thanks for the PR, @cool-RR !

PR #1443, opened 2020-06-20
PEP 387: typo fix

wimglenn:


Rosuav:
Thanks for catching that!

PR #1445, opened 2020-06-22
Update language on PEP-Delegate

willingc:
Add language about PEP-Delegate and self-nomination notification

### DIFF
--- a/pep-0001.txt
+++ b/pep-0001.txt
@@ -275,12 +275,17 @@ The final authority for PEP approval is the Steering Council. However, whenever
 a new PEP is put forward, any core developer that believes they are suitably
 experienced to make the final decision on that PEP may offer to serve as
 the BDFL-Delegate for that PEP, and they will then have the authority to approve
-(or reject) that PEP. Individuals taking on this responsibility are free to seek
+(or reject) that PEP. Under the Steering Council governance, the terms,
+PEP-Delegate (preferred) and BDFL-Delegate, may be used in discussion. 
+Individuals offering to serve as PEP-Delegate should notify the Steering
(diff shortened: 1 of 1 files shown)


brandtbucher:
Looks good, thanks! Do you mind also clarifying or removing the bit that says:

> However, content review and final acceptance of the PEP must be requested of the core developers, usually via an email to the
python-dev mailing list.

I don't think python-dev has final acceptance power.

Rosuav:
Note that #1672 changed the header to PEP-Delegate, so the wording here may need to be changed to reflect that.

JelleZijlstra:
Yes, there is a merge conflict now.

JelleZijlstra:
This has been open for a long time with little action and has a conflict now. I suggest we close it and open a new PR if the current SC feels that a clarification on the PEP-Delegate procedure is necessary.

gvanrossum:
Should probably just create an entry on the SC tracker, otherwise they will never make time for this. I think this is a good update to the PEP, too bad about the conflict.

pep-0001.txt
```diff
@@ -275,12 +275,17 @@ The final authority for PEP approval is the Steering Council. However, whenever
 a new PEP is put forward, any core developer that believes they are suitably
 experienced to make the final decision on that PEP may offer to serve as
 the BDFL-Delegate for that PEP, and they will then have the authority to approve
```
CAM-Gerlach, line 277: Just FYI, if this is re-opened, this line should probably be modified as well, since it mentions only `BDFL-Delegate` in the context of current processes.

willingc:
I'm happy to fix the merge conflict and reopen. The prior Steering Council had recommended the changes.

gvanrossum:
Go for it!

PR #1451, opened 2020-06-23
PEP 622: Structural Pattern Matching (Bucher, Kohn, Levkivskyi, GvR, Talin)

gvanrossum:
I am about to post about this to python-dev.


gvanrossum:
Pushing it now so I can reference it in the email to python-dev.

PR #1458, opened 2020-06-23
PEP 618: Final

brandtbucher:
CC @pitrou


gvanrossum:
Whoops, I hit return before editing the subject. :-(

PR #1463, opened 2020-06-24
PEP 622: Update Post-History

brandtbucher:


gvanrossum:
Wait, are you planning to do a revised post today?

brandtbucher:
No, I thought it should reflect the new thread you just started. 

I'm still a bit unsure of when exactly this field should be updated.

gvanrossum:
No, the new thread is just to focus the discussion (the old thread is 95 messages long and GMail will split it once it reaches 100). I think we should declare a new "posting" when we have substantially changed our position, or dealt with alternate ideas in the Rejected Ideas section. Right now, I see a few ways it can go:

1. We stick to our guns, and add text to Rejected Ideas about wildcards, `|`, `else`, and `?`
2. We add `else` but stick to our guns for everything else
3. We go for `?x` or `x?`, make plain `?` the wildcard, and add `else` (since `case?:` is ugly)

Once we're agreed on a choice here we can "post" a new version.

PR #1473, opened 2020-06-27
PEP 622: Fix wrong statements of Haskell

thautwarm:
The well-known languages with bound-once matching are Erlang and Prolog.

[Prolog is matching patterns with unifications](http://www.deransart.fr/prolog/unification.html), hence this technique is natural to achieve.

[Erlang is using scope to classify if variable is to be bound or loaded](https://erlang.org/doc/reference_manual/expressions.html#variables), this is closer to PEP 622.

### DIFF
--- a/pep-0622.rst
+++ b/pep-0622.rst
@@ -1551,7 +1551,7 @@ rarely.
 Algebraic matching of repeated names
 ------------------------------------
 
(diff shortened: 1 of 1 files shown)


thautwarm:
> Prolog is not a functional language

Yes.

> Does it perhaps also appear in elixir?

No, Elixir uses pin operators to explicitly indicate variable reuse.

PR #1475, opened 2020-06-27
PEP 622: Don't special-case complex or type

brandtbucher:


gvanrossum:
How are you implementing this?

brandtbucher:
Type-checks in the VM if no `__match_args__`.

Although we could do a flag on the type object if that feels too magical (also allows us to opt-in other extension types if needed).

gvanrossum:
Yeah, an undocumented flag seems better.

Of course a user can get the same behavior with a proxy and a private attribute, but for these built-ins a special case seems better.

PR #1476, opened 2020-06-28
PEP 432: Withdraw to make way for future subproposals

ncoghlan:

### DIFF
--- a/pep-0432.txt
+++ b/pep-0432.txt
@@ -25,7 +25,8 @@ For most of that time, the changes were maintained either in a separate feature
 branch, or else as underscore-prefixed private APIs in the main CPython repo.
 
 In 2019, PEP 587 migrated a subset of those API changes to the public CPython
-API for Python 3.8+.
+API for Python 3.8+ (specifically, the PEP updated the interpreter runtime to
+offer an explicitly multi-stage struct-based configuration interface).
 
 In June 2020, in response to a query from the Steering Council, the PEP authors
 decided that it made sense to withdraw the original PEP, as enough has changed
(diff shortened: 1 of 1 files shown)


pep-0432.txt
```diff
@@ -6,13 +6,48 @@ Author: Nick Coghlan <ncoghlan@gmail.com>,
...
+infrastructure changes needed to enable them, would likely still be worth
+exploring:
+
+* shipping an alternate Python executable that ignores all user level
+  settings and runs in isolated mode by default, and would hence be more
+  suitable for execution of system level Python applications than the default
+  interpreter
+* enhancing the zipapp module to support the creation of single-file executables
+  from pure Python scripts (and potentially even Python extension modules, given
+  the introduction of multi-phase extension module initialisation)
```
vstinner, line 50: Hum, one of the PEP 432 motivation was to rewrite the getpath.c code. IMHO it's still a valid point :-)

pep-0432.txt
```diff
@@ -6,13 +6,48 @@ Author: Nick Coghlan <ncoghlan@gmail.com>,
...
+
+From late 2012 to mid 2020, this PEP provided general background and specific
+concrete proposals for making the CPython startup sequence easier to maintain
+and the CPython runtime easier to embed as part of a larger application.
+
+For most of that time, the changes were maintained either in a separate feature
+branch, or else as underscore-prefixed private APIs in the main CPython repo.
+
+In 2019, PEP 587 migrated a subset of those API changes to the public CPython
+API for Python 3.8+.
```
vstinner, line 28: Hum, maybe make it more explicit that PEP 587 is mostly the concrete implementation of the "configuration" part of this PEP.

vstinner:
I would still be very interested by the ability to configure the Path Configuration https://docs.python.org/dev/c-api/init_config.html#init-path-config in Python, but I don't have the bandwidth to experiment an implementation.

PR #1479, opened 2020-06-29
PEP 622: Rename name pattern to capture pattern

gvanrossum:
Also make the section headings describing various types of pattern plural, e.g. "Capture patterns".

See https://github.com/gvanrossum/patma/issues/112

### DIFF
--- a/pep-0622.rst
+++ b/pep-0622.rst
@@ -893,7 +893,7 @@ variables. For example::
 Note about constants
 --------------------
 
-The fact that capture pattern is always an assignment target may create unwanted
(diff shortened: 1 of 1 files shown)


pep-0622.rst
```diff
@@ -893,7 +893,7 @@ variables. For example::
 Note about constants
 --------------------
 
-The fact that name pattern is always an assignment target may create unwanted
+The fact that capture pattern is always an assignment target may create unwanted
```
brandtbucher, line 896:
```suggestion
The fact that a capture pattern is always an assignment target may create unwanted
```

PR #1484, opened 2020-06-30
PEP 622: Drop __match__ from the spec

brandtbucher:
Just a rough draft to see what this would look like (I didn't bother with the rejected or deferred ideas sections).

I'll note that these changes naturally lent themselves to placing increased emphasis on `__match_args__`, which I think is good.

### DIFF
--- a/pep-0622.rst
+++ b/pep-0622.rst
@@ -100,12 +100,12 @@ the proposed code will work without any modifications to the definition of
 
 See the `syntax`_ sections below for a more detailed specification.
 
-Similarly to how constructing objects can be customized by a user-defined
-``__init__()`` method, we propose that destructuring objects can be customized
-by a new special ``__match__()`` method. As part of this PEP we specify the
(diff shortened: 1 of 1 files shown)


pep-0622.rst
```diff
@@ -414,26 +414,25 @@ example::
...
-and some builtin and standard library classes, custom classes are only
-required to follow the protocol specified in `runtime`_ section. After all,
-the authors of a class know best how to "revert" the logic of the
-``__init__()`` they wrote. The runtime will then chain these calls to allow
-matching against arbitrarily nested patterns.
+By default, sub-patterns may only be matched by keyword for
+user-defined classes. In order to suport positional sub-patterns, a
+custom ``__match_args__`` attribute is required. After all, the
+authors of a class know best how to "revert" the logic of the
+``__init__()`` they wrote. The runtime allows matching against
```
gvanrossum, line 433:
```suggestion
custom ``__match_args__`` attribute is required.
The runtime allows matching against
```

PR #1486, opened 2020-06-30
PEP 508: Add missing parentheses

y9k8i:
The definition of python_version was updated in #1123.
However, two closing parentheses are missing:
https://github.com/python/peps/blob/47fa1fb501121f6c8956493b28b30f4300fbd1df/pep-0508.txt#L280
https://github.com/python/peps/blob/47fa1fb501121f6c8956493b28b30f4300fbd1df/pep-0508.txt#L537
Note: It is not necessary to change the line 519 because a closing parenthesis exists.


Rosuav:
Seems the code in the original PR was right, the textual examples have the typo. Thanks for catching this.

PR #1497, opened 2020-07-06
PEP 624: Remove Py_UNICODE encoder APIs

methane:


pep-0624.rst
```diff
@@ -0,0 +1,233 @@
...
+
+Add ``Py_DEPRECATED(3.3)`` to these APIs.
+
+* ``PyUnicode_EncodeDecimal()``
+* ``PyUnicode_TransformDecimalToASCII()``.
+
+This change is committed already.
+https://github.com/python/cpython/commit/9c3840870814493fed62e140cfa43c2883e12181
+
+All other APIs have been marked ``Py_DEPRECATED(3.3)`` already.
```
vstinner, line 123: Currently, the documentation says ".. deprecated-removed:: 3.3 4.0". Can you please also update this documentation, and announce the removal in What's New in Python 3.9?

pep-0624.rst
```diff
@@ -0,0 +1,233 @@
...
+
+Python 3.10
+-----------
+
+Remove these APIs.
```
vstinner, line 128: I would prefer to have at least one release which emits DeprecationWarning *at runtime* before the removal. Maybe it's still acceptable for Python 3.9, if, as you wrote, these functions are basically unused by the top 4000 PyPI projects. But you will need the 3.9 release manager approval.

If it's too late for 3.9, the runtime warning should be added to 3.10.

I'm fine with only having a single release between introducing the warnings and the removal. Even if PEP 387 does/will require 2 releases (2 years), the PEP explains why for these functions, it's acceptable for have a shorter deprecation period.

pep-0624.rst
```diff
@@ -0,0 +1,233 @@
...
+
+Pros:
+
+* We can have more consistent API set.
+
+Cons:
+
+* We have more public stable APIs to maintain.
+* Existing public APIs are enough for most use cases, and
+  ``PyUnicode_AsEncodedString()`` can be used in other cases.
```
vstinner, line 185: I would add that the main advantage of specialized functions is performance, but so far, there was no explicit user request nor benchmark showing a significant speedup.

pep-0624.rst
```diff
@@ -0,0 +1,233 @@
...
+* User can encode UCS-4 string in C without creating Unicode object.
+
+Cons:
+
+* We have more public stable APIs to maintain.
+* Applications which uses UTF-8 or UTF-32 can not use these APIs
+  anyway.
+* Other Python implementations may not have builtin codec for UCS-4.
+* If we change the Unicode internal representation to UTF-8, we need
+  to keep UCS-4 support only for these APIs.
```
vstinner, line 208: Py_UNICODE is currently defined as wchar_t. So it's possible to use wide character literal L'c' and wide string literal L"string". It's not possible with Py_UCS4 which is a number, not a character.

Also, this change is likely to break existing C extensions relying on the assumption that Py_UNICODE is wchar_t.

pep-0624.rst
methane: I will update them after this PEP is accepted, because this PEP is written there is no consensus when it is removed.

> I'm not comfortable with removing functions without emitting warning **at runtime** during at least one release. Maybe:

I don't like runtime warning.  For example, `PyUnicode_EncodeUTF8(PyUnicode_AS_UNICODE(obj), ...)` may cause SEGV if obj is borrowed reference.

pep-0624.rst
```diff
@@ -0,0 +1,233 @@
...
+================
+
+There are alternative APIs to accept ``PyObject *unicode`` instead of
+``Py_UNICODE *``.
+
+
+========================================= ====================================================================
+Deprecated API                            Alternative APIs
+========================================= ====================================================================
+``PyUnicode_Encode()``                    ``PyUnicode_AsEncodedString()``
```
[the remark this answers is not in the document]
methane: No. I can not find any code using it.

```
$ rg -w PyUnicode_Encode
Cython-0.29.20/Cython/Includes/cpython/unicode.pxd
277:    object PyUnicode_Encode(Py_UNICODE *s, Py_ssize_t size,
```

methane:
> I'm not comfortable with removing functions without emitting warning **at runtime** during at least one release. Maybe:

I don't like runtime warning.  For example, `PyUnicode_EncodeUTF8(PyUnicode_AS_UNICODE(obj), ...)` may cause SEGV if obj is borrowed reference.

pep-0624.rst
methane: Deprecated APIs are much slower because they creates temporary Unicode object by `PyUnicode_FromWideChar`.

In case of utf-16-le, utf-16-be, utf-32-le, utf-32-be, `PyUnicode_AsEncodedString()` need to search codec from registry. We may be able to add fast paths for them.

pep-0624.rst
methane: Of course, this is breaking change like removing deprecated API.

pep-0624.rst
```diff
@@ -0,0 +1,256 @@
...
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 07-Jul-2020
+Python-Version: 3.11
+
+
+Abstract
+========
+
+This PEP proposes to remove deprecated Py_UNICODE encoder APIs:
```
vstinner, line 14:
```suggestion
This PEP proposes to remove deprecated ``Py_UNICODE`` encoder APIs in Python 3.11:
```

pep-0624.rst
```diff
@@ -0,0 +1,256 @@
+PEP: 624
+Title: Remove Py_UNICODE encoder APIs
+Author: Inada Naoki <songofacandy@gmail.com>
+Status: Draft
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 07-Jul-2020
```
vstinner, line 7: You could keep the original creation date ;-) (I don't ask you to change anything)

pep-0624.rst
vstinner: An extension module cannot look for the encode and decode functions once, and then use it? The io.TextIOWrapper does that for example.

vstinner:
> I don't like runtime warning. For example, PyUnicode_EncodeUTF8(PyUnicode_AS_UNICODE(obj), ...) may cause SEGV if obj is borrowed reference.

Do you mean that PyUnicode_AS_UNICODE() would return NULL? I don't understand.

Please add a section in the Alternatives to explain why you don't want to emit DeprecationWarning.

methane:
> Do you mean that PyUnicode_AS_UNICODE() would return NULL? I don't understand.

I meant obj may be borrowed reference from list, dict, or something else.

* Warning filter may be clear the dict or list.
* While executing warning filter, GIL can be switched to another thread and the dict/list will be changed.

So emitting DeprecationWarning from APIs which didn't execute arbitrary code (e.g. `__eq__`) and didn't release GIL is breaking change.

vstinner:
Let me take an example:

```
PyObject *
PyUnicode_EncodeUTF8(const Py_UNICODE *s,
                     Py_ssize_t size,
                     const char *errors)
{
    PyObject *v, *unicode;
    // (1)
    unicode = PyUnicode_FromWideChar(s, size);
    if (unicode == NULL)
        return NULL;
    // (2)
    v = _PyUnicode_AsUTF8String(unicode, errors);
    Py_DECREF(unicode);
    // (3)
    return v;
}
```

I understand that inserting a warning in (1) can trigger the bug you're talking about. But if you insert the warning in (2) and (3), it's fine, no?

methane:
Assume this code.

```c
PyObject *u = PyList_GET_ITEM(arr, i);  // u is borrowed reference.
// Assumes PyUnicode_EncodeUTF8 never release GIL
PyObject *b = PyUnicode_EncodeUTF8(PyUnicode_AS_UNICODE(u), PyUnicode_GET_SIZE(u), NULL);
// Assumes u is still living borrowed reference. <-- here
PyObject *t = PyTuple_Pack(2, u, b);
Py_DECREF(b);
return t;
```

This code is fine for now, because PyUnicode_EncodeUTF8 never release GIL.
`u` is borrowed but living reference after calling PyUnicode_EncodeUTF8.

But if PyUnicode_EncodeUTF8 release GIL, u may be not dangling reference now.

vstinner:
Well, if you consider that it's a no-go, please just explain in the PEP that it's the reason to not emit DeprecationWarning at runtime.

pep-0624.rst
```diff
@@ -0,0 +1,284 @@
...
+   return t;
+
+If we emit Python warning from ``PyUnicode_EncodeUTF8()``, warning
+filters and other threads may change the ``list`` and ``u`` can be
+a dangling reference after ``PyUnicode_EncodeUTF8()`` returned.
+
+Additionally, since we are not changing behavior but removing C APIs,
+runtime ``DeprecationWarning`` might not helpful for Python
+developers. We should warn to extension developers instead.
+
```
vstinner, line 257: Would you mind to add a section with links to previous discussions/issues? Something like:

```
Discussions
===

* `Plan to remove Py_UNICODE APis except PEP 623
  <https://mail.python.org/archives/list/python-dev@python.org/thread/S7KW2U6IGXZFBMGS6WSJB26NZIBW4OLE/#S7KW2U6IGXZFBMGS6WSJB26NZIBW4OLE>`_
* `bpo-41123: Remove Py_UNICODE APIs except PEP 623: <https://bugs.python.org/issue41123>`_
```

Are you aware of other past discussions?

Since PEP 623 is related but different, it may also help to mention it somewhere in this PEP, and maybe explain that why they are independent.

PR #1499, opened 2020-07-06
PEP 622: Ditch leading dots for name loads

brandtbucher:

### DIFF
--- a/pep-0622.rst
+++ b/pep-0622.rst
@@ -322,20 +322,18 @@ the matching expression (same as for literals)::
 
   from enum import Enum
 
-  class Color(Enum):
-      BLACK = 1
-      RED = 2
+  class Sides(str, Enum):
+      SPAM = "spam"
+      EGGS = "eggs"
+      ...
 
-  BLACK = 1
-  RED = 2
+  match entree[-1]:
+      case Sides.SPAM:  # Compares entree[-1] == Sides.SPAM.
+          response = f"Have you got anything without {Sides.SPAM}?"
+      case side:  # Assigns side == entree[-1].
+          response = f"Well, could I have their {Sides.SPAM} instead of the {side} then?"
 
-  match color:
-      case Color.BLACK:
-          print("Black suits every color")
-      case BLACK:  # This will just assign a new value to BLACK.
-          ...
-
-Note that it is not possible for unqualified names (such as ``BLACK``
+Note that it is not possible for unqualified names (such as ``side``
 in the example above) to be used as constant value patterns.  See
 `rejected ideas`_ for other syntactic alternatives that were
 considered for constant value patterns.
@@ -1131,14 +1129,12 @@ ambiguous with capture patterns. Four other alternatives were considered:
 * Use some implicit rules. For example if a name was defined in the global
   scope, then it refers to a constant, rather than represents a capture pattern::
 
-    FOO = 1
-    value = 0
+    assert "spam" in globals() and "spam" not in locals()
+    assert "side" not in globals()
 
-    match value:
-        case FOO:  # This would not be matched
-            ...
-        case BAR:  # This would be matched
-            ...
+    match entree[-1]:
+        case spam: ...  # Compares entree[-1] == spam.
+        case side: ...  # Assigns side = entree[-1].
 
   This however can cause surprises and action at a distance if someone
   defines an unrelated coinciding name before the match statement.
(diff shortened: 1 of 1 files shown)


brandtbucher:
Yeah, that's probably a good idea. I'll rewrite the existing `BLACK` and `MAX_INT` examples tomorrow and push them to this PR.

pep-0622.rst
```diff
@@ -318,9 +318,7 @@ Constant Value Patterns
 This is used to match against constants and enum values.
 Every dotted name in a pattern is looked up using normal Python name
 resolution rules, and the value is used for comparison by equality with
-the matching expression (same as for literals). As a special case to avoid
-ambiguity with capture patterns, simple names must be prefixed with a dot to be
-considered a reference::
+the matching expression (same as for literals)::
```
gvanrossum, line 321:
```suggestion
the match subject (same as for literals)::
```

pep-0622.rst
```diff
@@ -1186,20 +1179,40 @@ ambiguous with capture patterns. Four other alternatives were considered:
...
-  a given name is a constant to be matched against, not to be assigned to::
+* Introduce a special symbol, for example ``.``, ``$``, or ``^`` to
+  indicate that a given name is a constant to be matched against, not
+  to be assigned to.  An earlier version of this proposal used a
+  leading-dot rule::
 
     FOO = 1
     value = 0
 
     match value:
```
gvanrossum, line 1190: This whole example is confusing.

pep-0622.rst
```diff
@@ -1186,20 +1179,40 @@ ambiguous with capture patterns. Four other alternatives were considered:
...
+  converting them to ``Enum`` types, or enclosing them in their own
+  namespace (considered by the authors to be one honking great idea)::
+
+    class Consts:
+        FOO = 1
+        ...
+
+    value = 0
+
+    match value:
```
gvanrossum, line 1208: This too

gvanrossum:
(Waiting for your updates to the examples.)

brandtbucher:
No problem, I'm busier today so it will be this afternoon/evening most likely.

brandtbucher:
I've unified the examples for all constant value pattern alternatives with a new snippet that's (in my opinion) idiomatic and easier to follow.

pep-0622.rst
```diff
@@ -318,24 +318,22 @@ Constant Value Patterns
...
-  match color:
-      case Color.BLACK:
-          print("Black suits every color")
-      case BLACK:  # This will just assign a new value to BLACK.
-          ...
+  match entree[-1]:
+      case Sides.SPAM:  # Compares entree[-1] == Sides.SPAM.
+          response = f"Have you got anything without {Sides.SPAM}?"
+      case side:  # Assigns side == entree[-1].
+          response = f"Well, could I have their {Sides.SPAM} instead of the {side} then?"
```
gvanrossum, line 334: I'd suggest not distracting the reader with `{Sides.SPAM}` in the f-string.
```suggestion
          response = "Have you got anything without Spam?"
      case side:  # Assigns side == entree[-1].
          response = f"Well, could I have their Spam instead of the {side} then?"
```

pep-0622.rst
```diff
@@ -318,24 +318,22 @@ Constant Value Patterns
...
 
-Note that it is not possible for unqualified names (such as ``BLACK``
+Note that it is not possible for unqualified names (such as ``side``
 in the example above) to be used as constant value patterns.  See
```
gvanrossum: Somehow using `side` above feels distracting too. The original phrase was a bit awkward already. How about this?
```suggestion
Note that there is no way to use unqualified names as constant value
patterns (they always denote variables to be captured).  See
```

pep-0622.rst
```diff
@@ -1131,14 +1129,12 @@ ambiguous with capture patterns. Four other alternatives were considered:
 * Use some implicit rules. For example if a name was defined in the global
   scope, then it refers to a constant, rather than represents a capture pattern::
 
-    FOO = 1
-    value = 0
+    assert "spam" in globals() and "spam" not in locals()
+    assert "side" not in globals()
```
gvanrossum, line 1133: Maybe just state the conditions in English? I totally misread these the first one I saw them.

pep-0622.rst
```diff
@@ -1184,50 +1170,32 @@ ambiguous with capture patterns. Four other alternatives were considered:
...
-        case $BAR:  # This would be matched
-            ...
+    match entree[-1]:
+        case spam: ...   # Compares entree[-1] == spam.
+        case side?: ...  # Assigns side = entree[-1].
 
-  But the capture patterns are more common in typical code, so having special
-  syntax for common case would be weird.
+  But capture patterns are more common in typical code, so it is
+  undesirable to require special syntax for them.
```
gvanrossum, line 1198: There are more arguments against marking capture variables with sigils:
- None of the other languages surveyed do this (and there's no precedent in Python either).
- It would look confusing to see
  ```
  case (%x, %y):
      return Point(x, y)  # Why not Point(%x, %y)
  ```

PR #1502, opened 2020-07-07
PEP 622: Why the match statement is the correct feature right now

dmoisset:
Explain why the PEP proposes this and only this construct.

### DIFF
--- a/pep-0622.rst
+++ b/pep-0622.rst
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
 with pattern matching, since although it may be handy, it will likely be used
 rarely.
 
+Other pattern-based constructions
+---------------------------------
+
+Many other languages supporting pattern-matching use it as a base for multiple
+language constructs, including a matching operator, a generalized form 
+of assignment, a filter for loops, a method for synchronizing communication,
+or specialized if statements. Some of these were mentioned in the discussion
+of the first draft. Another question asked is why this particular form (joining
+binding and conditional selection) and not others were chosen.
+
+Introducing more uses of patterns would be too bold and premature for the
+experience we have using patterns and would make this proposal too
+complicated. The statement as presented provides a form of the feature that
+is sufficiently general to be useful while being self-contained and not
+having a massive impact in the syntax and semantics of the language as a whole.
+
+After some experience with this feature, the community may have a better
+knowledge of what other pattern matching constructs could be valuable in Python.
 
 Algebraic matching of repeated names
 ------------------------------------
--- a/pep-0622.rst
+++ b/pep-0622.rst
@@ -1767,7 +1767,7 @@ Many other languages supporting pattern-matching use it as a basis for multiple
 language constructs, including a matching operator, a generalized form 
 of assignment, a filter for loops, a method for synchronizing communication,
 or specialized if statements. Some of these were mentioned in the discussion
-of the first draft. Another question asked is why this particular form (joining
+of the first draft. Another question asked was why this particular form (joining
 binding and conditional selection) was chosen while other forms were not.
 
 Introducing more uses of patterns would be too bold and premature given the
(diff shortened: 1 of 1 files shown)


pep-0622.rst
```diff
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
...
+language constructs, including a matching operator, a generalized form 
+of assignment, a filter for loops, a method for synchronizing communication,
+or specialized if statements. Some of these were mentioned in the discussion
+of the first draft. Another question asked is why this particular form (joining
+binding and conditional selection) was chosen while other forms were not.
+
+Introducing more uses of patterns would be too bold and premature for the
+experience we have using patterns and would make this proposal too
+complicated. The statement as presented provides a form of the feature that
+is sufficiently general to be useful while being self-contained and not
```
gvanrossum, line 1776:
```suggestion
is sufficiently general to be useful while being self-contained, and without
```

pep-0622.rst
```diff
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
...
+having a massive impact in the syntax and semantics of the language as a whole.
```
gvanrossum, line 1777:
```suggestion
having a massive impact on the syntax and semantics of the language as a whole.
```

pep-0622.rst
```diff
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
...
+
+After some experience with this feature, the community may have a better
+knowledge of what other pattern matching constructs could be valuable in Python.
```
gvanrossum, line 1780:
```suggestion
feeling for what other uses of pattern matching could be valuable in Python.
```

pep-0622.rst
```diff
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
 with pattern matching, since although it may be handy, it will likely be used
 rarely.
 
+Other pattern-based constructions
+---------------------------------
+
+Many other languages supporting pattern-matching use it as a base for multiple
```
gvanrossum, line 1766:
```suggestion
Many other languages supporting pattern-matching use it as a basis for multiple
```

pep-0622.rst
```diff
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
...
+language constructs, including a matching operator, a generalized form 
+of assignment, a filter for loops, a method for synchronizing communication,
+or specialized if statements. Some of these were mentioned in the discussion
+of the first draft. Another question asked is why this particular form (joining
+binding and conditional selection) was chosen while other forms were not.
+
+Introducing more uses of patterns would be too bold and premature for the
```
gvanrossum, line 1773:
```suggestion
Introducing more uses of patterns would be too bold and premature given the
```

pep-0622.rst
```diff
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
...
+experience we have using patterns and would make this proposal too
```
gvanrossum, line 1774:
```suggestion
experience we have using patterns, and would make this proposal too
```

pep-0622.rst
```diff
@@ -1760,6 +1760,24 @@ an expression. Nor do we propose a ``while match`` construct present in some lan
...
 rarely.
 
+Other pattern-based constructions
+---------------------------------
+
+Many other languages supporting pattern-matching use it as a basis for multiple
+language constructs, including a matching operator, a generalized form 
+of assignment, a filter for loops, a method for synchronizing communication,
+or specialized if statements. Some of these were mentioned in the discussion
+of the first draft. Another question asked is why this particular form (joining
```
gvanrossum, line 1770:
```suggestion
of the first draft. Another question asked was why this particular form (joining
```

PR #1503, opened 2020-07-07
PEP622: Be more explicit about side-effects

dmoisset:
This addresses some concerns about possible methods being called by the `match` statement.

### DIFF
--- a/pep-0622.rst
+++ b/pep-0622.rst
@@ -796,8 +796,8 @@ Only match-by-name will work by default, and classes should define
 match-by-position.  Additionally, dataclasses and named tuples will
 support match-by-position out of the box. See below for more details.
 
-Exception semantics
--------------------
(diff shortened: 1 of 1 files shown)


pep-0622.rst
```diff
@@ -807,6 +807,15 @@ an ``AttributeError`` raised while trying to lookup an attribute while matching
 attributes of a Class Pattern; that case results in just a matching failure,
 and the rest of the statement proceeds normally.
 
+The only side-effect carried on explicitly by the matching process is the binding of
+names. However, the process relies on attribute access,
+instance checks, `len()`, equality and item access on the subject and some of
```
hugovk, line 812: Double backticks for RST (single for MD):
```suggestion
instance checks, ``len()``, equality and item access on the subject and some of
```

PR #1507, opened 2020-07-08
PEP 625: Delegate sdist format specification to PEP 517

uranusjr:
Following [the discussion on possible sdist formats](https://discuss.python.org/t/pep-625-file-name-of-a-source-distribution/4686/6), avoid adding any specifics on sdist formats at all in the PEP.


uranusjr:
Closing since we’ll need to decide which archive format to associate with the extension. PEP 517 as it stands wouldn’t work if we go with zip, so I’ll close this and send a new patch when there is a conclusion.

PR #1519, opened 2020-07-10
PEP 596: 3.9.0b4 was published last Friday

gvanrossum:
@ambv, is 3.9.0b5 still expected July 20, or is that pushed back now? How about the rest of the releases?


ambv:
Thanks for this. Rest of the dates stays as is unless strictly necessary. External distributions rely on us here.

gvanrossum:
Thanks! Because of the PEG parser I permanently have a tab open on this PEP... :-)

PR #1521, opened 2020-07-10
PEP 612: Polishing up

mrkmndz:
* Made indentation consistent in code snippets
* removed extraneous backslash in grammar
* tiny bit more detail in a confusing snippet

### DIFF
--- a/pep-0612.rst
+++ b/pep-0612.rst
@@ -148,8 +148,11 @@ type this more complex decorator.
 Specification
 -------------
 
-ParamSpec Declarations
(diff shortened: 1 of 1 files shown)


mrkmndz:
@gvanrossum I previously had a couple examples already on two spaces, as otherwise we have to do a lot of unsightly line breaking :/ 
@pradeep90 pushed back on doing so inconsistently, leading to this change.

@gvanrossum , between being consistent 2, inconsistent, or consistent 4 with lots of function breaking, what do you think is preferrable?

gvanrossum:
Okay, makes sense.

PR #1528, opened 2020-07-16
PEP 629: Add versioning to PyPI's Simple API

dstufft:

### DIFF
--- a/pep-0629.rst
+++ b/pep-0629.rst
@@ -1,0 +1,133 @@
+PEP: 629
+Title: Versioning PyPI's Simple API
+Author: Donald Stufft <donald@stufft.io>
+Discussions-To: https://discuss.python.org/
+Status: Draft
+Type: Standards Track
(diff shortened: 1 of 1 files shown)


brettcannon:
@gvanrossum not really. The clients will get upgraded to support new protocols a lot faster than the servers. Plus the Simple API is specifically designed to work plainly over HTTP and a directory served by `-m http.server`, so adding that kind of server-level smarts pushes the API out past design goal/constraint of "server serves up plain HTML over HTTP, client parses that for info".

gvanrossum:
Got it, thanks for the explanation.

dstufft:
Yea, I'm getting ready to post a discuss thread (just waiting on it to render) but Brett hit the nail on the head.

This is largely just to provide more information to the clients so that they can make more intelligent decisions rather than either relying on heuristics or falling back on broad assumptions.

PR #1538, opened 2020-07-27
PEP 604: I've been asked to take over as PEP-Delegate

gvanrossum:
Here I'm also retargeting this to Python 3.10 and proposing various
small changes to clarify the PEP. In addition, we have a new reference
implementation in the works.


gvanrossum:
Let's wait for the PEP author, @pprados to review this!

PR #1540, opened 2020-07-29
PEP 604: Remove a misleading sentence in PROS

methane:
`The implementation can use the tuple present in ``Union`` parameter, without create a new instance` doesn't seem real PROS, because `int | float | str` can not be faster than `(int, float, str)`.
It just only mitigate CONS (worse performance).


gvanrossum:
We took care of this in a different PR.

PR #1544, opened 2020-07-29
PEP 612: Updating PEP metadata

mrkmndz:

### DIFF
--- a/pep-0612.rst
+++ b/pep-0612.rst
@@ -8,7 +8,7 @@ Type: Standards Track
 Content-Type: text/x-rst
 Created: 18-Dec-2019
 Python-Version: 3.10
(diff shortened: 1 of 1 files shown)


pep-0612.rst
```diff
@@ -7,8 +7,8 @@ Status: Draft
 Type: Standards Track
 Content-Type: text/x-rst
 Created: 18-Dec-2019
-Python-Version: 3.9
-Post-History: 18-Dec-2019
+Python-Version: 3.10
+Post-History: 13-Jul-2020
```
brettcannon, line 11: If this has been posted twice, then you should list all dates, not just the latest one.

```suggestion
Post-History: 18-Dec-2019, 13-Jul-2020
```

PR #1550, opened 2020-08-04
PEP 604 - Updates and Edits

MaggieMoss:
- Added additional examples and details to the specification section.

cc: @gvanrossum @pprados

### DIFF
--- a/pep-0604.rst
+++ b/pep-0604.rst
@@ -111,6 +111,8 @@ If a metaclass implements the ``__or__`` operator, it will override this::
     typing.Union[__main__.C, int]
 
 
+
+
 Objections and responses
 ========================
 
@@ -124,7 +126,7 @@ For more details about discussions, see links below:
 
 PROS:
 
-- This syntax can be more readable, and is similary to others languages (Scala, ...)
+- This syntax can be more readable, and is similar to other languages (Scala, ...)
 - At runtime, ``int|str`` might return a simple object in 3.9, rather than everything that
   you'd need to grab from importing ``typing``
 
@@ -164,11 +166,11 @@ if we accept to not be compatible with the dynamic evaluation of annotations (``
 PROS:
 
 - If they were permitted, then instance checking could use an extremely clean-looking notation
-- The implementation can use the tuple present in ``Union`` parameter, without create a new instance
+- The implementation can use the tuple present in ``Union`` parameter, without creating a new instance
 
 CONS:
 
-- Must migrate all the ``typing`` module in ``builtin``
+- Must migrate all of the ``typing`` module in ``builtin``
 
 
 Reference Implementation
--- a/pep-0604.rst
+++ b/pep-0604.rst
@@ -58,7 +58,7 @@ We will also be able to write ``t | None`` or ``None | t`` instead of
     isinstance(42, None | int)
 
 Specification
-========
+=============
 
 The new union syntax should be accepted for function, variable and parameter annotations.
 
@@ -88,42 +88,36 @@ Simplified Syntax
 
 The new syntax should be accepted for calls to `isinstance` as long as the Union items are valid arguments to `isinstance` themselves.
 
-::isinstance
-::
-    # valid
-    isinstance("", int | str)
+isinstance::
+  # valid
+  isinstance("", int | str)
 
 
-    # invalid
-    isinstance(2, list[int]) # TypeError: isinstance() argument 2 cannot be a parameterized generic
-    isinstance(1, int | list[int])
+  # invalid
+  isinstance(2, list[int]) # TypeError: isinstance() argument 2 cannot be a parameterized generic
+  isinstance(1, int | list[int])
 
-::subclass
-::
-    # valid
-    assert issubclass(bool, int | float)
+subclass::
+  # valid
+  assert issubclass(bool, int | float)
 
-    # invalid
+  # invalid
 
-The existing `typing.Union` and `|` syntax should be equal.
-::
+The existing `typing.Union` and `|` syntax should be equal.::
   int | str == typing.Union[int, str]
 
   typing.Union[int, int] == int
   int | int == int
 
-The order of the items in the Union should not matter for equality.
-::
+The order of the items in the Union should not matter for equality.::
   (int | str) == (str | int)
   (int | str | float) == typing.Union[str, float, int]
 
 
-Optional values should be equal to the new union syntax, for example:
-::
+Optional values should be equal to the new union syntax, for example: ::
   None | t == typing.Optional[t]
 
-A new Union.__str__() method should be implemented.
-::
+A new Union.__str__() method should be implemented.::
   (int | list[str]).__str__()
   # int | list[str]
 
--- a/pep-0604.rst
+++ b/pep-0604.rst
@@ -73,9 +73,6 @@ Simplified Syntax
 
     f([1, "abc"], None)
 
-    assert str | int == Union[str,int]
-    assert str | int | float == Union[str, int, float]
-
     # Instead of typing.List[typing.Union[str, int]]
     typing.List[str | int]
 
@@ -116,10 +113,10 @@ A new Union.__repr__() method should be implemented.
   str(int | int)
   # int
 
-isinstance
-----------
+isinstance and issubclass
+-------------------------
 
-The new syntax should be accepted for calls to ``isinstance`` as long as the Union items are valid arguments to ``isinstance`` themselves.
+The new syntax should be accepted for calls to ``isinstance`` and ``issubclass`` as long as the Union items are valid arguments to ``isinstance`` and ``issubclass`` themselves.
 
 ::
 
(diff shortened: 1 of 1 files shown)


pep-0604.rst
```diff
@@ -8,6 +8,7 @@ Type: Standards Track
 Content-Type: text/x-rst
 Created: 28-Aug-2019
 Python-Version: 3.10
+Discussions-To: typing-sig@python.org
```
hugovk, line 11: The CI failed because the headers are out of order: `Discussions-To` should follow `BDFL-Delegate`:

https://github.com/python/peps/blob/82a3d1f3dfd3ee293d722131e4bb7d4de4d551e2/pep-0001.txt#L512-L532

MaggieMoss: Thank you!

pep-0604.rst
```diff
@@ -73,13 +74,59 @@ Here are some examples of what we can do with this feature.
     assert str | int == Union[str,int]
     assert str | int | float == Union[str, int, float]
 
-    assert isinstance("", int | str)
-    assert issubclass(bool, int | float)
+    # Instead of
+    # typing.List[typing.Union[str, int]]
+    typing.List[str | int]
+
+    # list[typing.Union[str, int]] (Currently TypeError: 'type' object has no attribute '__getitem__')
```
gvanrossum, line 81: I don't think that error is relevant. PEP 585 makes it work starting in 3.9. You might merge this with the previous example like you did for dict/Dict below.

pep-0604.rst
```diff
@@ -73,13 +74,59 @@ Here are some examples of what we can do with this feature.
...
+    list[int | str]
+
+    # Instead of typing.Dict[str, typing.Union[int, float]]
+    typing.Dict[str, int | float]
+    dict[str, int | float]
+
+The new syntax should be accepted for calls to `isinstance` as long as the Union items are valid arguments to `isinstance` themselves.
```
gvanrossum, line 88: Use double backticks (also in various places below).

```suggestion
The new syntax should be accepted for calls to ``isinstance`` as long as the Union items are valid arguments to ``isinstance`` themselves.
```

pep-0604.rst
```diff
@@ -73,13 +74,59 @@ Here are some examples of what we can do with this feature.
...
+
+isinstance::
```
gvanrossum, line 90: You need a blank line hereafter for the markup to work. (Also below.)

pep-0604.rst
```diff
@@ -73,13 +74,59 @@ Here are some examples of what we can do with this feature.
...
+  isinstance(2, list[int]) # TypeError: isinstance() argument 2 cannot be a parameterized generic
+  isinstance(1, int | list[int])
+
+subclass::
+  # valid
+  assert issubclass(bool, int | float)
+
+  # invalid
+
+The existing `typing.Union` and `|` syntax should be equal.::
```
gvanrossum, line 105: equal -> equivalent? Also drop the period before the colon (again below).

pep-0604.rst
```diff
@@ -73,13 +74,59 @@ Here are some examples of what we can do with this feature.
...
+The order of the items in the Union should not matter for equality.::
+  (int | str) == (str | int)
+  (int | str | float) == typing.Union[str, float, int]
+
+
+Optional values should be equal to the new union syntax, for example: ::
+  None | t == typing.Optional[t]
+
+A new Union.__str__() method should be implemented.::
+  (int | list[str]).__str__()
```
gvanrossum, line 120: I would spell this using either `str(...)` or `print(...)`. But actually, it should be a new `__repr__` -- by defining just `__repr__` both `str()` and `repr()` will produce the same output. (And the REPL uses `repr()`.)

MaggieMoss:
@gvanrossum Thank you so much for the feedback, I really appreciate it. 

I've implemented your suggestions. Happy to continue iterating :) 


> Also, let's get approval from @pprados to add your name to the PEP authors.

Apologies, I thought you were cc'd on the email where he mentioned he was okay with this. I removed my name and I'll let @pprados confirm here.

pep-0604.rst
```diff
@@ -74,61 +76,72 @@ Simplified Syntax
...
+    # Instead of typing.List[typing.Union[str, int]]
     typing.List[str | int]
 
-    # list[typing.Union[str, int]] (Currently TypeError: 'type' object has no attribute '__getitem__')
-    list[int | str]
-
     # Instead of typing.Dict[str, typing.Union[int, float]]
-    typing.Dict[str, int | float]
     dict[str, int | float]
+    list[str | int]
```
gvanrossum, line 84: Looks like this line belongs in the previous group, and you accidentally dropped `typing.Dict[str, int | float]` here.

pep-0604.rst
```diff
@@ -74,61 +76,72 @@ Simplified Syntax
     assert str | int == Union[str,int]
     assert str | int | float == Union[str, int, float]
```
gvanrossum, line 77: I think the call example and the asserts are unnecessary (what the asserts say is repeated better below).

pep-0604.rst
```diff
@@ -74,61 +76,72 @@ Simplified Syntax
...
+The new syntax should be accepted for calls to ``issubclass`` as long as the Union items are valid arguments to ``issubclass`` themselves.
+
+::
+
+  # valid
+  assert issubclass(bool, int | float)
+
+  # invalid
+  assert issubclass(bool, bool | list[int])
+
```
gvanrossum, line 144: There's a lot of duplication for isinstance and subclass -- can you combine the two sections?

pep-0604.rst
```diff
@@ -74,61 +76,72 @@ Simplified Syntax
...
 
 Once the Python language is extended, mypy [1]_ and other type checkers will
 need to be updated to accept this new syntax.
```
gvanrossum, line 147: This seems to belong in the Implementation section below. You could also combine that section with the Reference Implementation section further below.

MaggieMoss:
Thanks again for the review. :)

pep-0604.rst
```diff
@@ -130,30 +128,12 @@ The new syntax should be accepted for calls to ``isinstance`` as long as the Uni
...
-  assert issubclass(bool, int | float)
-
-  # invalid
-  assert issubclass(bool, bool | list[int])
-
-
-Once the Python language is extended, mypy [1]_ and other type checkers will
-need to be updated to accept this new syntax.
+    # valid
+    assert issubclass(bool, int | float)
```
gvanrossum, line 132:
```suggestion
    issubclass(bool, int | float)
```

pep-0604.rst
```diff
@@ -130,30 +128,12 @@ The new syntax should be accepted for calls to ``isinstance`` as long as the Uni
...
 
-Implementation
-==============
+    # invalid
+    assert issubclass(bool, bool | list[int])
```
gvanrossum, line 135:
```suggestion
    issubclass(bool, bool | list[int])
```

pep-0604.rst
```diff
@@ -123,7 +169,7 @@ For more details about discussions, see links below:
 
 PROS:
 
-- This syntax can be more readable, and is similary to others languages (Scala, ...)
+- This syntax can be more readable, and is similar to other languages (Scala, ...)
 - At runtime, ``int|str`` might return a simple object in 3.9, rather than everything that
```
gvanrossum, line 173:
```suggestion
- At runtime, ``int|str`` might return a simple object in 3.10, rather than everything that
```

pep-0604.rst
```diff
@@ -123,7 +169,7 @@ For more details about discussions, see links below:
```
gvanrossum, line 169: The headline just above here says `Union[type1|type2]`. I think that `|` is a mistake and the intention was to write `Union[type1, type2]`.

pep-0604.rst
```diff
@@ -163,16 +209,26 @@ if we accept to not be compatible with the dynamic evaluation of annotations (``
 PROS:
 
 - If they were permitted, then instance checking could use an extremely clean-looking notation
-- The implementation can use the tuple present in ``Union`` parameter, without create a new instance
+- The implementation can use the tuple present in ``Union`` parameter, without creating a new instance
```
gvanrossum, line 212: In #1540, @methane proposes to strike this sentence, as it is not a real example. (Probably `int|float|str` is actually slower than `(int, float, str)`, which maybe should be listed under CONS.)

pep-0604.rst
```diff
@@ -163,16 +209,26 @@ if we accept to not be compatible with the dynamic evaluation of annotations (``
...
 
 CONS:
 
-- Must migrate all the ``typing`` module in ``builtin``
+- Must migrate all of the ``typing`` module in ``builtin``
```
gvanrossum, line 216: This is not correct (as your implementation proves :-).

pep-0604.rst
```diff
@@ -118,13 +164,13 @@ For more details about discussions, see links below:
 - `Discussion in python-ideas <https://mail.python.org/archives/list/python-ideas@python.org/thread/FCTXGDT2NNKRJQ6CDEPWUXHVG2AAQZZY/>`_
 - `Discussion in typing-sig <https://mail.python.org/archives/list/typing-sig@python.org/thread/D5HCB4NT4S3WSK33WI26WZSFEXCEMNHN/>`_
 
-1. Add a new operator for ``Union[type1|type2]``?
+1. Add a new operator for ``Union[type1, type2]``?
 -------------------------------------------------
```
gvanrossum, line 168:
```suggestion
--------------------------------------------------
```

PR #1555, opened 2020-08-07
PEP 626:Guarantee stability of line table during release and define API for out-of-process tools.

markshannon:

### DIFF
--- a/pep-0626.rst
+++ b/pep-0626.rst
@@ -186,7 +186,11 @@ Out of process debuggers and profilers
 ''''''''''''''''''''''''''''''''''''''
 
 Out of process tools, such as py-spy [1]_, cannot use the C-API, and must parse the line number table themselves.
-To reduce the work required to implement these tools, the following C struct and functions are provided.
+Although the line number table format may change without warning,
+it will not change during a release unless absolutely necessary for a bug fix.
+
+To reduce the work required to implement these tools, the following C struct and utility functions are provided.
+Note that these functions are not part of the C-API, so will be need to be linked into any code that needs to use them.
 
 ::
 
--- a/pep-0626.rst
+++ b/pep-0626.rst
@@ -182,6 +182,53 @@ Access to the ``f_lineno`` attribute of frame objects through C API functions is
 
 Accessing ``f_lineno`` directly through the underlying data structure is forbidden.
 
+Out of process debuggers and profilers
+''''''''''''''''''''''''''''''''''''''
+
+Out of process tools, such as py-spy [1]_, cannot use the C-API, and must parse the line number table themselves.
+To reduce the work required to implement these tools, the following C struct and functions are provided.
+
+::
+
+    typedef struct addressrange {
+        int ar_start;
+        int ar_end;
+        int ar_line;
+        int opaque1;
+        void *opaque2;
+    } PyCodeAddressRange;
+
+    void PyLineTable_InitAddressRange(char *linetable, int firstlineno, PyCodeAddressRange *range);
+    int PyLineTable_NextAddressRange(PyCodeAddressRange *range);
+    int PyLineTable_PreviousAddressRange(PyCodeAddressRange *range);
+
+``PyLineTable_InitAddressRange`` initializes the ``PyCodeAddressRange`` struct from the line number table and first line number.
+
+``PyLineTable_NextAddressRange`` advances the range to the next entry, returning non-zero if valid.
+
+``PyLineTable_PreviousAddressRange`` retreats the range to the previous entry, returning non-zero if valid.
+
+For example, the following code prints out all the address ranges:
+
+::
+
+    void print_address_ranges(char *linetable, int firstlineno)
+    {
+        PyAddressRange range;
+        PyLineTable_InitAddressRange(linetable, firstlineno, &range);
+        while (PyLineTable_NextAddressRange(&range)) {
+            printf("Bytecodes from %d (inclusive) to %d (exclusive) ",
+                   range.start, range.end);
+            if (range.line < 0) {
+                /* line < 0 means no line number */
+                printf("have no line number\n");
+            }
+            else {
+                printf("have line number %d\n", range.line);
+            }
+        }
+    }
+
 
 Performance Implications
 ========================
@@ -193,13 +240,19 @@ Code with long sequences of ``pass`` statements will probably become a bit slowe
 Reference Implementation
 ========================
 
-Under development.
+https://github.com/markshannon/cpython/tree/new-linetable-format-version-2
 
 Copyright
 =========
 
 This document is placed in the public domain or under the
 CC0-1.0-Universal license, whichever is more permissive.
+
+References
+==========
+
+.. [1] py-spy: Sampling profiler for Python programs
+   (https://github.com/benfred/py-spy)
 
 
 
--- a/pep-0626.rst
+++ b/pep-0626.rst
@@ -218,7 +218,7 @@ For example, the following code prints out all the address ranges:
 
     void print_address_ranges(char *linetable, int firstlineno)
     {
-        PyAddressRange range;
+        PyCodeAddressRange range;
         PyLineTable_InitAddressRange(linetable, firstlineno, &range);
         while (PyLineTable_NextAddressRange(&range)) {
             printf("Bytecodes from %d (inclusive) to %d (exclusive) ",


pep-0626.rst
```diff
@@ -182,6 +182,53 @@ Access to the ``f_lineno`` attribute of frame objects through C API functions is
...
+
+``PyLineTable_NextAddressRange`` advances the range to the next entry, returning non-zero if valid.
+
+``PyLineTable_PreviousAddressRange`` retreats the range to the previous entry, returning non-zero if valid.
+
+For example, the following code prints out all the address ranges:
+
+::
+
+    void print_address_ranges(char *linetable, int firstlineno)
```
pablogsal, line 215: This won't work as `py-spy` cannot execute code in the process being traced

markshannon: That doesn't make sense to me. 
`py-spy` must execute code; it's software.

Or are you saying that `py-spy` cannot use `PyLineTable_InitAddressRange`, etc. for licensing reasons, or similar?

gvanrossum: Presumably py-spy is something like gdb attaching to a different process. It can easily read any data in the process, but it is not so easy to call a function. (I think gdb can, but I presume it requires borrowing a thread, and given that py-spy is a profiling tool I think it would rather not do so.)

pablogsal: Exactly: these tools attach to a process using `ptrace` and only read memory from it using the `ptrace` or `process_vm_readv` system calls. They will not execute code in the remote process due to security concerns (and also because is orders of magnitude more difficult than just reading memory). The setup is basically you have process A running Python and process B running the profiler. Process B can be implemented in any other language and will not be linked against `libpython.so`. Then process B attaches to process A and reads its memory (the location of the memory and the layout are obtaining using DWARF information or hardcoded offsets, there are many techniques). 

These tools can also analyze core files to get the python stack trace by reading the core file memory and in these core files, you cannot execute functions either (because is a core file not a living process).

markshannon: I believe you that these tools can't execute code in a different process.

But why can't they execute code in their own process, like all other programs? That's the bit I don't get.

pablogsal: > But why can't they execute code in their own process, like all other programs? That's the bit I don't get.

How they will execute `PyLineTable_NextAddressRange`? These programs do not link against any `libpython` and can be implemented in any language. You say:

> To reduce the work required to implement these tools, the following C struct and functions are provided.

but when you mention "are provided"...where is this provided? If this is provided as part of the C-API then it won't work because these programs do not link against libpython (and even if they do, they need to analyze multiple versions of python so they would need to link against multiple different libpythons simultaneously which is not possible). Or are you  meaning that these programs should implement analogous of those functions on their own?

benfred: Yeah, I can’t easily call python c-api functions from within py-sy =(.  Since py-spy copies memory from the profiled python process, that means any pointers we read won’t point to valid memory in the py-spy process - so passing these copied structs to libpython would just segfault right away.  Py-spy also supports multiple versions of python from a single py-spy binary which would make linking harder.

I’d rather not have py-spy run code in the profiled process. Its [technically possible](https://eklitzke.org/ptrace-continued), but modifying the target process like this is dangerous, and the steps are different on each OS/ CPU architecture. 

Fwiw, I’m perfectly happy to implement new code with py-spy to handle the new line format - I don’t want py-spy compatibility to slow down core python development =). However that will involve parsing the new co_linetable attribute directly - and I saw here ```the format is opaque, unspecified and may be changed without notice``` which sounds like this is discouraged.

markshannon: Please stop with all the stuff about running code in another process. It's irrelevant.
I know that an out of process debugger can't (easily) use the API of the running Python process, it says so in this PR.
https://github.com/python/peps/pull/1555/files#diff-e5a7e9acca6a66fb142d5a627bd16cabR188

You can run `PyLineTable_NextAddressRange` is any process you like. Run it in the debugger process; run it in the Python process; run it in a webapp on a Raspberry Pi, if you want. It doesn't matter.
`PyLineTable_NextAddressRange` is portable C code, you can run it anywhere you like.

Regarding the format. Yes, it might be changed without notice, which is why the `PyLineTable` functions are provided. 
I will clarify that the format won't change during a release, unless absolutely necessary for a bug fix.

So, just use the `PyLineTable` functions and you don't need to worry about the internal format. Copy the data across, as is already done for the `co_lnotab` data, and use the provided functions.

`py-spy`, and AFAICT other profilers and debuggers, need additional code for each release, so linking in the latest version of the `PyLineTable` functions shouldn't be a problem. Just make your own dll/so for each version and load that once you know which version of Python you're attached to.

benfred: > So, just use the PyLineTable functions and you don't need to worry about the internal format. Copy the data across, as is already done for the co_lnotab data, and use the provided functions.

Thanks for the explanation! If I'm understanding you, I think what you're proposing could work for py-spy - but I have some questions:

1) Can I treat co_linetable like a PyVarObject to figure out the number of bytes to copy? The text in this PEP made it sound like co_linetable was an opaque pointer and wasn't clear that I could  get the size of the struct. I don't need the exact format, but do need to know how many bytes to copy over.

2) Will the PyLineTable functions access any pointer values inside of co_linetable? If they are only accessing data stored inside co_linetable, and don't follow any pointers from that struct we should be good. Otherwise we need to add a callback to do the memory access to the PyLineTable API's (any pointers will point to the memory space of the profiled program - so accessing in py-spy will segfault).

3) Will the PyLineTable functions have any other dependencies on other code in the python runtime? If I can just include the C source for only these functions and then bundle with py-spy that would work pretty well. I can't use dlopen to load a dll/so built per python version like you suggested (py-spy statically links to musl libc on linux), but can statically link in these functions by mangling the names or wrapping in a c++ namespace easily enough.

Btw, thanks for thinking about py-spy when designing this - I really appreciate it .

pep-0626.rst
```diff
@@ -186,7 +186,11 @@ Out of process debuggers and profilers
 ''''''''''''''''''''''''''''''''''''''
 
 Out of process tools, such as py-spy [1]_, cannot use the C-API, and must parse the line number table themselves.
-To reduce the work required to implement these tools, the following C struct and functions are provided.
+Although the line number table format may change without warning,
+it will not change during a release unless absolutely necessary for a bug fix.
+
+To reduce the work required to implement these tools, the following C struct and utility functions are provided.
+Note that these functions are not part of the C-API, so will be need to be linked into any code that needs to use them.
```
pablogsal, line 193: I think this line clarifies the proposal much better 👍

pep-0626.rst
pablogsal: > 3\. Will the PyLineTable functions have any other dependencies on other code in the python runtime? If I can just include the C source for only these functions and then bundle with py-spy that would work pretty well.

I think this is the best approach, like many other tools other than `py-spy` may have further constrains. For instance, if one of these tools is already linked against a given Python runtime (I maintain such a tool myself that for instance is implemented itself in Python as a C extension) it won't be able to link against any other runtimes so including the source and manging it will be verbatim. Given the new clarification:

> Note that these functions are not part of the C-API, so will be need to be linked into any code that needs to use them.

I assume this means that these functions won't need to use the C-API so IIUC vendoring the source code of these functions and operating over a copied blob of data will be enough.

pablogsal: > 1. Can I treat co_linetable like a PyVarObject to figure out the number of bytes to copy? The text in this PEP made it sound like co_linetable was an opaque pointer and wasn't clear that I could  get the size of the struct. I don't need the exact format, but do need to know how many bytes to copy over.

I assume `co_linetable` must be either a Python object (in which case you can get the size as if it were a `PyVarObject `), a structure (in which case you can use DWARF info, a hardcoded offset or `offsetof` over a vendored copy of the struct or an opaque pointer + a size parameter somewhere.

pep-0626.rst
```diff
@@ -182,6 +182,57 @@ Access to the ``f_lineno`` attribute of frame objects through C API functions is
...
+
+::
+
+    typedef struct addressrange {
+        int ar_start;
+        int ar_end;
+        int ar_line;
+        int opaque1;
+        void *opaque2;
```
pablogsal, line 202: This only works if `PyLineTable_InitAddressRange` and `PyLineTable_InitAddressRange` does not itself access the `opaque2` as this pointer will be in the remote process, not in the profiler one.

markshannon: Now you're just making stuff up, to object to.
Why would you assume that a pointer would point to invalid memory?

pep-0626.rst
markshannon: `co_linetable` is a Python object, for now.
The layout of code objects is not part of the C-API, so might well change in future versions.

markshannon: > 1. Can I treat co_linetable like a PyVarObject to figure out the number of bytes to copy?

Yes, for now. It may change in future versions.

> 2. Will the PyLineTable functions access any pointer values inside of co_linetable?

No
> 3. Will the PyLineTable functions have any other dependencies on other code in the python runtime?

No.

markshannon: Note that `PyLineTable_InitAddressRange` expects a `char *. It knows nothing about object layout.

pep-0626.rst
markshannon: Unless, of course, you pass an invalid pointer to `PyLineTable_InitAddressRange`. Not much I can do about that.

pep-0626.rst
markshannon: Just to clarify, the `PyLineTable functions` will only access memory inside the array of bytes passed to `PyLineTable_InitAddressRange`, but nothing outside of that array. It knows nothing of the original code object.

pep-0626.rst
pablogsal: I am not objecting, I do like the proposal, I am just trying to make sure this mechanism will work as intended.

> Now you're just making stuff up, to object to.
> Why would you assume that a pointer would point to invalid memory?

If you copy the structure from process A to process B, then the opaque pointer in the structure will contain a memory address that refers to process A. Trying to dereference that pointer in process B will segfault because that address is invalid in process B. In order to make it work you need to copy the structure pointed by that pointer from the memory of process A. 

This means, when executing the function in process B you would need to substitute `opaque_pointer->some_attr` with `copy_memory(opaque_ptr, num_bytes, &dest); dest->some_attr`.

pablogsal: To make this more clear, let's say you have process A and process B. The structure in process A (the python process) looks like this (not the ex:

{ 1, 2, 3, 0xa12345678 }

The memory address that corresponds to the opaque pointer, 0xa12345678 is valid and works fine in process A because that is where is intended to be used. Dereferencing this pointer using `opaque->attr` will work.

Now you copy this struct to process B. In process B the memory address 0xa12345678 is invalid because it points to something in process A. Trying to treat this as a pointer and doing `opaque->attr` will segfault. What process B needs to do is copy from process A whatever is pointed by the address 0xa12345678 and then accessing the field `attr` from the copied memory.

With this in mind, what I am trying to explain of that if these functions that we are proposing cannot dereference pointers if we intend people to use them 'as it is' in their profilers/debuggers.

markshannon: All `PyCodeAddressRange` structs would be in the debugger.
Why would it be in the Python process? That makes no sense.

pablogsal: > All `PyCodeAddressRange` structs would be in the debugger.
> Why would it be in the Python process? That makes no sense.

Because that's where the debugger copied them from, no?

pablogsal: When the debugger copies the structs from the Python process, they live in the debugger, yes, but the values of the pointers that were copied still point to memory on the Python process.

markshannon: The line number table, doesn't contain any pointers, just bytes representing offsets. There are no pointers.
The `PyCodeAddressRange` is just a structure for iterating over the line number table. There is no concept of local or remote here, just plain old C code.

pablogsal: > The `PyCodeAddressRange` is just a structure for iterating over the line number table. There is no concept of local or remote here, just plain old C code.

Ohhhhhhh! So you are saying that the `PyLineTable_InitAddressRange` and `PyLineTable_NextAddressRange` won't access the `opaque2` pointer in the `PyCodeAddressRange` struct?

markshannon: No. `PyLineTable_InitAddressRange` initializes it, and `PyLineTable_NextAddressRange` uses. It's marked opaque, so that user code shouldn't modify it, or assume that it's value is meaningful.

I'm genuinely perplexed why of all the many millions of pointers in C code, you think this one will be so problematic?

If you pass invalid values to `PyLineTable_InitAddressRange`, then either it or `PyLineTable_NextAddressRange` will crash or return incorrect results. Pass valid data and it will work (assuming it is bug free).  Just like all C code.

pablogsal: > I'm genuinely perplexed why of all the many millions of pointers in C code, you think this one will be so problematic?

I think the confusion here is that I was assuming that the `PyCodeAddressRange` struct was something that the debugger will need to copy from the Python process and therefore the pointers would be pointing to memory in the remote, but (correct me if I am wrong) it seems that this is something that lives entirely on the debugger and is only used for iteration, no?

markshannon: Yes. In the example, `PyCodeAddressRange range` is allocated on the debugger's stack.

pablogsal: Fantastic, all is good now! Thanks for the patience, @markshannon :)

markshannon: It's rare that I'm called patient :smile: 
Thanks for the thorough review.

pep-0626.rst
benfred: awesome - that sounds good thanks!

PR #1560, opened 2020-08-12
Fix incorrect backticks

hugovk:
Like https://github.com/python/peps/pull/1559, for:

* PEP 445
* PEP 467
* PEP 548
* PEP 621

Found with:

```console
$ git grep "[A-Za-z0-9]\`\`[A-Za-z0-9]"
pep-0445.txt:``PyMem_Realloc()`` indirectly call``PyObject_Malloc()`` and
pep-0467.txt:This PEP proposes that ``bytes`` and ``bytearray``gain an optimised
pep-0548.rst:Reading this we first see the``break``, which obviously applies to
pep-0621.rst:for using``long_description`` and a corresponding
```


hugovk:
Yes, I was about to add one to avoid the single backticks, and it's based on grep, so I'll see about adding a rule for this too!

PR #1564, opened 2020-08-15
PEP 622: More details on 'marker prefix for lookups' rejected idea

ncoghlan:
The "Rejected Ideas" section didn't clearly cover the perceived
benefits of using a marker prefix on value lookup match patterns.

This expands on that section, while still making it clear that
the listed PEP authors don't agree that the proposal would actually
deliver the benefits that its proponents believe it would provide.


gvanrossum:
I'm sorry, I don't want the PEP to be a soapbox for this misguided idea. I'll adopt a few of the clarifications you are proposing at the beginning of the PR, but I don't feel your idea deserves this much airtime in the PEP.

PR #1569, opened 2020-08-15
Add sphinx support - pydotorg compat (ref #1385)

AA-Turner:
See #2, #1385 for context.

This is the core Sphinx support part, comprising 
- core
- package.py

Note - this is a temporary compat layer to check that both pydotorg and Sphinx/GH-pages can be driven from the same back-end, this isn't intended to be used as an end product
Note: this is unlikely to successfully build, I am splitting into logical chunks as per [Victor's comment](https://github.com/python/peps/pull/1385#pullrequestreview-432196609).


AA-Turner:
Closing (xref #1385 reason for closing)

PR #1580, opened 2020-09-03
PEP 627: Mark as Accepted

encukou:
Paul Moore accepted this PEP on [Discourse](https://discuss.python.org/t/pep-627/4126/42).

This should ideally be merged immediately after [the PyPA spec PR](https://github.com/pypa/packaging.python.org/pull/756).


pfmoore:
Once CI has completed here, I'll commit both the spec PR and this one.

PR #1585, opened 2020-09-07
PEP 597: Add io.LOCALE_ENCODING constant

methane:
See discussion in https://discuss.python.org/t/3880/16 .

### DIFF
--- a/pep-0597.rst
+++ b/pep-0597.rst
@@ -21,6 +21,9 @@ This PEP proposes:
 * Add ``encoding="locale"`` option to ``TextIOWrapper``. It behaves
   like ``encoding=None`` but don't raise a warning.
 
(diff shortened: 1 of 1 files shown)


pep-0597.rst
```diff
@@ -21,6 +21,9 @@ This PEP proposes:
 * Add ``encoding="locale"`` option to ``TextIOWrapper``. It behaves
   like ``encoding=None`` but don't raise a warning.
 
+* Add ``io.LOCALE_ENCODING = "locale"`` constant to avoid confusing
+  ``LookupError``. unknown encoding: locale``.
```
tk0miya, line 25: It seems backticks are inserted to the tail unexpectedly.

methane: Thank you. I fixed it in #1589.

PR #1587, opened 2020-09-07
Fix backwards compatible update logic

ofek:


pep-0631.rst
```diff
@@ -58,7 +58,9 @@ Every element SHOULD be an `entry <#entries>`_.
     ]
 
 Any element MAY be an inline table to support possible future functionality.
-Backends MUST silently ignore these until the specification is extended.
+Backends MUST error when encountering these until the specification is extended.
+Other parsers SHOULD do the same, though they MAY consider the entire enclosing
```
pganssle, line 62: This is still not right. The ideal situation is that every parser would mark the relevant field as dynamic (which basically means "ask the backend" — the backend, in turn, will raise an error unless it's coded against a newer version of the spec).

Saying that parsers SHOULD raise an exception defeats the entire purpose of the backwards compatibility component, so this should read:

> Backends MUST error when encountering these until the specification is extended; other parsers SHOULD consider the entire enclosing field dynamic.

Possibly a version that makes the rationale a bit clearer would be:

> Backends MUST error when encountering these unless the specification is extended; other parsers — particularly those not intended to be listed in a project's `build-system.requires` — SHOULD consider the entire enclosing field dynamic.

pep-0631.rst
```diff
@@ -58,7 +58,9 @@ Every element SHOULD be an `entry <#entries>`_.
     ]
 
 Any element MAY be an inline table to support possible future functionality.
-Backends MUST silently ignore these until the specification is extended.
+Backends MUST error when encountering these until the specification is extended.
```
pganssle, line 61: Regarding the "Any element MAY be an inline table"... language, I fear this may be confusing to users who don't understand that the audience for this table is tools that are consumers of the `[project]` table, _not_ people writing the `[project]` table.

I'm not sure if there's better wording, though. Possibly:

> Future versions of this specification that allow for any element to be an inline table should be considered backwards compatible, and consumers of the `[project]` table should take this into consideration.

ofek:
Corrected https://github.com/python/peps/pull/1588

PR #1590, opened 2020-09-08
Proposal: Ignore all empty entries within PYTHONPATH

david-a-wheeler:
Currently, common ways to set ``$PYTHONPATH`` also unexpectedly add the
current directory as an entry.  As a result, users have an increased
risk of unintentionally running malicious code.

This PEP proposes that all empty entries within ``$PYTHONPATH`` be ignored,
instead of being treated as the current directory.

I proposed this to <python-list@python.org>
on Thu, 27 Aug 2020 19:33:13 -0400 (EDT), with subject
"Proposal: Ignore all empty entries within $PYTHONPATH".
I received no comments of any kind. So I thought I'd propose
it here as a PEP.

Implementing this is trivial. I have created this as a PEP because
this is a behavioral change in a long-present mechanism, so there
is a relevant concern about backwards compatibility.
That said, I believe those concerns are easily addressed.
See the proposed PEP for details.

My apologies if this is the wrong process, just let me know
what you'd like me to do instead. Thank you!

Signed-off-by: David A. Wheeler <dwheeler@dwheeler.com>

### DIFF
--- a/pep-9999.rst
+++ b/pep-9999.rst
@@ -1,0 +1,116 @@
+PEP: 9999
+Title: Ignore all empty entries within ``$PYTHONPATH``
+Author: David A. Wheeler <dwheeler@dwheeler.com>
+Status: Draft
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 27-Aug-2020
+Python-Version: 3.7.10
+Post-History: 27-Aug-2020
+
+
+Abstract
+========
+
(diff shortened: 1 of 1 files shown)


gvanrossum:
Sorry, that's not how the PEP process works. Can you please start a discussion on python-ideas? Also, please read PEP 1 -- you're going to need a sponsor. Good luck!

PR #1592, opened 2020-09-08
Optional Syntax PEP 

MaggieMoss:
Initial Pull Request for optional syntax PEP.

### DIFF
--- a/pep-9999.rst
+++ b/pep-9999.rst
@@ -1,0 +1,154 @@
+PEP: 12
+Title: Allow writing optional types as ``x?``
+Author: Maggie Moss <maggiebmoss@gmail.com>
+Status: Draft
+Type: Process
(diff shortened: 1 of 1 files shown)


Rosuav:
Has this been discussed on any of the relevant mailing lists? A PEP is a late step in the process, not the beginning of an idea. Start by discussing the idea on python-ideas or typing-sig, and if you have general consensus that this needs to be written up as a PEP, you should also be able to find a sponsor.

MaggieMoss:
It's been discussed

- https://mail.python.org/archives/list/typing-sig@python.org/thread/SWAY6V7WZLVPGYQEMHXJROT2747OXSRX/
- https://github.com/python/typing/issues/429

And in person at typing meetups I've attended. I can close the PR until I've found a sponsor.

PR #1610, opened 2020-09-21
Travis CI: Only deploy once

hugovk:
Right now, the deploy to S3 is made for both the `3.7` and `3.7-dev` jobs, for example:

* https://travis-ci.com/github/python/peps/jobs/388374276#L737
* https://travis-ci.com/github/python/peps/jobs/388374277#L741

Instead, only deploy for a single CI job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,4 @@ deploy:
     on:
         branch: master
         repo: python/peps
+        python: "3.7"


hugovk:
🚀  PR no longer needed now GitHub Actions is used instead of Travis CI: https://github.com/python/peps/pull/1669.

PR #1617, opened 2020-09-24
PEP 637: Fix rST errors caused by incorrect indentation

eric-wieser:
Without this:

* blocks of prose render as code
* blocks of code randomly render inside blockquotes


pep-0637.txt
```diff
@@ -328,13 +328,13 @@ The following old semantics are preserved:
...
 
-    ::
+   ::
 
       obj[]  # SyntaxError
 
 2. A single index value remains a single index value when passed:
 
-    ::
+   ::
```
eric-wieser, line 337: If not aligned with the previous line of text, `::` becomes a blockquote

PR #1621, opened 2020-09-26
PEP 0603 - Fix figure 2

adamchainz:
Images are uploaded into per-PEP directories so using those from another PEP doesn't work (any more?). This PR duplicates the used image from PEP 0550.


gvanrossum (approved):

gvanrossum:
Thanks for noticing and fixing! I have no idea either whether this is new or was always broken, but it will be fixed now.

PR #1620, opened 2020-09-25
PEP 637: Integrate feedback from GvR

stefanoborini:
Integrates feedback from GvR. 

Note: this PR is built on top of #1619

PR #1622, opened 2020-09-26
PEP 637 allow for unpacking of starargs

stefanoborini:
As from @eric-wieser idea on python-ideas, *args are now unpacked.


stefanoborini:
PR built on top of #1620 . Need merge of that one first.

pep-0637.rst
```diff
@@ -347,17 +347,12 @@ The successful implementation of this PEP will result in the following behavior:
    - but if no ``**kwargs`` parameter is defined, it is an error.
 
 
-7. Sequence unpacking remains a syntax error inside subscripts::
+7. Sequence unpacking is allowed inside subscripts::
```
eric-wieser, line 350: Note that this requires some specification for what `obj[*()]` should do, which likely has similar semantics to `obj[]`. One option would be an error, another would be to handle it like `obj[kwargs=...]` is handled.

eric-wieser: Also would probably want to include `obj[*(x,)] == obj[x]` and `obj[*(x,),] == obj[x,]` as corner cases.

gvanrossum:
@stefanoborini Are you going to respond to @eric-wieser's latest comments or should I just merge like this?

stefanoborini:
@gvanrossum no he has a good point. I'll add those cases (and solve conflicts)

pep-0637.rst
```diff
@@ -354,12 +354,34 @@ The successful implementation of this PEP will result in the following behavior:
    This allows notations such as ``[:, *args, :]``, which could be treated 
    as ``[(slice(None), *args, slice(None))]``.
 
+   The following notation equivalence should be honored::
+
+       obj[*()]
```
gvanrossum, line 359: Did you discuss this on python-ideas? I actually think the reason for rejecting `obj[]` (that it looks like a typo) doesn't apply here. Ditto for `obj[**{}]`.

eric-wieser: I'm in favor of allowing that too - my point above was mainly that _something_ needed to be specified, and then python-ideas could argue for a better specification.

gvanrossum (commented):
Okay, LGTM. I'll land when tests pass.

gvanrossum:
That test is just hanging. I'm going to land.

pep-0637.rst
stefanoborini: I was not sure about it... maybe we should let it pass, and treat it like obj[()], same for obj[**{}] is valid and equivalent to obj[()]

PR #1626, opened 2020-09-27
PEP 637 Default sentinel value for no value passed as index

stefanoborini:
Added discussion on the sentinel value when nothing is passed as an index.

### DIFF
--- a/pep-0637.rst
+++ b/pep-0637.rst
@@ -833,11 +833,11 @@ This option sounds appealing because:
        >>> foo(k=3)
        () {'k': 3}
 
-   Although we do accept the following asymmetry in behavior when a single
-   value is passed, but that ship has sailed::
+   Although we do accept the following asymmetry in behavior compared to functions 
+   when a single value is passed, but that ship has sailed::
 
-       >>> foo(1, k=3)
-       (1,) {'k': 3}   # for indexes, a plain 1 is passed
+       >>> foo(5, k=3)
+       (5,) {'k': 3}   # for indexing, a plain 5, not a 1-tuple is passed
 
 For option 2, using ``None``, it was objected that NumPy uses it to indicate
 inserting a new axis/dimensions (there's a ``np.newaxis`` alias as well)::
--- a/pep-0637.rst
+++ b/pep-0637.rst
@@ -801,7 +801,7 @@ A proposed hack would be to let the user specify which entity to use when an
 index is not specified, by specifying a default for the ``index``, but this
 forces necessarily to also specify a (never going to be used, as a value is
 always passed by design) default for the ``value``, as we can't have
-non-default arguments after defaulted one.::
+non-default arguments after defaulted one::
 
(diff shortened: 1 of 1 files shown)


pep-0637.rst
```diff
@@ -772,44 +775,125 @@ make the dictionary type accept it automatically, to insert or refer to the valu
...
+
+    obj[k=3] = 5  # would call type(obj).__setitem__(???, 5, k=3)
+
+A proposed hack would be to let the user specify which entity to use when an
+index is not specified, by specifying a default for the ``index``, but this
+forces necessarily to also specify a (never going to be used, as a value is
+always passed by design) default for the ``value``, as we can't have
+non-default arguments after defaulted one.::
+
+    def __setitem__(index=SENTINEL, value=NEVERUSED, *, k)
```
gvanrossum, line 806: I think the stronger argument against this is that some classes may want to use `**kwargs` in the method definition, and they would end up with
```
def __setitem__(self, index, value, **kwargs): ...
```
However this would cause problems if the caller uses `index` or `value` as a keyword (or `self`!):
```
x[value=1] = 0
```
The intention is that this would end up calling
```
x.__setitem__(SENTINEL, 0, **{"value": 1})
```
but if you try such a call you'll get
```
TypeError: __setitem__() got multiple values for argument 'value'
```
The accepted solution is to use positional-only arguments (PEP 570), like this:
```
def __setitem__(self, index, value, /, **kwargs): ...
```
but now there is no way to pass something for `value` without also passing something for `index` -- we can provide defaults but that doesn't help:
```
def __setitem__(self, index=SENTINEL, value=..., /, **kwargs): ...
```
but if you call this with one positional argument (supposedly the value) that argument ends up as `index`.

This was discussed in the list (by you? by Stephen? I don't recall) but it's not properly represented by the text you're proposing here for the PEP.

stefanoborini: @gvanrossum I'll make it more explicit (tomorrow though, I am falling asleep). 

Using index or value as a keyword and obtaining a multiple values error is probably expected behavior. Whoever attempts this is intentionally trying to break the code. 

On the second part of your post, D'Aprano commented on range() being the odd example we currently have, where the end value is the first or the second depending on how it's called. I think it's the only case I can recall, and it makes sense from the outside, but I would not expect it to happen in setitem. The likelihood of incorrect implementation out there is too high. In range, it's one and done.

gvanrossum: > Using index or value as a keyword and obtaining a multiple values error is probably expected behavior. Whoever attempts this is intentionally trying to break the code.

Please read PEP 570 -- it is specifically written to address this anomaly. The user isn't supposed to know or care what the actual names of these arguments is (some other class may use `i` and `v`) and if an error was raised if you hit such a conflict, it would prevent the user from creating e.g. columns named `index` or `value` (or `self`).

Agreed that we shouldn't make classes that implement `__setitem__` have to deal with the horrors of the `range()` signature.

PR #1627, opened 2020-09-29
Record Pablo as the PEP delegate for PEP 626.

Yhg1s:
The Steering Council have asked Pablo to be the PEP delegate for PEP 626, and he's agreed.


pablogsal (approved):

markshannon:
Thanks @pablogsal

PR #1630, opened 2020-09-30
PEP 484: Clarify that 'import ... as ...' means 'X as X'

gvanrossum:
See
https://mail.python.org/archives/list/typing-sig@python.org/thread/YLJPWECBNPD2K4TRIBRIPISNUZJCRREY/
esp.
https://mail.python.org/archives/list/typing-sig@python.org/message/RDJECZOIFRLU2IQZWU276MZDWVJQZOAY/

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -1651,6 +1651,9 @@ Additional notes on stub files:
 * Modules and variables imported into the stub are not considered
   exported from the stub unless the import uses the ``import ... as
   ...`` form or the equivalent ``from ... import ... as ...`` form.
+  (*UPDATE:* To clarify, the intention here is that only names using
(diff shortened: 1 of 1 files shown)


pep-0484.txt
```diff
@@ -1651,6 +1651,9 @@ Additional notes on stub files:
 * Modules and variables imported into the stub are not considered
   exported from the stub unless the import uses the ``import ... as
   ...`` form or the equivalent ``from ... import ... as ...`` form.
+  (*UPDATE:* To clarify, the intention here is that only names using
```
hauntsaninja, line 1654: possible nit:
```suggestion
  (*UPDATE:* To clarify, the intention here is that only names imported using
```

PR #1633, opened 2020-10-02
PEP 634: The continuing saga

gvanrossum:


gvanrossum:
This is my cleanup of PEP 634. I plan no further revisions to this before sending it to the SC (unless we come to a decision on walrus vs AS).

I'll do PEP 635 next.

CC: @Tobias-Kohn, @brandtbucher, @viridia, @dmoisset

PR #1635, opened 2020-10-02
PEP 569: Python 3.8.6 was released on 2020-09-24

hugovk:
* https://www.python.org/downloads/release/python-386/
* https://discuss.python.org/t/python-3-8-6-is-now-available/5299


Mariatta (approved):
Thanks!

PR #1640, opened 2020-10-06
PEP637 setter notation getitem -> setitem

gerritholl:
In PEP-0637, in the example discussing setter notation, change the example from `__getitem__` to `__setitem__`, as the former appears to be a mistake in this context.


Rosuav:
I believe you're right.

PR #1641, opened 2020-10-07
PEP 596, 619: Update based on released versions

The-Compiler:
https://www.python.org/downloads/release/python-390/
https://www.python.org/downloads/release/python-3100a1/


pablogsal (approved):

pablogsal:
Thanks a lot for the PR @The-Compiler !

PR #1643, opened 2020-10-09
PEP 584: Fix order in specification examples

pochmann:
Updating an existing key's value doesn't move it to the back.


gvanrossum (approved):
Good eye!

PR #1659, opened 2020-10-19
PEP 621: Fix typos

domdfcoding:
In PEP 621 rather than saying that values "cannot be dynamic" in source distributions it actually says "cannot **by** dynamic". I'm sure that's a typo.


brettcannon (approved):

brettcannon:
Thanks!

PR #1661, opened 2020-10-19
PEP 634: change walrus pattern to AS pattern

gvanrossum:


gvanrossum:
Whoops, the PEP 635 half of this conflicts with another branch. I'll remove that for now.

pep-0634.rst
```diff
@@ -640,7 +622,7 @@ Other notation used beyond standard EBNF:
   mapping_pattern: '{' [items_pattern] '}'
   items_pattern: ','.key_value_pattern+ ','?
   key_value_pattern:
-      | (literal_pattern | value_pattern) ':' or_pattern
+      | (literal_pattern | value_pattern) ':' as_pattern
```
brandtbucher, line 625:
```suggestion
      | (literal_pattern | value_pattern) ':' pattern
```

pep-0634.rst
```diff
@@ -415,7 +400,7 @@ Syntax::
...
```
brandtbucher, line 403:
```suggestion
      | (literal_pattern | value_pattern) ':' pattern
```

pep-0634.rst
```diff
@@ -463,10 +448,7 @@ Syntax::
       | keyword_patterns
   positional_patterns: ','.pattern+
   keyword_patterns: ','.keyword_pattern+
-  keyword_pattern: NAME '=' or_pattern
-
-(Note that positional patterns may be unparenthesized walrus patterns,
-but keyword patterns may not.)
+  keyword_pattern: NAME '=' as_pattern
```
brandtbucher, line 451:
```suggestion
  keyword_pattern: NAME '=' pattern
```

pep-0634.rst
```diff
@@ -651,7 +633,7 @@ Other notation used beyond standard EBNF:
       | keyword_patterns
   positional_patterns: ','.pattern+
   keyword_patterns: ','.keyword_pattern+
-  keyword_pattern: NAME '=' or_pattern
+  keyword_pattern: NAME '=' as_pattern
```
brandtbucher, line 636:
```suggestion
  keyword_pattern: NAME '=' pattern
```

pep-0634.rst
```diff
@@ -177,8 +177,8 @@ Patterns
 The top-level syntax for patterns is as follows::
 
   patterns: open_sequence_pattern | pattern
-  pattern: walrus_pattern | or_pattern
-  walrus_pattern: capture_pattern ':=' or_pattern
+  pattern: as_pattern | or_pattern
+  as_pattern: or_pattern 'as' capture_pattern
```
brandtbucher, line 181: I forget, do we still need/want to unify the grammars in the implementation and spec? As it currently stands, `as_pattern` isn't a rule, it's just folded into `pattern`:

```
pattern[expr_ty]:
    | value=or_pattern 'as' target=NAME { _Py_MatchAs(value, target->v.Name.id, EXTRA) }
    | or_pattern
```

brandtbucher (changes requested):
Mostly good. Just a few fixes to the grammar changes:

pep-0634.rst
brandtbucher: (this isn't the only difference, though)

gvanrossum: I'd like the implementation eventually to use the same grammar as the PEP. I'd rather not change the PEP (unless we find bugs) because I use the rule names (e.g. "as pattern") in the text without explanation.

brandtbucher (approved):

PR #1664, opened 2020-10-19
PEP 640: Update post history and fix a bit of markup.

Yhg1s:


pep-0640.rst
```diff
@@ -132,7 +132,8 @@ Backwards Compatibility
 Introducing a new token means there are no backward compatibility concerns.
 No valid syntax changes meaning.
 
-"?" is not considered an identifier, so str.isidentifier() does not change.
+`"?"` is not considered an identifier, so `str.isidentifier()` does not
```
brandtbucher, line 135:
```suggestion
``"?"`` is not considered an identifier, so ``str.isidentifier()`` does not
```

hugovk: btw CI linting would catch this: PR https://github.com/python/peps/pull/1570 :)

Yhg1s: Fixed, thanks :)

PR #1665, opened 2020-10-19
PEP 635: More arguments against '?' wildcards

gvanrossum:
Now that PEP 640 exists we might as well mention another argument against "?".

@Tobias-Kohn

### DIFF
--- a/pep-0635.rst
+++ b/pep-0635.rst
@@ -782,9 +782,26 @@ items, capturing the first and last values.  While that may be the
 ultimate "wildcard", it does not convey the desired semantics.
 
 An alternative that does not suggest an arbitrary number of items
-would be ``?``.  However, this would require changes in the tokenizer,
-and it would put Python in a rather unique position:
+would be ``?``.  This is even being proposed independently from
+pattern matching in PEP 640.  We feel however that using ``?`` as a
+special "assignment" target is likely more confusing to Python users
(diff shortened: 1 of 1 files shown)


brandtbucher (approved):
Looks fine to me. I'm not sure I feel as strongly about it as you... but then again, it's not my PEP! 🙂

gvanrossum:
If you squint it's almost the same argument as "_" is used in all other languages. :-)

We'll live regardless of what happens with PEP 640.

Tobias-Kohn:
Yes, this looks good.  The situation where the discussion for or against `?` is held distributed over two PEPs is slightly strange (PEP 640 mentions our PEPs quite prominently and repeatedly and this seems a bit of an answer to that ^_^), but I fully agree with every bit you wrote here :-).

Your argument here goes further than the "`_` is used in all other languages" I would say; the question mark is obviously used with so many different meanings that it does not seem to have an agreed upon inherent meaning.  This seems quite relevant to me, as it translates to "the question mark _seems_ intuitive, but experience shows it really is _not_."

> We'll live regardless of what happens with PEP 640.

Yes, absolutely, but it just is nicer without a question mark :-).

gvanrossum:
There's also a discussion about this going on in python-dev, under "PEP 640: Unused variable syntax."

PR #1675, opened 2020-10-22
PEP-634: Describe how keys are compared in mapping patterns

dmoisset:
Clarifies slightly better how mapping patterns work, in particular it covers better if a ``True`` key in a pattern matches a ``1`` value in the subject.


gvanrossum (approved):
Thanks for catching this!

PR #1694, opened 2020-10-28
Sequence patterns cannot match iterators

brandtbucher:
Whoops.

### DIFF
--- a/pep-0634.rst
+++ b/pep-0634.rst
@@ -361,7 +361,7 @@ otherwise it is a variable-length sequence pattern.
 
 A sequence pattern fails if the subject value is not an instance of
 ``collections.abc.Sequence``.  It also fails if the subject value is
-an instance of ``str``, ``bytes`` or ``bytearray``.
+an iterator, or an instance of ``str``, ``bytes``, or ``bytearray``.
 
 A fixed-length sequence pattern fails if the length of the subject
 sequence is not equal to the number of subpatterns.


pep-0634.rst
```diff
@@ -361,7 +361,7 @@ otherwise it is a variable-length sequence pattern.
 
 A sequence pattern fails if the subject value is not an instance of
 ``collections.abc.Sequence``.  It also fails if the subject value is
-an instance of ``str``, ``bytes`` or ``bytearray``.
+an iterator, or an instance of ``str``, ``bytes``, or ``bytearray``.
```
gvanrossum, line 364: This is redundant -- iterators are not instances of `Sequence`.

brandtbucher: Hm, okay. I mean, it *is* possible for an object to implement both protocols (and the implementation protects against that by rejecting anything with `__next__`), but I agree that that’s probably obscure and uncommon enough to leave out of the spec.

gvanrossum: Why does the implementation need to protect against this? Are there specific that are `Sequence` instances and have a `__next__`? Would anything go horribly wrong if a user defined such a type? (Who knows, it might really behave like a `Sequence` but have a `__next__` for historical reasons that always raises.)

brandtbucher: Ah, I think this originates from back in April or so when we still weren't sure how sequence patterns would work. We knew iterators were out, but hadn't settled on `collections.abc.Sequence` yet. It looks like the the prohibition made it as far as PEP 622...

> To match a sequence pattern the subject must be an instance of `collections.abc.Sequence`, and it cannot be any kind of string (`str`, `bytes`, `bytearray`). It cannot be an iterator.

...but was dropped when we split them up.

You're absolutely right that the implementation doesn't need to screen out iterators, since we only ever try length checks and nonnegative integer indices anyways... iteration isn't even used. I'll update it tonight.

PR #1695, opened 2020-10-29
PEP 626: Clarify lifetimes of helper functions and struct.

markshannon:
As requested by @pablogsal

### DIFF
--- a/pep-0626.rst
+++ b/pep-0626.rst
@@ -219,6 +219,7 @@ Note that these functions are not part of the C-API, so will be need to be linke
   For reliable operation, ``linetable`` should be copied into a local buffer before calling ``PyLineTable_InitAddressRange``.
 
   Although these functions are not part of C-API, they will provided by all future versions of CPython.
+  The ``PyLineTable_`` functions do not call into the C-API, so can be safely copied into any tool that needs to use them.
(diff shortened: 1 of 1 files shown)


pep-0626.rst
```diff
@@ -214,6 +214,13 @@ Note that these functions are not part of the C-API, so will be need to be linke
 
 ``PyLineTable_PreviousAddressRange`` retreats the range to the previous entry, returning non-zero if valid.
 
+.. note::
+  The data in ``linetable`` is immutable, but its lifetime depends on its code object.
+  For reliable operation, ``linetable`` should be copied into a local buffer before calling ``PyLineTable_InitAddressRange``.
+
+  Although these functions are not part of C-API, they will provided by all future versions of CPython.
```
pablogsal, line 221:
```suggestion
  Although these functions are not part of C-API, they will provided by all future versions of CPython and they will remain self-contained.
```

pablogsal (commented):
You can also mark the PEP as accepted if you wish

pep-0626.rst
markshannon: What does "self contained" mean in this context? I've never heard that term applied to a C function.

pablogsal: That it does not require the rest of the C-API (so copying all these functions to a different process will suffice).

pablogsal (approved):

PR #1718, opened 2020-11-16
PEP 635: Add link to Guido's copy of DLS '20 paper.

gvanrossum:


pxeger:
The linked paper has a typo in the Acknowledgments section - it says "eferences" not "references". May be too late to change now?

PR #1724, opened 2020-12-01
PEP 643: Mark as accepted

pganssle:
As BDFL-Delegate, I am happy to accept PEP 643: https://discuss.python.org/t/pep-643-metadata-for-package-source-distributions/5577/53


pfmoore (approved):
I'll get the packaging.python.org PR merged over the next day or two.

PR #1745, opened 2020-12-24
PEP 612: Clean up code examples, link to CPython implementation

Fidget-Spinner:
An unused variable was removed from the code example.

I also added a line to point to the [CPython implementation](https://github.com/python/cpython/pull/23702) for posterity.


gvanrossum:
Can you also fix the typo on line 302?

gvanrossum (approved):

PR #1751, opened 2020-12-29
PEP 646: Various updates

mrahtz:
Naming:
* Rename Tensor -> Array (it's less jargony, and I think it's the more general term)
* Rename Expand -> Unpack (to be consistent with the terminology for normal tuples)
* Rename ArgTs -> Ts, ReturnT -> R
* Rename 'type tuple variable' -> 'type variable tuple'

Semantic:
* Explicitly state that a TypeVarTuple can hold *zero* or more types
* Explicitly state that TypeVarTuple doesn't support variance or bound yet
* Support Union[*Ts]
* Support concatenating multiple unpacked TypeVarTuple when there's no ambiguity
* Support aliases
* Remove support for class overloads; replace with overloads of individual methods

Pedagogical:
* Reorder introductory material to make it clearer how TypeVarTuples behave when not unpacked, and to make it clearer that using them without unpacking them is a perfectly valid thing to do
* Remove example of unpacking being used with a regular Tuple rather than a TypeVarTuple (my main reason for wanting to include it was to emphasise that a TypeVarTuple behaves like a Tuple, but I think this is emphasised better in previous sections now, and since I don't think it has any use cases, it seems better to remove it to keep things shorter and more to the point)
* Replace Tuple[*Ts] with just Ts (now that we've settled on Ts definitely meaning "A Tuple filled with types", writing Tuple[*Ts] is redundant - it's exactly the same as Ts, but with more keystrokes)
* State explicitly that TypeVarTuple can be used with Callable
* Changes args_to_tuples example to args_to_lists (so it's clearer where the Tuple comes from)
* Show more examples of where Map can be used
* Move section on nesting Map to the 'Rationale and Rejected Ideas' section (it's complicated enough to be too distracting if it were in the main section, and since there isn't an obvious use-case, we leave it as an optional feature)
* Add section on a full example of an array type
* Remove ideas for future PEPs (to reduce length)
* Add more detail on the range of type concatenations that are allowable

Other:
* Add Pradeep to the authors list, since he's contributed so much :)
* Add Eric Traut in the acknowledgements
* Update Post-History
* Fix some references
* Fix first Array example to remove the need for a cast
* Various wording tweaks

### DIFF
--- a/pep-0646.rst
+++ b/pep-0646.rst
@@ -2,6 +2,7 @@ PEP: 0646
 Title: Variadic Generics
 Author: Mark Mendoza <mendoza.mark.a@gmail.com>,
         Matthew Rahtz <mrahtz@google.com>,
+        Pradeep Kumar Srinivasan <gohanpra@gmail.com>,
         Vincent Siles <vsiles@fb.com>
 Sponsor: Guido van Rossum <guido@python.org>
 Status: Draft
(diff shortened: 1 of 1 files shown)


gvanrossum (approved):
Rather than reviewing this carefully and filing nits, I'll just land your changes and eventually I'll do another full pass.

PR #1691, opened 2020-10-27
PEP 635: Add a rebuttal of PEP 642

brandtbucher:
CC @Tobias-Kohn

PR #1757, opened 2021-01-04
Revert "PEP 635: Add a rebuttal of PEP 642 (GH-1691)"

gvanrossum:
This reverts commit aafca24693112ae2513f04169f355eba6ef4cd77.

PEP 642 was updated and the commentary here is out of date.
IMO we needn't address it here.


brandtbucher (approved):
Yup, I agree.

PR #1762, opened 2021-01-14
PEP 650: Specifying Installer Requirements for Python Projects

brettcannon:


brettcannon:
@di FYI; once this is merged I will do a separate PR to update the Post-History and Discussions-To to discuss.python.org.

PR #1773, opened 2021-01-20
PEP 651: Change title to clarify topic

gvanrossum:
@markshannon When I first saw your email I thought it was about numeric overflow.


markshannon:
:+1:

PR #1774, opened 2021-01-21
PEP 637: Added notation use case for star arguments on PEP 646

stefanoborini:
Adds a use case in the proposal for star arguments as needed in PEP 646 Variadic Generics.


gvanrossum:
Thanks!

PR #1775, opened 2021-01-22
PEP 651: Clarifications, as suggested by GvR.

markshannon:


gvanrossum (approved):
Do you have a work-in-progress prototype implementation yet? That would be good to link to from the PEP.

Thanks for doing this! This has bugged me forever. At this point I think you should just submit to the SC for acceptance.

markshannon:
No implementation as yet, but I've got the C-stack overflow checks working (on linux at least).
The implementation of Python-to-python calls should be reasonably straightforward once https://bugs.python.org/issue42990 is done.

PR #1783, opened 2021-01-27
PEP 558: Make it clear tracing mode is being eliminated

ncoghlan:
The initial version of the PEP entrenched the tracing mode distinction,
and was worded accordingly.

Now that tracing mode is only relevant as a historical artifact that is
being eliminated, the wording should make that clear.


markshannon:
:+1: That's clearer

PR #1797, opened 2021-02-05
PEP 621, 635, 637, 639, 642, 644, 645, 647, 650: Fix single backticks

hugovk:
This is the same as https://github.com/python/peps/pull/1668, but without changes to code blocks, and without the linter.


gvanrossum (approved):
Thanks!

PR #1799, opened 2021-02-06
PEP 597: Update

methane:
* Add Backward/Forward Compatibility section
* Add How to teach this section
* Remove io.LOCALE_ENCODING constant


orsenthil:
I am a core-dev, but not an expert in this domain. The changes proposed looks good to me.

vstinner (commented):
I like most changes but I'm not sure on how to make the warning quiet (use "locale" on purpose) without losing support for Python 3.9 and older. See my email on python-dev.

vstinner:
Thanks, I like the updated PEP ;-)

PR #1805, opened 2021-02-07
Normalize the 'Created' field

brettcannon:
All PEPs now follow the format specified in PEP 1: `dd-mmm-yyyy` which corresponds to `%d-%b-%Y`.

### DIFF
--- a/pep-0004.txt
+++ b/pep-0004.txt
@@ -6,7 +6,7 @@ Author: Brett Cannon <brett@python.org>, Martin von Löwis <martin@v.loewis.de>
 Status: Active
 Type: Process
 Content-Type: text/x-rst
-Created: 1-Oct-2000
+Created: 01-Oct-2000
 Post-History:
 
 
--- a/pep-0013.rst
+++ b/pep-0013.rst
@@ -4,7 +4,7 @@ Author: The Python core team and community
 Status: Active
 Type: Informational
 Content-Type: text/x-rst
-Created: 2018-12-16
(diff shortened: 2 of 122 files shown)


gvanrossum (approved):
Note that there was a typo in your description (`yyy` -> `yyyy`).

Are you going to do the other date fields too?

encukou:
> `dd-mmm-yyyy` which corresponds to `%d-%b-%Y`.

Only with some locales. Sometimes, `%d-%b-%Y` gives me `08-úno-2021`.

Would it be reasonable to change PEP 1 to format that's independent of locale, like `%Y-%m-%d` (`2021-02-08`)?

gvanrossum:
Can we first land this? There are too many pending PRs relating to this minor issue.

Re: the idea of using a different format needs much more consideration before it can be made to work -- personally I'd prefer to keep the current format and just state it has to be the English locale (like the rest of the text has to be in English).

encukou:
Sure – if you've considered the locale issue, go ahead!

(I'm just raising the point because, IMO, this is the time it should be considered. The more automation is built around it, the harder it'll be to change.)

brettcannon:
@gvanrossum I only did this to mine the PEPs for creation date details, so my itch is scratched at this point. 😉

Jeffkeen7 (commented):
JeffreyKeen7:normalize-creation-date-creator-owner-developer-composer=jeffreykeen7

PR #1825, opened 2021-02-20
Update PEP 646 draft

mrahtz:
Semantic changes:
* Unparameterised variadics are compatible with parameterised variadics in both direction
* Allow suffixing

Readability changes:
* De-emphasise unparameterised `Array` behaving like `Array[Any, ...]` to avoid the suggestion that `Array[Any, ...]` might be supported
* Explicitly forbid `Array[Any, ...]` and similar
* Clarify that only `*args` can be annotated as `*Ts`; other arguments must use `Tuple[*Ts]`
* Remove 'Final Notes' section (I was hoping we'd have more edge cases to add to this, but in the end there haven't been any, and the only example we do have is obscure enough that I think we can just remove it)

Also some minor typo fixes.


gvanrossum (approved):
Didn't really read it carefully, will just merge.

PR #1828, opened 2021-02-21
Update PEP 646 draft

mrahtz:
Readability changes:
* Tweak framing to also acknowledge that variadic generics have been wanted for other things too (though as far as I've been able to tell there isn't a consistent enough pattern to the other use cases to justify including them in the PEP)
* Move the Array example much earlier (both because reading a detailed example is going to be a faster way of getting the gist for people in a hurry, and because we want to make it clear to numerical computing folks from the start that we can also specify a datatype here). (We still have a more detailed Array example later; I'm leaving that there for the time being pending some discussion with Pradeep, but eventually I guess it'll be removed and absorbed into this earlier example)
* Make it clear that other types can appear in `Union` alongside `TypeVarTuple` instances.

Semantic changes:
* Disallow a `Union` of more than one type variable tuple (I thought this would be easier than concatenation of more than one type variable tuple, but Pradeep points out it could still get hairy for type checkers, so disallow it for now to be on the safe side)

Also another typo fix (`add_batch_axis` -> `del_batch_axis`).

CC @pradeep90

### DIFF
--- a/pep-0646.rst
+++ b/pep-0646.rst
@@ -18,7 +18,8 @@ Abstract
 PEP 484 introduced ``TypeVar``, enabling creation of generics parameterised
 with a single type. In this PEP, we introduce ``TypeVarTuple``, enabling parameterisation
 with an *arbitrary* number of types - that is, a *variadic* type variable,
-enabling *variadic* generics. This allows the type of array-like structures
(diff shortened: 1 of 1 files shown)


gvanrossum (approved):
I'll merge this now, trying to avoid long-pendign PRs.

PR #1830, opened 2021-02-22
A couple PEP 654 edits

1st1:
* Fix a few markup issues & typos; use REsT references.
* Mention Task Groups in Motivation; add an example of the ``trio.MultiError`` API

### DIFF
--- a/pep-0654.rst
+++ b/pep-0654.rst
@@ -38,11 +38,16 @@ together as the stack unwinds. Several real world use cases are listed below.
   exceptions in the results list.  The Trio [2]_
   library has a ``MultiError`` exception type which it raises to report a
   collection of errors. Work on this PEP was initially motivated by the
-  difficulties in handling ``MultiErrors``, which are detailed in a design
+  difficulties in handling ``MultiErrors`` [9]_, which are detailed in a design
   document for an improved version, ``MultiError2`` [3]_.
   That document demonstrates how difficult it is to create an effective API
   for reporting and handling multiple errors without the language changes we
-  are proposing.
+  are proposing (see also the `Programming Without 'except \*'`_ section.)
+
+  Implementing a better task spawning API in asyncio, inspired by Trio
+  nurseries, was the main motivation for this PEP.  That work is currently
+  blocked on Python not having native language level support for exception
+  groups.
 
 * **Multiple failures when retrying an operation.** The Python standard
   library's ``socket.create_connection`` function may attempt to connect to
@@ -90,9 +95,9 @@ Rationale
 Grouping several exceptions together can be done without changes to the
 language, simply by creating a container exception type.
 Trio [2]_ is an example of a library that has made use of this technique in its
-``MultiError`` [9]_ type. However, such approaches require calling code to catch
-the container exception type, and then inspect it to determine the types of
-errors that had occurred, extract the ones it wants to handle and reraise the
+``MultiError`` [9]_ type. However, such an approach requires calling code to catch
+the container exception type, and then to inspect it to determine the types of
+errors that had occurred, extract the ones it wants to handle, and reraise the
 rest.
 
 Changes to the language are required in order to extend support for
@@ -1054,6 +1059,49 @@ are handling ``ExceptionGroups`` of ``T`` rather than only naked ``Ts`` should b
 specified in the same place where we state ``T``.
 
 
+Programming Without 'except \*'
+===============================
+
+Consider the following simple example of the ``except *`` syntax (pretending
+Trio natively supported this proposal):
+
+.. code-block::
+
+   try:
+       async with trio.open_nursery() as nursery:
+           # Make two concurrent calls to child()
+           nursery.start_soon(child)
+           nursery.start_soon(child)
+   except *ValueError:
+       pass
+
+Here is how this code would look in Python 3.9:
+
+.. code-block::
+
+   def handle_ValueError(exc):
+       if isinstance(exc, ValueError):
+           return None
+       else:
+           return exc   # reraise exc
+
+   with MultiError.catch(handle_ValueError):
+       async with trio.open_nursery() as nursery:
+           # Make two concurrent calls to child()
+           nursery.start_soon(child)
+           nursery.start_soon(child)
+
+This example clearly demonstrates how unintuitive and cumbersome handling
+of multiple errors is in current Python.  The exception handling logic has
+to be in a separate closure and is fairly low level, requiring the writer to
+have non-trivial understanding of both Python exceptions mechanics and the
+Trio APIs.  Instead of using the ``try..except`` block we have to use a
+``with`` block.  We need to explicitly reraise exceptions we are not handling.
+Handling more exception types or implementing more complex
+exception handling logic will only further complicate the code to the point
+of it being unreadable.
+
+
 See Also
 ========
 
@@ -1094,6 +1142,7 @@ References
 
 .. [12] https://github.com/python/exceptiongroups/issues/4
 
+.. [13] https://trio.readthedocs.io/en/stable/reference-core.html#nurseries-and-spawning
 
 Copyright
 =========
--- a/pep-0654.rst
+++ b/pep-0654.rst
@@ -1,6 +1,8 @@
 PEP: 654
 Title: Exception Groups and except*
-Author: Irit Katriel <iritkatriel@gmail.com>, Yury Selivanov <yury@edgedb.com>, Guido van Rossum <guido@python.org>
+Author: Irit Katriel <iritkatriel@gmail.com>,
+        Yury Selivanov <yury@edgedb.com>,
+        Guido van Rossum <guido@python.org>
 Status: Draft
 Type: Standards Track
 Content-Type: text/x-rst
@@ -23,8 +25,7 @@ Motivation
 ==========
 
 The interpreter is currently able to propagate at most one exception at a
-time. The chaining features introduced in
-`PEP 3134 <https://www.python.org/dev/peps/pep-3134/>`_ link together
+time. The chaining features introduced in :pep:`3134` link together
 exceptions that are related to each other as the cause or context, but
 there are situations where multiple unrelated exceptions need to be propagated
 together as the stack unwinds. Several real world use cases are listed below.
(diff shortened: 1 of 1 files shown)


pep-0654.rst
```diff
@@ -1048,72 +1060,103 @@ clauses. We rejected this because it would be less obvious. The fact that we
...
+       nursery.start_soon(child)
+       nursery.start_soon(child)
 
-* The ``asyncio`` standard library
+This example clearly demonstrates how unintuitive and cumbersome handling
+of multiple errors is in current Python.  The exception handling logic has
+to be in a separate closure and is fairly low level, requiring the writer to
+have non-trivial understanding of both Python exceptions mechanics and the
+Trio APIs.  Instead of using the ``try..except`` block we have to use a
+``with`` block.  Handling more exception types or implementing more complex
```
iritkatriel, line 1101:
```suggestion
``with`` block.  We need to explicitly reraise exceptions we are not handling. Handling more exception types or implementing more complex
```

pep-0654.rst
```diff
@@ -1048,72 +1060,103 @@ clauses. We rejected this because it would be less obvious. The fact that we
...
 
-  PR:  https://github.com/iritkatriel/cpython/pull/10
+.. code-block::
 
-* PEP 3134: Exception Chaining and Embedded Tracebacks
+   def handle_ValueError(exc):
+     if isinstance(exc, ValueError):
+       return None
+     else:
+       return exc
```
iritkatriel, line 1088:
```suggestion
       return exc   # reraise exc
```

iritkatriel (approved):
Great! I just added the point about having to explicitly reraise exception you're not handling.

1st1:
Thanks Irit!

@gvanrossum Guido, does the PR look good to you?

pep-0654.rst
```diff
@@ -23,49 +25,52 @@ Motivation
...
   for reporting and handling multiple errors without the language changes we
-  are proposing.
+  are proposing (see also the `Programming Without except*`_ section.)
+
+* **Implementing better Task spawning API in asyncio**.
+  The current asyncio API for spawning child Tasks -- the ``asyncio.gather()``
+  function [1]_ -- has a few shortcomings, the most notable of which is
+  how it handles exceptions.  A better API, inspired by the Trio's "Nursery"
+  primitive [13]_, has been a long-awaited feature of asyncio, blocked on
+  Python not having native language level support for exception groups.
```
gvanrossum, line 52: This needs some more elaboration. Maybe just reference the complexity of Trio's solution?

pep-0654.rst
```diff
@@ -23,49 +25,52 @@ Motivation
...
   multiple tasks and return their results in aggregate. There isn't currently
   a good way for such libraries to handle situations where multiple tasks
-  raise exceptions. The Python standard library's
-  `asyncio.gather() <https://docs.python.org/3/library/asyncio-task.html#asyncio.gather>`_
+  raise exceptions. The Python standard library's ``asyncio.gather()`` [1]_
   function provides two options: raise the first exception, or return the
-  exceptions in the results list.  The `Trio <https://trio.readthedocs.io/en/stable/>`_
+  exceptions in the results list.  The Trio [2]_
   library has a ``MultiError`` exception type which it raises to report a
   collection of errors. Work on this PEP was initially motivated by the
```
gvanrossum, line 40: Maybe also link to `[9]` here?

pep-0654.rst
```diff
@@ -90,12 +96,11 @@ Rationale
...
-that has made use of this technique in its
-`MultiError <https://trio.readthedocs.io/en/stable/reference-core.html#trio.MultiError>`_
-type. However, such approaches require calling code to catch the container
-exception type, and then inspect it to determine the types of errors that had
-occurred, extract the ones it wants to handle and reraise the rest.
+Trio [2]_ is an example of a library that has made use of this technique in its
+``MultiError`` [9]_ type. However, such approaches require calling code to catch
+the container exception type, and then inspect it to determine the types of
+errors that had occurred, extract the ones it wants to handle and reraise the
+rest.
```
gvanrossum, line 103:
```suggestion
``MultiError`` [9]_ type. However, such an approache requires calling code to catch
the container exception type, and then to inspect it to determine the types of
errors that had occurred, extract the ones it wants to handle, and reraise the
rest.
```

pep-0654.rst
iritkatriel, line 52: Is there an open issue or something else you could link to here?

pep-0654.rst
```diff
@@ -1048,72 +1060,103 @@ clauses. We rejected this because it would be less obvious. The fact that we
 are handling ``ExceptionGroups`` of ``T`` rather than only naked ``Ts`` should be
 specified in the same place where we state ``T``.
 
-See Also
-========
 
-* An analysis of how exception groups will likely be used in asyncio
-  programs:
-  https://github.com/python/exceptiongroups/issues/3#issuecomment-716203284
+Programming Without except*
```
gvanrossum, line 1064:
```suggestion
Programming Without ``except *``
```
(Not sure if RsT likes code blocks in titles though.)

pep-0654.rst
```diff
@@ -1048,72 +1060,103 @@ clauses. We rejected this because it would be less obvious. The fact that we
...
+===========================
 
-* The issue where the ``except*`` concept was first formalized:
-  https://github.com/python/exceptiongroups/issues/4
+Consider the following simple example of the ``except *`` syntax (suppose
+Trio natively supports this proposal):
```
gvanrossum, line 1068:
```suggestion
Consider the following simple example of the ``except *`` syntax (pretending
Trio natively supported this proposal):
```

(Also: it's a little unconventional to see the "efficient" version first and the "inefficient" version second. Usually we do before/after, where before is the old way and after is the new, improved way. :-)

pep-0654.rst
```diff
@@ -1048,72 +1060,103 @@ clauses. We rejected this because it would be less obvious. The fact that we
...
+.. code-block::
 
-* Reference implementation:
+   try:
+     async with trio.open_nursery() as nursery:
+       # Make two concurrent calls to child()
+       nursery.start_soon(child)
+       nursery.start_soon(child)
+   except *ValueError:
+     pass
```
gvanrossum, line 1078: Please use 4-space indents.

pep-0654.rst
```diff
@@ -1048,72 +1060,103 @@ clauses. We rejected this because it would be less obvious. The fact that we
...
+       nursery.start_soon(child)
 
-* The ``asyncio`` standard library
+This example clearly demonstrates how unintuitive and cumbersome handling
+of multiple errors is in current Python.  The exception handling logic has
+to be in a separate closure and is fairly low level, requiring the writer to
+have non-trivial understanding of both Python exceptions mechanics and the
+Trio APIs.  Instead of using the ``try..except`` block we have to use a
+``with`` block.  We need to explicitly reraise exceptions we are not handling. Handling more exception types or implementing more complex
+exception handling logic will only further complicate the code to the point
```
gvanrossum, line 1102:
```suggestion
``with`` block.  We need to explicitly reraise exceptions we are not handling.
Handling more exception types or implementing more complex
exception handling logic will only further complicate the code to the point
```

gvanrossum (commented):
(Sorry I seem to be a bit hung up on tenses. :-)

pep-0654.rst
1st1: It should handle them OK, but they will look awkward when rendered to HTML. So I personally try to avoid using them in titles.

gvanrossum:
(Please use a more precise commit message when you're applying a suggestion or using the wed editor. "Update <document>" says nothing.)

pep-0654.rst
gvanrossum: Maybe use single quotes then?

pep-0654.rst
1st1: Yeah! I had the same thought, but then opted for my variant because of just how arcane the "inefficient" way looks like compared to `except*`. I'm afraid that readers would be lost in it and skip the whole section, not getting why we added it in the first place. :/

1st1:
> (Please use a more precise commit message when you're applying a suggestion or using the wed editor. "Update " says nothing.)

Yeah, I'm always folding all suggestion commits manually before merging the PR.

pep-0654.rst
1st1: Alright

pep-0654.rst
1st1: @iritkatriel @gvanrossum I can add a section to the PEP about TaskGroups API. Or you guys think that creating a bpo with the proposed design would be a better move?

pep-0654.rst
1st1: Done

1st1:
I've forced pushed a rebased version with all suggestion commits squashed.

pep-0654.rst
iritkatriel: Adding that design here means we are asking the steering council to approve it as part of this PEP, and if it's not straightforward I think it makes this PEP too big in scope. To Guido's point - isn't this really the same problem/use case as Trio's?

gvanrossum: Yeah, I think we could shorten this parargaph so it just says "the same applies for asyncio".

gvanrossum:
Also, maybe just land it and iterate in a new PR.

pep-0654.rst
1st1: My problem with removing every notion of TaskGroups from this PEP weakens its motivation.  Ultimately adding task groups to asyncio were the main reason why we started working on this PEP.

gvanrossum: So say exactly that.

iritkatriel: I don't think you should remove it, just roll it into the same paragraph with Trio because it's the same kind of use case.  Like the paragraph about multiple use callbacks has two different use cases (atexit and pytest).

1st1: I've pushed an update, basically adding this paragraph to the first bullet point:

>   Implementing a better, inspired by Trio nurseries, task spawning API in
  asyncio was the main motivation for this PEP. That work is currently blocked
  on Python not having native language level support for exception groups.

I propose to merge this PR with that, but if you guys want to phrase it differently please open a follow up PR.

1st1:
> "Implementing a better task spawning API in asyncio, inspired by Trio nurseries, was the main motivation for this PEP."

Done, @iritkatriel

1st1:
Merged! Thanks for reviews and suggestions @iritkatriel and @gvanrossum. Please feel free to edit the new content.

iritkatriel:
Cool!   I think there are a bunch of section headings that appear at the same level as except* but should be under it. Will fix.

1st1:
> Cool! I think there are a bunch of section headings that appear at the same level as except* but should be under it. Will fix.

Also need to make sure we use consistent indentation in our code snippets. Right now it's 2 spaces in some -- need to change that to 4.  Do you want me to submit a PR?

iritkatriel:
> Also need to make sure we use consistent indentation in our code snippets. Right now it's 2 spaces in some -- need to change that to 4. Do you want me to submit a PR?

Yes please

PR #1836, opened 2021-02-23
PEP 654: reference 12029 re: using the same check as except for a match

graingert:


gvanrossum:
I'm going to leave it up to @iritkatriel and/or @1st1 to approve this addition -- I'm not sure this issue is relevant for the readers of the PEP (although it is the reason of this particular phrasing).

iritkatriel:
If anything, shouldn't we reference the language reference?  
https://docs.python.org/dev/reference/compound_stmts.html#the-try-statement

gvanrossum:
I don't think we need to point to the language reference -- it's implied.
:-)

1st1:
I'm not sure I see the point of linking that issue. Perhaps @graingert could explain?

PR #1839, opened 2021-02-24
PEP 634: Fix error typo in mapping patterns (TypeError ->ValueError)

Fidget-Spinner:
I don't think mapping patterns are supposed to raise TypeError on duplicate keys, instead it's probably ValueError.  A few paragraphs down the PEP mentions:

> If duplicate keys are detected in the mapping pattern, the pattern is considered invalid, and a ValueError is raised.

So it seems like a contradiction.

The implementation seems to only raise TypeError for class patterns too:
https://github.com/python/cpython/pull/22917/files

### DIFF
--- a/pep-0634.rst
+++ b/pep-0634.rst
@@ -412,7 +412,7 @@ and it must be last.
 A mapping pattern may not contain duplicate key values.
 (If all key patterns are literal patterns this is considered a
 syntax error; otherwise this is a runtime error and will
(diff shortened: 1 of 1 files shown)


brandtbucher (approved):
[Yep.](https://github.com/python/cpython/blob/ca6e332b917342ee67b36aa7315efa4d203be112/Python/ceval.c#L927-L929) Thanks!

PR #1840, opened 2021-02-25
PEP 427: Point to spec on packaging.python.org

takluyver:
Counterpart to https://github.com/pypa/packaging.python.org/pull/839 - cc @pfmoore


pfmoore (approved):
I'll wait to merge this until the packaging.python.org PR is merged.

PR #1844, opened 2021-02-26
PEP-654: remove cause, context from traverse()

iritkatriel:


iritkatriel:
This was a middle of the night silliness. 

We don't need to pass the cause and context from the root. The ones for the leaf are on the leaf exception itself.

PR #1880, opened 2021-03-20
PEP 646: Be more explicit about grammar changes needed

mrahtz:
Since PEP 637 was rejected.

### DIFF
--- a/pep-0646.rst
+++ b/pep-0646.rst
@@ -690,9 +690,13 @@ Backwards Compatibility
 In order to use the star operator for unpacking of ``TypeVarTuple`` instances,
 we would need to make two grammar changes:
 
-1. Star expressions must be made valid in index operations,
-   e.g. ``Tuple[*Ts]`` (but *not* in slice operations in general - e.g.
-   ``Tuple[*Ts:*Ts]`` is nonsensical and should remain invalid).
+1. Star expressions must be made valid in at least index operations.
+   For example, ``Tuple[*Ts]`` and ``Tuple[T1, *Ts, T2]`` would both
+   be valid. (This PEP does not allow multiple unpacked ``TypeVarTuple``
+   instances to appear in a single parameter list, so ``Tuple[*Ts1, *Ts2]``
+   would be a runtime error. Also note that star expressions would *not*
+   be valid in slice expressions -  e.g. ``Tuple[*Ts:*Ts]`` is
+   nonsensical and should remain invalid.)
 2. We would need to make '``*args: *Ts``' valid.
 
 In both cases, at runtime the star operator would call ``Ts.__iter__()``.
--- a/pep-0646.rst
+++ b/pep-0646.rst
@@ -695,6 +695,10 @@ we would need to make two grammar changes:
    ``Tuple[*Ts:*Ts]`` is nonsensical and should remain invalid).
 2. We would need to make '``*args: *Ts``' valid.
 
+In both cases, at runtime the star operator would call ``Ts.__iter__()``.
+This would, in turn, return an instance of a helper class, e.g.
+``UnpackedTypeVarTuple``, whose ``repr`` would be ``*Ts``.
+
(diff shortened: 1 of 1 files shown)


pep-0646.rst
```diff
@@ -683,11 +687,17 @@ these type annotations.
...
-is '``*args: *Ts``', which would require an additional change. (This
-change could be avoided by simply mandating this be written
-as ``Unpack[Ts]``, however.)
+In order to use the star operator for unpacking of ``TypeVarTuple`` instances,
+we would need to make two grammar changes:
+
+1. Star expressions must be made valid in index operations,
+   e.g. ``Tuple[*Ts]`` (but *not* in slice operations in general - e.g.
+   ``Tuple[*Ts:*Ts]`` is nonsensical and should remain invalid).
+2. We would need to make '``*args: *Ts``' valid.
```
JelleZijlstra, line 696: You'll also have to discuss what this actually does at runtime. Does it call `Ts.__iter__()` or do something else? What will show up in `__annotations__`?

mrahtz: Ah, good point - I've added some details on this. (I haven't explicitly mentioned `__annotations__`, but I've mentioned what the `repr` of the object returned by `*Ts` is, which I hope should be sufficient.)

pep-0646.rst
```diff
@@ -683,11 +687,21 @@ these type annotations.
...
-the star operator are almost all covered by PEP 637. The one exception
-is '``*args: *Ts``', which would require an additional change. (This
-change could be avoided by simply mandating this be written
-as ``Unpack[Ts]``, however.)
+In order to use the star operator for unpacking of ``TypeVarTuple`` instances,
+we would need to make two grammar changes:
+
+1. Star expressions must be made valid in index operations,
+   e.g. ``Tuple[*Ts]`` (but *not* in slice operations in general - e.g.
+   ``Tuple[*Ts:*Ts]`` is nonsensical and should remain invalid).
```
gvanrossum, line 695: Also clarify that `Tuple[x, *y, z]` is supported.

we probably need to clarify that this syntax would be accepted anywhere, with semantics as described in PEP 637 (there are subtle edge cases that were reasoned through for PEP 637).

mrahtz: What kinds of place do you have in mind when you say "accepted anywhere"? Aren't starred expressions still only valid in the context of indexing in PEP 637?

In terms of "subtle edge cases" - I see there are these examples in PEP 637:
```python
obj[*()]
obj[*(x,)]
obj[*(x,),]
```
Are these the ones you had in mind?

More generally, I'm started to wonder whether it would be better to do a from-scratch implementation of PEP 646 in CPython (rather than basing it on the PEP 637 implementation) ato get a better sense of exactly what would be required. Do you think this would be worth it - would this make a big difference to the SC? Or would it be better to get the SC's early opinion before spending time on implementation?

pep-0646.rst
```diff
@@ -683,11 +687,25 @@ these type annotations.
...
+we would need to make two grammar changes:
+
+1. Star expressions must be made valid in at least index operations.
+   For example, ``Tuple[*Ts]`` and ``Tuple[T1, *Ts, T2]`` would both
+   be valid. (This PEP does not allow multiple unpacked ``TypeVarTuple``
+   instances to appear in a single parameter list, so ``Tuple[*Ts1, *Ts2]``
+   would be a runtime error. Also note that star expressions would *not*
+   be valid in slice expressions -  e.g. ``Tuple[*Ts:*Ts]`` is
+   nonsensical and should remain invalid.)
+2. We would need to make '``*args: *Ts``' valid.
```
gvanrossum, line 700:
```suggestion
2. We would need to make '``*args: *Ts``' valid in function definitions.
```

gvanrossum (approved):
With my small change I now declare this perfect. I will merge this PR and send the PEP to the SC on your behalf.

mrahtz:
Great, thanks Guido!

PR #1918, opened 2021-04-10
PEP 626: typo: should be f_lineno, not f_lineo.

ztane:
I noticed this because the typo even found its way to the [What's new listing in CPython](https://github.com/python/cpython/blame/master/Doc/whatsnew/3.10.rst#L176)


Rosuav:
Thanks, yep, that looks like a typo.

PR #1921, opened 2021-04-13
Bisect CODEOWNERS to try and figure out why it isn't working

brettcannon:
Part of #1920

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,600 +3,600 @@
 # active core developers.
 
 # The PEP editors are the fallback "owners" for everything in this repository.
-*             @python/pep-editors
+# *             @python/pep-editors
 
-pep-0001.txt  @warsaw @ncoghlan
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -3,600 +3,600 @@
...
-pep-0414.txt  @mitsuhiko @ncoghlan
-pep-0415.txt  @benjaminp
-pep-0416.txt  @vstinner
-pep-0417.txt  @voidspace
-pep-0418.txt  @vstinner
-pep0418/      @vstinner
-# pep-0419.txt
-pep-0420.txt  @ericvsmith
-pep-0421.txt  @ericsnowcurrently
-pep-0422.txt  @ncoghlan Daniel Urban
```
JelleZijlstra, line 256: This line looks like it might have been the problem

JelleZijlstra: I'll submit a PR reverting this diff except for the bad line to see if it works

PR #1946, opened 2021-05-05
PEP 0013: Update Current Steering Council for 2021 term

matrixise:


warsaw (approved):
Thanks!  I'm not sure how this didn't get updated after the election :smile:

PR #1944, opened 2021-05-02
PEP 660: Editable installs for PEP 517 style build backends

sbidoul:

PR #1955, opened 2021-05-10
PEP 658: Static Distribution Metadata in the Simple Repository API

uranusjr:
Spawned from pypa/warehouse#8254. Missing fields to be filled later after a number is assigned.

### DIFF
--- a/pep-0658.rst
+++ b/pep-0658.rst
@@ -1,8 +1,8 @@
 PEP: 658
 Title: Static Distribution Metadata in the Simple Repository API
 Author: Tzu-ping Chung <uranusjr@gmail.com>
-Sponsor:
-PEP-Delegate:
+Sponsor: Brett Cannon <brett@python.org>
+PEP-Delegate: Donald Stufft <donald@stufft.io>
 Discussions-To: https://discuss.python.org/t/8651
 Status: Draft
 Type: Standards Track
--- a/pep-0658.rst
+++ b/pep-0658.rst
@@ -51,17 +51,18 @@ contains the necessary information for common use cases. The metadata
 served must be completely static, i.e. identical to the ``METADATA``
 file in the ``.dist-info`` directory [dist-info]_ if the distribution
 is installed. The repository can provide this for any distributions,
-but it is expected they will only provide them for wheels [wheel]_,
-since an sdist [sdist]_ does not currently have a way to promise the
-metadata will stay the same after it is built.
+but it is expected they will only provide them for wheels [wheel]_
+at the current time, since an sdist [sdist]_ does not yet have a way
+to promise the metadata will stay the same after it is built.
 
 Since not all distributions have static metadata, an HTML attribute
 on the distribution file's anchor link is needed to indicate whether a
 client is able to choose the separately served metadata file instead.
-The attribute can also be used denote whether the metadata file can be
-downloaded. If the attribute is missing from an anchor link, static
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
You can take PEP 658; #1944 will need to be renumbered anyway.

pfmoore:
This looks broadly good to me. I spotted a couple of typos, I'll do a proper review when I have more time.

pep-0658.rst
```diff
@@ -0,0 +1,178 @@
...
+Content-Type: text/x-rst
+Created: 10-May-2021
+Post-History: 10-May-2021
+Resolution:
+
+
+Abstract
+========
+
+This PEP proposes adding anchor tag to expose the ``METADATA`` file
```
JelleZijlstra, line 18:
```suggestion
This PEP proposes adding an anchor tag to expose the ``METADATA`` file
```

pep-0658.rst
```diff
@@ -0,0 +1,178 @@
...
+
+
+Motivation
+==========
+
+Package management workflows made popular by recent tooling increase
+the need to inspect distribution metadata without intending to install
+the distribution, and download multiple distributions of a project to
+choose from based on their metadata. This means they end up discarding
+much downloaded data, which is inefficient and results in bad user
```
JelleZijlstra, line 31:
```suggestion
much downloaded data, which is inefficient and results in a bad user
```

uranusjr:
Thanks, I’ve assigned the number and added the newly-created Discourse link to the document. I admit I didn’t re-read the document myself after writing it 😓

pep-0658.rst
```diff
@@ -0,0 +1,178 @@
+PEP: 658
+Title: Static Distribution Metadata in the Simple Repository API
+Author: Tzu-ping Chung <uranusjr@gmail.com>
+Sponsor:
```
pradyunsg, line 4: @pfmoore here, and in code owners?

uranusjr: I’m not sure whether Paul or Donald is better to put in which fields TBH

pfmoore: As this is package index stuff, @dstufft is PEP-delegate by default. He can get someone else to handle it if he prefers, and I'd be willing to do it if he wants to pass the job on.

I'd rather not be sponsor, TBH. With the new CODEOWNERS workflow, it looks like that means I'd now be responsible for merging any PRs to the PEP and I don't really have the time to handle that right now...

JelleZijlstra: > With the new CODEOWNERS workflow, it looks like that means I'd now be responsible for merging any PRs to the PEP

I'm happy to remain as chief button-pusher to merge uncontroversial PRs :). I think the main point of the CODEOWNERS change is to make it easier for sponsors to stay up to date on changes to their PEPs, but they don't necessarily have to take care of all changes.

pradyunsg: Donald is gonna be the delegate IIUC, based on the SC's standing delegation for PyPI.

Sponsor is... well, whoever ends up sponsoring this.

JelleZijlstra (approved):

uranusjr:
The sponsor has been set to @brettcannon ([ref](https://discuss.python.org/t/8651/18)) and the PEP delegate @dstufft ([as the standing delegate](https://github.com/python/steering-council/blob/main/process/standing-delegations.md#pypa-delegations)).

I also made some changes to the proposal according to early feedback.

PR #1967, opened 2021-05-22
small change

iritkatriel:


Rosuav:
You're suggesting a change to a branch that isn't the repository's primary branch. I'd advise negotiating with the PEP author, as there's no guarantee that the branch you're looking at is the primary copy of that document.

PR #1977, opened 2021-06-03
PEP 662: Editable installs via virtual wheels

gaborbernat:
Signed-off-by: Bernát Gábor <gaborjbernat@gmail.com>

### DIFF
--- a/pep-0662.rst
+++ b/pep-0662.rst
@@ -1,7 +1,7 @@
-PEP: 9999
-Title: Editable installs
+PEP: 662
+Title: Editable installs via virtual wheels
 Author: Bernát Gábor <gaborjbernat@gmail.com>
-Sponsor: TBD
+Sponsor: Brett Cannon <brett@python.org>
 Discussions-To: https://discuss.python.org/t/discuss-tbd-editable-installs-by-gaborbernat/9071
 Status: Draft
 Type: Standards Track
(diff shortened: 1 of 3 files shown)


brettcannon:
Would you mind word wrapping this? See https://www.python.org/dev/peps/pep-0012/#general for guidance.

gaborbernat:
Uh, ok doing that makes a PITA to make amendments to the PEP because you have to reindent multiple lines manually after every edit. Ideally we should do that just before merging not?

brettcannon:
If you're explicitly asking for the rest of the PEP to be reviewed before reflowing that's fine, but the reflow will need to happen before the PEP can be accepted.

JelleZijlstra:
I don't think every edit needs to trigger a complete reflow. If you need to add a few words to the middle of a paragraph later, you can just have a shorter line in the middle and leave the rest of the paragraph alone. The thing to aim for is that every line stays within the length limit.

pep-9999.rst
```diff
@@ -0,0 +1,205 @@
+PEP: 9999
+Title: Editable installs
+Author: Bernát Gábor <gaborjbernat@gmail.com>
+Sponsor: TBD
+Discussions-To: https://discuss.python.org/t/draft-pep-editable-installs-for-pep-517-style-build-backends/8510
```
pradyunsg, line 5: Use the topic about this specific PEP draft, rather than PEP 660's topic?

```suggestion
Discussions-To: https://discuss.python.org/t/9071
```

pep-9999.rst
```diff
@@ -0,0 +1,340 @@
...
+
+   {
+     "metadata_for_build_editable": "<dir to dist-info>",
+     "purelib": {"<project dir>": "<project dir>"}
+   }
+
+The frontend then could either:
+
+-  Add the source directory onto the target interpreters ``sys.path`` during
+   startup of it. This is done by creating a ``pth`` file into the target
```
brettcannon, line 244:
```suggestion
   startup of the interpreter. This is done by creating a ``pth`` file into the target
```

pep-9999.rst
```diff
@@ -0,0 +1,340 @@
...
+   interpreters ``purelib`` folder. setuptools_ does this today and is what `pip
+   install -e <project_directory>`_ translate too. This solution is fast and
```
brettcannon, line 246:
```suggestion
   install -e <project_directory>`_ translates to. This solution is fast and
```

brettcannon (changes requested):
LGTM overall, but I want to double-check that the `:pep:` directives actually turn into hyperlinks as I didn't think they did thanks to us not using sphinx.

brettcannon (changes requested):
LGTM! Can you add me as the sponsor, take the next PEP number (662), and add me to `.github/CODEOWNERS`? I will then approve and merge this.

gaborbernat:
Sure, I'll ask another day or two to do some clarifications that came up on the discussion topic 🙏 thanks!

gaborbernat:
> LGTM! Can you add me as the sponsor, take the next PEP number (662), and add me to `.github/CODEOWNERS`? I will then approve and merge this.

All done now 👍 thanks!

pradyunsg:
You'd wanna retitle the PR. :)

pep-0662.rst
```diff
@@ -0,0 +1,387 @@
...
+       """protocol version of the editable metadata, this PEP defines version 1"""
+
+       metadata_for_build_editable: str
+       """distribution information of the package as defined by PEP-491"""
+
+       paths: SchemaPaths
+       """files to expose into the target interpreter"""
+
+
+The schema paths map from project source absolute paths to target directory
```
layday, line 201: Nit: should be "scheme", not "schema".

brettcannon (approved):

PR #1978, opened 2021-06-05
Updates to PEP 660

sbidoul:
A few updates to PEP 660 following ongoing discussions.

### DIFF
--- a/pep-0660.rst
+++ b/pep-0660.rst
@@ -88,16 +88,8 @@ build_wheel_for_editable
 
 :: 
 
-  def build_wheel_for_editable(
-      wheel_directory,
-      scheme=scheme, 
-      config_settings=None):
+  def build_wheel_for_editable(wheel_directory, config_settings=None):
       ...
-
-``scheme``: a dictionary of installation categories ``{ 'purelib':
-'/home/py/.../site-packages', 'platlib': '...'}``. This makes it possible to
-use relative paths to the source code, which might help the interpreter find
-the package after the root path changes with ``chroot`` or similar.
 
 Must build a ``.whl`` file, and place it in the specified ``wheel_directory``.
 It must return the basename (not the full path) of the .whl file it creates, as
--- a/pep-0660.rst
+++ b/pep-0660.rst
@@ -160,13 +160,14 @@ install. This section provides examples and is not normative.
   can achieve a higher level of fidelity than path-based methods.
 * Symbolic links are another useful mechanism to realize editable installs.
   Since, at the time this writing, the ``wheel`` specification does not support
-  symbolic links, they are not directly usable. It is however possible for the
-  backend to create an symlink structure in a build directory of the source
-  tree such as ``.editable/$COMPATIBILITY_TAGS``, and add that directory to the
-  python path via a ``.pth`` file in the "editable" wheel. Alternatively it is
-  also possible to defer the creation of symbolic links to an import hook or
-  executable ``.pth`` that the backend adds in the "editable" wheel (although
-  this approach requires special care to keep the RECORDS file up-to-date).
+  symbolic links, they are not directly usable to set-up symbolic links in the
+  target environment. It is however possible for the backend to create a
+  symlink structure in some ``build`` directory of the source tree, and add
+  that directory to the python path via a ``.pth`` file in the "editable"
+  wheel. If some files linked in this manner depend on python implementation or
+  version, ABI or platform, care must be taken to generate the link structure
+  in different directories depending on compatibility tags, so the same project
+  tree can be installed in editable mode in multiple environments.
 
 Frontend requirements
 ---------------------
--- a/pep-0660.rst
+++ b/pep-0660.rst
@@ -1,5 +1,5 @@
 PEP: 660
-Title: Editable installs for PEP 517 style build backends
+Title: Editable installs for pyproject.toml based builds
 Author: Daniel Holth <dholth@gmail.com>, Stéphane Bidoul <stephane.bidoul@gmail.com>
 Sponsor: Paul Moore <p.f.moore@gmail.com>
 Discussions-To: https://discuss.python.org/t/draft-pep-editable-installs-for-pep-517-style-build-backends/8510
--- a/pep-0660.rst
+++ b/pep-0660.rst
@@ -200,6 +200,16 @@ Frontends must not expose the wheel obtained from ``build_wheel_for_editable``
 to end users. The wheel must be discarded after installation and must not be
 cached nor distributed.
 
+Limitations
+===========
+
+This PEP does not support making the content of the wheel .data directory
+"editable" (scripts, C header files, and other data files).
+
+Scripts in the .data directory are therefore not supported by this PEP either,
+and package authors are encouraged to migrate to ``console_scripts`` entry
+points, or make their scripts tiny wrapper around library functionality.
+
 Rejected ideas
 ==============
 
--- a/pep-0660.rst
+++ b/pep-0660.rst
@@ -165,7 +165,8 @@ install. This section provides examples and is not normative.
   tree such as ``.editable/$COMPATIBILITY_TAGS``, and add that directory to the
   python path via a ``.pth`` file in the "editable" wheel. Alternatively it is
   also possible to defer the creation of symbolic links to an import hook or
-  executable ``.pth`` that the backend adds in the "editable" wheel.
+  executable ``.pth`` that the backend adds in the "editable" wheel (although
+  this approach requires special care to keep the RECORDS file up-to-date).
 
 Frontend requirements
 ---------------------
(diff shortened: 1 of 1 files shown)


pep-0660.rst
```diff
@@ -189,12 +195,44 @@ cached nor distributed.
...
 
+Virtual wheel
+-------------
+
+`Another approach was proposed <https://github.com/pypa/pip/pull/8215>`_, where
+the build backend returns a mapping from source files and directories to the
+installed layout. It is then up to the installer frontend to realize the
+editable installation by whatever mean it deems adequate for its users.
+
+In terms of capabilities, both proposals seem equivalent, at least in
```
gaborbernat, line 215: If you're going with this bold statement do clarify how you're planning to support headers, data files and scripts.

sbidoul: Why should I? The whole point is that we don't know (in either proposal) and can let backend authors experiment. 

Regarding headers, for instance, assuming there is a use case to install them in editable mode, one can use symlinks. But is there such a use case ? This PEP explicitly says only python files must be "edtiables" and changes to other file may require an additional installation step.

pep-0660.rst
```diff
@@ -157,11 +158,16 @@ install. This section provides examples and is not normative.
   a path importable, often including the project's own ``setup.py`` and other
   scripts that would not be part of a normal installation. The proxy strategy
   can achieve a higher level of fidelity than path-based methods.
+* Symbolic links are another useful mechanism to realize editable installs.
+  Since, at the time this writing, the ``wheel`` specification does not support
+  symbolic links, they are not directly usable. It is however possible to defer
```
gaborbernat, line 163: This seems very magical, and like a hack to me. As an end-user would definitely be surprised to discover this unexpected side-effect when first running my application. Also, I think this would run into issues with resource files where the user is not using importlib.resources but pkg_resources 🤔 Because then you might need to trigger the symlink before a module alongside is actually imported, not?

sbidoul: Sure it's a hack. With an executable .pth it works reliably though. To be honest and blunt, all concepts we are discussing for editables sound like hacks to me (except a .pth that exposes a whole directory).

pep-0660.rst
gaborbernat: What about scripts? Those can be python files.

> Regarding headers, for instance, assuming there is a use case to install them in editable mode, one can use symlinks. But is there such a use case?

Check astropy project that ships headers. The wheel spec allows it so there is a such use case.

> Why should I? The whole point is that we don't know (in either proposal) and can let backend authors experiment.

We should at least hypothetically provide a way to achieve this, just to prove that the PEP allows that functionality.

pep-0660.rst
gaborbernat: The backend generating files into the target interpreter post-installation seems like a nightmare for a frontend. All of a sudden has no idea what files it needs to uninstall. This would be a no go for me as a frontend developer. If anything I'd prefer to explicitly ban this in the PEP to provide some sanity for the frontend developers.

sbidoul: I have added another approach to handle symbolic links.

sbidoul:
@dholth have you seen this one ? Anything you'd like to change or add here ?

pep-0660.rst
dholth: https://github.com/pypa/pip/pull/8215 doesn't seem related to this PEP.

We also rejected my prototype that uses the wheel structure but skips creating an archive. The archive is easy to produce; no one has to worry about cleaning a build directory (full of the unpacked wheel); we avoid describing exactly what an unpacked wheel is.

I think packages that link against other packages that provide headers are used to being able to find those headers in source checkouts, as well as in the installed location? I don't think the way installed headers work in Python is particularly great.

Most scripts are entry_points scripts handled by the standard wheel install.

`setup.py develop` is about making your code importable by other code, and providing your metadata so that your editable distribution is "installed" according to "pip list".

But old-fashioned scripts are standalone. They need to import library code, but almost never need to be "imported by" dependents. There is less pressure to copy them to the central scripts location during an editable install. If you really are editing, you may run those scripts directly from the checkout. They will work fine.

Data files are almost always relative to the source code (part of purelib/platlib) since unfortunately the data_files feature has been broken for a long time. It does not install into a predictable location given varying virtualenv / --user / global install targets.

This is a long way to say that we think the most important use case is covered without worrying about scripts, headers, and data. It's hard to manually do an in-place build with correct .dist-info metadata on the .dist-info search path. `setup.py develop` or a replacement editable system solves that.

Could line 218 be reworded to say something more like "both proposals could provide the core 'editable' feature".

My biggest concern with the virtual wheel proposal is that build systems do not work by generating the necessary mapping as an intermediate step. Providing it to the hook could be burdensome and restrict the design of 'editable-capable' build systems. And that while it's pretty obvious to see how the simple case would work, it would still be the installer's responsibility to make sense of whatever crazy mapping was provided to it by the backend - if the structure of the mapping did not resemble the structure of what was on disk?

In this PEP and in `setup.py develop` the publisher has an incentive to lay out their source code in a way that would (mostly) work with a normal `.pth` file.

pep-0660.rst
```diff
@@ -157,11 +158,19 @@ install. This section provides examples and is not normative.
...
   scripts that would not be part of a normal installation. The proxy strategy
   can achieve a higher level of fidelity than path-based methods.
+* Symbolic links are another useful mechanism to realize editable installs.
+  Since, at the time this writing, the ``wheel`` specification does not support
+  symbolic links, they are not directly usable. It is however possible for the
+  backend to create an symlink structure in a build directory of the source
+  tree such as ``.editable/$COMPATIBILITY_TAGS``, and add that directory to the
+  python path via a ``.pth`` file in the "editable" wheel. Alternatively it is
+  also possible to defer the creation of symbolic links to an import hook or
+  executable ``.pth`` that the backend adds in the "editable" wheel.
```
dholth, line 168: The symbolic links section is not very inspiring. We could own our lack of support for symlinks or finally add symlinks to wheel (or encourage someone to add them in the future).

I don't understand the $COMPATIBILITY_TAGS variable.

I wouldn't complain if the backend produced build/editable/ with one or two links to the top-level packages we export; add that to .pth. Solving the 'accidentally import setup.py' problem. I would be confused if a complete tree of links to all our .py was in there.

The deferred idea is best removed. It's possible but weird.

pep-0660.rst
```diff
@@ -114,11 +114,11 @@ Build-backends must produce wheels that have the same dependencies
 with the exception that they can add dependencies necessary for their editable
 mechanism to function at runtime (such as `editables`_).
 
-The filename for the “editable” wheel needs to be PEP 427 compliant too. It
+The filename for the "editable" wheel needs to be PEP 427 compliant too. It
```
dholth, line 117: Goodbye unicode quotes. I notice a lack of emoji in the PEP as well.

pep-0660.rst
```diff
@@ -189,12 +198,44 @@ cached nor distributed.
...
+identifier. In other words ``pkg==1.0+local`` is not satisfied by version
 ``1.0+local.editable``.
 
+Virtual wheel
+-------------
+
+`Another approach was proposed <https://github.com/pypa/pip/pull/8215>`_, where
+the build backend returns a mapping from source files and directories to the
+installed layout. It is then up to the installer frontend to realize the
+editable installation by whatever mean it deems adequate for its users.
```
dholth, line 216: means

sbidoul:
@dholth I have handled your comments, please have a look. 

To the best of my knowledge all feedback received in the discussion threads has been incorporated.

sbidoul:
I also removed the controversial scheme argument.

pep-0660.rst
```diff
@@ -186,15 +192,73 @@ Frontends must not expose the wheel obtained from ``build_wheel_for_editable``
 to end users. The wheel must be discarded after installation and must not be
 cached nor distributed.
 
+Limitations
+===========
+
+This PEP does not support making the content of the wheel .data directory
+"editable" (scripts, C header files, and other data files).
```
dholth, line 199: How about

This PEP focuses on making the ``purelib|platlib`` categories (installed into ``site-packages``) "editable".  It does not make special provision for the other categories ``headers|scripts|data``.  Package authors are encouraged to use ``console_scripts``, make their scripts tiny wrappers around library functionality, or manage these from the source checkout during development.

Because:

`distribution-1.0.data/(purelib|platlib|headers|scripts|data)`.

`(purelib|platlib|headers|scripts|data)` can all appear in the `.data` directories. These are called `install paths` or `categories`.

sbidoul: Thanks. Update done.

sbidoul:
@pfmoore do you think it would be useful to add links to know prototype in the PEP ?

Other than that this should be ready to merge.

pfmoore:
I think that would be useful, yes.

pfmoore (approved):

pfmoore:
Would you prefer to update this PR to add the links when you have them, or do them as a separate PR? Either way works for me.

sbidoul:
@pfmoore feel free to merge. I'll do another one when I have collected all the links if this one is merged.

PR #1995, opened 2021-06-19
Sphinx - epub

AA-Turner:
A stab at ePub support (xref #1994).

Previewing on a desktop app (adobe digital editions) everything looks fine apart from code blocks (e.g. page 66), so will have a look at overriding the CSS.

Test: `make sphinx-epub` and look at `build/PEPs.epub`. This is the one massive file for all PEPs approach, which is the default from the Sphinx ePub 3 writer -- if a per-PEP solution were wanted, I imagine this may be tricker.

(Note that this builds on the sphinx-theming branch -- the relevant commits are the last three (387b19b...37d9975), or changes in 
`build.py`, `contents-epub.rst`, and `Makefile`)

cc: @froi, @Mariatta

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,3 +79,6 @@ fail-warning:
 
 check-links:
 	$(SPHINX_BUILD) --check-links
+
+sphinx-epub:
+	$(SPHINX_BUILD) --build-epub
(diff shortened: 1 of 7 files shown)


AA-Turner:
Closing this, there've been no updates in a while. Perhaps the best way of doing this if epub support were wanted is to generate one per PEP. Happy to review something if this is revived but likely won't advance this myself.

PR #1996, opened 2021-06-21
Only restrict to deploying on pushes to `master` in the python/peps repo

AA-Turner:
cc: @pablogsal 

Relaxes restriction from #1987, builds on PRs from forks.

### DIFF
--- a/.github/workflows/deploy-gh-pages.yml
+++ b/.github/workflows/deploy-gh-pages.yml
@@ -1,9 +1,6 @@
 name: Deploy to GitHub Pages
 
-on:
-  push:
-    branches: [main]
-  workflow_dispatch:
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   deploy-to-pages:
@@ -34,6 +31,8 @@ jobs:
         run: rm -r build/.doctrees/
 
       - name: 🚀 Deploy to GitHub pages
+        # This allows CI to build branches for testing
+        if: github.ref == 'refs/heads/main'
         uses: JamesIves/github-pages-deploy-action@4.1.1
         with:
           branch: gh-pages # The branch to deploy to.


.github/workflows/deploy-gh-pages.yaml
```diff
@@ -3,11 +3,20 @@ name: Deploy to GitHub Pages
 on:
   push:
     branches: [master]
+  pull_request:
 
 jobs:
   deploy-to-pages:
     runs-on: ubuntu-latest
 
+    # If on the python/peps repo, only deploy on pushes to `master`. Otherwise, build
```
pablogsal, line 12: I am not a fan of this, unfortunately. I don't like all these hardcoded values. If users want to deploy from forks they can maybe do it from other branches

AA-Turner: It is hardcoded, but shouldn't change? (barring branch renames of course)

Can have a think of other ways to do it but this was the cleanest I came up with last night to restrict deploys to master on this repo, whilst being liberal on forks (only had the time to test just now)

AA-Turner: Also noting the `on:` target will only deploy on master currently for this repo and forks.

pablogsal: > whilst being liberal on forks (only had the time to test just now)

But forks can do whatever they want on their master branch, can't they?

AA-Turner: True, perhaps I'm overthinking this in trying to make it easy (and people have coped up until now without rendered previews I guess). 

This is less important than the theming work in 1933 though anyway, and was a (perhaps poor) attempt at a quality of life improvement :)

hugovk: In general, I would like my fork to behave as closely as possible to upstream, so I have a better idea of the effect of my changes would have on upstream before opening a PR.

So it would be nice to have fork deploys too, as long as they work out of the box without needing to set any deploy keys, and don't fail my CI.

Some projects use for example secret credentials to deploy to S3, which would fail for forks and would need skipping. Not the case with gh-pages?

AA-Turner: Yeah, the idea here is that it works out of the box with the github integration - no external services / arcanery are neeeded.

E.g. I've been deploying previews all the way through the sphinx integration on https://aa-turner.github.io/peps/ using something similar to what I proposed here (but more hacky). 

Three caveats I can think of: 
- (a) I don't know if when GH Pages is enabled on this repo, if it will pull through automatically to forks (otherwise people would need to go to the pages tab of repo settings to enable)
- (b) Only one PR would preview at any one time, and it would be the latest pushed
- (c) Unlike e.g. the netifly bot, there is currently no nice automatic comment to point out the preview deploy (I think that this could be done though via the same actions script -- haven't tried)

As I said above, this is a much more trivial píece of work than the final Sphinx PR, and is definitely non-core work in the Sphinx transition process.

A

AA-Turner:
Hijacking this PR to test Sphinx on PRs, but not deploy.

AA-Turner:
Forgot to update the PR title (not the end of the world but mildly annoying) -- as shown here the Sphinx job now follows the `build` job more closely and runs on PRs, but skips the deploy step

A

PR #2001, opened 2021-06-24
PEP 634: Clarify late-registration edge-case

brandtbucher:
...as requested by @markshannon in https://github.com/python/cpython/pull/26864.

Also, add myself as a CODEOWNER of PEPs 635 and 636.


brettcannon (approved):

gvanrossum (commented):
Nice corner case! LGTM.

PR #2008, opened 2021-06-28
PEP-662: C'est la vie

gaborbernat:
https://discuss.python.org/t/pronouncement-on-peps-660-and-662-editable-installs/9450/1


pfmoore (changes requested):
Can you add a "Resolution" header linking to the resolution posting as well, please (see [here](https://www.python.org/dev/peps/pep-0001/#pep-header-preamble))?

pfmoore (approved):

brettcannon (approved):

PR #2022, opened 2021-07-04
PEP 657: use 0-indexed offsets

isidentical:
For the background of this change, see the discussion on: https://github.com/python/cpython/pull/26958#issuecomment-873480427

### DIFF
--- a/pep-0657.rst
+++ b/pep-0657.rst
@@ -218,12 +218,10 @@ and ``-1`` for the ``PyCode_Addr2Location`` API. The availability of the informa
 on whether the offsets fall under the range, as well as the runtime flags for the interpreter
 configuration.
 
-Although the AST nodes use ``int`` types to store these values, ``uint8_t`` types will be used
-for storage in the new map to minimize storage impact. This decision allows offsets to go from
-0 to 255, while offsets bigger than these values will be treated as missing (value of 0).
-We believe this is an acceptable compromise as line lengths in Python tend to
-be much lower than this limit (a query of the top 100 packages in PyPI shows
-that less than 0.01% of lines were longer than 255 characters).
+The AST nodes use ``int`` types to store these values. The current implementation, however,
+utilizes ``uint8_t`` types as an implementation detail to minimize storage impact. This decision
+allows offsets to go from 0 to 255, while offsets bigger than these values will be treated as
+missing (value of 0).
 
 As specified previously, the underlying storage of the offsets should be
 considered an implementation detail, as the public APIs to obtain this values
--- a/pep-0657.rst
+++ b/pep-0657.rst
@@ -221,7 +221,7 @@ configuration.
 The AST nodes use ``int`` types to store these values. The current implementation, however,
 utilizes ``uint8_t`` types as an implementation detail to minimize storage impact. This decision
 allows offsets to go from 0 to 255, while offsets bigger than these values will be treated as
-missing (value of 0).
+missing (returning ``-1`` on the ``PyCode_Addr2Location`` and ``None`` API in the ``co_positions()`` API).
 
 As specified previously, the underlying storage of the offsets should be
 considered an implementation detail, as the public APIs to obtain this values
--- a/pep-0657.rst
+++ b/pep-0657.rst
@@ -210,11 +210,17 @@ Offset semantics
 ^^^^^^^^^^^^^^^^
 
 These offsets are propagated by the compiler from the ones stored currently in
-all AST nodes. They are 1-indexed and a value of 0 will mean that the
-information is not available. Although the AST nodes use ``int`` types to store
-these values, ``uint8_t`` types will be used for storage in the new map to
-minimize storage impact. This decision allows offsets to go from 0 to 255,
-while offsets bigger than these values will be treated as missing (value of 0).
+all AST nodes. The output of the public APIs (``co_positions`` and ``PyCode_Addr2Location``)
+that deal with these attributes use 0-indexed offsets (just like the AST nodes), but the underlying
+implementation is free to represent the actual data in whatever form they choose to be most efficient.
+The error code regarding information not available is ``None`` for the ``co_positions()`` API,
+and ``-1`` for the ``PyCode_Addr2Location`` API. The availability of the information highly depends
+on whether the offsets fall under the range, as well as the runtime flags for the interpreter
+configuration.
+
+Although the AST nodes use ``int`` types to store these values, ``uint8_t`` types will be used
+for storage in the new map to minimize storage impact. This decision allows offsets to go from
+0 to 255, while offsets bigger than these values will be treated as missing (value of 0).
 We believe this is an acceptable compromise as line lengths in Python tend to
 be much lower than this limit (a query of the top 100 packages in PyPI shows
 that less than 0.01% of lines were longer than 255 characters).


isidentical:
CC: @ammaraskar

pep-0657.rst
```diff
@@ -210,11 +210,14 @@ Offset semantics
...
 
 These offsets are propagated by the compiler from the ones stored currently in
-all AST nodes. They are 1-indexed and a value of 0 will mean that the
-information is not available. Although the AST nodes use ``int`` types to store
-these values, ``uint8_t`` types will be used for storage in the new map to
-minimize storage impact. This decision allows offsets to go from 0 to 255,
-while offsets bigger than these values will be treated as missing (value of 0).
+all AST nodes. The output of the public APIs (``co_positions`` and ``PyCode_Addr2Location``)
+that deal with these attributes use 0-indexed offsets (just like the AST nodes) and designates
+``-1`` as the return code for information not available, but the underlying implementation is
```
isidentical, line 215: This is also more consistent with the `PyCode_Addr2Line`, since it also returns `-1` on error but for the sake of consistency we used to convert `-1` to `0` on `PyCode_Addr2Location`. Now that cast won't be needed.

pep-0657.rst
```diff
@@ -210,11 +210,17 @@ Offset semantics
...
+that deal with these attributes use 0-indexed offsets (just like the AST nodes), but the underlying
+implementation is free to represent the actual data in whatever form they choose to be most efficient.
+The error code regarding information not available is ``None`` for the ``co_positions()`` API,
+and ``-1`` for the ``PyCode_Addr2Location`` API. The availability of the information highly depends
+on whether the offsets fall under the range, as well as the runtime flags for the interpreter
+configuration.
+
+Although the AST nodes use ``int`` types to store these values, ``uint8_t`` types will be used
+for storage in the new map to minimize storage impact. This decision allows offsets to go from
+0 to 255, while offsets bigger than these values will be treated as missing (value of 0).
```
terryjreedy, line 223: If 0 is used as an actual offset for the first char of the line, it cannot also be used as missing.  255 could be used to mean 255 or more, to be turned into None.  Or am I missing something?

pablogsal: In the public API missing will return -1 in C and None in Python. The internal values are an implementation detail

pablogsal, line 223: Please, add a sentence mentioning that this is an implementation detail for the first version and that the design allows this to be lifted in the future

isidentical: See the second next paragraph. 

> As specified previously, the underlying storage of the offsets should be
considered an implementation detail, as the public APIs to obtain this values
will return either C ``int`` types or Python ``int`` objects, which allows to
implement better compression/encoding in the future if bigger ranges would need
to be supported.  This PEP proposes to start with this simpler version and
defer improvements to future work.

pablogsal: Yeah, I know, but reading this sentence is still a bit confusing because it seems that is dictating the format. We should clarify what is an implementation detail and what is a public API clearly to avoid confusion

isidentical: Any suggestion on phrasing this while not repeating that paragraph?

ammaraskar: Maybe something along the lines of:

> The AST nodes use ``int`` types to store these values. The current implementation, however, utilizes `uint8_t` types
> as an implementation detail to minimize storage impact. This decision allows offsets to go from
> 0 to 255, while offsets bigger than these values will be treated as missing (value of 0).

ammaraskar (approved):
Looks good with Pablo's suggestion incorporated.

pablogsal (approved):

pep-0657.rst
pablogsal: Yeah, this is great

pablogsal (approved):

pep-0657.rst
```diff
@@ -210,14 +210,18 @@ Offset semantics
...
+implementation is free to represent the actual data in whatever form they choose to be most efficient.
+The error code regarding information not available is ``None`` for the ``co_positions()`` API,
+and ``-1`` for the ``PyCode_Addr2Location`` API. The availability of the information highly depends
+on whether the offsets fall under the range, as well as the runtime flags for the interpreter
+configuration.
+
+The AST nodes use ``int`` types to store these values. The current implementation, however,
+utilizes ``uint8_t`` types as an implementation detail to minimize storage impact. This decision
+allows offsets to go from 0 to 255, while offsets bigger than these values will be treated as
+missing (value of 0).
```
pablogsal, line 224:
```suggestion
missing (returning ``-1`` on the ``PyCode_Addr2Location`` and ``None`` in ``co_positions()``).
```

terryjreedy:
The merged version looks good to me.

gvanrossum:
Whoa, don’t we use 1-based offsets in SyntaxError attributes?

ammaraskar:
There's a bit of discussion around that here: https://github.com/python/cpython/pull/26958#issuecomment-873480483

I think the consensus ended up being that `SyntaxError` is the odd one out here and we'd rather simplify the places we have the 1 to 0 conversion logic and stay consistent with the ast.

gvanrossum:
Hm, possibly, but we chose 1-based because editors like vim and emacs interpret column numbers as 1-based. Where else do we use 0-based offsets te refer to source locations? (In Python code. Of course in C code it’s 0-based all the way down. :-)

pablogsal:
> Where else do we use 0-based offsets te refer to source locations?

In the AST, all offsets are 0 based:

```
>>> import ast
>>> ast.dump(ast.parse("x"), include_attributes=True)
"Module(body=[Expr(value=Name(id='x', ctx=Load(), 
    lineno=1, col_offset=0, end_lineno=1, end_col_offset=1
), 
    lineno=1, col_offset=0, end_lineno=1, end_col_offset=1
)], type_ignores=[])"
```

so one of the advantages of 0-based is that every tool that currently deals with these can be used without modifications to deal with the new positions. We asked around some authors and seems that everyone frefers 0-based positions (so they can also do `line[offset:end_offset]` directly.

terryjreedy:
I not sure exact what "In Python code." is meant to say, but both Python and tk (and hence tkinter and IDLE) are 0-based for both indexing and slicing.

gvanrossum:
What I meant is when you print column offsets for end-user tools like editors to use, 1-based seems to be the norm. OTOH when you are using this to index into an array of lines, 0-based is more convenient and seems conventional. (Everybody agrees that line numbers start at 1 though. :-)

I'm not yet sure whether PEP 657 is more faced towards end users and their editors, or towards developers of Python libraries that will need to slice and dice the source code.

I do observe that in traceback.py the TraecbackException object has a documented offset field that's 1-based (since it is copied from SyntaxError).

I guess there will be a few traps no matter what we decide...

pablogsal:
> What I meant is when you print column offsets for end-user tools like editors to use, 1-based seems to be the norm. OTOH when you are using this to index into an array of lines, 0-based is more convenient and seems conventional. (Everybody agrees that line numbers start at 1 though. :-)

Oh, i see what you mean. Notice that we are not showing the numbers at the moment (and we don't plan to) so users will not see "error in column x". These numbers are mainly focused for tools to use them to grab the correct piece of source code that is associated with an instruction. If we ever show these numbers, we can easily make them 1-based only for the display.

> I guess there will be a few traps no matter what we decide...

Right, after asking around seems that there is an opportunity to avoid a slightly bigger subset of tracks for libraries consuming these numbers. So this decision would prioritize reusing existing code and may give some small surprises when developing new code. I think is a good balance but I totally understand if someone sees the other way :)

gvanrossum:
Okay, sounds good!

terryjreedy:
I see the essence of PEP 657 as making the full AST position info available on the code object as iterable co_positions.  bpo-43950 translates the position info, within the existing traceback.py traceback construction method, into a new caret line added under the code line. 

bpo-44569, just opened, will factor the per-frame formatting code into a separate method that users can replace.  Users can then handle the position info as they wish while customizing frame formatting, all without giving up other traceback features (like truncating tracebacks resulting from runaway recursion).  IDLE, for instance. will omit the caret line and instead include the position info so that Shell can highlight the error slice within the code line.

PR #2028, opened 2021-07-07
PEP 101: #python-dev channel is now on Libera Chat

jack1142:
Per the announcement on the python-dev mailing list:
https://mail.python.org/archives/list/python-dev@python.org/thread/FQW44V4DZFX4GAGOAMAZAN5KDQXCASDU/


gvanrossum (approved):
Is it likely to stay there for the coming years?

PR #2030, opened 2021-07-10
PEP 558: adjustments after updating implementation

ncoghlan:
* `pdb` stores `__return__` and `__exception__`` entries on
  arbitrary frames, so that feature needs to be preserved
* Some fast locals proxy operations will implicitly update the
  underlying shared mapping on the frame
* Anyone explicitly calling `LocalsToFast` is going to want the
  read/write proxy, not any of the read-only options (plus it's
  hard to fit an essay in an error message)
* Remove lingering reference to the removed `PyLocals_RefreshViews()`

### DIFF
--- a/pep-0558.rst
+++ b/pep-0558.rst
@@ -363,7 +363,10 @@ API:
 
 ``__getitem__`` operations on the proxy will populate the ``fast_refs`` mapping
 (if it is not already populated), and then either return the relevant value
-(if the key is found in the ``fast_refs`` mapping), or else raise ``KeyError``.
+(if the key is found in either the ``fast_refs`` mapping or the ``f_locals``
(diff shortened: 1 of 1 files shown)


ncoghlan:
@njsmith @markshannon The reference implementation is largely back to only needing "turn the crank" type work to flesh out the Python and C APIs - the core technical issues (including harmonisation with the 3.11 frame implementation changes) have all been resolved. There are some performance problems with the proxy due to the need to support extra keys to keep `pdb` happy, but consumers can mitigate those with a call to `f_locals.copy()` to get a true snapshot that won't redundantly refresh itself when it doesn't need to.

My current plan is to post this version to python-dev next week, requesting pronouncement from Nathaniel.

PR #2035, opened 2021-07-17
PEP 558: Update for fast locals proxy caching design changes

ncoghlan:
Further implementation work on the fast locals proxy resulted
in treating the "f_locals" frame storage as an implicitly
updated cache, moreso than treating it as a dynamic snapshot.


ncoghlan:
@markshannon @njsmith As an FYI, I'm planning to send this version to python-dev tomorrow (hence the added publication date)

PR #2038, opened 2021-07-18
PEP 558: Incorporate review comments from python-dev posting

ncoghlan:
* Propose a PyLocals_Kind enum rather than PyLocals_GetReturnsCopy
* Reference python-dev threads from early 2021
* Remove an outdated mention of possibly dropping support for storing
  extra f_locals keys in optimised scopes
* Make it more explicit that accessing frame.f_locals in Python
  refreshes the f_locals cache on that frame
* Note that read-only views on optimised frames have the same cache
  consistency limitations as the read/write proxy
* Fix miscellaneous typos

### DIFF
--- a/pep-0558.rst
+++ b/pep-0558.rst
@@ -25,8 +25,14 @@ presence or absence of tracing functions.
 In addition, it proposes that the following functions be added to the stable
 Python C API/ABI::
 
+    typedef enum {
+      PyLocals_UNDEFINED = -1,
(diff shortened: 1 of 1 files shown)


ncoghlan:
Ah, I should have put the in-progress label on this one - I was going to wait until @njsmith had provided comments on the posted version before publishing the update :)

No worries though, as I doubt replacing the clunky boolean query API with an enum is going to prove controversial.

PR #2039, opened 2021-07-20
PEP 646: Add section on grammar changes

mrahtz:
@gvanrossum @pradeep90 @fylux

### DIFF
--- a/pep-0646.rst
+++ b/pep-0646.rst
@@ -850,64 +850,80 @@ However, note that slices involving starred expressions are still invalid:
 ``*args`` as a TypeVarTuple
 ---------------------------
 
-The second change enables use of ``*args: *Ts``.
+The second change enables use of ``*args: *Ts`` in function definitions.
 
 Before:
 
 ::
 
-    star_etc[StarEtc*]:
-    | '*' a=param_no_default b=param_maybe_default* c=[kwds] {
-        _PyPegen_star_etc(p, a, b, c) }
-    | '*' ',' b=param_maybe_default+ c=[kwds] {
-        _PyPegen_star_etc(p, NULL, b, c) }
-    | a=kwds { _PyPegen_star_etc(p, NULL, NULL, a) }
+    star_etc:
+    | '*' param_no_default param_maybe_default* [kwds]
+    | '*' ',' param_maybe_default+ [kwds]
+    | kwds
 
 After:
 
 ::
 
-    star_etc[StarEtc*]:
-    | '*' a=param_no_default b=param_maybe_default* c=[kwds] {
-        _PyPegen_star_etc(p, a, b, c) }
-    | '*' a=param_no_default_star_annotation b=param_maybe_default* c=[kwds] {
-        _PyPegen_star_etc(p, a, b, c) }
-    | '*' ',' b=param_maybe_default+ c=[kwds] {
-        _PyPegen_star_etc(p, NULL, b, c) }
-    | a=kwds { _PyPegen_star_etc(p, NULL, NULL, a) }
+    star_etc:
+    | '*' param_no_default param_maybe_default* [kwds]
+    | '*' param_no_default_star_annotation param_maybe_default* [kwds]
+    | '*' ',' param_maybe_default+ [kwds]
+    | kwds
 
 Where:
 
 ::
 
-    param_no_default_star_annotation[arg_ty]:
-    | a=param_star_annotation ',' tc=TYPE_COMMENT? { _PyPegen_add_type_comment_to_arg(p, a, tc) }
-    | a=param_star_annotation tc=TYPE_COMMENT? &')' { _PyPegen_add_type_comment_to_arg(p, a, tc) }
+    param_no_default_star_annotation:
+    | param_star_annotation ',' TYPE_COMMENT?
+    | param_star_annotation TYPE_COMMENT? &')'
 
-    param_star_annotation[arg_ty]: a=NAME b=star_annotation { _PyAST_arg(a->v.Name.id, b, NULL, EXTRA) }
+    param_star_annotation: NAME star_annotation
 
-    star_annotation[expr_ty]: ':' a=star_expression { _PyAST_Tuple(_PyPegen_singleton_seq(p, a), Load, EXTRA) }
+    star_annotation: ':' star_expression
 
-This accomplishes the desired outcome (making ``*args: *Ts`` not be a syntax error)
-while matching the behaviour of a starred TypeVarTuple in an index: at runtime,
-``__iter__`` is called on the TypeVarTuple, and a tuple containing the resulting
-unpacked version of the TypeVarTuple is set as the type annotation for ``args``:
+This accomplishes the desired outcome (making ``*args: *Ts`` not be a syntax
+error) while matching the behaviour of star-unpacking in other contexts:
+at runtime, ``__iter__`` is called on the starred object, and a tuple
+containing the items of the resulting iterator is set as the type annotion
+for ``args``. In other words, ``*args: *foo`` is equivalent to
+``*args: tuple(foo)``.
 
 ::
 
     >>> Ts = TypeVarTuple('Ts')
     >>> def foo(*args: *Ts): pass  # Equivalent to `*args: tuple(Ts)`
     >>> foo.__annotations__
-    {'args': (UnpackedTs,)}  # Where UnpackedTs is Ts._unpacked
+    {'args': (*Ts,)}
+    # *Ts is the repr() of Ts._unpacked, an instance of UnpackedTypeVarTuple
 
-Note that the only scenario in which this grammar change allows ``*Ts`` to be used as a direct
-annotation (rather than being wrapped in e.g. ``Tuple[*Ts]``) is ``*args``. Other uses are still
-invalid:
+Note that the only scenario in which this grammar change allows ``*Ts`` to be
+used as a direct annotation (rather than being wrapped in e.g. ``Tuple[*Ts]``)
+is ``*args``. Other uses are still invalid:
 
 ::
 
     x: *Ts                 # Syntax error
     def foo(x: *Ts): pass  # Syntax error
+
+Implications
+''''''''''''
+
+As with the first grammar change, this change also has a number of side effects.
+In particular, the annotation of ``*args`` could be set to a starred object
+other than a ``TypeVarTuple`` - for example, the following nonsensical
+annotation is possible:
+
+::
+
+    >>> l = [1, 2, 3]
+    >>> def bar(*args: *l): pass  # Equivalent to `*args: tuple(l)`
+    >>> bar.__annotations__
+    {'args': (1, 2, 3)}
+
+Again, prevention of such annotations will need to be done by, say, static
+checkers, rather than at the level of syntax.
 
 Alternatives
 ------------
--- a/pep-0646.rst
+++ b/pep-0646.rst
@@ -808,7 +808,8 @@ Third, more than one instance of a star-unpack can occur within an index:
     array[*idxs_to_select, *idxs_to_select]
 
 The requirement that only one unpacked TypeVarTuple may be used per list of type parameters
-would therefore be implemented in e.g. ``typing.py`` rather than at the syntax level.
+would therefore be implemented only in type checking tools themselves rather than at the
+syntax level.
 
 Finally, note that star expressions within slices remain invalid:
 
@@ -858,12 +859,15 @@ Where:
 
     star_annotation[expr_ty]: ':' a=star_expression { _PyAST_Tuple(_PyPegen_singleton_seq(p, a), Load, EXTRA) }
 
-This accomplishes the desired outcome:
+This accomplishes the desired outcome (making ``*args: *Ts`` not be a syntax error)
+while matching the behaviour of a starred TypeVarTuple in an index: at runtime,
+``__iter__`` is called on the TypeVarTuple, and a tuple containing the resulting
+unpacked version of the TypeVarTuple is set as the type annotation for ``args``:
 
 ::
 
     >>> Ts = TypeVarTuple('Ts')
-    >>> def foo(*args: *Ts): pass
+    >>> def foo(*args: *Ts): pass  # Equivalent to `*args: tuple(iter(Ts))`
     >>> foo.__annotations__
     {'args': (UnpackedTs,)}  # Where UnpackedTs is Ts._unpacked
 
--- a/pep-0646.rst
+++ b/pep-0646.rst
@@ -753,9 +753,6 @@ within square brackets), necessary to support star-unpacking of TypeVarTuples:
     class Array(Generic[DType, *Shape]):
         ...
 
-(Note that ``from future import annotations`` does not help here, since no annotations
-are involved.)
-
 Before:
 
 ::
@@ -780,7 +777,7 @@ passed to ``__getitem__``. For example, if we do ``foo[a, *b, c]``, and
 
 To put it another way, note that ``x[..., *a, ...]`` produces the same result
 as ``x[(..., a*, ...)]``` (with any slices ``i:j`` in ``...`` replaced with
-``slice(i, j)``, with the one edge case that ``x[*a]`` becomes ``x[(*a,)]``.
+``slice(i, j)``, with the one edge case that ``x[*a]`` becomes ``x[(*a,)]``).
 
 TypeVarTuple Implementation
 '''''''''''''''''''''''''''
@@ -867,8 +864,7 @@ After:
 
     star_etc:
     | '*' param_no_default param_maybe_default* [kwds]
-    # This is the new alternative. Other alternatives are unchanged.
-    | '*' param_no_default_star_annotation param_maybe_default* [kwds]
+    | '*' param_no_default_star_annotation param_maybe_default* [kwds]  # New
     | '*' ',' param_maybe_default+ [kwds]
     | kwds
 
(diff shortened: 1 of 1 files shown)


pep-0646.rst
```diff
@@ -733,28 +733,159 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+    array[0, *idxs_to_select, -1]
+
+Third, more than one instance of a star-unpack can occur within an index:
+
+::
+
+    array[*idxs_to_select, *idxs_to_select]
+
+The requirement that only one unpacked TypeVarTuple may be used per list of type parameters
+would therefore be implemented in e.g. ``typing.py`` rather than at the syntax level.
```
JelleZijlstra, line 809: I'd suggest it should be implemented only by static type checkers, not necessarily by typing.py.

pep-0646.rst
```diff
@@ -733,28 +733,159 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+
+    param_star_annotation[arg_ty]: a=NAME b=star_annotation { _PyAST_arg(a->v.Name.id, b, NULL, EXTRA) }
+
+    star_annotation[expr_ty]: ':' a=star_expression { _PyAST_Tuple(_PyPegen_singleton_seq(p, a), Load, EXTRA) }
+
+This accomplishes the desired outcome:
+
+::
+
+    >>> def foo(*args: *b): pass
```
JelleZijlstra, line 863: So at runtime this is equivalent to `def foo(*args: tuple(b)): pass`? I think it would be good to say so explicitly.

pep-0646.rst
mrahtz: Right; in the future, when type annotations are not evaluation at runtime at all (maybe this is the case for 3.11 already, I've lost track of when this was supposed to happen?), this will only be validation by static type checkers. For the time being, though, when type annotations _are_ still being evaluated at runtime, it seems more consistent with runtime-validation of other typing-related stuff to also implement some checking in typing.py.

pep-0646.rst
mrahtz: Not quite, but I can see how this is confusing. I've added some extra detail - does this clarify?

pep-0646.rst
JelleZijlstra: It's still up in the air what will happen in 3.11; we may yet go with https://www.python.org/dev/peps/pep-0649/ .

Regardless, there are many invalid annotations that typing.py doesn't reject, and my general experience is that such checks tend to cause problems with future enhancements to the type system.

pep-0646.rst
JelleZijlstra: Thanks, but I think this still doesn't tell me what happens in terms of Python code execution. I would like to see an equivalent in terms of existing syntax; if it's not `tuple(bs)` what is it?

pep-0646.rst
mrahtz: Fair enough; I've amended this paragraph to state this should be checked by type checkers, and removed reference to typing.py.

pep-0646.rst
mrahtz: Ohhh, sorry - for some reason I interpreter `tuple(b)` as `(b,)`. Yes, `tuple(b)` is exactly right. Is this clear in the PEP now? (Sorry for the commit spam; a forced push didn't behave as I expected...)

JelleZijlstra (approved):
Thanks, this is clear to me now!

This is just my opinion and I'm not going to be the one making decisions here, but to me change (1) (* in subscripts) feels stronger than change (2) (* in *args annotations). (1) is a fairly natural extension of the grammar that could be independently useful. (2) is a special case for one very specific situation, so it is more ad hoc.

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
-   For example, ``Tuple[*Ts]`` and ``Tuple[T1, *Ts, T2]`` would both
-   be valid. (This PEP does not allow multiple unpacked ``TypeVarTuple``
-   instances to appear in a single parameter list, so ``Tuple[*Ts1, *Ts2]``
-   would be a runtime error. Also note that star expressions would *not*
-   be valid in slice expressions -  e.g. ``Tuple[*Ts:*Ts]`` is
-   nonsensical and should remain invalid.)
-2. We would need to make '``*args: *Ts``' valid in function definitions.
+There are three things to note about this grammar change. First, for consistency with
+star-unpacking in other contexts, ``_UnpackedTypeVarTuple`` would always be received
+by ``__class_getitem__`` as part of a tuple:
```
gvanrossum, line 779: To describe the semantics of this grammar, we need neither `_UnpackedTypeVarTuple` nor `__class_getitem__`. Rather, we need to describe what the instance method `__getitem__` will receive when the argument has various forms involving `*`.

Fortunately it seems to be that in all cases where the substitution is valid, `x[..., *a, ...]` produces the same as `x[(..., *a, ...)]`, with the one edge case that `x[*a]` becomes `x[(*a,)]`. Of course if the dots involve a slice, the substitution is not valid, but we can then replace slices with calls to `slice(...)` to make things valid, for example `x[1:2, *a]` will become `x[(slice(1, 2), *a)]`. And then we can just defer to the existing semantics for `*a` and for `x[y]`.

We probably will need to defend why we even allow `x[1:2, *a]`, since we don't need it for variadic types. Our case is somewhat weak here (and we can state that we'd be fine if the SC were to strike this), but to me it seems that it's the simplest grammar change. After all that's also why we chose to allow e.g. `x[*a, *b]`.

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+    C[*Ts, int]  # __class_getitem__ receives x=(Ts._unpacked, int)
+
+
+Second, the change would also enable star-unpacking of other structures within
+indexing operations:
+
+::
+
+    idxs_to_select = (1, 2, 3)
+    array[0, *idxs_to_select, -1]
```
gvanrossum, line 802: Yes it would, but PEP 637 was rejected because there weren't enough perceived benefits. And honestly the example doesn't make a lot of sense to me. (Is it equivalent to `array[0, 1, 2, 3, -1]`?)

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+
+Third, more than one instance of a star-unpack can occur within an index:
+
+::
+
+    array[*idxs_to_select, *idxs_to_select]
+
+The requirement that only one unpacked TypeVarTuple may be used per list of type parameters
+would therefore be implemented only in type checking tools themselves rather than at the
+syntax level.
```
gvanrossum, line 812: This should mention why we propose it, if we don't need it. As I said, it comes down to the simpler syntax, and consistency with `(*a, *b)`.

gvanrossum (commented):
(I'm running out of time to review the second half of the diff, here's what I have for the first half.)

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+    # Syntax error
+    array[*idxs_start:*idxs_end]
 
-If these grammar changes are considered too burdensome, we could instead
-simply use ``Unpack`` - though in this case it might be better for us to
-first decide whether there's a better option.
+``*args`` as a TypeVarTuple
+---------------------------
+
+The second change enables use of ``*args: *Ts``.
```
gvanrossum, line 824:
```suggestion
The second change enables use of ``*args: *Ts`` in function definitions.
```

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+
+Before:
```
gvanrossum, line 826: I suggest the same syntax cleanup.

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+
+After:
+
+::
+
+    star_etc[StarEtc*]:
+    | '*' a=param_no_default b=param_maybe_default* c=[kwds] {
+        _PyPegen_star_etc(p, a, b, c) }
+    | '*' a=param_no_default_star_annotation b=param_maybe_default* c=[kwds] {
+        _PyPegen_star_etc(p, a, b, c) }
```
gvanrossum, line 845: Maybe point out that this is the new alternative, whereas the two other alternatives are unchanged.

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+while matching the behaviour of a starred TypeVarTuple in an index: at runtime,
+``__iter__`` is called on the TypeVarTuple, and a tuple containing the resulting
+unpacked version of the TypeVarTuple is set as the type annotation for ``args``:
+
+::
+
+    >>> Ts = TypeVarTuple('Ts')
+    >>> def foo(*args: *Ts): pass  # Equivalent to `*args: tuple(Ts)`
+    >>> foo.__annotations__
+    {'args': (UnpackedTs,)}  # Where UnpackedTs is Ts._unpacked
```
gvanrossum, line 872: You don't define or explain `UnpackedTs`. How does it relate to `_UnpackedTypeVarTuple`? Is this literally what it will print?

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+
+    param_no_default_star_annotation[arg_ty]:
+    | a=param_star_annotation ',' tc=TYPE_COMMENT? { _PyPegen_add_type_comment_to_arg(p, a, tc) }
+    | a=param_star_annotation tc=TYPE_COMMENT? &')' { _PyPegen_add_type_comment_to_arg(p, a, tc) }
+
+    param_star_annotation[arg_ty]: a=NAME b=star_annotation { _PyAST_arg(a->v.Name.id, b, NULL, EXTRA) }
+
+    star_annotation[expr_ty]: ':' a=star_expression { _PyAST_Tuple(_PyPegen_singleton_seq(p, a), Load, EXTRA) }
+
+This accomplishes the desired outcome (making ``*args: *Ts`` not be a syntax error)
```
gvanrossum, line 862: I think you also need to explain the semantics when the thing after the star is not a TypeVarTuple. If I have
```
class X: ...
x = X()
def f(*args: *x): ...
print(f.__annotations__)
```
What will it print? What calls to methods of X are made?

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+    x: *Ts                 # Syntax error
+    def foo(x: *Ts): pass  # Syntax error
+
+Alternatives
+------------
+
+If these two grammar changes are considered too burdensome, there are two alternatives:
+
+1. **Support change 1 but not change 2**. Starred expressions within indexes are
+   more important to us than the ability to annotation ``*args``.
```
gvanrossum, line 889: Agreed.

gvanrossum (commented):
Here's the rest. Not too much new stuff.

pep-0646.rst
mrahtz: Great. Resolving.

pep-0646.rst
```diff
@@ -733,28 +733,166 @@ hopefully enable a thriving ecosystem of tools for analysing and verifying
...
+
+::
 
-In order to use the star operator for unpacking of ``TypeVarTuple`` instances,
-we would need to make two grammar changes:
+    class TypeVarTuple:
+        def __init__(self, name):
+            self._unpacked = _UnpackedTypeVarTuple(name) 
+        def __iter__(self):
+            yield self._unpacked
```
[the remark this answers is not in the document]
mrahtz: Ideally I think dynamic checkers _would_ be able to recognise `_UnpackedTypeVarTuple`. I've added what I imagine `_UnpackedTypeVarTuple` to look like (and removed the underscore, since in fact it's meant to be visible to things outside `typing.py`) - is this enough?

pep-0646.rst
mrahtz: Mm, great points. I've rewritten this section to explain these points (incorporating your wording, which I thought was very clear). What do you think?

pep-0646.rst
mrahtz: I've rewritten this section to make it clearer this is a side effect rather than something we necessarily endorse, and to clarify what this example is equivalent to - better?

pep-0646.rst
mrahtz: Done.

pep-0646.rst
mrahtz: Done.

pep-0646.rst
mrahtz: Have clarified this - better?

pep-0646.rst
mrahtz: Better?

pep-0646.rst
mrahtz: Done.

pep-0646.rst
mrahtz: Done.

mrahtz:
Thanks for the detailed comments, Guido - these are very helpful!

gvanrossum:
I’m on vacation for two weeks, will be sluggish to review again.

mrahtz:
Sure - enjoy!

gvanrossum:
Ugh, I lost track of this. Will review later today (after the Tensor Typing meeting).

pep-0646.rst
```diff
@@ -736,35 +736,45 @@ shape properties of numerical computing programs.
...
+
+::
+
+    DType = TypeVar('DType')
+    Shape = TypeVarTuple('Shape')
+    class Array(Generic[DType, *Shape]):
+        ...
+
+(Note that ``from future import annotations`` does not help here, since no annotations
+are involved.)
```
gvanrossum, line 757: This remark is a bit mysterious. Why would a reader make the mistake of thinking that would help? What might they think it helps for?

Also, it's `__annotations__`.

pep-0646.rst
```diff
@@ -850,65 +850,81 @@ However, note that slices involving starred expressions are still invalid:
...
-This accomplishes the desired outcome (making ``*args: *Ts`` not be a syntax error)
-while matching the behaviour of a starred TypeVarTuple in an index: at runtime,
-``__iter__`` is called on the TypeVarTuple, and a tuple containing the resulting
-unpacked version of the TypeVarTuple is set as the type annotation for ``args``:
+This accomplishes the desired outcome (making ``*args: *Ts`` not be a syntax
+error) while matching the behaviour of star-unpacking in other contexts:
+at runtime, ``__iter__`` is called on the starred object, and a tuple
+containing the items of the resulting iterator is set as the type annotion
+for ``args``. In other words, ``*args: *foo`` is equivalent to
+``*args: tuple(foo)``.
```
gvanrossum, line 892: I'd add "at runtime" somewhere in this last sentence, to make it absolutely clear that this is not an equivalence for static type checkers.

pep-0646.rst
```diff
@@ -850,65 +850,81 @@ However, note that slices involving starred expressions are still invalid:
...
+
+As with the first grammar change, this change also has a number of side effects.
+In particular, the annotation of ``*args`` could be set to a starred object
+other than a ``TypeVarTuple`` - for example, the following nonsensical
+annotation is possible:
+
+::
+
+    >>> l = [1, 2, 3]
+    >>> def bar(*args: *l): pass  # Equivalent to `*args: tuple(l)`
```
gvanrossum, line 922: I'd use a different variable name than `l`, which is easily confused with either `1` or `I` depending on the font.

pep-0646.rst
```diff
@@ -867,6 +867,7 @@ After:
 
     star_etc:
     | '*' param_no_default param_maybe_default* [kwds]
+    # This is the new alternative. Other alternatives are unchanged.
```
gvanrossum, line 870: With whole-line comments it's always a bit ambiguous where they point. Maybe just add `# New` to the end of the new alternative?

gvanrossum (commented):
Basically LGTM -- I only have very small nits left!

Thanks for your patience.

pep-0646.rst
pxeger: `from __future__ import annotations`

mrahtz: I added this based on one of your earlier comments:

> > The first grammar change enables use of star expressions in index operations (that is,
> > within square brackets), necessary to support star-unpacking of TypeVarTuples.
> 
> I think this should summarize the reason for this change here. The first Summary example shows it well:
> 
> >    class Array(Generic[DType, *Shape]):
> 
> That this is the very first example shows how fundamental this is. (And note that "from future import annotations" doesn't help in this position, since it is not technically an annotation.)

I thought you were gesturing towards the fact that with future annotations, new syntax is unnecessary for being able to do whatever manner of exotic things in `x: Tensor[A, B, C, D]` because of annotations not being evaluated at runtime - so a naive reader might think the same was true for everything else suggested in the PEP.

Anyway, I also thought it was a bit unclear in the end, so I'm happy to remove it.

pep-0646.rst
mrahtz: Done.

pep-0646.rst
mrahtz: Done.

pep-0646.rst
mrahtz: Done.

mrahtz:
Thanks, Guido!

gvanrossum (approved):
Yay! I'll merge this now, it's a very nice update.

Before we go back to the SC, let's make sure we check these boxes:
- [ ] Have we motivated sufficiently why we prefer the `*Ts` notation over `Unpack[Ts]`? (I think it's much more readable.)
- [ ] Have we collected enough endorsements from potential users? We should probably link to issues or messages (e.g. in the mailing list archives) in a brief section labeled "Endorsements" -- the  current SC seems to like such things. (We should also be transparent about the non-endorsement from PyTorch.)

PR #2045, opened 2021-08-02
Inline variance operators for generic type variables: ``+T`` and ``-T``

jorenham:
See https://github.com/python/typing/issues/813 for the discussion


JelleZijlstra:
Excited to see this! Per https://www.python.org/dev/peps/pep-0001/#submitting-a-pep, you'll need to find a core developer to sponsor this PEP. Core devs who are active in the typing area include @gvanrossum, @ilevkivskyi, and @ambv, but I don't know if any of them are interested in serving as a sponsor.

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+        def receive(event: T) -> None: ...
+
+
+
+Rationale
+=========
+
+Although defining the variance of generic type by using type variables
+that are constructed with the ``covariant=True`` or ``contravariant=True``
+keyword arguments, it has downsides.
```
JelleZijlstra, line 54: Incomplete sentence

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+
+
+Rationale
+=========
+
+Although defining the variance of generic type by using type variables
+that are constructed with the ``covariant=True`` or ``contravariant=True``
+keyword arguments, it has downsides. 
+
+This is why type checkers do no allow using co- or contravariant type
```
JelleZijlstra, line 56: do *not*.

Also, it's not clear what the "This is why" refers to.

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+variables outside of generic classes. 
+
+
+- In PEP 484 it states:
+
+    Covariance or contravariance is not a property of a type variable,
+    but a property of a generic class defined using this variable. 
+
+  Instead of defining variance on the type variable itself, it would
+  make more sense to mark the type variable as variant when defining
```
JelleZijlstra, line 66:
```suggestion
  make more sense to mark the type variable's variance when defining
```

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+  the generic type, and only there. 
+
+- Because variance is a property of the generic type itself, type 
+  checkers do no allow using co- or contravariant type variables
```
JelleZijlstra, line 70:
```suggestion
  checkers do not allow using co- or contravariant type variables 
```

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+    but a property of a generic class defined using this variable. 
+
+  Instead of defining variance on the type variable itself, it would
+  make more sense to mark the type variable as variant when defining
+  the generic type, and only there. 
+
+- Because variance is a property of the generic type itself, type 
+  checkers do no allow using co- or contravariant type variables 
+  outside of generic classes. But as the variance is currently a
+  property of the type variables, the must be used within variant
```
JelleZijlstra, line 72:
```suggestion
  property of the type variables, they must be used within
```

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+  generic classes.
+
+- Creating separate type variables for different variances often
```
JelleZijlstra, line 75: Not sure this is really common. It would be useful to have examples from open source projects where you see this duplication.

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Differences with current syntax
+-------------------------------
+
+The new typevar operators return a transparent wrapper around the 
+original type variable, which can be accessed with the `__orig__` 
+attribute on the returned wrapper. e.g.::
+
+    (+T).__orig__ is T
+    (+T).__covariant__ == not T.__covariant
```
JelleZijlstra, line 140:
```suggestion
    (+T).__covariant__ == not T.__covariant__
```
Although why not just say `(+T).__covariant__ is True`?

pep-9999.rst
```diff
@@ -0,0 +1,213 @@
...
+Thus, type variables defined with `covariant=True` and 
+`contravariant=True`, are not equivalent to `+T` and `-T`.
+
+
+`+T` and `-T` are not valid type annotations, and should only be used
+within the generic type parameter list of generic classes. e.g.:: 
+
+    class Spam(typing.Generic[+KT]): ...
+    class Eggs(typing.Protocol[KT, +VT]): ...
+    class OrderedSet(typing.AbstractSet[+T]): ...
```
JelleZijlstra, line 156: This is invalid according to the specification, which says you can only put it in Generic or Protocol. I was actually going to ask whether this should be allowed.

jorenham: Ah that's good to know. I can see how relaxing that restriction can be desirable, but I guess that that's out of scope here.

JelleZijlstra: Is it a relaxing? This syntax is definitely legal right now without the `+`. If you allow it with your new syntax, you'll have to decide what it means. For example, maybe it should be possible to have an invariant subclass of a covariant class.

JelleZijlstra: Actually, thinking about it a little more I believe you should allow this, because invariant subclasses of covariant classes are commonplace. My suggestion would be that you have to specify the `+` again for every subclass, and if there's no `+` the subclass is invariant.

For example, we have this in the standard library:
```
class AbstractSet(Generic[+T]): ...  # covariant

class set(Sequence[T]): ...  # invariant

class frozenset(Sequence[+T]): ... # covariant
```
To express that, we need what I suggested above.

Gobot1234: I also agree with Jelle here, I think it should be allowed anywhere the the old forms are allowed.
On another note typing.AbstractSet is deprecated in favour of collections.abc.Set, so it might be a good idea to change that.

JelleZijlstra: Another edge case is this:
```
class WhatIsMyVariance(Generic[+T], Sequence[-T]): ...
```
That should probably be disallowed, but this should be spelled out.

jorenham: I agree with your suggestion @JelleZijlstra . But I think that In your example `set` and `frozenset` also need to subclass `Generic`, i.e.

```python

class set(Sequence[T], Generic[T]): ...  # invariant

class frozenset(Sequence[T], Generic[+T]): ... # covariant
```

Then the restriction that `+T` and `-T` can only be used within the type parameter list of `Generic` and `Protocol` still holds. If you were to instead write `class frozenset(Sequence[+T], Generic[+T]): ...`, to me it looks as if `Sequence` itself is marked as covariant under `T`, whereas the covariance only applies to the specific definition of `frozenset`; the variance information does not need to be passed to the `Sequence` superclass. 
This also avoids the possibility for edge cases, like in `WhatIsMyVariance`.

pep-9999.rst
```diff
@@ -0,0 +1,211 @@
...
+
+    class SupportsPartialOrder(typing.Protocol[-T]):
+        def __le__(self, other: T): ...
+
+
+Differences with current syntax
+-------------------------------
+
+The new typevar operators return a transparent wrapper around the 
+original type variable, which can be accessed with the ``__orig__``
```
Gobot1234, line 134: Should this not be ``__origin__`` to match the other generic aliases for run time inspection?

jorenham: Ah, yes, good catch, that is what I was going for. I'll correct it.

pep-9999.rst
JelleZijlstra: That's not how things currently work, though: writing `class set(Sequence[T]):` is legal and accepted by typecheckers.

Rosuav:
Can you sort out matters of process (like getting a sponsor, a PEP number, and a build that passes), and then discuss details of the content of the PEP afterwards?

pep-9999.rst
jorenham: Ah I see, somehow I thought that in that case you couldn't do e.g. `spam: set[str]`, and that it required an explicit `Generic` as additional superclass. 

Anyway, I now indeed see how this behavior has to be explicitly defined. Requiring the same variance for the same type variable in the same generic class should do the trick. So no mixing; only allow e.g.
- `class LinkedList(Iterable[T], Container[T]): ... `
- `class LinkedList(Iterable[+T], Container[+T]): ... ` 
- `class LinkedList(Iterable[-T], Container[-T]): ... ` 

This could be checked within `typing.Generic.__init_subclass__`.

gvanrossum:
Do we have a sponsor yet?

jorenham:
@gvanrossum No not yet. Would you like to be the sponsor?

gvanrossum:
> @gvanrossum No not yet. Would you like to be the sponsor?

Hm, I am supporting this proposal but I don't really have the time to act as a good sponsor. According to [this section in PEP 1](https://www.python.org/dev/peps/pep-0001/#submitting-a-pep) a sponsor should also act as a mentor, and while I would like to, realistically I just don't have the time. If @JelleZijlstra wants to be a sponsor, we can ask the SC to approve this (and I think they would).

JelleZijlstra:
I'd be happy to do that!

If so my first recommendation would be to go through another round of review before formally submitting the PEP, since a few of the issues I brought up above are fairly large.

gvanrossum:
Great, let me know when you need me to do anything.

brettcannon:
Since it sounds like this PEP isn't ready to be merged I'm going to close it for now. Feel free to either re-open it or open a new one when the PEP is ready.

PR #2068, opened 2021-09-09
PEP 467: Drop the bchr builtin proposal

ncoghlan:
Also make it clearer that memoryview isn't an open issue for the PEP,
it's explicitly outside the PEP's scope.

### DIFF
--- a/pep-0467.txt
+++ b/pep-0467.txt
@@ -29,10 +29,6 @@ This PEP proposes five small adjustments to the APIs of the ``bytes`` and
 * Add ``bytes.getbyte`` and ``bytearray.getbyte`` byte retrieval methods
 * Add ``bytes.iterbytes`` and ``bytearray.iterbytes`` alternative iterators
 
-And one built-in::
-
-* ``bchr``
-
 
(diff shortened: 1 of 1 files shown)


pep-0467.txt
```diff
@@ -189,11 +174,28 @@ that ``bytes(x)`` (where ``x`` is an integer) should behave like the
...
+
 
-Open Issue: memoryview
-======================
+Scope limitation: memoryview
+----------------------------
 
-Updating ``memoryview`` with these new methods is outside the scope of this PEP.
+Updating ``memoryview`` with the new item retrieval methods is outside the scope
+of this PEP.
```
ncoghlan, line 198: This section looked weird to me when I was adding the new design discussion entry, so I tweaked it a bit:
- it's a deliberate scope limitation, not an open issue
- only the item retrieval methods (`getbyte()`, `iterbytes()`) potentially make sense for memory views, the alternate constructors would never be relevant

ethanfurman (approved):

ethanfurman (approved):

PR #2079, opened 2021-09-19
PEP 0000: Fix LGTM.com warning: Unnecessary 'else' clause in loop

DimitriPapadopoulos:
This '`while`' statement has a redundant '`else`' as no '`break`' is present in the body.
This '`for`' statement has a redundant '`else`' as no '`break`' is present in the body.

Fixes two of these LGTM.com warnings:
https://lgtm.com/projects/g/python/peps/?severity=warning

While the existing code is correct and runs without errors, I did find the `else:` doesn't help understand the logic of the code. But then this is mainly a question of tastes and habits, so feel free to disregard this issue.

### DIFF
--- a/pep0/pep.py
+++ b/pep0/pep.py
@@ -196,14 +196,13 @@ class PEP(object):
                                    pep_file.name)
         except StopIteration:
             raise PEPError("headers missing or out of order",
-                                pep_file.name)
+                           pep_file.name)
         required = False
         try:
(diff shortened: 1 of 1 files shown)


brettcannon:
Thanks for the PR, but `else` clauses can be there to clarify things as well, so I don't think changing things just to drop an `else` that is expected to always be hit is worth it.

Issue #2100, opened 2021-10-05
PEP 636: Minor typo: being->been 


raulcd:
While looking on the bug tracker I've found that someone opened an issue for a typo (https://bugs.python.org/issue45370)

There is a typo in https://www.python.org/dev/peps/pep-0636/#composing-patterns

The second sentence in the first paragraph : "[...] we have being doing that implicitly in the examples above."

When it should be "[...] we have been doing that implicitly in the examples above."

Opening an issue here and creating a PR for it.

PR #2101, opened 2021-10-05
PEP 636: Fix typo: being --> been

raulcd:
Seen originally at (https://bugs.python.org/issue45370).

Fixing typo `have being doing` to `have been doing`. 

Fixes https://github.com/python/peps/issues/2100


Rosuav:
LGTM.

PR #2110, opened 2021-10-17
PEP 663: Fix typo and creation date

jacobtylerwalls:
I found the creation date confusing at first, and was then motivated to fix when spotting the other typos.


Rosuav:
Interesting. My best guess is that the creation date was copied from PEP 435, so I'll let Ethan comment on whether that's deliberate (implying that this is reposting something that originated in that document) or irrelevant (in which case a date of Jul 1st or Jun 30th is correct - I presume you did the same thing I did and looked at the commit history).

The other changes are simple and clear. Thanks for creating the PR.

ethanfurman (approved):

ethanfurman:
Thanks!

PR #2123, opened 2021-10-23
PEP 671: Create a PEP for late-bound argument defaults

Rosuav:
Spinning up a first draft of a new proposal.


gvanrossum:
Did you want to co-author with Steven D’Aprano?

Rosuav:
Would be absolutely open to doing so. I'm just waiting for it to reach the web site and then I'll make the first post, and also invite coauthors. (It's been most of an hour - how often does the site get rebuilt?)

PR #2132, opened 2021-11-05
[PEP-654] add None check in subgroup example

iritkatriel:
Joe Gottman alerted us to an error in this example. Thanks Joe!


gvanrossum (approved):
I’d add the name of the reporter to the commit message body.

PR #2135, opened 2021-11-07
PEP 636: repeated patterns as enums

jsaponara:
In the previous section we are omitting a repeated pattern for brevity, so here is the perfect place to point out that commonly used patterns may be expressed most compactly as enums.  This is a technique that is not obvious from any other part of the tutorial.

I don't see any natural place to add it to the appendix, and would suggest leaving that unmodified.


jsaponara:
The needed tokenization really makes the change much less compelling.  What was I smoking?!?

PR #2138, opened 2021-11-08
PEP 3119: re: sets, "subclass/superclass" should be "subset/superset"

joseville1001:
re: sets, "subclass/superclass" should be "subset/superset" when referring to the inequality operations on sets.


gvanrossum (approved):
Thanks for your keen eye!

PR #2146, opened 2021-11-15
PEP 648: repeated word

DimitriPapadopoulos:


JelleZijlstra:
This one seems incorrect, it's saying the file is checked into version control in the reference implementation.

PR #2147, opened 2021-11-15
PEP 577: repeated word

DimitriPapadopoulos:


hugovk:
This repeated "that that" is correct and likely intentional.

Technical explanation: https://english.stackexchange.com/a/3426/9001

DimitriPapadopoulos:
Ah, it's "that framing" and not another one, OK.

JelleZijlstra:
Single "that" is also correct here though and seems more readable to me.

DimitriPapadopoulos:
Yes, I too find it more readable. However, since `that that` is correct too, perhaps the original author @ncoghlan could have a look at PEP 577 and decide if it's OK to change `that that` into `that`.

gvanrossum:
I think the original "that that" is correct here. Please don't believe everything your tools tell you.

DimitriPapadopoulos:
This PR is closed.

By now I think I have understood your point about tools (a variant of "A Foolish Consistency is the Hobgoblin of Little Minds"?) and I agree with you. This was a honest error. More of interest might be the reasons behind this error: certainly I did not work hard enough before submitting the PR, perhaps the initial text was convoluted and could be improved.

gvanrossum:
I surmise that another reason behind the error may be that mass changes to “fix” some common issue found using a tool are more risky than the perpetrators tend to think. (Decades ago, Eric Raymond added something new to Python and then fixed all places in the stdlib he could find where it could be applied. Except he also introduced some errors that weren’t discovered until much later — neither he nor the reviewers caught the mistakes. “Leave well enough alone” is a wise rule.)

DimitriPapadopoulos:
Because documentation is for humans and not executed by machines, improving it might have more benefits than "improving" code.

PR #2152, opened 2021-11-16
PEP 8103: Update with nomination results and initial Voter Roll

ewdurbin:


Yhg1s:
This isn't the final voter roll yet, correct? People still have a chance to realise they haven't indicated they are still active before we finalise the list and send out ballots? (If this is the final list, we should give MAL or someone else a chance to ping inactive core devs first.)

pablogsal:
> This isn't the final voter roll yet, correct? People still have a chance to realise they haven't indicated they are still active before we finalise the list and send out ballots? (If this is the final list, we should give MAL or someone else a chance to ping inactive core devs first.)

AFAIK those emails have been sent already and people are answering them, so this is not definite, but @ewdurbin can confirm.

ewdurbin:
That's correct this voter roll is not final, see the PR to [python/voters](https://github.com/python/voters/pull/59) for more info.

I have already begun notifying those who would be marked inactive.

Yhg1s:
Ah, yes, sorry, should've scrolled through the rest of my email first :) @ewdurbin do you want to keep the PR pending until you get all responses, or should we just merge this?

Yhg1s (approved):

ewdurbin:
@Yhg1s I say merge for now, handle all the updates in the voters repo, then send a final PR with the roll as it stands on 2021-11-30 before vote starts.

Yhg1s:
Merge now, let you sort out the rest. Sounds good to me :)

PR #2153, opened 2021-11-16
PEP 508: Fix grammar for 'not in' operator

DanielSanchez97:
Fix bug in grammar where the 'not in' operator for 'marker_op' was not properly being parsed. Add test to confirm behavior.


JelleZijlstra:
@dstufft could you review this? I'm not familiar with this grammar.

PR #2161, opened 2021-11-24
PEP 670: LTO+PGO benchmark

vstinner:
* Don't convert l-value macros
* Add "Examples of duplication of side effects" section
* Remove the gcc -O0 versus gcc -Og benchmark: it's not relevent for
  this PEP.


vstinner:
cc @erlend-aasland @encukou

erlend-aasland (approved):

PR #2167, opened 2021-11-30
PEP 675: Arbitrary Literal Strings

pradeep90:
+ Move in draft from the Google Doc.
+ Turn unicode to ASCII.

cc @gbleaney @JelleZijlstra

### DIFF
--- a/pep-0675.rst
+++ b/pep-0675.rst
@@ -1,0 +1,950 @@
+PEP: 675
+Title: Arbitrary Literal Strings
+Version: $Revision$
+Last-Modified: $Date$
+Author: Pradeep Kumar Srinivasan <gohanpra@gmail.com>, Graham Bleaney <gbleaney@gmail.com>
+Sponsor: Jelle Zijlstra <jelle.zijlstra@gmail.com>
+Discussions-To: Typing-Sig <typing-sig@python.org>
+Status: Draft
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 30-Nov-2021
+Python-Version: 3.11
+Post-History:
+
+Abstract
+========
+
+There is currently no way to specify that a function parameter can be
+of any literal string type; we have to specify the precise literal
+string, such as ``Literal["foo"]``. This PEP introduces a supertype of
+literal string types: ``Literal[str]``. This allows a function to
+accept arbitrary literal string types such as ``Literal["foo"]`` or
+``Literal["bar"]``.
+
+Motivation
+==========
+
+A common security vulnerability is for a program to include
+user-controlled data in a command it executes. For example, a naive
+way to look up a user record from a database is to accept a user id
+and insert it into a predefined SQL query:
+
+::
+
+    def query_user(conn: Connection, user_id: str) -> User:
+        query = f"SELECT * FROM data WHERE user_id = {user_id}"
+        conn.execute(query)
+
+    query_user(conn, "user123")  # OK.
+
+However, the user-controlled data ``user_id`` is being mixed with the
+SQL command string, which means a malicious user could run arbitrary
+SQL commands:
+
+::
+
+    # Delete the table.
+    query_user(conn, "user123; DROP TABLE data;")
+
+    # Fetch all users (since 1 = 1 is always true).
+    query_user(conn, "user123 OR 1 = 1")
+
+
+To prevent such SQL injection attacks, SQL APIs offer parameterized
+queries, which separate the executed query from user-controlled data
+and make it impossible to run arbitrary queries. For example, with
+`sqlite3 <https://docs.python.org/3/library/sqlite3.html>`_, our
+original function would be written safely as a query with parameters:
+
+::
+
+    def query_user(conn: Connection, user_id: str) -> User:
+        query = "SELECT * FROM data WHERE user_id = ?"
+        conn.execute(query, (user_id,))
+
+
+The problem is that there is no way to enforce this
+discipline. sqlite3's own `documentation
+<https://docs.python.org/3/library/sqlite3.html>`_ can only admonish
+the reader to not dynamically build the ``sql`` argument from external
+input; the API's authors cannot express that through the type
+system. Users can (and often do) still use a convenient f-string as
+before and leave their code vulnerable to SQL injection.
+
+Existing tools, such as the popular security linter `Bandit
+<https://github.com/PyCQA/bandit/blob/aac3f16f45648a7756727286ba8f8f0cf5e7d408/bandit/plugins/django_sql_injection.py#L102>`_,
+attempt to detect unsafe external data used in SQL APIs, by inspecting
+the AST or by other semantic pattern-matching. These tools, however,
+preclude common idioms like storing a large multi-line query in a
+variable before executing it, adding literal string modifiers to the
+query based on some conditions, or transforming the query string using
+a function. (We survey existing tools in the "Rejected Alternatives"
+section.) For example, many tools will detect a false positive issue
+in this benign snippet:
+
+
+::
+
+    def query_data(conn: Connection, user_id: str, limit: bool) -> None:
+        query = """
+            SELECT
+                user.name,
+                user.age
+            FROM data
+            WHERE user_id = ?
+        """
+        if limit:
+            query += " LIMIT 1"
+
+        conn.execute(query, (user_id,))
+
+We want to forbid harmful execution of user-controlled data while
+still allowing benign idioms like the above and not requiring extra
+user work.
+
+To meet this goal, we introduce the ``Literal[str]`` type, which only
+accepts string values that are known to be made of literals. This is a
+generalization of the ``Literal["foo"]`` type from `PEP 586
+<https://www.python.org/dev/peps/pep-0586/>`_. A string of type
+``Literal[str]`` cannot contain user-controlled data. Thus, any API
+that only accepts ``Literal[str]`` will be immune to injection
+vulnerabilities (with pragmatic `limitations <Appendix B:
+Limitations_>`_).
+
+Since we want the ``sqlite3`` ``execute`` method to disallow strings
+built with user input, we would make its `typeshed stub
+<https://github.com/python/typeshed/blob/1c88ceeee924ec6cfe05dd4865776b49fec299e6/stdlib/sqlite3/dbapi2.pyi#L153>`_
+accept a ``sql`` query that is of type ``Literal[str]``:
+
+::
+
+    def execute(self, sql: Literal[str], parameters: Iterable[str] = ...) -> Cursor: ...
+
+
+This successfully forbids our unsafe SQL example. The variable
+``query`` below is inferred to have type ``str``, since it is created
+from a format string using ``user_id``, and cannot be passed to
+``execute``:
+
+::
+
+    def query_user(conn: Connection, user_id: str) -> User:
+        query = f"SELECT * FROM data WHERE user_id = {user_id}"
+        conn.execute(query)
+        # Error: Expected Literal[str], got str.
+
+The method remains flexible enough to allow our more complicated
+example:
+
+::
+
+    def query_data(conn: Connection, user_id: str, limit: bool) -> None:
+        # This is a literal string.
+        query = """
+            SELECT
+                user.name,
+                user.age
+            FROM data
+            WHERE user_id = ?
+        """
+
+        if limit:
+            # Still has type Literal[str] because we added a literal string.
+            query += " LIMIT 1"
+
+        conn.execute(query, (user_id,))  # OK
+
+Notice that the user did not have to change their SQL code at all. The
+type checker was able to infer the literal string type and complain
+only in case of violations. The ``Literal[str]`` type is also useful
+in other cases where we want strict command-data separation, such as
+when building shell commands or when rendering a string into an HTML
+response without escaping (eg. via Django's ``mark_safe``
+function). Overall, this combination of strictness and flexibility
+makes it easy to enforce safer API usage in sensitive code without
+burdening users.
+
+Usage statistics
+----------------
+
+In a sample of open-source projects using ``sqlite3``, we found that
+``conn.execute`` was called `~67%
+<https://grep.app/search?q=conn%5C.execute%5C%28%5Cs%2A%5B%27%22%5D&regexp=true&filter[lang][0]=Python>`_
+of the time with a safe string literal and `~33%
+<https://grep.app/search?current=3&q=conn%5C.execute%5C%28%5Ba-zA-Z_%5D%2B%5C%29&regexp=true&filter[lang][0]=Python>`_
+of the time with an unsafe, dynamically-built local variable. Using
+this PEP's literal string type along with a type checker would have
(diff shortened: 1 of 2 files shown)


gvanrossum:
@JelleZijlstra When you land this, make sure the commit title is more descriptive than "Initial commit." :-)

JelleZijlstra:
I'll request some changes and then merge it as "subsequent commit".

Thanks for sending the PR! I'll review it later today and hopefully merge it.

pradeep90:
> @JelleZijlstra When you land this, make sure the commit title is more descriptive than "Initial commit." :-)

Hehe, changed the commit title.

pep-0675.rst
```diff
@@ -0,0 +1,954 @@
...
+checker already propagates types across variable assignments or
+function calls.
+
+In the current type system itself, if the SQL or shell command
+execution function only accepted three possible input strings, our job
+would be done. We would just say:
+
+::
+
+    def execute(query: Literal["foo"] | Literal["bar"] | Literal["baz"]) -> None: ...
```
JelleZijlstra, line 216:
```suggestion
    def execute(query: Literal["foo", "bar", "baz"]) -> None: ...
```
Use more idiomatic syntax

pep-0675.rst
```diff
@@ -0,0 +1,954 @@
...
+
+``Literal[str]`` is compatible with the type ``str``. It inherits all
+methods from ``str``. So, if we have a variable ``s`` of type
+``Literal[str]``, it is safe to write ``s.startswith("hello")``.
+
+Note that, beyond the few composition rules mentioned above, this PEP
+doesn't change inference for other ``str`` methods such as
+``literal_string.upper()``. If this PEP is accepted, we could also
+overload the typeshed stubs for ``str`` to preserve literal-ness in a
+broader set of scenarios where it makes sense. For example,
```
JelleZijlstra, line 337: Could there be more of a specification of "where it makes sense"? How should typeshed decide when to use Literal[str]?

pep-0675.rst
```diff
@@ -0,0 +1,954 @@
...
+<https://github.com/PyCQA/bandit/blob/aac3f16f45648a7756727286ba8f8f0cf5e7d408/bandit/plugins/django_sql_injection.py#L102>`_
+has a plugin to warn when SQL queries are not literal
+strings. `pyanalyze
+<https://github.com/quora/pyanalyze/blob/afcb58cd3e967e4e3fea9e57bb18b6b1d9d42ed7/README.md#extending-pyanalyze>`_
+has a similar extension. The problem is that many perfectly safe SQL
+queries are dynamically built out of string literals, as shown in the
+`Motivation`_ section. At the AST level, the resultant SQL query is
+not going to appear as a string literal anymore and is thus
+indistinguishable from a potentially malicious string. To use these
+tools would require significantly restricting developers ability to
```
JelleZijlstra, line 593:
```suggestion
tools would require significantly restricting developers' ability to
```

pep-0675.rst
```diff
@@ -0,0 +1,954 @@
...
+Focusing solely on the example of preventing SQL injection, tooling to
+catch this kind of issue seems to come in three flavors: AST based,
+function level analysis, and taint flow analysis.
+
+**AST based tools include Bandit and pyanalyze**: `Bandit
+<https://github.com/PyCQA/bandit/blob/aac3f16f45648a7756727286ba8f8f0cf5e7d408/bandit/plugins/django_sql_injection.py#L102>`_
+has a plugin to warn when SQL queries are not literal
+strings. `pyanalyze
+<https://github.com/quora/pyanalyze/blob/afcb58cd3e967e4e3fea9e57bb18b6b1d9d42ed7/README.md#extending-pyanalyze>`_
+has a similar extension. The problem is that many perfectly safe SQL
```
JelleZijlstra, line 588: Pyanalyze in fact handles the example in the motivation:
```
Revealed type is 'Literal['xy', 'x']' (code: inference_failure)
In striadd.py at line 5
   2:     x = "x"
   3:     if cond:
   4:         x += "y"
   5:     reveal_type(x)
                      ^
```
It's true though that it doesn't handle the full range of what `Literal[str]` would allow (e.g., having a function of type `(Literal[str]) -> Literal[str]`). In that regard it looks more like semgrep than bandit; I haven't used either of those other tools though.

pep-0675.rst
```diff
@@ -0,0 +1,954 @@
...
+        value_clause = ", ".join(["?"] * len(insert_columns))
+
+        sql += f" ({column_clause}) VALUES ({value_clause})"
+        return sql
+
+    def insert_data(
+        conn: Connection,
+        kvs_to_insert: Dict[Literal[str], str]
+    ) -> None:
+        query = build_query("data", kvs_to_insert.keys())
```
JelleZijlstra, line 627:
```suggestion
        query = build_insert_query("data", kvs_to_insert.keys())
```

pep-0675.rst
```diff
@@ -0,0 +1,954 @@
...
+vulnerabilities (with pragmatic `limitations <Appendix B:
+Limitations_>`_).
+
+Since we want the ``sqlite3`` ``execute`` method to disallow strings
+built with user input, we make it accept a ``sql`` query that is of
+type ``Literal[str]``:
+
+::
+
+    def execute(self, sql: Literal[str], parameters: Iterable[str] = ...) -> Cursor: ...
```
JelleZijlstra, line 121: Where would you make this change? Would typeshed's stubs for sqlite change?

JelleZijlstra (changes requested):

pep-0675.rst
pradeep90: Thanks for confirming! We weren't sure about `pyanalyze`. Given your explanation yesterday, I realize that `pyanalyze` is more runtime-based than AST-based. Moved it to the Semgrep bullet.

pep-0675.rst
pradeep90: Yes, we would have to update its typeshed stub. Linked to it.

pep-0675.rst
pradeep90: Good question. We faced a bit of a dilemma here. We wanted to limit the specification, which is why we enumerated the 4 most-frequently used means of composing literal strings. Listing all of the `str` methods seemed like overkill and we didn't want the PEP to be rejected because of too many changes to typeshed.

But users might ask for more convenient changes in the future - such as `my_literal_str.upper()`. So, we wanted to leave this open.

I've replaced "where it makes sense" with "where all inputs are literals" and clarified that this is merely for convenience and is not required by the PEP.

What do you suggest?

JelleZijlstra: I don't have a concrete suggestion, but as a typeshed maintainer I'd have to come up with some rules, so I'd appreciate if the PEP offered more guidance on when exactly a function should use `Literal[str]` annotations.

gbleaney: Here's my attempt to make it more concrete. I've sorted all the `str` functions from typeshed into categories:

```
# Overload Literal[str] output if Literal[str] inputs
# ===================================================
def __new__(cls, object: Literal[str] = ...) -> Literal[str]: ...
def capitalize(self: Literal[str]) -> Literal[str]: ...
def casefold(self: Literal[str]) -> Literal[str]: ...
def center(self: Literal[str], __width: SupportsIndex, __fillchar: Literal[str] = ...) -> Literal[str]: ...
if sys.version_info >= (3, 8):
    def expandtabs(self: Literal[str], tabsize: SupportsIndex = ...) -> Literal[str]: ...
else:
    def expandtabs(self: Literal[str], tabsize: int = ...) -> Literal[str]: ...
def format(self: Literal[str], *args: Literal[str], **kwargs: Literal[str]) -> Literal[str]: ...
def format_map(self: Literal[str], map: Mapping[str, Literal[str]]) -> Literal[str]: ...
def join(self: Literal[str], __iterable: Iterable[Literal[str]]) -> Literal[str]: ...
def ljust(self: Literal[str], __width: SupportsIndex, __fillchar: Literal[str] = ...) -> Literal[str]: ...
def lower(self: Literal[str]) -> Literal[str]: ...
def lstrip(self: Literal[str], __chars: Literal[str] | None = ...) -> Literal[str]: ...
def partition(self: Literal[str], __sep: Literal[str]) -> tuple[Literal[str], Literal[str], Literal[str]]: ...  # Biasing to overly restrictive: '__sep'
def replace(self: Literal[str], __old: Literal[str], __new: Literal[str], __count: SupportsIndex = ...) -> Literal[str]: ...  # Biasing to overly restrictive: '__old'
if sys.version_info >= (3, 9):
    def removeprefix(self: Literal[str], __prefix: Literal[str]) -> Literal[str]: ...  # Biasing to overly restrictive: '__prefix'
    def removesuffix(self: Literal[str], __suffix: Literal[str]) -> Literal[str]: ...  # Biasing to overly restrictive: '__suffix'
def rjust(self: Literal[str], __width: SupportsIndex, __fillchar: Literal[str] = ...) -> Literal[str]: ...
def rpartition(self: Literal[str], __sep: Literal[str]) -> tuple[Literal[str], Literal[str], Literal[str]]: ...  # Biasing to overly restrictive: '__sep'
def rsplit(self: Literal[str], sep: Literal[str] | None = ..., maxsplit: SupportsIndex = ...) -> list[Literal[str]]: ...  # Biasing to overly restrictive: 'sep'
def rstrip(self: Literal[str], __chars: Literal[str] | None = ...) -> Literal[str]: ...  # Biasing to overly restrictive: '__chars'
def split(self: Literal[str], sep: Literal[str] | None = ..., maxsplit: SupportsIndex = ...) -> list[Literal[str]]: ...  #
def splitlines(self: Literal[str], keepends: bool = ...) -> list[Literal[str]]: ...
def strip(self: Literal[str], __chars: Literal[str] | None = ...) -> Literal[str]: ...  #
def swapcase(self: Literal[str]) -> Literal[str]: ...
def title(self: Literal[str]) -> Literal[str]: ...
def upper(self: Literal[str]) -> Literal[str]: ...
def zfill(self: Literal[str], __width: SupportsIndex) -> Literal[str]: ...
def __add__(self: Literal[str], __s: Literal[str]) -> Literal[str]: ...
def __iter__(self: Literal[str]) -> Iterator[str]: ...
def __mod__(self: Literal[str], __x: Union[Literal[str], Tuple[Literal[str], ...]]) -> str: ...
def __mul__(self: Literal[str], __n: SupportsIndex) -> Literal[str]: ...
def __repr__(self: Literal[str]) -> Literal[str]: ...
def __rmul__(self: Literal[str], n: SupportsIndex) -> Literal[str]: ...
def __str__(self: Literal[str]) -> Literal[str]: ...
def __getnewargs__(self: Literal[str]) -> tuple[Literal[str]]: ...

# Explicitly ruing out override
# =============================
# Can only be made safe if `Literal[int]` were supported for table
def translate(self, __table: Mapping[int, int | str | None] | Sequence[int | str | None]) -> str: ...
# Allows selecting arbitrary strs (constrained by the contents of self)if given arbitrary index / slice
def __getitem__(self: Literal[str], __i: SupportsIndex | slice) -> str: ...

# Override doesn't make sense (ie. doesn't return str). Note that some of these could make sense for `Literal[int]`, `Literal[bytes]`, etc. if we choose to support those.
# =============================
def count(self, x: str, __start: SupportsIndex | None = ..., __end: SupportsIndex | None = ...) -> int: ...
def encode(self, encoding: str = ..., errors: str = ...) -> bytes: ...
def endswith(
    self, __suffix: str | Tuple[str, ...], __start: SupportsIndex | None = ..., __end: SupportsIndex | None = ...
) -> bool: ...
def find(self, __sub: str, __start: SupportsIndex | None = ..., __end: SupportsIndex | None = ...) -> int: ...
def index(self, __sub: str, __start: SupportsIndex | None = ..., __end: SupportsIndex | None = ...) -> int: ...
def isalnum(self) -> bool: ...
def isalpha(self) -> bool: ...
if sys.version_info >= (3, 7):
    def isascii(self) -> bool: ...
def isdecimal(self) -> bool: ...
def isdigit(self) -> bool: ...
def isidentifier(self) -> bool: ...
def islower(self) -> bool: ...
def isnumeric(self) -> bool: ...
def isprintable(self) -> bool: ...
def isspace(self) -> bool: ...
def istitle(self) -> bool: ...
def isupper(self) -> bool: ...
def rfind(self, __sub: str, __start: SupportsIndex | None = ..., __end: SupportsIndex | None = ...) -> int: ...
def rindex(self, __sub: str, __start: SupportsIndex | None = ..., __end: SupportsIndex | None = ...) -> int: ...
def startswith(
    self, __prefix: str | Tuple[str, ...], __start: SupportsIndex | None = ..., __end: SupportsIndex | None = ...
) -> bool: ...
@staticmethod
@overload
def maketrans(__x: dict[int, _T] | dict[str, _T] | dict[str | int, _T]) -> dict[int, _T]: ...
@staticmethod
@overload
def maketrans(__x: str, __y: str, __z: str | None = ...) -> dict[int, int | None]: ...
def __contains__(self, __o: str) -> bool: ...  # type: ignore[override]
def __eq__(self, __x: object) -> bool: ...
def __ge__(self, __x: str) -> bool: ...
def __gt__(self, __x: str) -> bool: ...
def __hash__(self) -> int: ...
def __le__(self, __x: str) -> bool: ...
def __len__(self) -> int: ...
def __lt__(self, __x: str) -> bool: ...
def __ne__(self, __x: object) -> bool: ...
```

@pradeep90 perhaps we incorporate the suggested overrides from the first section as an appendix?

pradeep90: > I don't have a concrete suggestion, but as a typeshed maintainer I'd have to come up with some rules, so I'd appreciate if the PEP offered more guidance on when exactly a function should use Literal[str] annotations.

@gbleaney Thanks for classifying the `str` methods.

@JelleZijlstra For now, I've removed the paragraph about typeshed and other `str` methods. We need to discuss some tradeoffs for those scenarios. Basically, adding `Literal[str]` overload to a method in `str` would affect any class that subclasses `str` - users could see spurious override errors ([snippet](https://mypy-play.net/?mypy=latest&python=3.10&gist=7c103d627b7b23f2c342cb2be00f58b0&flags=strict)) and could violate type safety by returning a value of non-literal type. I'll make that a separate PR since we might indeed decide not to  change `str` in typeshed.

Does that unblock this initial PR?

JelleZijlstra: Sounds good, let's just land the PEP first. I have some concerns about how this would interact with typeshed but that's probably better discussed elsewhere.

JelleZijlstra (approved):

NotWearingPants:
> perfectly distinguishing good and bad queries reduces to the halting problem

@pradeep90 I'd love to see a proof of this, can you provide the reference?

gbleaney:
@NotWearingPants we don't have a formalized proof for you, only the intuitive comparison: Similarly to how pathological program can decide whether or not to halt based on what the analyzing program will predict, a pathological program could decide to perform a malicious operation or not based on whether or not the analyzing program predicted it would. Perhaps we should have just invoked Rice's Theorem, or made a less strong statement. The key is just that it's impossible to know for some programs if they are going to do something malicious or not

PR #2186, opened 2021-12-08
PEP 669: Add an enum and namespace the interface.

markshannon:
From suggestion from @vstinner: put all the events into an enum and namespace the whole thing.

### DIFF
--- a/pep-0669.rst
+++ b/pep-0669.rst
@@ -20,6 +20,10 @@ on CPython that will enable monitoring at low cost.
 Although this PEP does not specify an implementation, it is expected that
 it will be implemented using the quickening step of PEP 659 [1]_.
 
+A ``sys.monitoring`` namespace will be added, which will contain
+the relevant functions and enum.
+
+
 Motivation
 ==========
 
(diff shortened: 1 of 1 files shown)


vstinner:
I suggested adding a new module because the sys module already contains 97 symbols. It seems like the sys module is the place where new functions are added when no better could be found. Maybe the sys module is a good place for these new event monitoring functions.

Maybe tomorrow, more high-level functions can be added in such hypothetical module, to have a more advanced and more high-level API. For example, the profile module has a command line interface and provides Profile and Stats classes. The timeit module also has a command line interface.

Moreover, if it's a separared module, it becomes easier to add more documentation about how it can be used, add recipes in the doc. See for example the tracemalloc module:
https://docs.python.org/dev/library/tracemalloc.html#examples

markshannon:
The original version of this had a separate module, but @zooba suggested that it should go in the `sys` modules.
https://github.com/python/peps/pull/2070#discussion_r712575895

I don't much care either way, so I'm happy to go along with the majority decision.

PR #2216, opened 2022-01-08
PEP 677: Use PEP 646 syntax when discussing unpack

stroxler:
I'd intended to make this update everywhere in an earlier PR, but missed this line.

### DIFF
--- a/pep-0677.rst
+++ b/pep-0677.rst
@@ -468,8 +468,8 @@ evaluate the AST form::
         Name("bool")
     )
 
-and be treated by type checkers as equivalent to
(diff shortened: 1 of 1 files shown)


pep-0677.rst
```diff
@@ -468,8 +468,8 @@ evaluate the AST form::
         Name("bool")
     )
 
-and be treated by type checkers as equivalent to
-``Callable[[int, Unpack[Ts]], bool]``.
+and be treated by type checkers as equivalent to or ``Callable[[int,
```
pxeger, line 471: > to or

PR #2219, opened 2022-01-09
PEP 677: Include more rejected alternatives

stroxler:
Adding more options that came up in typing-sig and python-dev discussion.

### DIFF
--- a/pep-0677.rst
+++ b/pep-0677.rst
@@ -907,6 +907,39 @@ We rejected this change because:
   and autoformatters than to bake it into the parser and throw syntax errors.
 
 
+Making ``->`` bind tighter than ``|``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to allow both ``->`` and ``|`` tokens in type expressions we
+had to choose precedence. In the current proposal, this is a function
+returning an optional boolean::
+
+    (int, str) -> bool | None  # equivalent ot (int, str) -> (bool | None)
+
+We considered having ``->`` bind tighter so that instead the expression
+would parse as ``((int, str) -> bool) | None``. There are two advantages
+to this:
+- It means we no would longer have to treat ``None | (int, str) ->
+  bool`` as a syntax error.
+- Looking at typeshed today, optional callable arguments are very common
+  because using ``None`` as a default value is a standard python idiom.
+  Having ``->`` bind tighter would make these easier to write.
+
+We decided against this for a few reasons:
+- The function header ``def f() -> int | None: ...`` is legal
+  and indicates a function returning an optional int. To be consistent
+  with function headers, callable types should do the same.
+- TypeScript is the other popular language we know of that uses both
+  ``->`` and ``|`` tokens in type expressions, and they have ``|`` bind
+  tighter. While we do not have to follow their lead, we prefer to do
+  so.
+- We do acknowledge that optional callable types are common and
+  having ``|`` bind tighter forces extra parentheses, which makes these
+  types harder to write. But code is read more often than written, and
+  we believe that requiring the outer parentheses for an optional callable
+  type like ``((int, str) -> bool) | None`` is preferable for readability.
+
+
 Introducing type-strings
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


pep-0677.rst
```diff
@@ -826,6 +826,27 @@ replacement for callable types:
...
   ``Callable``.
 
+Hybrid keyword-arrow Syntax
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the Rust language, a keyword ``fn`` is used to indicate functions
+in much the same way as Python's ``def``, and callable types are
+indicated using a hybrid arrow syntax ``Fn(i64, String) -> bool``.
+
+We could use the ``def`` keyword in callable types for python, for
```
JelleZijlstra, line 836:
```suggestion
We could use the ``def`` keyword in callable types for Python, for
```

pep-0677.rst
```diff
@@ -850,6 +955,36 @@ from the Steering Council on ensuring that type expressions do not
...
+much.
+
+In order to reduce the number of brackets needed in complex callable
+types, it would be possible to allow tuples for the argument list::
+
+    callable[(int, str), bool]
+
+This actually is a significant readability improvement for multi-argument
+functions, but the problem is that it makes callables with no arguments, which
+are *very* common in typeshed, hard to write. Because ``()`` is not a legal
```
JelleZijlstra, line 980: That's not true
```
In [6]: len(())
Out[6]: 0

In [7]: type(())
Out[7]: tuple
```

stroxler: Oh oops, I hadn't had my coffee yet.

That's supposed to be talking about one-argument functions!

pep-0677.rst
```diff
@@ -826,6 +826,27 @@ replacement for callable types:
...
+``def(int, str) -> bool``. But we think this might confuse readers
+into thinking ``def(A, B) -> C`` is a lambda, particularly because
+Javascript's ``function`` keyword is used in both named and anonymous
+functions.
+
+We could alternatively use ``Def(int, str) -> bool``, where the
+capital letter strongly suggests a type rather than a lambda. But
+that would require a new keyword, and we do not see any benefits that
+justify adding a new keyword or making the types more verbose than
+the current proposal.
```
gvanrossum, line 847: I would drop this paragraph (on the 'Def' keyword), unless someone other than you actually brought that up on a public forum. It's just not something anyone would propose. (but using 'def' totally would be).

pep-0677.rst
```diff
@@ -839,6 +860,90 @@ existing function header syntax. Moreover, it is visually similar to
...
 y``.
 
+Requiring Outer Parentheses
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A concern with the current proposal is readability, particularly
+when callable types are used in return type position which leads to
+multiple top-level ``->`` tokens, for example::
+
+    def f() -> (int) -> (int) -> int
```
gvanrossum, line 870: Missing colon.

```suggestion
    def f() -> (int) -> (int) -> int:
```

pep-0677.rst
```diff
@@ -839,6 +860,90 @@ existing function header syntax. Moreover, it is visually similar to
...
+
+We rejected this change because:
+
+- The outer parentheses only help readability in some cases, mostly when a
+  callable type is used in return position. In many other cases they hurt
+  readability rather than helping.
+- We agree that it might make sense to encourage outer parentheses in several
+  cases, particularly callable types in function return annotations. But we
+  believe it is more appropriate to encourage this in style guides, linters,
+  and autoformatters than to bake it into the parser and throw syntax errors.
```
gvanrossum, line 909: Also, this entire example looks unPythonic, and even if such a function would make sense, I'd probably use a type alias to make the return type less look like line noise.

gvanrossum (commented):
Small nits.

pep-0677.rst
stroxler: Simplified the example a bit and added a bullet about type aliasing as a solution.

Although I have some skepticism about type aliases as a general answer to complex types. Based on the production typed-python code I see I, developers seem hesitant to use them (at least at Meta).

I can understand why. Most code forms a graph where you constantly have to jump-to-source to understand it. One nice thing about types is that they normally speak for themselves. Once we use type aliases, that becomes less true - readers might have to jump hundreds of lines to the top of a module to see what a type actually represents.

This isn't necessarily a bad thing since type aliases provide abstraction and one kind of readability, but there is definitely a cost and it seems to me that many developers don't want to pay it.

gvanrossum (approved):

pep-0677.rst
gvanrossum: Yeah, it depends on lots of things. But I don’t want to feel that we have to do something just because Facebook coders prefer it. There’s a whole community here, and some people are already concerned that Python is more and more being controlled by enterprise developers.

PR #2224, opened 2022-01-11
PEP 677: Add Discussions-To and Post-History headers

stroxler:
It isn't entirely clear whether Discussions-To should be typing-sig or python-dev, but PEPs 604 and 585 have a similar purpose and both say typing-sig so I went with that (especially since neither Pradeep nor I generally follow python-dev)


gvanrossum:
I feel it should be python-dev, since this introduces significant (and controversial!) syntax. Sorry!

PR #2226, opened 2022-01-11
PEP 680: link discuss post, minor edits

hauntsaninja:


JelleZijlstra (approved):

pep-0680.rst
```diff
@@ -2,13 +2,13 @@ PEP: 680
...
 Sponsor: Petr Viktorin <encukou@gmail.com>
-Discussions-To: https://discuss.python.org/t/adopting-recommending-a-toml-parser/4068
+Discussions-To: https://discuss.python.org/t/pep-680-tomllib-support-for-parsing-toml-in-the-standard-library/13040
 Status: Draft
 Type: Standards Track
 Content-Type: text/x-rst
 Created: 01-Jan-2022
 Python-Version: 3.11
-Post-History: 1900-01-01
+Post-History: 2022-01-11
```
CAM-Gerlach, line 11:
```suggestion
Post-History: 11-Jan-2022
```

This should be the same format as `Created`, per PEP 1. While I would of course prefer ISO 8601, this is what we're stuck with.

pep-0680.rst
```diff
@@ -2,13 +2,13 @@ PEP: 680
 Title: tomllib: Support for parsing TOML in the Standard Library
 Author: Taneli Hukkinen, Shantanu Jain <hauntsaninja at gmail.com>
 Sponsor: Petr Viktorin <encukou@gmail.com>
-Discussions-To: https://discuss.python.org/t/adopting-recommending-a-toml-parser/4068
+Discussions-To: https://discuss.python.org/t/pep-680-tomllib-support-for-parsing-toml-in-the-standard-library/13040
```
CAM-Gerlach, line 5:
```suggestion
Discussions-To: https://discuss.python.org/t/13040
```

The shortlink fits better and will not break if you change the title

pep-0680.rst
```diff
@@ -2,13 +2,13 @@ PEP: 680
 Title: tomllib: Support for parsing TOML in the Standard Library
```
CAM-Gerlach, line 2:
```suggestion
Title: tomllib: Support for Parsing TOML in the Standard Library
```

or

```suggestion
Title: tomllib: Support for parsing TOML in the standard library
```

I suggest either use Title Case or Sentence case consistently (title case seems to predominate for PEP titles, though both can be found). To note, per the [standard library docs](https://docs.python.org/3/library/), it isn't considered a proper noun as it is not capitalized there, if that was where the confusion came from.

pep-0680.rst
```diff
@@ -2,13 +2,13 @@ PEP: 680
...
 Author: Taneli Hukkinen, Shantanu Jain <hauntsaninja at gmail.com>
```
CAM-Gerlach, line 3:
```suggestion
Author: Taneli Hukkinen, Shantanu Jain <hauntsaninja at gmail.com>
```

Would be good to include @hukkin 's email...

pep-0680.rst
```diff
@@ -191,31 +191,36 @@ Basing on another TOML implementation
 
 Potential alternatives include:
 
-* ``tomlkit``.
+* ``tomlkit``
+
   ``tomlkit`` is well established, actively maintained and supports TOML v1. An
```
CAM-Gerlach, line 196: This is better than how it was, but since all items but one begin with the project name and then repeat it as the first word of the first sentence, perhaps consider just doing:

```suggestion
* ``tomlkit`` is well established, actively maintained and supports TOML v1. An
```

(and the same for the others, perhaps with a tweak to the opening line if you think its important to be strictly correct, such as `Several potential alternative implementations exist:`

pep-0680.rst
```diff
@@ -356,22 +361,14 @@ Other considered names include:
...
 * https://mail.python.org/pipermail/python-dev/2019-May/157405.html
-* https://github.com/hukkin/tomli/issues/141
+* https://mail.python.org/archives/list/python-ideas@python.org/thread/IWJ3I32A4TY6CIVQ6ONPEBPWP4TOV2V7/
 * https://discuss.python.org/t/adopting-recommending-a-toml-parser/4068/84
-
-Useful https://grep.app searches (note, ignore vendored):
-
-* toml.load[s] usage https://grep.app/search?q=toml.load&filter[lang][0]=Python
-* toml.dump[s] usage https://grep.app/search?q=toml.dump&filter[lang][0]=Python
-* TomlEncoder subclasses https://grep.app/search?q=TomlEncoder%29%3A&filter[lang][0]=Python
```
CAM-Gerlach, line 374: These were indeed very useful; instead of removing them completely, could you link them inline where relevant (discussion of write support, and in the differences from `toml` appendix)?

CAM-Gerlach (commented):
Here's a few tweaks relevant to the specific changes here. Once this is merged, I'll submit a followup PR with some suggested copyediting fixes/improvements. Thanks!

pep-0680.rst
hukkin: I can if necessary, but also not a huge fan of sharing any of my email addresses publicly on the Internet. Is there a good reason for having it, or a policy requiring it?

pep-0680.rst
hauntsaninja: A previous version of this PEP had a lot more text on a) reasons to include a write API, b) an appendix with discussion on the design space, but encukou removed this in https://github.com/hauntsaninja/peps/pull/3/files . 

https://github.com/hauntsaninja/peps/blob/9d497a22a755479eef02ee406458bf93e5766127/pep-9999.rst#appendix-b-designing-a-write-api
https://github.com/hauntsaninja/peps/blob/9d497a22a755479eef02ee406458bf93e5766127/pep-9999.rst#including-an-api-for-writing-toml

If we were to include these grep.app links in the PEP, it would probably make sense to restore at least some of that text. However, I wouldn't want to do that without encukou's approval, since he removed the discussion originally.

Maybe let's address this in a follow up, and not block the copyediting changes in this + your future PR on that.

hauntsaninja:
Thank you for the review! Made most of the changes :-)

pep-0680.rst
CAM-Gerlach: Actually, email address is explicitly stated to be optional optional, not recommended or required as I thought, [per PEP 1](https://www.python.org/dev/peps/pep-0001/#pep-header-preamble). My mistake; sorry about that!

pep-0680.rst
CAM-Gerlach: Hmm, thanks. While potentially useful, I suppose if a write API is considered out of scope, the PEP doesn't need to be belabored with lengthy material about it (though personally, I find much of the rationale compelling), so I can understand why Petr might have made the change.

In any case, I see this as rather orthogonal to inline-linking the links where they are relevant, as there are still sections (just not long ones) that refer to them. In particular, I suggest linking `significant fraction` in Appendix A - 1. to the `dump` link and `users` to the `load` link, and linking `two instances` (assuming that's the relevant reference) under Appendix A - 6. to the `TomlEncoder` link. I suggest also linking `one instance` under Appendix A - 5. to the relevant grep.app link.

Or, if you prefer, we can go ahead with this, and I can re-add them in my copyediting suggestions PR...up to you.

hauntsaninja: Okay, yes. I think I focussed a little too hard on your suggestion of linking to this in "discussion of write support", but it definitely makes sense to cite these in the appendix. I've added links to justify use of "significant fraction".

I don't have clean grep.app links for the more niche stuff, I just went through hits for a couple search terms. I could link straight to Github for the one-two instance things, if you think that's worth it.

CAM-Gerlach: Looks good; thanks @hauntsaninja !

CAM-Gerlach (approved):
LGTM now; thanks @hauntsaninja for responding to the feedback!

encukou (approved):
Looks good, thanks!

PR #2230, opened 2022-01-12
PEP 677: Format the flat_map example consistently

stroxler:
When Pradeep reread the PEP yesterday he noticed that all of `Callable[...]` based versions of `flat_map` are spread across several lines, whereas our demonstration of the new syntax is a one-liner.

Keeping the formatting consistent helps make comparisons apples-to-apples.


gvanrossum (approved):
Looks much clearer, thanks!

PR #2242, opened 2022-01-15
PEP 673: minor fixes to the code samples

JelleZijlstra:
- Use `...` instead of a unicode ellipsis
- Fix dataclass definition with defaulted before non-defaulted field


JelleZijlstra:
cc @pradeep90 @Gobot1234

Gobot1234 (approved):

PR #2243, opened 2022-01-16
Fix function signatures in PEP 677 

tusharsadhwani:
The `flat_map` function is passing the function first in its calls, so the definition should reflect the same.


JelleZijlstra (approved):
cc @stroxler

stroxler:
Looks good to me, thanks @tusharsadhwani!

PR #2245, opened 2022-01-16
Fix Sphinx build

AA-Turner:
@JelleZijlstra 

Annoyingly simple fix, as usual!

xref https://github.com/AA-Turner/peps/runs/4834683325 for a successful build

A


AA-Turner:
And green 🎉https://github.com/python/peps/actions/runs/1705537380 

A

PR #2250, opened 2022-01-19
PEP 646: Mark as Accepted

mrahtz:
Huge thank you to everyone who helped get this across the finish line, especially @gvanrossum! Well done to all of us, @mrkmndz @pradeep90 @vsiles @fylux!


gvanrossum:
There’s also a header that adds a link to the acceptance notice.

At this point I am also comfortable with small clarifications.

PR #2252, opened 2022-01-20
Relax CODEOWNERS rule to allow PEP editors to "own" PEPs

AA-Turner:
xref https://github.com/python/peps/pull/2228#discussion_r788349937

A

### DIFF
--- a/pep-0001.txt
+++ b/pep-0001.txt
@@ -188,9 +188,9 @@ The standard PEP workflow is:
   "Informational", or "Process" as appropriate, and for the "Status:"
   field enter "Draft".  For full details, see `PEP Header Preamble`_.
 
-* Update `.github/CODEOWNERS` [7]_ such that any core developer co-author(s) or
-  sponsor are listed for your new file such that any future pull requests will
-  be assigned to them.
+* Update `.github/CODEOWNERS` [7]_ such that any co-author(s) or sponsors
+  with the commit bit on the `PEP repository`_are listed for your new file
+  such that any future pull requests will be assigned to them.
 
 * Push this to your GitHub fork and submit a pull request.
 
@@ -696,7 +696,7 @@ For each new PEP that comes in an editor does the following:
 
 * The file name extension is correct (i.e. ``.rst``).
 
-* Make sure the appropriate core developer(s) is added to
+* Make sure the appropriate sponsor or co-author(s) are added to
   `.github/CODEOWNERS` [7]_.
 
 * Skim the PEP for obvious defects in language (spelling, grammar,


CAM-Gerlach:
Yeah, it certainly seems like a simple oversight to me; PEP 1 may not have envisioned PEP editors who weren't core developers, like you, me and @JelleZijlstra are today. This affects you too on PEP 676, and @JelleZijlstra is already listed as CODEOWNER on PEP 673 and PEP 675, so it seems worth clarifying (even if, I think, we all understand the intent of it).

AA-Turner:
Build failed due to reST typo, now fixed.

JelleZijlstra (approved):
#2112 is where we generally approved PEP editors as sponsors. This is consistent with that change.

pep-0001.txt
```diff
@@ -188,9 +188,9 @@ The standard PEP workflow is:
   "Informational", or "Process" as appropriate, and for the "Status:"
   field enter "Draft".  For full details, see `PEP Header Preamble`_.
 
-* Update `.github/CODEOWNERS` [7]_ such that any core developer co-author(s) or
-  sponsor are listed for your new file such that any future pull requests will
-  be assigned to them.
+* Update `.github/CODEOWNERS` [7]_ such that any co-author(s) or sponsors
+  with the commit bit on the `PEP repository`_ are listed for your new file
+  such that any future pull requests will be assigned to them.
```
CAM-Gerlach, line 193:
```suggestion
  with write access to the `PEP repository`_ are listed for your new file,
  such that any future pull requests will be assigned to them.
```

Add missing comma (from original). Also, I suggest using the actual term here as used in the GitHub UI, as its meaning is clearer and more obvious than old-school git jargon that not every user may be familiar with.

pep-0001.txt
```diff
@@ -696,7 +696,7 @@ For each new PEP that comes in an editor does the following:
 
 * The file name extension is correct (i.e. ``.rst``).
 
-* Make sure the appropriate core developer(s) is added to
+* Make sure the appropriate sponsor or co-author(s) are added to
```
CAM-Gerlach, line 699:
```suggestion
* Make sure the appropriate sponsor and co-author(s) are added to
```

Per the above text, both should be added, not one or the other as "or" might imply

CAM-Gerlach (approved):
Two small textual suggestions; otherwise :heavy_check_mark: from me

pep-0001.txt
JelleZijlstra: While we're here, maybe we should reword the sentence such that it does not say "such that" twice.

pep-0001.txt
AA-Turner: Here I disagree (from logic this is ∨ not ⊻) -- we don't say "either sponsor or co-author(s)", and there are many PEPs without sponsors (core dev author). "or" in this context is therefore one of P, Q, P+Q.

pep-0001.txt
AA-Turner: Also applied to useages of "git push privileges" for consistency

pep-0001.txt
```diff
@@ -188,9 +188,9 @@ The standard PEP workflow is:
   "Informational", or "Process" as appropriate, and for the "Status:"
   field enter "Draft".  For full details, see `PEP Header Preamble`_.
 
-* Update `.github/CODEOWNERS` [7]_ such that any core developer co-author(s) or
-  sponsor are listed for your new file such that any future pull requests will
-  be assigned to them.
+* Update `.github/CODEOWNERS` [7]_ such that any co-author(s) or sponsors
+  with write access to the `PEP repository`_ are listed for your new file.
+  This means that any future pull requests will be assigned to them.
```
CAM-Gerlach, line 193:
```suggestion
* Update `.github/CODEOWNERS` [7]_ such that any co-author(s) or sponsors
  with write access to the `PEP repository`_ are listed for your new file.
  This ensures any future pull requests touching the file will be assigned
  to them.
```

Better wording

AA-Turner: "changing" over "touching"?

pep-0001.txt
CAM-Gerlach: Huh...I (AmE) at least read this as an exclusive or, i.e. implying that either P, Q but not P + Q, whereas with "and" I read it as the union of the sets P and Q intersected with the set of users with write access to `python/peps` (even if set P may be empty and the result is logically equivalent to only Q). But if others read it differently and "or" reads clearer for the intended meaning, then I guess I'm the odd CAM out, heh.

AA-Turner:
We've three PEP editors approving, I'll wait until tomorrow afternoon (~4pm UTC, 20/01) and merge there's no negative comments

A

Rosuav:
I wonder if the last sentence might be less awkward as "sponsor/co-author(s) are added"? The singular/plural discrepancy is hard to handle, but maybe combining them might feel better.

CAM-Gerlach:
Great suggestion @Rosuav ; this also bypasses the "and"/"or" issue I was haggling over

pep-0001.txt
```diff
@@ -696,7 +697,7 @@ For each new PEP that comes in an editor does the following:
 
 * The file name extension is correct (i.e. ``.rst``).
 
-* Make sure the appropriate core developer(s) is added to
+* Make sure the appropriate sponsor or co-author(s) are added to
```
CAM-Gerlach, line 700:
```suggestion
* Make sure the appropriate sponsor/co-author(s) are added to
```

Implement @Rosuav 's suggestion

AA-Turner:
```diff
- * Make sure the appropriate core developer(s) is added to
-   `.github/CODEOWNERS` [7]_.
+ * Ensure that everyone listed as a sponsor or co-author of the PEP with write 
+   access to the `PEP repository`_ is added to `.github/CODEOWNERS` [7]_.
```

A larger rephrasing, but avoids a slash in prose. @Rosuav / @CAM-Gerlach ? I understand the plural agreement issue now (we could also solve it by saying sponsor(s), but I'm not sure it makes sense for there to be more than one sponsor on a PEP)

A

Rosuav: Seems decent. It runs on a bit, though; what about "who has write access" to clarify that clause?

AA-Turner:
Thanks all for the reviews, merged
A

PR #2264, opened 2022-01-24
Exclude `.codespell` from Sphinx

AA-Turner:
RE #2151; fixes Sphinx build.

A

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -41,6 +41,8 @@ exclude_patterns = [
     "README.rst",
     "CONTRIBUTING.rst",
     "pep_sphinx_extensions/LICENCE.rst",
+    # Miscellaneous
+    ".codespell",
 ]
 
 # -- Options for HTML output -------------------------------------------------


AA-Turner:
Good god I'm bad on getting things mixed up in PR titles -- first RST vs RFC, now codeowners vs codespell 😁

Thanks for merging Jelle

CAM-Gerlach:
Happens to all of us :laughing: Luckily I fixed it with a few seconds to spare :)

This is going to keep happening whenever anyone add `.txt` files to the repo, whether as config or as a PEP supplement. Is there any way we can specifically match only `.+.rst|^/pep-[0-9]{4}\.txt$`, or at least only `.txt` (but _not_ `.rst`) in the root?

AA-Turner:
Ideally I'd whitelist to `include_patterns = ["pep-????"]`, but Sphinx doesn't support that. (Need to get a second to submit a feature request & implementation upstream)

The build on PRs going forwards will make it much easier to catch in the future, though.

A

PR #2274, opened 2022-01-25
PEP 594: remove modules from the list that are already gone or deprecated

brettcannon:


pep-0594.rst
```diff
@@ -632,9 +571,6 @@ Discussions
   of modules like ``imp`` until Python 3.10. Version 3.8 will be released
   shortly before Python 2 reaches end-of-life. A delay reduced churn for
   users that migrate from Python 2 to 3.8.
```
CAM-Gerlach, line 573: @brettcannon Just to make sure, you still want to keep this one?

AA-Turner: `imp` is already deprecated (https://docs.python.org/3/library/imp.html), and slated for removal in Python 3.12 (this doesn't seem to be in the docs though, so a issue might need to be raised).

https://github.com/python/cpython/blob/072f4a473e861c6c987650f08990c0ed1f76715f/Lib/imp.py#L31-L34

A

CAM-Gerlach: > imp is already deprecated (https://docs.python.org/3/library/imp.html)

Yeah, we're both aware :slightly_smiling_face: In this PR, @brettcannon removed `imp` from the list above for that reason; I was just checking in case he wanted to remove it here as well.

> this doesn't seem to be in the docs though, so a issue might need to be raised).

At the top:

> Deprecated since version 3.4: The imp module is deprecated in favor of importlib.

AA-Turner: Ahh! Sorry for the noise then. 

> At the top

That doesn't note the pending removal, though. Rather minor, regardless.

A

PR #2295, opened 2022-02-02
PEP 682: Format Specifier for Signed Zero

belm0:


pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+========
+
+An optional, literal ``z`` is added to the
+`Format Specification Mini-Language`_ following ``sign``::
+
+    [[fill]align][sign][z][#][0][width][grouping_option][.precision][type]
+
+where ``z`` is allowed only for float and Decimal types.  The new specifier
+is supported by both f-strings and ``str.format()``, but not %-formatting.
+
```
ericvsmith, line 82: It would be more correct to say that support of 'z' is provided by `float.__format__`, `decimal.__format__`, and I guess `complex.__format__`. A consequence of that is that only f-strings, built-in format(), and str.format() will allow 'z' to be used, and not %-formatting.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+numbers.
+
+The proposed extension is intentionally ``[sign][z]`` rather than
+``[sign[z]]``, since the latter would defeat ``sign``'s default (``-``)
+when using the ``z`` option.  Precisely because it is the default, ``-`` is
+not well known, and requiring it would increase the difficulty for those
+using or encountering the new option.
+
+There is a long-standing decision not to enhance the old %-formatting, so
+only f-strings and ``str.format()`` support the new option.
```
ericvsmith, line 135: I think this should say "only f-strings, built-in format(), and str.format() will benefit from the new option".

ericvsmith (commented):
I think this looks good, with a few small corrections.

pep-9999.rst
belm0: I guess `int.__format__` too?   Even though there isn't an int -0, we don't want an exception when someone hands an int to a format like `z.1f`.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
+PEP: 9999
```
JelleZijlstra, line 1: You can take PEP 682

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+Title: Format Specifier for Signed Zero
+Author: John Belmonte <john@neggie.net>
+Sponsor: Mark Dickinson
+Discussions-To: https://bugs.python.org/issue45995
```
JelleZijlstra, line 5: This should be a mailing list (probably python-dev; example: PEP 677) or Discuss thread (example: PEP 680).

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+Created: 29-Jan-2022
+Python-Version: 3.11
+Post-History:
+
+
+Abstract
+========
+
+Though float and Decimal types can represent `signed zero`_, in many fields
+of mathematics negative zero is surprising or unwanted-- especially in the
```
JelleZijlstra, line 18:
```suggestion
of mathematics negative zero is surprising or unwanted -- especially
```
"in the" is repeated on the next line, and add spacing

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+    ValueError: unsupported format character 'z' (0x7a) at index 1
+
+.. _`Format Specification Mini-Language`: https://docs.python.org/3/library/string.html#format-specification-mini-language
+
+
+Rationale
+=========
+
+There are use cases where negative zero is unwanted in formatted number
+output-- arguably it's the more common case.  Expanding the format
```
JelleZijlstra, line 108:
```suggestion
output -- arguably it's the more common case.  Expanding the format
```

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+    >>> for x in (.002, -.001, .060):
+    ...     print(f'{round(x, 1) + 0.: .1f}')
+     0.0
+     0.0
+     0.1
+
+The behavior of number formatting with regard to negative zero is not
+unique to Python.  The first low-level implementations of ``printf()``
+and such likely wired the sign bit to ``-`` in the output without special
+consideration of the zero mantissa case.  In hindsight, it would have
```
JelleZijlstra, line 65: This sounds a bit speculative; the PEP would be stronger with more discussion of how other languages handle this. I see below that there's a pending C++ proposal, but has this never come up in C or Java or other languages?

Feel free to disregard this though if you feel the PEP is strong enough as is.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+feature request prompting this PEP was argued without knowledge of the
+C++ proposal.)
+
+.. _`proposed for C++ std::format()`: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1496r2.pdf
+
+
+Reference Implementation
+========================
+
+A reference implementation exits at `pull request #30049`_.
```
JelleZijlstra, line 153:
```suggestion
A reference implementation exists at `pull request #30049`_.
```

JelleZijlstra (changes requested):
Thanks! A couple of typos and clerical suggestions, and one bigger suggestion.

pep-9999.rst
ericvsmith: Hmm. Good question. I believe that the `f` forces the int to be converted to a float, so that `z.1f` would automatically be supported for ints.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+    >>> x
+    -0.0
+
+When formatting a number, negative zero can result from rounding.  Assuming
+the user's intention is truly to discard precision, the distinction between
+negative and positive zero of the rounded result might be considered an
+unwanted artifact::
+
+    >>> for x in (.002, -.001, .060):
+    ...     print(f'{x: .1f})
```
hugovk, line 41: Fix syntax error:
```suggestion
    ...     print(f'{x: .1f}')
```

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+The solution must be opt-in, because we can't change the behavior of
+programs that may be expecting or relying on negative zero when formatting
+numbers.
+
+The proposed extension is intentionally ``[sign][z]`` rather than
+``[sign[z]]``, since the latter would defeat ``sign``'s default (``-``)
+when using the ``z`` option.  Precisely because it is the default, ``-`` is
+not well known, and requiring it would increase the difficulty for those
+using or encountering the new option.
```
CAM-Gerlach, line 132: My knowlege of string format specifiers and the format spec mini-language clearly isn't as deep as yours, but as a "regular" Python developer I was a bit confused on this point—could you perhaps clarify a bit what you mean here?

CAM-Gerlach (changes requested):
Thanks, @belm0 , this looks great! I have a handful of comments from the RST technical side, as well as a few copyediting suggestions.

Before renaming the PEP to its number, to ensure they and those of the other reviewers don't get lost, I strongly recommend applying them (using Github's <kbd>Add to batch</kbd> button and then the <kbd>Commit</kbd> button at the top on the ``Files`` tab, making things easier for both you and reviewers) or not, as the case may be.

There are a few missing sections you probably want to include; they can be very brief (even just a line or two, in some cases) but its likely worth at least explicitly mentioning them:

* [ ] Backwards Compatibility: Presumably none, but it would be good to state this explicitly if so (and explain otherwise). You could at least briefly highlight how this proposal avoids the backward compatibility implications of changing the default behavior.
* [ ] Security Implications: If this is used in code, are there any foreseen implications for security? If so, you should describe them, and if not, you can simply state such explicitly. String formatting bugs (Log4Shell) have been all the rage lately, though this seems pretty safe.
* [ ] How to Teach This: It shouldn't need to be long, but it would be nice to at least briefly describe how you would suggest educating a newer user on when and how to use this feature.
* [ ] Rejected Ideas: If there have been any alternate ways of implementing this that have come up, or you think might, it would be worth at least briefly mentioning them here and why they were not pursued—e.g. change the default behavior? Name it something different from `z`? Make it a different part of the format specifier/format string mini-language?

Thanks again, and looking forward to seeing this as an official PEP!

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+An optional, literal ``z`` is added to the
+`Format Specification Mini-Language`_ following ``sign``::
+
+    [[fill]align][sign][z][#][0][width][grouping_option][.precision][type]
+
+where ``z`` is allowed only for float and Decimal types.  The new specifier
+is supported by both f-strings and ``str.format()``, but not %-formatting.
+
+Synopsis::
```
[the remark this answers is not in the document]
JelleZijlstra: I do like "Examples" better, but to me "Synopsis" sounds fine too.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
+PEP: 9999
+Title: Format Specifier for Signed Zero
+Author: John Belmonte <john@neggie.net>
+Sponsor: Mark Dickinson
```
brettcannon, line 4: Mark's email is missing.

CAM-Gerlach: To note, there's a bit of an inconsistency between PEP-0001 and PEP-0012 here, as they both have header blocks but say somewhat different things in them (and your PEP template inside the `pep-0012` directory uses the PEP 1 header block). The former says only `<real name of sponsor>` for the sponsor field, and `<list of authors' real names and optionally, email addrs>` for the `Author` field, while the latter states `[Full Name <email at example.com>]` for both.

Probably something we should decide on and resolve, and maybe reduce the substantial duplication between the two (which besides being non-DRY leads to this sort of conflict and confusion) when it comes to header information. Want me to open an issue?

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+
+    >>> x = -.00001
+    >>> '%z.1f' % x  # not supported on %-formatting
+    ValueError: unsupported format character 'z' (0x7a) at index 1
+
+.. _`Format Specification Mini-Language`: https://docs.python.org/3/library/string.html#format-specification-mini-language
+
+
+Rationale
+=========
```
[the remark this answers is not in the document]
belm0: I wanted a section to concisely state the specification (currently Proposal), and another section to explain the rationale of the proposal.  It seems strange to explain why something is before stating what the thing is.  For PEP's, is it that "Rationale" implies "rationale of the motivation" rather than "rationale of the specification"?

Are there perhaps some other sections I can employ to match my intention?  Or do I explain why the specification is as it is within the Specification section itself?

pep-9999.rst
belm0: If the language was `[sign[z]]`, and an existing spec string in a program was `".1f"`, to suppress negative zero you'd have to change it to `"-z.1f"`.  In other words, you'd have to know that 1) `-` was implied in the original spec (most Python users never use this option explicitly and have not encountered it), and 2) when `z` is used `sign` must be explicit.

I'll try to improve the explanation.

pep-9999.rst
JelleZijlstra: You can see it in the python-dev archives at https://groups.google.com/g/dev-python/c/sdd97gYcf0Y dickinsm at gmail.

We should also get approval from him before merging this PR; I saw that he offered to sponsor on the bug, but he should get a chance to review the whole PEP too.

pep-9999.rst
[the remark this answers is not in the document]
CAM-Gerlach: Looks like the suggested change was implemented, thanks!

pep-9999.rst
CAM-Gerlach: Ah, now it all makes sense, thanks—it didn't help that my eyes were not really parsing it correctly.

pep-9999.rst
```diff
@@ -0,0 +1,173 @@
...
+of mathematics negative zero is surprising or unwanted-- especially in the
+in the context of displaying an (often rounded) numerical result.  This PEP
+proposes an extension to the string format specification, allowing negative
+zero to be normalized to positive zero.
+
+.. _`signed zero`:  https://en.wikipedia.org/wiki/Signed_zero
+
+
+Background
+==========
```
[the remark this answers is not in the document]
CAM-Gerlach: (And the fact that your content is essentially a perfect match for what's supposed to go in `Motivation`)

pep-9999.rst
CAM-Gerlach: On further thought, I might suggest firming up the speculative language here, and moving the rest of the discussion of other languages to the Rationale section, which is a good place to lead in to your high-level design decisions with this proposal.

pep-9999.rst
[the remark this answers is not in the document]
CAM-Gerlach: Hmm, I see how that can be potentially a little perplexing, and keep in mind that the rules aren't "hard and fast".

PEP 1 (PEP-0001) describes it this way:

> The rationale fleshes out the specification by describing why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages.
>
> The rationale should provide evidence of consensus within the community and discuss important objections or concerns raised during discussion.

And PEP 12 (PEP-0012) and the template say,

> [Describe why particular design decisions were made.]

Based on that and my impression of what is generally done in successful PEPs (others may differ):

* The Rationale section leads in to the specification by outlining the prior art/how the problem has been solved in other languages, linking/summarizing the consensus of any community discussions leading up to your PEP, outlining the high-level design decisions you made, and justifying how they were arrived at. For a short PEP like this, it needn't be too long.
* The Specification section then describes the full normative spec in explicit detail, which implements the design decisions discussed in the Rationale; it can sometimes briefly mention in context why a choice was made, though that is usually best kept fairly minimal and with clear reason (which you did).
* The Rejected Ideas section enumerates potential alternatives you considered or others suggested, and why they were not adopted.

So, what I would suggest for your leading Rationale section is:

* Your `Similar work` section would be a good fit near the top of the rationale section, as well as further detail you might want to add about prior art in other languages
* If you were involved in any public discussions that helped lead to the PEP, you might want to link and briefly summarize them
* Then, you could give a very high-level summary (even a sentence or two) of your design: a new formatting specifier to control this behavior, off by default (or you could move that latter part and its paragraph to a rejected idea), which the first three paragraphs, and the example, of your current rationale section lay out the reasoning for that high-level design choice. 

Then, you have your specification section, as it is now.

Finally, you can make your last couple paragraphs separate subsections in the Rejected Ideas section (and maybe the opt-out too), since they each discuss why a particular alternative was not adopted, along with any you might add based on my suggestions above, and any others suggest.

Does that make sense?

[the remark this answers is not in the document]
belm0: > The rationale fleshes out the specification by describing why particular design decisions were made. 

This seems to reinforce that rationale is secondary to the specification.  So I wonder why the "fleshing out" is prescribed to come before the specification itself.  Is it worth reconsidering that order?

But I'll revise the document to confirm to the current guidelines-- thank you for the suggestions.

pep-9999.rst
AA-Turner: I'd agree with Eric here -- if you do keep the current text, "long-standing decision" should cite something.

pep-9999.rst
```diff
@@ -0,0 +1,188 @@
...
+
+
+Similar Work
+============
+
+The same formatting option for negative zero has been `proposed for
+C++ std::format()`_.  While the proposal was withdrawn for C++20, a
+consensus proposal is promised for C++23.  (For what it's worth, the original
+feature request prompting this PEP was argued without knowledge of the
+C++ proposal.)
```
AA-Turner, line 160: This should probably go into Rationale (if you use that section)

pep-9999.rst
```diff
@@ -0,0 +1,188 @@
...
+
+.. _`proposed for C++ std::format()`: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1496r2.pdf
+
+
+Reference Implementation
+========================
+
+A reference implementation exists at `pull request #30049`_.
+
+.. _`pull request #30049`: https://github.com/python/cpython/pull/30049
```
AA-Turner, line 170: Whilst this is in draft I think this is fine, but we recently changed several links to permalinks -- that might be considered at a later stage (@CAM-Gerlach may have a view here)

pep-9999.rst
```diff
@@ -0,0 +1,188 @@
...
+
+.. code-block:: pycon
+
+    >>> x = -.00001
+    >>> f'{x:z.1f}'
+    '0.0'
+
+    >>> x = decimal.Decimal('-.00001')
+    >>> '{}'.format('+z.1f')
+    '+0.0'
```
AA-Turner, line 104: From a desk review I don't think as written this works -- did you mean:
```suggestion
    >>> x = decimal.Decimal('-.00001')
    >>> '{:+z.1f}'.format(x)
    '+0.0'
```

pep-9999.rst
CAM-Gerlach: I'm afraid I don't follow, sorry—that's a link to a pull request, which already is a permalink (that PR will always map to that number), not a link to a specific file as per the previous discussion.

AA-Turner: The contents of the branch that the PR represents might change, though? 

A

pep-9999.rst
CAM-Gerlach: Couldn't this be tested and confirmed on your reference implementation?

pep-9999.rst
CAM-Gerlach: Yes, but that's intentional, as the link is to the current reference implementation, not to an out of date snapshot of it (which would need to constantly be updated every time the user made a commit), just like if it were the link to a GitHub/GitLab repository that contained a reference implementation independent of CPython itself, as on many PEPs. Additionally, being a PR, even if the fork branch backing it is moved, deleted or modified further, once it is merged (or closed, etc) it will perpetually display the commits, files, etc. from that point, as it isn't actually showing the user's branch but rather the ref stored under `pull/$ID/head:$BRANCHNAME` of the upstream repo.

AA-Turner: Sounds good -- flagged as a potential issue but seems no such issue exists, which is always nice!
A

CAM-Gerlach: Better safe than sorry, as we say across the pond!

pep-9999.rst
belm0: > I believe that the f forces the int to be converted to a float, so that z.1f would automatically be supported for ints.

Indeed, this is correct.  I was confusing it with having a integer type in the format string.  The implementation behaves as expected:
```python
>>> f"{0:z.1f}"
'0.0'
>>> f"{0:zd}"
ValueError: Negative zero coercion (z) not allowed in integer format specifier
```

pep-9999.rst
belm0: If it's OK, I'll merge with Discussions-To as blank and fill it in with a follow-up PR after the PEP is posted.

pep-9999.rst
belm0: I've removed the speculative stuff.  I think the rationale stands on its own.  There doesn't seem to be other libraries offering a negative zero option in their numeric formatting functions, and I've noted that.

pep-0682.rst
```diff
@@ -0,0 +1,183 @@
...
+Created: 29-Jan-2022
+Python-Version: 3.11
+Post-History:
+
+
+Abstract
+========
+
+Though ``float`` and ``Decimal`` types can represent `signed zero`_, in many fields
+of mathematics, negative zero is surprising or unwanted -- especially
```
AA-Turner, line 18: Line wrapping per PEP 12

```suggestion
Though ``float`` and ``Decimal`` types can represent `signed zero`_, in many 
fields of mathematics, negative zero is surprising or unwanted -- especially
```

pep-0682.rst
```diff
@@ -0,0 +1,183 @@
...
+        return f"[{','.join(f'{term:{format_spec}}' for term in v)}]"
+
+The solution must be opt-in, because we can't change the behavior of
+programs that may be expecting or relying on negative zero when formatting
+numbers.
+
+The proposed extension is intentionally ``[sign][z]`` rather than
+``[sign[z]]``.  The default for ``sign`` (``-``) is not widely known or
+explicitly written, so this avoids everyone having to learn it just to use
+the ``z`` option.
```
AA-Turner, line 105: This is a matter of draughting, so you're well within your rights as author to ignore. It does feel a slight jump in topic though -- the previous paragraphs discuss the format specification in the round, and why the solution must be invisible to those not using it, etc. This paragraph though skips detail of the current format specification or grammar of the mini-langauge itself, going straight into the nested vs sibiling approach. 

A possible solution would be to combine this paragraph with the preceeding, and say something along the lines of the solution must be opt in, and given that the negative option is the default we can't link the two, moving the detail of the grammar to the specification.

Another solution is to move this entire paragraph to the specification -- this could be argued to be explanation of the choices in the specification rather than rationale for the chage itself.

pep-0682.rst
```diff
@@ -0,0 +1,183 @@
...
+new option, %-formatting cannot.  There is already precedent for not
+extending %-formatting with new options, as was the case for the
+``,`` option (:pep:`378`).
+
+To date, there doesn't appear to be other widely-used languages or libraries
+providing such a formatting option for negative zero.  However, the same
+``z`` option syntax and semantics has been `proposed for C++ std::format()`_.
+While the proposal was withdrawn for C++20, a consensus proposal is promised
+for C++23.  (For what it's worth, the original `feature request`_ prompting
+this PEP was argued without knowledge of the C++ proposal.)
```
AA-Turner, line 117: I'm not sure "for what it's worth..." adds anything substantive, although noting the independent evolution of exactly the same solution can't hurt.

pep-0682.rst
```diff
@@ -0,0 +1,183 @@
...
+
+An optional, literal ``z`` is added to the
+`Format Specification Mini-Language`_ following ``sign``:
+
+.. code-block:: text
+
+    [[fill]align][sign][z][#][0][width][grouping_option][.precision][type]
+
+where ``z`` is allowed for numerical types other than integer.  Support for
+``z`` is provided by the ``__format__()`` method of each numeric type,
```
AA-Turner, line 134: For methods I have seen `.` being used as a lead, although I don't know if there is a style guide reccomendation.

```suggestion
``z`` is provided by the ``.__format__()`` method of each numeric type,
```

AA-Turner (commented):
One formatting comment (line-wrapping), three draughting comments.

A

pep-0682.rst
belm0: Thank you.  This is related to the discussion with CAM on section ordering: https://github.com/python/peps/pull/2295#discussion_r798041177.  I originally wrote the rationale section intending that it support the decisions of the specification, and it was placed after the specification.

> this could be argued to be explanation of the choices in the specification rather than rationale for the change itself.

As CAM cited from PEP 1:

> "The rationale fleshes out the specification by describing why particular design decisions were made."

This suggests that the rational supports the specification, and is not merely a rationale for the change itself.  If it does support the specification, logically it should follow that section.  I wish that the guidelines on section ordering gave me such leeway.

Possible resolutions, with my preference first:
   1. move Rationale after Specification (perhaps naming it "Specification Rationale" for clarity)
   2. move some rationale text into the specification, as you suggest.  (My view is that it makes the specification less concise, and I was happy to detail these things in another section.)

AA-Turner: Thanks for the extracts, I missed that discussion.

My own opinion is that clarity of expression is more important than adherence to the "rules". (That said, the "rules" exist for a reason!). I've also never been entirely sure of the difference between Motivation and Rationale.

I'd say flip the order if it makes the most sense. 

@CAM-Gerlach -- would you have a strong objection to this?

A

pep-0682.rst
```diff
@@ -0,0 +1,183 @@
...
+What we would like instead is a first-class option to normalize negative
+zero, on top of everything else that numerical string formatting already
+offers.
+
+
+Rationale
+=========
+
+There are use cases where negative zero is unwanted in formatted number
+output -- arguably it's the more common case.  Expanding the format
```
CAM-Gerlach, line 80:
```suggestion
output -- arguably, not wanting it is more common  Expanding the format
```

Make phrasing more clear as to exactly what "it" refers to (more common to want it, or not want it?) and add missing comma

pep-0682.rst
CAM-Gerlach: My actual suggestion was to not include this paragraph, the one following, and optionally the one preceding it in the Rationale at all:

> Finally, you can make your last couple paragraphs separate subsections in the Rejected Ideas section (and maybe the opt-out too), since they each discuss why a particular alternative was not adopted, along with any you might add based on my suggestions above, and any others suggest.

They are rather detailed and implementation-level (as @AA-Turner noticed), and refer to why specific potential alternatives were not adopted, which is exactly what belongs in the Rejected Ideas section (with subheadings for each separate idea)—which, indeed, goes _after_ the Specification, where you want them.

This just leaves the current mention of prior art (and any you add), the brief high-level rationale of the core design decision, and the counterpoint to why pre-rounding instead wouldn't work (the latter of which, if you really wanted, you _could_ consider moving to Motivation or as a separate rejected idea), which makes much more sense as an introduction to the specification. If you prefer, you could also consider re-organizing those sub-parts a bit to put the pre-rounding approach first (why the status quo doesn't work), prior art section second (how did other languages tackle the problem) and the high-level description of your design last, which then smoothy outlines the rationale that led to the overall idea of the proposal, which then leads into the specification itself.

CAM-Gerlach (commented):
My recommendation for how to deal with the structure issue, and one small copyedit on the added text.

pep-9999.rst
mdickinson: +1 for using https://discuss.python.org/c/peps/ for the discussion; adding a link to the thread in a follow-up PR sounds fine to me

pep-0682.rst
```diff
@@ -0,0 +1,183 @@
...
+
+    >>> for x in (.002, -.001, .060):
+    ...     print(f'{round(x, 1) + 0.: .1f}')
+     0.0
+     0.0
+     0.1
+
+What we would like instead is a first-class option to normalize negative
+zero, on top of everything else that numerical string formatting already
+offers.
```
mdickinson, line 73: Should we be providing evidence that the feature is needed / frequently requested here, or do you think that can be saved for the PEP discussion? It's a reasonably frequently-asked question on Stack Overflow, for example. Some samples are below; there are plenty more where those came from:

One answer (to a question specific to printing negative zeros) showing the use of a post-formatting regex: https://stackoverflow.com/a/36604981/270986

Another question that's more clearly about handling negative zeros resulting from the rounding that's implicit in float formatting:
https://stackoverflow.com/questions/41564311/ironpython-formatting-issue-with-modulo-operator-negative-zero

Another one, with me proposing a non-solution in comments:
https://stackoverflow.com/questions/66283651/how-to-handle-rounding-to-negative-zero-in-python-docstring-tests

A Java variant (accepted solution also involves a post-formatting regex replacement):
https://stackoverflow.com/questions/11929096/negative-sign-in-case-of-zero-in-java

And another one, for Objective C:
https://stackoverflow.com/questions/10969399/prevent-small-negative-numbers-printing-as-0

pep-0682.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. code-block:: pycon
+
+    >>> for x in (.002, -.001, .060):
+    ...     print(f'{round(x, 1) + 0.: .1f}')
+     0.0
+     0.0
+     0.1
+
+What we would like instead is a first-class option to normalize negative
```
mdickinson, line 71: FWIW, I wrote some notes a couple of weeks ago on this and had a couple of domain-oriented examples of my own. I'd be happy for those to be included if the author feels the might be useful or persuasive; if not, I'll save them for the discussion threads. (The first example would inevitably lead to a not-useful tangent of when using binary floating-point for representing monetary calculations is acceptable.)

```python
>>> previous_balance = 3.30
>>> payments = [1.10, 2.20]
>>> new_balance = previous_balance - sum(payments)
>>> print("Amount owed: {:.2f}".format(new_balance))
Amount owed: -0.00
```

and

```python
>>> celcius = -17.8
>>> fahrenheit = 32.0 + 1.8*celcius
>>> print(f"Today's temperature is {fahrenheit:.1f}°F.")
Today's temperature is -0.0°F.
```

mdickinson (commented):
@belm0 Many thanks for writing this. It looks to me as though this is ready to go as soon as we resolve discussion on the ordering of the Motivation / Specification / Rationale pieces.

One question: do we want to add a "How to Teach This" section? I've taught many scores of Python classes, and feature creep in the language _is_ a real concern for instructors trying to squeeze coverage of the core language into 1.5-2 days so that they can move onto Pandas / tensorflow / building GUIs / using Cython / whatever else.  But from that perspective, this seems like a benign change: for those courses that cover the formatting specification language in detail, this would be an extra line with an example. For courses not going into detail it would make essentially no difference.

The other perspective is that of a developer encountering this in someone else's code and needing to understand what it's doing, but that also seems straightforward - they'd simply go to the appropriate section of the library manual, which would (presumably) contain the details of the use of `z`. (It's probably too esoteric a feature to include those details the tutorial as well as the library reference.)

On other missing sections:

- I don't think we need a "Security Implications" section - this is pretty clearly not security-adjacent, and all the section would say would be that there aren't any security implications that we're aware of.

- I dare say we'll need a "Rejected Ideas" section as soon as discussion gets going, but that can wait for now.

Some other minor comments inline:

mdickinson:
Possible wording for "How to Teach This", based on how I'd approach teaching it in one of our Python courses:

## How to Teach This

A typical introductory Python course will not cover string formatting in full detail. For such a course, no adjustments would need to be made. For a course that does go into details of the string formatting mini-language, a single example demonstrating the effect of the `z` on a negative value that's rounded to zero by the formatting should be enough. For an independent developer encountering the feature in someone else's code, reference to the [Format Specification Mini Language](https://docs.python.org/3/library/string.html#format-specification-mini-language) section of the library reference manual should suffice.

pep-0682.rst
CAM-Gerlach: > (2) the reasons for choosing the particular solution that's being proposed from amongst other possible solutions.

As I've been mentioning, the high level reasoning that led to the solution goes in the Rationale section, but the specific reasons why other possible solutions were not selected, and those dealing with the details of the specification, belong in the Rejected Ideas section, which goes _after_ the Specification section. Is there a reason you'd rather not use that section for the more specific alternatives and details of the specification, while still keeping a short high-level rationale and comparison of prior art (which, certainly, should go before proposing what we're doing, its "prior" art after all) before? That seems like a good compromise that moves the stuff that you and Adam identified that doesn't really belong before the specification, after it, while still keeping the stuff that leads into it before and following PEP 1 and PEP 12 on the Rationale section itself?

mdickinson: > Is there a reason you'd rather not use that section for the more specific alternatives and details of the specification

No - I'll step out of this and leave the wrangling to you, Adam and John. I don't really care much about the exact order of the sections or whether there's strict adherence to PEP 12 or not - I care much more that the resulting text is coherent and is meaningful when read from top to bottom.

AA-Turner: I also don't have a strong opinion, having said my piece above!

> I don't really care much about the exact order of the sections or whether there's strict adherence to PEP 12 or not - I care much more that the resulting text is coherent and is meaningful when read from top to bottom.

Seconded, although perhaps verboten for a pep editor to say :P

A

belm0: >  the high level reasoning that led to the solution goes in the Rationale section, but the specific reasons why other possible solutions were not selected, and those dealing with the details of the specification, belong in the Rejected Ideas section

I would use the Rejected Ideas section if there was an idea that had legs and was actually rejected.  To apply an opposite polarity to every design decision and render it as a rejected idea seems a little contrived.  The items currently documented in Rationale, for the most part, had no reasonable alternatives-- and I touch on why that is for each.  A (fictitious) example of something I'd put into Rejected Ideas would be "We explored automatically inferring suppression of negative zero based on whether the input was rounded, but it was found to not be reliable and still may break existing programs."

I appreciate that the PEP guidelines are an established thing, and I'm merely a PEP author.  On the other hand, my eyes are fresh to this, and there may be ways to refine the guidelines.  At the least, I've come upon some contradiction in the documentation as to whether Rational is that of the entire PEP or of the Specification.

What I'd like to do from here is keep the current text, but as a "Specification Rationale" subsection within "Specification", so as to make the aim clear.  I hope it's an acceptable compromise, given that the sanctioned "Rationale" section is optional.

The attention from everyone towards making a good document has been impressive and welcome-- even if we have different views on it-- thank you.

pep-0682.rst
belm0: I think full stop was omitted, so I'll apply the change manually.

pep-0682.rst
belm0: Thank you for the suggestion.  I think the existing rounding example gets at the heart of it by showing how precision intended to be discarded is affecting the output-- let's see if that's enough.

Yes, the `Decimal` type would certainly be warranted for an account that can only represent units of .01.

Specifically on the temperature example, one defense of displaying "-0" I came across was regarding temperature, where people really want to distinguish between below vs. precisely at 0 (especially in Celsius).

pep-0682.rst
belm0: added most of these (and How To Teach This) -- thank you!

pep-0682.rst
belm0: I think I've addressed what Adam raised originally by moving relevant text into Specification.

AA-Turner:
I'll merge this now as the basic text seems done & to allow posting to Discourse / python-dev. Improvements can be made in follow-up PRs if need be.

A

AA-Turner:
Thanks @belm0!

A

mdickinson:
Thanks, all! I guess the next steps are to:

- Open a discussion thread on Discourse (discuss.python.org), or on python-dev if people prefer.
- If discussion happens on Discourse, we need at least a post to python-dev pointing to that discussion thread.
- Create a small PR to update the "Discussions to" link.

@belm0 Are you happy to do all of the above? I can open the discussion thread or send the python-dev post if that's helpful.

pep-0682.rst
```diff
@@ -0,0 +1,217 @@
...
+To normalize negative zero when formatting, it is necessary to perform
+a redundant (and error-prone) pre-rounding of the input:
+
+.. code-block:: pycon
+
+    >>> for x in (.002, -.001, .060):
+    ...     print(f'{round(x, 1) + 0.: .1f}')
+     0.0
+     0.0
+     0.1
```
belm0, line 69: By the way, when `code-block:: pycon` is used, I noticed that the official rendering applies a proportional font to `>>>` and `...` specifically, which looks rather odd and affects indent alignment.

Screen Shot 2022-02-09 at 8 35 43 AM

AA-Turner: Hmm, this isn't ideal. It's a pythondotorg styling problem, I'd imagine - and hard to fix currently.

A

belm0: Grep shows only one other PEP using `code-block:: pycon`, so there's no safety in numbers.  I'll remove the usage for now, as part of #2317.

ericvsmith: Why would it be "pycon", and not "python"? It seems like all of the other PEPs don't specify a language, so I think you're right in removing it.

AA-Turner: > Why would it be "pycon", and not "python"

`pycon` is the Pygments lexer for python console text, versus `python` which is for scripts. Syntactically `pycon` is correct here, although I don't know the exact difference between the two.

> It seems like all of the other PEPs don't specify a language

Support for specifying languages was only added very recently (by adding Pygments to the requirements file) -- as a PEP editor I would strongly encourage explicitness in language as it makes reading PEPs easier with correctly highlighted code.

A

AA-Turner: (`pycon` main differences seem to be that it doesn't highlight console output as Python code, and also highlights tracebacks)

CAM-Gerlach: A bit late now, sorry (I was invited to visit the SpaceX Starship production and launch site in Texas and thus was unavailable for the past week) but I can fully confirm everything @AA-Turner said, which was the basis of my initial implemented suggestion. In addition to the reasons he mentioned, being explicit about the language in the code block is also ensures the block will be highlighted correct regardless of what we decide to do in terms of the default syntax highlighter (none, auto, Python, etc).

Also, I cannot reproduce the reported issue on Firefox/Windows the [deployed PEP](https://www.python.org/dev/peps/pep-0682/) on Python.org rendered via the legacy docutils build system. If you're seeing it there, this is likely due to platform/OS/version-specific font/rendering differences given the current "system font stack" approach to monospace fonts, which is why I advocate using a webfont instead to guarantee consistent, correct rendering on all platforms, OS versions and user configurations to avoid a plethora of possible edge-case display issues like this with the high number of possible combinations of monospace, sans and serif fonts that could be selected depending on the OS, version, browser and user config. NB @AA-Turner @hugovk

belm0: Thank you-- while we did revert the pycon block annotations for now, I agree that correct block annotation is desirable and we'd like all the PEP documents to be doing that.

Summary from the time of the decision (https://github.com/python/peps/pull/2317#discussion_r804252471):
to summarize:
  * the console-specific formatting looks great with the new rendering system for peps.python.org
  * however, rendering on the standard pep site is botched.  @AA-Turner notes it may be hard to fix.  (I'd be happy to help if it's feasible.)
  * there is only one other PEP so far using `pycon`
  * it seems we'll need a script to convert console blocks in existing PEP's anyway, so perhaps no need to put this PEP on the bleeding edge

CAM-Gerlach: Yes, but as I noted there, all this is specific to the `pycon` highlighter; simply using the `python` highlighter instead should fix the issue (unless you've confirmed it doesn't, but I didn't find any mention of that here nor there). And I am not aware of plans for batch-converting all old PEPs to use the new explicit language markup, except in cases on active/process PEPs where the highlighting is widely off (wrong language, etc) but there's no reason not to be explicit and correct in new ones as @AA-Turner says, unless it is not compatible with both renderers.

AA-Turner: > batch-converting all old PEPs to use the new explicit language markup

I have avoided this topic as I don't want to spark a discussion. My current opinion is we should leave implicit literal block syntax as it is in most places, but update all cases where the language is not Python. This opinion is subject to change several times per arbitrary time period.

This probably means we wouldn't update to `pycon` here on a retroactive sweep. 

> typefaces

another debate for another time ;)

A

CAM-Gerlach: > I have avoided this topic as I don't want to spark a discussion. My current opinion is we should leave implicit literal block syntax as it is in most places, but update all cases where the language is not Python. This opinion is subject to change several times per arbitrary time period.

I share this opinion as well, and am willing to help implement if and when the time comes for this, though I think it could be limited to not include withdrawn, superceded, and otherwise inactive PEPs. For new PEPs I would favor the `.. code-block:: python` syntax for all code blocks, which I believe is your viewpoint as well.

PR #2296, opened 2022-02-02
PEP 681: Fix type annotation of field_descriptors param

debonte:
Fix field_descriptors type annotation to match in-use alternate form.

### DIFF
--- a/pep-0681.rst
+++ b/pep-0681.rst
@@ -556,7 +556,7 @@ following declaration within their type stubs or source files:
       eq_default: bool = True,
       order_default: bool = False,
       kw_only_default: bool = False,
-      field_descriptors: Tuple[Union[type, Callable[..., Any]], ...] = (()),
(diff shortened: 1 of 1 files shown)


pep-0681.rst
```diff
@@ -203,7 +203,7 @@ customization of default behaviors:
       eq_default: bool = True,
       order_default: bool = False,
       kw_only_default: bool = False,
-      field_descriptors: Tuple[type, ...] = (()),
+      field_descriptors: Tuple[Union[type, Callable[..., Any]], ...] = (()),
```
JelleZijlstra, line 206:
```suggestion
      field_descriptors: tuple[type | Callable[..., Any], ...] = (),
```
Let's use modern syntax.

Also `()` is enough to specify an empty tuple.

JelleZijlstra: We can't use that form in typing-extensions because it needs to support older versions, but the PEP targets 3.11 so we should use annotations that would be idiomatic in 3.11.

PR #2297, opened 2022-02-02
PEP 675: remove unnecessary reference to Self type PEP

pradeep90:
We don't strictly rely on Self type PEP and people on typing-sig asked that we mention it in the "Requires" field. Better to just remove the dependency.

Also link to Perl's taint (runtime taint-checking) and thank David and Arie.

cc @gbleaney

### DIFF
--- a/pep-0675.rst
+++ b/pep-0675.rst
@@ -734,8 +734,9 @@ type checker.
 This extension to the ``LiteralString`` concept would dramatically
 increase the scope of the proposal by requiring changes to one of the
 most fundamental types in Python. While runtime taint checking on
-strings has been `considered <https://bugs.python.org/issue500698>`_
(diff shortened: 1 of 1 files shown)


pep-0675.rst
```diff
@@ -562,8 +562,23 @@ annotating it as ``x: LiteralString``:
...
 ================
 
-We propose an implementation for ``typing.LiteralString`` similar to that for
-``typing.Self`` from :pep:`673`.
+We propose an implementation for ``typing.LiteralString`` similar to
+``typing.NoReturn``:
+
+::
+
+    @_SpecialForm
```
JelleZijlstra, line 570: I'd rather not put this in a PEP as it's an implementation detail. I think you can just say that it will behave like NoReturn at runtime and leave it at that.

PR #2312, opened 2022-02-07
PEP 517: Remove "Provisional acceptance" section now the PEP is Final

domdfcoding:
PEP 517 was made final in https://github.com/python/peps/pull/1712, so the section is outdated and can be removed.


AA-Turner:
Thanks!

A

PR #2320, opened 2022-02-11
PEP 683: Add a PEP for immortal objects.

ericsnowcurrently:
I plan on merging/posting this as soon as I get a review from my co-author.

### DIFF
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -144,6 +144,12 @@ can drop their workarounds.  For the per-interpreter GIL project,
 immortal objects greatly simplifies the solution for existing static
 types, as well as objects exposed by the public C-API.
 
+In general, a strong immutability guarantee for objects enables Python
+applications to scale like never before.  This is because they can
+then leverage multi-core parallelism without a tradeoff in memory
+usage.  This is reflected in most of the above cases.
+
+
 Performance
 -----------
 
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -312,9 +312,10 @@ Open Issues
 References
 ==========
 
-.. [python-dev] This was discussed in December 2021 on python-dev:
-   https://mail.python.org/archives/list/python-dev@python.org/thread/7O3FUA52QGTVDC6MDAV5WXKNFEDRK5D6/#TBTHSOI2XRWRO6WQOLUW3X7S5DUXFAOV
-   https://mail.python.org/archives/list/python-dev@python.org/thread/PNLBJBNIQDMG2YYGPBCTGOKOAVXRBJWY
+This was discussed in December 2021 on python-dev:
+
+* https://mail.python.org/archives/list/python-dev@python.org/thread/7O3FUA52QGTVDC6MDAV5WXKNFEDRK5D6/#TBTHSOI2XRWRO6WQOLUW3X7S5DUXFAOV
+* https://mail.python.org/archives/list/python-dev@python.org/thread/PNLBJBNIQDMG2YYGPBCTGOKOAVXRBJWY
 
 
 Copyright
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -17,9 +17,10 @@ Abstract
 Under this proposal, any object may be marked as immortal.
 "Immortal" means the object will never be cleaned up (at least until
 runtime finalization).  Specifically, the `refcount`_ for an immortal
-object is set to a sentinel value, and that refcount is never
-changed by ``Py_INCREF()``, ``Py_DECREF()``,
-or ``Py_SET_REFCNT()``.
+object is set to a sentinel value, and that refcount is never changed
+by ``Py_INCREF()``, ``Py_DECREF()``, or ``Py_SET_REFCNT()``.
+For immortal containers, the ``PyGC_Head`` is never
+changed by the garbage collector.
 
 Avoiding changes to the refcount is an essential part of this
 proposal.  For what we call "immutable" objects, it makes them
@@ -39,7 +40,10 @@ Motivation
 Without immortal objects, all objects are effectively mutable.  That
 includes "immutable" objects like ``None`` and ``str`` instances.
 This is because every object's refcount is frequently modified
-as it is used during execution.
+as it is used during execution.  In addition, for containers
+the runtime may modify the object's ``PyGC_Head``.  These
+runtime-internal state currently prevent
+full immutability.
 
 This has a concrete impact on active projects in the Python community.
 Below we describe several ways in which refcount modification has
@@ -216,6 +220,7 @@ The approach involves these fundamental changes:
 * update ``Py_INCREF()`` and ``Py_DECREF()`` to no-op for objects with
   the magic refcount (or its most significant bit)
 * do the same for any other API that modifies the refcount
+* stop modifying ``PyGC_Head`` for immortal containers
 * ensure that all immortal objects are cleaned up during
   runtime finalization
 
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -202,8 +202,6 @@ Non-Obvious Consequences
   be incref'ed enough to reach ``_Py_IMMORTAL_REFCNT`` and then
   be treated as immortal
 
-??? immortal != immutable
-
 
 Specification
 =============
@@ -277,6 +275,19 @@ Other API that might benefit from such notes are currently undocumented.
 
 We wouldn't add a note anywhere else (including for ``Py_INCREF()`` and
 ``Py_DECREF()``) since the feature is otherwise transparent to users.
+
+
+Rejected Ideas
+==============
+
+Equate Immortal with Immutable
+------------------------------
+
+Making a mutable object immortal isn't particularly helpful.
+The exception is if you can ensure the object isn't actually
+modified again.  Since we aren't enforcing any immutability
+for immortal objects it didn't make sense to emphasis
+that relationship.
 
 
 Reference Implementation
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -283,12 +283,6 @@ We wouldn't add a note anywhere else (including for ``Py_INCREF()`` and
 ``Py_DECREF()``) since the feature is otherwise transparent to users.
 
 
-How to Teach This
-=================
-
-This is not a user-facing change.
-
-
 Reference Implementation
 ========================
 
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -1,7 +1,6 @@
 PEP: 683
 Title: Immortal Objects, Using a Fixed Refcount
 Author: Eric Snow <ericsnowcurrently@gmail.com>, Eddie Elizondo <eelizondo@fb.com>
-PEP-Delegate: <TBD>
 Discussions-To: python-dev@python.org
 Status: Draft
 Type: Informational
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -84,10 +84,6 @@ otherwise immutable objects.  With immortal objects,
 support for a per-interpreter GIL
 becomes much simpler.
 
-(We also looked into just setting the refcount really high and not
-worrying about races, but the issue of cache invalidation described
-above becomes more pronounced without the GIL to smooth things out.)
-
 Avoiding Copy-on-Write
 ----------------------
 
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -17,7 +17,7 @@ Abstract
 
 Under this proposal, any object may be marked as immortal.
 "Immortal" means the object will never be cleaned up (at least until
-runtime finalization).  Specifically, the refcount for an immortal
+runtime finalization).  Specifically, the `refcount`_ for an immortal
 object is set to a sentinel value, and that refcount is never
 changed by ``Py_INCREF()`` or ``Py_DECREF()``.
 
@@ -29,6 +29,8 @@ would otherwise be prohibitive.
 
 This proposal is CPython-specific and, effectively, describes
 internal implementation details.
+
+.. _refcount: https://docs.python.org/3.11/c-api/intro.html#reference-counts
 
 
 Motivation
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -142,7 +142,7 @@ types, as well as objects exposed by the public C-API.
 Performance
 -----------
 
-A naive implementation shows a 4% slowdown. [benchmarks]_
+A naive implementation shows `a 4% slowdown`_.
 Several promising mitigation strategies will be pursued in the effort
 to bring it closer to performance-neutral.
 
@@ -150,6 +150,8 @@ On the positive side, immortal objects save a significant amount of
 memory when used with a pre-fork model.  Also, immortal objects provide
 opportunities for specialization in the eval loop that would improve
 performance.
+
+.. _a 4% slowdown: https://github.com/python/cpython/pull/19474#issuecomment-1032944709
 
 Backward Compatibility
 -----------------------
@@ -310,11 +312,8 @@ Open Issues
 References
 ==========
 
-.. [benchmarks]
-   https://github.com/python/cpython/pull/19474#issuecomment-1032944709
-.. [python-dev]
+.. [python-dev] This was discussed in December 2021 on python-dev:
    https://mail.python.org/archives/list/python-dev@python.org/thread/7O3FUA52QGTVDC6MDAV5WXKNFEDRK5D6/#TBTHSOI2XRWRO6WQOLUW3X7S5DUXFAOV
-.. [python-dev-alt]
    https://mail.python.org/archives/list/python-dev@python.org/thread/PNLBJBNIQDMG2YYGPBCTGOKOAVXRBJWY
 
 
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -209,15 +209,6 @@ Non-Obvious Consequences
 ??? immortal != immutable
 
 
-Concerns
-========
-
-(This topic has been discussed previously on python-dev. [python-dev_])
-
-Concerns have centered around the performance penalty.
-Mitigation is discussed above.
-
-
 Specification
 =============
 
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -293,7 +293,9 @@ that relationship.
 Reference Implementation
 ========================
 
-A PR has been posted. [elizondo]_
+The implementation is proposed on GitHub:
+
+https://github.com/python/cpython/pull/19474
 
 
 Open Issues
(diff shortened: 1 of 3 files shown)


ericsnowcurrently:
@eduardo-elizondo

AA-Turner:
You can take 683
A

pep-06XX-immortal-objects.rst
```diff
@@ -0,0 +1,342 @@
+PEP: 6XX
+Title: Immortal Objects, Using a Fixed Refcount
+Author: Eric Snow <ericsnowcurrently@gmail.com>, Eddie Elizondo <eelizondo@fb.com>
+PEP-Delegate: <TBD>
```
AA-Turner: For rendering it would be better to leave the `<TBD>` fields out whilst still undecided

pep-06XX-immortal-objects.rst
```diff
@@ -0,0 +1,342 @@
...
+Post-History:
+Resolution: <TBD>
+
+
+Abstract
+========
+
+Under this proposal, any object may be marked as immortal.
+"Immortal" means the object will never be cleaned up (at least until
+runtime finalization).  Specifically, the refcount for an immortal
```
AA-Turner: Minor: You never link / explain what the refcount is, although it may be assumed knowledge for the type of people reading this PEP

pep-06XX-immortal-objects.rst
```diff
@@ -0,0 +1,342 @@
...
+The alternative is to ensure that all shared objects are truly immutable.
+There would be no races because there would be no modification.  This
+is something that the immortality proposed here would enable for
+otherwise immutable objects.  With immortal objects,
+support for a per-interpreter GIL
+becomes much simpler.
+
+(We also looked into just setting the refcount really high and not
+worrying about races, but the issue of cache invalidation described
+above becomes more pronounced without the GIL to smooth things out.)
```
AA-Turner: Minor: This might be more appropriate for "Rejected Ideas", but you don't have that section currently.

pep-06XX-immortal-objects.rst
```diff
@@ -0,0 +1,342 @@
...
+
+Avoiding Copy-on-Write
+----------------------
+
+For some applications it makes sense to get the application into
+a desired initial state and then fork the process for each worker.
+This can result in a large performance improvement, especially
+memory usage.  Several enterprise Python users (e.g. Instagram,
+YouTube) have taken advantage of this.  However, the above
```
AA-Turner: It might be useful have a cite for this

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
+PEP: 683
+Title: Immortal Objects, Using a Fixed Refcount
+Author: Eric Snow <ericsnowcurrently@gmail.com>, Eddie Elizondo <eelizondo@fb.com>
+PEP-Delegate: <TBD>
```
AA-Turner, line 4: For rendering it would be better to leave the `<TBD>` fields out whilst still undecided

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+Post-History:
+Resolution: <TBD>
+
+
+Abstract
+========
+
+Under this proposal, any object may be marked as immortal.
+"Immortal" means the object will never be cleaned up (at least until
+runtime finalization).  Specifically, the refcount for an immortal
```
AA-Turner, line 19: Minor: You never link / explain what the refcount is, although it may be assumed knowledge for the type of people reading this PEP.

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+The alternative is to ensure that all shared objects are truly immutable.
+There would be no races because there would be no modification.  This
+is something that the immortality proposed here would enable for
+otherwise immutable objects.  With immortal objects,
+support for a per-interpreter GIL
+becomes much simpler.
+
+(We also looked into just setting the refcount really high and not
+worrying about races, but the issue of cache invalidation described
+above becomes more pronounced without the GIL to smooth things out.)
```
AA-Turner, line 88: Minor: This might be more appropriate for "Rejected Ideas", but you don't have that section currently.

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+
+Avoiding Copy-on-Write
+----------------------
+
+For some applications it makes sense to get the application into
+a desired initial state and then fork the process for each worker.
+This can result in a large performance improvement, especially
+memory usage.  Several enterprise Python users (e.g. Instagram,
+YouTube) have taken advantage of this.  However, the above
```
AA-Turner, line 97: It might be useful have a citation/link for this.

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+Most notably, the cases described in the two examples above stand
+to benefit greatly from immortal objects.  Projects using pre-fork
+can drop their workarounds.  For the per-interpreter GIL project,
+immortal objects greatly simplifies the solution for existing static
+types, as well as objects exposed by the public C-API.
+
+Performance
+-----------
+
+A naive implementation shows a 4% slowdown. [benchmarks]_
```
AA-Turner, line 147: Please may you change this (and the other footnotes) to inline links.

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+
+* immortal containers effectively immortalize each contained item
+* the same is true for objects held internally by other objects
+  (e.g. ``PyTypeObject.tp_subclasses``)
+* an immortal object's type is effectively immortal
+* though extremely unlikely (and technically hard), any object could
+  be incref'ed enough to reach ``_Py_IMMORTAL_REFCNT`` and then
+  be treated as immortal
+
+??? immortal != immutable
```
AA-Turner, line 208: Is this line finished?

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+
+
+Concerns
+========
+
+(This topic has been discussed previously on python-dev. [python-dev_])
+
+Concerns have centered around the performance penalty.
+Mitigation is discussed above.
```
AA-Turner, line 217: I'm not sure this section adds much, as it redirects to other sections where you address concerns in-line.

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+Other API that might benefit from such notes are currently undocumented.
+
+We wouldn't add a note anywhere else (including for ``Py_INCREF()`` and
+``Py_DECREF()``) since the feature is otherwise transparent to users.
+
+
+How to Teach This
+=================
+
+This is not a user-facing change.
```
AA-Turner, line 297: This section can be removed.

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
...
+
+
+Reference Implementation
+========================
+
+A PR has been posted. [elizondo]_
```
AA-Turner, line 303: Could you change to something like "The reference implementation is contained in CPython PR #spam` please (and make it an inline link).

pep-0683.rst
```diff
@@ -0,0 +1,342 @@
+PEP: 683
+Title: Immortal Objects, Using a Fixed Refcount
+Author: Eric Snow <ericsnowcurrently@gmail.com>, Eddie Elizondo <eelizondo@fb.com>
+PEP-Delegate: <TBD>
+Discussions-To: python-dev@python.org
+Status: Draft
+Type: Informational
+Created: 10-02-2022
```
AA-Turner, line 8:
```suggestion
Created: 10-Feb-2022
```

AA-Turner: @ericsnowcurrently this is needed to make the build pass
A

pep-0683.rst
```diff
@@ -0,0 +1,343 @@
+PEP: 683
+Title: Immortal Objects, Using a Fixed Refcount
+Author: Eric Snow <ericsnowcurrently@gmail.com>, Eddie Elizondo <eelizondo@fb.com>
```
ericsnowcurrently, line 3: @eduardo-elizondo, would you rather use a different email?

brettcannon:
@ericsnowcurrently please update `CODEOWNERS` so you get asked to review any PRs for the PEP.

pep-0683.rst
```diff
@@ -0,0 +1,343 @@
...
+Specification
+=============
+
+The approach involves these fundamental changes:
+
+* add ``_Py_IMMORTAL_REFCNT`` (the magic value) to the internal C-API
+* update ``Py_INCREF()`` and ``Py_DECREF()`` to no-op for objects with
+  the magic refcount (or its most significant bit)
+* do the same for any other API that modifies the refcount
+* ensure that all immortal objects are cleaned up during
```
brettcannon, line 230: How would this work? Some list of pointers that you can iterate through to free at runtime finalization?

ericsnowcurrently: I'll have to check with Eddie, but I expect the plan was to walk through the GC objects.

pep-0683.rst
```diff
@@ -0,0 +1,343 @@
...
+
+* (public) ``Py_INCREF()``
+* (public) ``Py_DECREF()``
+* (public) ``Py_SET_REFCNT()``
+* (private) ``_Py_NewReference()``
+
+API that exposes refcounts (unchanged but may now return large values):
+
+* (public) ``Py_REFCNT()``
+* (public) ``sys.getrefcount()``
```
brettcannon, line 262: Is it worth tweaking this to always return `sys.maxsize()` for immortal objects so people can test if an object is immortal?

pep-0683.rst
ericsnowcurrently: Why an inline link?

pep-0683.rst
ericsnowcurrently: For now I'd rather folks avoid even thinking about immortal objects and keep this as internal as possible.  We can deal with it later if it looks like it might actually be useful.

pep-0683.rst
AA-Turner: Detail is in #2130 et al, the precis is that footnotes add an extra step of indirection. 

To keep all links in one place, you could use `target_` (no square brackets) and change 

```
.. [target]
   https://target.example/spam/spam/ham/eggs/and/spam
```

to

```
.. _target
   https://target.example/spam/spam/ham/eggs/and/spam
```

A

pep-0683.rst
ericsnowcurrently: I don't really have any.  However, it's fairly common knowledge on the core team.

pep-0683.rst
```diff
@@ -0,0 +1,343 @@
+PEP: 683
+Title: Immortal Objects, Using a Fixed Refcount
+Author: Eric Snow <ericsnowcurrently@gmail.com>, Eddie Elizondo <eelizondo@fb.com>
+PEP-Delegate: <TBD>
+Discussions-To: python-dev@python.org
+Status: Draft
+Type: Informational
```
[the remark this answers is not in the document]
ericsnowcurrently: There's no API, no (real) user-facing changes, not much controversy, and not a whole lot to decide.  Basically, it is only internal implementation details.  Normally this wouldn't be a PEP but the relevant info is spread around in several email threads, BPO issues, and GH issues.  @encukou asked for an informational PEP to bring the details together for a bit more clarity.

pep-0683.rst
ericsnowcurrently: Sorry, I meant is there any problem with spelling out the URL directly?

pep-0683.rst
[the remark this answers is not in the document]
JelleZijlstra: Interesting, I'd assume that a change to core refcounting semantics that causes a 4% slowdown would be controversial enough to require SC approval, but I'll trust your judgment.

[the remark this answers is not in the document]
encukou: I've added it to the SC agenda.

> @encukou asked for an informational PEP

Hm, it might turn out to need a standards track one. Anyway, IMO the text is most important.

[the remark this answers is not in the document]
encukou: Please make it a normal standards track PEP. Thanks!

— Petr, *on behalf of the SC*

pep-0683.rst
eduardo-elizondo: @brettcannon at a high level, there are three different kinds of cases that we need to care about when immortalizing objects to make sure that we correctly clean them up:

1) Runtime globals: Py_None, Py_True, Py_False, Small Ints, and types defined by: `PyTypeObject Foo =`. None of these instances are not cleaned up today, thus, the behavior will remain the same even after they are made immortal.

2) Immortalizing non-containers during runtime: Examples of these include interned strings. All of these objects have to be found during runtime shutdown, so they must be tracked somewhere. Fortunately, the biggest use case of this would be interned strings which are already tracked in a python dictionary. We should just refer to those at runtime shutdown to properly clean them up. But yes, the generic strategy is to track and deallocate at shutdown.

3) Immortalizing containers during runtime: Example of these include immortalizing the heap after initialization. To correctly track these instances, we'll leverage the GC's permanent generation by pushing all immortalized containers there. During runtime shutdown, the strategy will be to first let the runtime try to do its best effort of deallocating these instances normally. Most of the module deallocation will now be handled by `pylifecycle.c:finalize_modules` which cleans up the remaining modules as best as we can. It will change which modules are available during `__del__` but that's already defined as undefined behavior by the [docs](https://docs.python.org/3/reference/datamodel.html?highlight=__del__#object.__del__). Optionally, we could do some topological disorder to guarantee that user modules will be deallocated first before the stdlib modules. Finally, anything leftover (if any) can be found through the permanent generation gc list which we can clear after `finalize_modules`.

ericsnowcurrently:
@eduardo-elizondo, unless you have any objections, I'm going to merge this first thing tomorrow morning and post to python-dev.

pep-0683.rst
brettcannon: > 1. None of these instances are cleaned up today, thus, the behavior will remain the same even after they are made immortal.

Does it make sense to continue to do that, or should we fix the situation?

ericsnowcurrently: Static objects should not be freed.

AA-Turner (approved):

pep-0683.rst
```diff
@@ -0,0 +1,335 @@
...
+* is there any other impact on GC?
+
+
+References
+==========
+
+.. [python-dev] This was discussed in December 2021 on python-dev:
+   https://mail.python.org/archives/list/python-dev@python.org/thread/7O3FUA52QGTVDC6MDAV5WXKNFEDRK5D6/#TBTHSOI2XRWRO6WQOLUW3X7S5DUXFAOV
+   https://mail.python.org/archives/list/python-dev@python.org/thread/PNLBJBNIQDMG2YYGPBCTGOKOAVXRBJWY
+
```
AA-Turner, line 318: @ericsnowcurrently I noticed on my last read through that this footnote isn't referenced anywhere anymore. If intentional, I'd suggest making it body text in the references section and not a footnote.

A

pep-0683.rst
```diff
@@ -0,0 +1,335 @@
...
+
+Abstract
+========
+
+Under this proposal, any object may be marked as immortal.
+"Immortal" means the object will never be cleaned up (at least until
+runtime finalization).  Specifically, the `refcount`_ for an immortal
+object is set to a sentinel value, and that refcount is never
+changed by ``Py_INCREF()`` or ``Py_DECREF()``.
```
eduardo-elizondo, line 21: `Py_SET_REFCNT`  won't change the refcount either! https://github.com/python/cpython/pull/19474/files#r776528822

pep-0683.rst
```diff
@@ -0,0 +1,335 @@
...
+can drop their workarounds.  For the per-interpreter GIL project,
+immortal objects greatly simplifies the solution for existing static
+types, as well as objects exposed by the public C-API.
+
+Performance
+-----------
+
+A naive implementation shows `a 4% slowdown`_.
+Several promising mitigation strategies will be pursued in the effort
+to bring it closer to performance-neutral.
```
eduardo-elizondo, line 147: Probably not needed to include here but these are four different strategies that I'll try:
* Immortalized Interned Strings (including Py_IDENTIFIERs)
* Immortalizing heap after startup
* GC improvements for large applications
* Immortalizing Module-level globals at import time

pep-0683.rst
```diff
@@ -0,0 +1,335 @@
...
+* (public) ``Py_INCREF()``
+* (public) ``Py_DECREF()``
+* (public) ``Py_SET_REFCNT()``
+* (private) ``_Py_NewReference()``
+
+API that exposes refcounts (unchanged but may now return large values):
+
+* (public) ``Py_REFCNT()``
+* (public) ``sys.getrefcount()``
+* (public) ``sys.gettotalrefcount()``
```
eduardo-elizondo, line 251: This won't change! Or rather, it will change but not in the expected way.

This is something that I really looked into to make sure that the refleak tooling wouldn't break from this change. `gettotalrefcount ` relies on `_Py_RefTotal` which is updated independently from the object's refcount field. Thus, an object's immortality will not make the value of `gettotalrefcount` be something really large.

Currently, I've placed the `_Py_RefTotal` update after the immortalization check. Thus, this value now reflects all the increfs/decrefs that did take place. If we move the check to before the immortalization check, it will now reflect all the attempts to incref and decrefs that took place (but they could have been a no-op). I chose the former since it's a bit unintuitive to have a returned value that's larger than the actual effective operations. This is enough to make the refcount leak tooling to work as is. However, if we choose the latter, it will work even in cases where we can upgrade a once mortal object to be immortal (which is beyond of the scope of any of the current changes in the PR). In this case, `_Py_RefTotal` will always be updated regardless of the the immortalization status of the instance, but the returned value might make less sense as a whole.

The final consideration is that we need to make sure that `_PySet_Dummy` is not set to be immortal, otherwise, it will affect the implementation of `_Py_GetRefTotal`. Perhaps we should just remove the use of `_PySet_Dummy` and fix the gdb plugin in another way?

eduardo-elizondo (approved):
@ericsnowcurrently I've added a bunch of comments. Up to you if you think we need to update the PEP to reflect what I mention there. Otherwise, looks good to me!

ericsnowcurrently:
@eduardo-elizondo Great!  I'll make some updates and merge it.  Thanks for the help.

pep-0683.rst
ericsnowcurrently: I've mentioned that `sys.gettotalrefcount()` is unaffected but didn't include any of your explanation.  If you think any of it would be helpful to PEP readers then we can add it in a follow-up.

CAM-Gerlach:
Oops, too late on that change—glad I didn't start a full review. Just remember for the future, though—titles don't have periods in English, and neither do commit message summary lines (which is what the PR title maps to).

pep-0683.rst
```diff
@@ -0,0 +1,373 @@
+PEP: 683
+Title: Immortal Objects, Using a Fixed Refcount
+Author: Eric Snow <ericsnowcurrently@gmail.com>, Eddie Elizondo <eduardo.elizondorueda@gmail.com>
+Discussions-To: python-dev@python.org
```
CAM-Gerlach, line 4: Please update the PEP with a link to the actual thread once you create it, so it is easy for others to find without a lot of guessing and Googling.

Also, there are a smattering of relatively non-critical editing issues (aside from the title containing a comma, which is a bit irregular), but as this was merged just as I got to it, I'm a bit late to the party now so if needed, I can make a separate PR fixing those (though they aren't terribly critical).

PR #2322, opened 2022-02-13
PEP 682: small additions

belm0:
* clarify supported presentation types
  * cite Rust survey
  * note C99 use of `z`

### DIFF
--- a/pep-0682.rst
+++ b/pep-0682.rst
@@ -169,7 +169,7 @@ C99 ``printf`` already uses the ``z`` option character for another
 purpose:  qualifying the unsigned type (``u``) to match the length of
 ``size_t``.  However, since the signed zero option specifically disallows
 ``z`` for integer presentation types, it's possible to disambiguate the two
-uses, should C99 want to adopt this new option.
+uses, should C want to adopt this new option.
 
 
 Backwards Compatibility
--- a/pep-0682.rst
+++ b/pep-0682.rst
@@ -111,8 +111,13 @@ std::format()`_.  While the proposal was withdrawn for C++20, a consensus
 proposal is promised for C++23.  (The original `feature request`_ prompting
 this PEP was argued without knowledge of the C++ proposal.)
 
+When Rust developers debated whether to suppress negative zero in ``print``
+output, they took a small `survey of other languages`_.  Notably, it didn't
+mention any language providing an option for negative zero handling.
+
 .. _`proposed for C++ std::format()`: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1496r2.pdf
 .. _`feature request`: https://bugs.python.org/issue45995
+.. _`survey of other languages`: https://github.com/rust-lang/rfcs/issues/1074#issuecomment-718243936
 
 
 Specification
--- a/pep-0682.rst
+++ b/pep-0682.rst
@@ -165,6 +165,12 @@ the new option, %-formatting cannot.  There is already precedent for not
 extending %-formatting with new options, as was the case for the
 ``,`` option (:pep:`378`).
 
+C99 ``printf`` is already employing the ``z`` option character for another
+purpose:  qualifying the unsigned type (``u``) to match the length of
+``size_t``.  However, since the signed zero option specifically disallows
+``z`` for integer presentation types, it's possible to disambiguate the two
+uses, should C99 want to adopt this new option.
+
 
 Backwards Compatibility
 =======================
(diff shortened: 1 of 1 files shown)


mdickinson (approved):
LGTM, modulo the linter complaints about single backticks.

belm0:
@gpshead would you take a look, as the changes reflect your suggestions?

JelleZijlstra (approved):

pep-0682.rst
```diff
@@ -125,10 +130,11 @@ An optional, literal ``z`` is added to the
 
     [[fill]align][sign][z][#][0][width][grouping_option][.precision][type]
 
-where ``z`` is allowed for numerical types other than integer.  Support for
-``z`` is provided by the ``.__format__()`` method of each numeric type,
-allowing the specifier to be used in f-strings, built-in ``format()``, and
-``str.format()``.  The %-formatting style will not support the new option.
+where ``z`` is allowed for floating-point presentation types (``f``, etc.).
```
AA-Turner, line 133: > floating-point presentation types

I think the previous wording is clearer here -- what is a "floating-point presentation type"? The example of `f` etc also doesn't help, as f-strings are a mechanism by which to format things to strings, not themselves a type.

Is there a reason we shouldn't say "numerical types other than integer"?

pep-0682.rst
```diff
@@ -159,6 +165,12 @@ the new option, %-formatting cannot.  There is already precedent for not
 extending %-formatting with new options, as was the case for the
 ``,`` option (:pep:`378`).
 
+C99 ``printf`` is already employing the ``z`` option character for another
```
AA-Turner, line 168: simpler to read. I'd also avoid the colon in the next line (169) if possible, but it is easier to parse.
```suggestion
C99 ``printf`` already uses the ``z`` option character for another
```

pep-0682.rst
```diff
@@ -159,6 +165,12 @@ the new option, %-formatting cannot.  There is already precedent for not
...
+purpose:  qualifying the unsigned type (``u``) to match the length of
+``size_t``.  However, since the signed zero option specifically disallows
+``z`` for integer presentation types, it's possible to disambiguate the two
+uses, should C99 want to adopt this new option.
```
AA-Turner, line 172: This may be ignorance of how the C standard works -- I thought when standardised to a year (C 1999) the specification didn't change following that, so "should C99 want to adopt this new option" can never be true. If not please dismiss this comment.

pep-0682.rst
belm0: "presentation type" refers to the `type` field of the format specification, and is used extensively in the format specification documentation to avoid confusion with the type of the format input value.  `f` is an example presentation type.  Would "(`f`, `g`, etc.) be more clear?

belm0: > Would "(f, g, etc.)" be more clear?

I've made that change for now

AA-Turner: "floating-point presentation types as defined by the format specification"? If the list of floating point presentation types is small enough, it might be better to just list them all exhaustivley -- i.e. `f`, `g`, `blah` as Python 3.11.

I didn't pick up that this was a reference to the format spec usage from the docs -- when I see `type` I read that as a object type by default. It also might be worth adding a sentence on the fact that input types get coerced depending on the presentation type, or just redirect to the docs here -- the distinction isn't obvious right away to those who haven't spent a lot of time with the formatting innards.

A

belm0: I'm wondering if it's enough as is.  This PEP extends the Format Specification Mini-Language, and there is already a link to those docs.  The phrase "floating point presentation types" is used twice there, e.g. in defining the recently-added `'_'` option.

Even if the person evaluating the PEP isn't familiar with the mini-language documentation, "presentation type" seems a good hint that it's regarding the output type, and to further clarify I've given a few examples (`f`, `g`).  There are 9 floating point presentation types-- I'd rather not list them exhaustively.

belm0: I've changed the wording to:

> (``f``, ``g``, etc.,  as defined by the format specification documentation)

AA-Turner: nine is too many, yes. I think your latest change is sufficient to point to the right direction -- thank you.

A

AA-Turner:
@belm0 Happy to merge, or can wait if you still want @gpshead to take a look.

A

belm0:
Thank you-- please go ahead and merge.  If there are follow-up suggestions I'll do another PR.

PR #2323, opened 2022-02-13
PEP 655: Integrate feedback from S Pradeep Kumar

davidfstr:


pep-0655.rst
```diff
@@ -138,6 +148,35 @@ for TypedDict also supports
...
+Interaction with ``total=False``
+--------------------------------
+
+Any :pep:`589`-style TypedDict declared with ``total=False`` is equivalent
+to a TypedDict with an implicit ``total=True`` definition with all of its
+keys marked as ``NotRequired[]``.
+
+Therefore:
+
+::
```
AA-Turner, line 160:
```suggestion
.. code-block: python
```

JelleZijlstra (approved):

JelleZijlstra:
Happy to merge this whenever you want me to (unless you want Guido or Pradeep to have a look first).

pep-0655.rst
davidfstr: According to [Sphinx documentation](https://www.sphinx-doc.org/en/1.4.9/markup/code.html), `.. code-block:: python` is equivalent to just `::` in regular RST documents. And indeed using the former form appears to have no effect.

Strangely, _neither_ form actually syntax-highlights Python code in color, at least in PEPs.

hugovk: Will `::` default to Python?

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#showing-code-examples

davidfstr:
I'd be interested in leaving this open a few days in case @pradeep90 has any comments here, or the [related typing-sig post](https://mail.python.org/archives/list/typing-sig@python.org/message/JMEOFNSFNVU3TKFWZRGY444LFKNO34SL/), say until Thu 2/17 AoE.

pep-0655.rst
hugovk: Coloured syntax highlighting is planned via PEP 676, you can see a preview:
* https://python.github.io/peps/pep-0008/
* https://python.github.io/peps/pep-0655/

So a TypeScript code block would be useful for that particular snippet.

AA-Turner: Seconding Hugo -- the default for literal blocks is Python, but explicitness is always better, and it allows highlighting different languages, such as the TypeScript code in this PEP.

A

pradeep90 (approved):

AA-Turner:
@davidfstr given Pradeep's approval are you happy for this to be merged?

A

davidfstr:
Yes, I’m happy merging this

pep-0655.rst
CAM-Gerlach: Seconding Hugo and Adam as well. Furthermore, as that page also mentions, that default is fully configurable on a per-project and per-file basis, and we have discussed doing so in the past.

davidfstr: I'll make a separate PR that marks the language for non-Python blocks

Presumably the PEP repo will be configured to use Python as the default syntax-highlighting language, so I won't bother to explicitly set the default language to Python in this specific PEP.

CAM-Gerlach: Its still a good idea to explicitly specify the language in new PEPs, as it is more clear and consistent and we've discussed using a different default (though I think our current thinking, at least between @AA-Turner and myself, is to stick with `python` and modify any others accordingly), but it isn't critical.

hugovk: Explicit isn't *always* better than implicit :) 

Because virtually all code is Python, and there are a lot of Python code blocks, I think we should allow the default `::` for Python (and specify the language for other code blocks). 

RST isn't easy. Just yesterday I read of someone ditching it for MyST: https://twitter.com/hynek/status/1493509644228173830

We should make writing documents fluent for PEP authors and not require them to stop and look up the syntax (at least I can't always remember the syntax off the top of my head) when the double colon shortcut is so easy.

(If there's a PEP with a lot of C, I think it would be fine switching the default for that PEP to C and then using `::`.)

CAM-Gerlach: I don't think there's harm in _encouraging_ users to be explicit, but now that it seems we've at least informally settled on an approach going forward (use `python` as the default and update any non-`python` code blocks at least on active, process, etc PEPs), I agree with the sentiment that its overstepping to require all new PEP authors to explicitly annotate every code block (its what I would do myself in my own PEPs, but I can get a bit too carried away with purity at the expense of practicality sometimes).

Using `.. language` at the top of the file for PEPs like PEP 7 makes sense as well (particularly on old PEPs, to allow fixing their syntax highlighting in as little as one line rather than modifying every code block, reducing noise and churn), so long as PEP authors and editors are cognizant of it.

PR #2339, opened 2022-02-18
PEP 668: Change title to better summarise the content

pradyunsg:
See https://discuss.python.org/t/graceful-cooperation-between-external-and-python-package-managers-pep-668/10302/26?u=pradyunsg

### DIFF
--- a/pep-0668.rst
+++ b/pep-0668.rst
@@ -1,5 +1,5 @@
 PEP: 668
-Title: Graceful cooperation between external and Python package managers
+Title: Marking Python base environments as “externally managed”
 Author: Geoffrey Thomas <geofft@ldpreload.com>,
         Matthias Klose <doko@ubuntu.com>,
         Filipe Laíns <lains@riseup.net>,


pep-0668.rst
```diff
@@ -1,5 +1,5 @@
 PEP: 668
-Title: Graceful cooperation between external and Python package managers
+Title: Marking Python base environments as “externally managed”
```
CAM-Gerlach, line 2:
```suggestion
Title: Marking Python base environments as "externally managed"
```

Might be best to avoid getting too "smart" with smart quotes here?

AA-Turner: +1 -- I've even considered suggesting an anti-smart-quote linter, but I'm not quite that bad yet! Docutils' smartquotes.py handles this, an adaptation of Gruber's smartypants.

A

CAM-Gerlach: I've noticed this in a few other recent PEPs and also considered adding one; there's an existing pre-commit hook that checks for and replaces them with their "dumb" equivalents, but its third-party and doesn't correctly respect EOL chars so if we do go with that, the best approach is just a simple pygrep rule. On the other hand, I'm not certain Unicode typographic quotes are _that_ bad in reST source outside of code and other places where it is syntactically meaningful.

pradyunsg: Is this a blocking concern?

encukou: I'm not aware of any rule against using proper quotes in the source.

AA-Turner: I suppose an initial concern might've been restricting PEP sources to the ASCII codepage, but that ship has sailed.

There is no rule (no English style guide could truly claim to set rules!), but there is an element of consistency is a useful thing to have.

A

CAM-Gerlach: I admit I have a bit of an instinctual aversion to typographical quotes in source files, due to the fact that in most source languages (code, etc) they are syntactically meaningful and thus can have serious yet often difficult to spot effects on the output. The benefit to disallowing them here, though, is less clear-cut—it would be optimal to use only use standard quotes, both for consistency and to allow linting for them, since there are common places in PEPs (in code, etc) where typographical quotes _aren't_ valid but can easily sneak in, but not critical unlike elsewhere, since in most places they don't affect the syntax validity or the rendered output.

Here its just a matter of clicking one button to apply the changes, so there isn't a real reason not to, but I'm not sure if its worth having a broader discussion of whether or not to prefer this, since really not _that_ important, at the end of the day.

pradyunsg: OK, it seems like the consensus is that it's an optional change.

I prefer the current form as it stands, but if someone feels strongly about this + has the ability to push to this PR, please feel free to do so. I'm going to mark this as resolved.

CAM-Gerlach: I don't want to drag out this discussion, but as someone also experienced relatively with Sphinx (and themeing), I'd be particularly interested to hear your rationale for using literal typographic quotes over letting Sphinx's automatic smart quotes support handle it, as I don't believe you've yet actually mentioned it and I'm personally not aware of any, aside from reports of some corner-cases that the latter doesn't always handle. So I'm certainly curious what I'm missing here.

For me, consistency aside (since using literal typographic quotes in the source is a strictly greater burden on authors, for no obvious benefit aside from a few corner cases) , my main concern is being able to lint for them in the places they are syntactically erroneous (code, etc) with a simple, cheap single-character pygrep regex as opposed to having to parse reST syntax to determine what's a code block and what's not (with all half-dozen different ways to define one), since I have seen a number of instances in other Sphinx projects where they sneak into code examples and silently invalidate them.

CAM-Gerlach (approved):
I'll go ahead and merge this since we can always change the quote characters later, to avoid our rather nitpicky debate from holding up @pradyunsg 's new title

I do believe this is the first instance of literal typographic quotes being used in a PEP title, so hopefully nothing breaks...

pep-0668.rst
CAM-Gerlach: (I also merged the PR, so our discussion doesn't hold up the new title while we bikeshed)

encukou: IMO this would make a good topic for Discourse, but:
> rationale for using literal typographic quotes over letting Sphinx's automatic smart quotes support

I'd like to train my fingers to use proper quotes for the language (I use “English”, „Czech“ and "programming languages"), and I don't want to have to think about which text processors are in place.

CAM-Gerlach: Hmm, interesting! I guess my general philosophy on such things is the opposite—I tend to prefer, when possible, training my fingers to take the one correct (if not always purely optimal) action that works everywhere, rather than bearing the cognitive overhead of constant context switching, having to spend mental energy on every instance what the most optimal syntax I can get away with is, at the risk of making a mistake. Same thing with `.. code-block:: python`, using ` `` ` instead of `` ` `` and ` * ` and ` ** ` consistently in both reST and Markdown, etc.

PR #2342, opened 2022-02-19
PEP 683: Implement minor typographical changes

JelleZijlstra:
Some things I noticed while reading through #2341.

### DIFF
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -184,7 +184,7 @@ Backward Compatibility
 
 This proposal is meant to be completely compatible.  It focuses strictly
 on internal implementation details.  It does not involve changes to any
-public API, other a few minor changes in behavior related to refcounts
+public API, other than a few minor changes in behavior related to refcounts
 (but only for immortal objects):
 
 * code that inspects the refcount will see a really, really large value
@@ -503,11 +503,11 @@ to do its best effort of deallocating these instances normally.  Most
 of the module deallocation will now be handled by
 ``pylifecycle.c:finalize_modules()`` which cleans up the remaining
 modules as best as we can.  It will change which modules are available
-during __del__ but that's already defined as undefined behavior by the
-docs.  Optionally, we could do some topological disorder to guarantee
+during ``__del__``, but that's already defined as undefined behavior by the
+docs.  Optionally, we could do some topological sorting to guarantee
 that user modules will be deallocated first before the stdlib modules.
 Finally, anything leftover (if any) can be found through the permanent
-generation gc list which we can clear after finalize_modules().
+generation GC list which we can clear after ``finalize_modules()``.
 
 For non-container objects, the tracking approach will vary on a
 case-by-case basis.  In nearly every case, each such object is directly
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -503,7 +503,7 @@ to do its best effort of deallocating these instances normally.  Most
 of the module deallocation will now be handled by
 ``pylifecycle.c:finalize_modules()`` which cleans up the remaining
 modules as best as we can.  It will change which modules are available
-during ``__del__``, but that's already defined as undefined behavior by the
+during ``__del__``, but that's already explicitly undefined behavior in the
 docs.  Optionally, we could do some topological sorting to guarantee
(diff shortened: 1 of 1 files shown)


pep-0683.rst
```diff
@@ -503,11 +503,11 @@ to do its best effort of deallocating these instances normally.  Most
 of the module deallocation will now be handled by
 ``pylifecycle.c:finalize_modules()`` which cleans up the remaining
 modules as best as we can.  It will change which modules are available
-during __del__ but that's already defined as undefined behavior by the
-docs.  Optionally, we could do some topological disorder to guarantee
+during ``__del__``, but that's already defined as undefined behavior by the
+docs.  Optionally, we could do some topological sorting to guarantee
```
JelleZijlstra, line 507: I don't think "topological disorder" is a term outside solid-state physics, but might be wrong.

pep-0683.rst
```diff
@@ -503,11 +503,11 @@ to do its best effort of deallocating these instances normally.  Most
...
 that user modules will be deallocated first before the stdlib modules.
 Finally, anything leftover (if any) can be found through the permanent
```
CAM-Gerlach, line 509:
```suggestion
Finally, anything left over (if any) can be found through the permanent
```

"Leftover" is a noun, "left over" is an adjective

pep-0683.rst
```diff
@@ -503,11 +503,11 @@ to do its best effort of deallocating these instances normally.  Most
 of the module deallocation will now be handled by
 ``pylifecycle.c:finalize_modules()`` which cleans up the remaining
 modules as best as we can.  It will change which modules are available
-during __del__ but that's already defined as undefined behavior by the
-docs.  Optionally, we could do some topological disorder to guarantee
+during ``__del__``, but that's already defined as undefined behavior by the
```
CAM-Gerlach, line 506:
```suggestion
during ``__del__``, but that's already declared to be undefined behavior by the
```

Avoid confusing and repetitive "defined to be undefined" phrasing.

CAM-Gerlach (commented):
A few further textual fixes

AA-Turner (approved):

pep-0683.rst
JelleZijlstra: Not sure about this one, will leave to @ericsnowcurrently to decide.

CAM-Gerlach: Would you suggest a different word than "declared"? There are certainly other options ("documented", "stated", etc), but the point is "defined as undefined" is inherently oxymoronic, and there's no strong compelling need to use the very specific diction "defined" rather than one of its synonyms here that would justify this.

JelleZijlstra: To me "defined as undefined" sounds odd but not wrong, since "undefined behavior" is a well-defined (sorry) concept in programming language design. I don't think it matters much, so I'll just let Eric make the call when he reviews this PR.

CAM-Gerlach: Sure, but as a copyeditor, I don't see why it is necessary for it to sound "odd" (and potentially confusing) when there are many equally- or more-appropriate synonyms for the first "defined" (unlike the second usage, which is part of the well-defined term "undefined behavior").

Here are a few other alternatives of the top of my head for @ericsnowcurrently to consider, if he prefers them, that would also avoid the problem:

```suggestion
during ``__del__``, but that's already stated to be undefined behavior by the
```

```suggestion
during ``__del__``, but that's already described as undefined behavior by the
```

```suggestion
during ``__del__``, but that's already characterized as undefined behavior by the
```

JelleZijlstra:
ping @ericsnowcurrently

pep-0683.rst
ericsnowcurrently: I'd rather the following:

```suggestion
during ``__del__``, but that's already explicitly undefined behavior in the
```

ericsnowcurrently (approved):
LGTM

Thanks!

pep-0683.rst
eduardo-elizondo: Oops, that was a typo on my end! I mostly was referring to a topological 'destruction' not 'disorder' under the guarantee that we have some sort of topological sorting.

PR #2351, opened 2022-02-21
PEP 676: Display mailing list names for thread links in Discussions-To

AA-Turner:
xref https://github.com/python/peps/pull/2344#issuecomment-1046438174

A

### DIFF
--- a/pep_sphinx_extensions/pep_processor/transforms/pep_headers.py
+++ b/pep_sphinx_extensions/pep_processor/transforms/pep_headers.py
@@ -39,12 +39,12 @@ class PEPHeaders(transforms.Transform):
         # Extract PEP number
         value = pep_field[1].astext()
         try:
-            pep = int(value)
+            pep_num = int(value)
         except ValueError:
             raise PEPParsingError(f"'PEP' header must contain an integer. '{value}' is invalid!")
 
         # Special processing for PEP 0.
-        if pep == 0:
+        if pep_num == 0:
             pending = nodes.pending(pep_zero.PEPZero)
             self.document.insert(1, pending)
             self.document.note_pending(pending)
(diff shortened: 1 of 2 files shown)


AA-Turner:
Only affects the new build system, so merging now.

A

CAM-Gerlach (commented):
@AA-Turner Good thinking, but there's several issues I spotted with it—

* The email obfuscation (`_mask_email is getting (inconsistently) applied for the thread link display text, making it uglier and more difficult to visually parse, and this explicitly contradicts PEP 1 (even though it appears to be what the legacy code is currently doing, which appears to be unintentional as the rather outdated allow-list is not respected):

  > Note that email addresses in the Discussions-To header will not be obscured.

  Furthermore, given the usages are merely the link text for a `mailto` link, which is even easier for spambots to programmatically parse, it makes little sense to do so here.

* The fact that the visible link text is just an email address (obfuscated or not) lends itself to a clear point of confusion and uncertainty—as a user, when I see a linked email address, I expect it to be a `mailto:` link, not a HTTP URL to a thread on the web. Indeed, some links are `mailto:` (for those listing email lists) while others are `https:` (those to threads), with no way of users telling them apart and knowing what they go to, aside from hovering over each link and carefully reading it.

* While this is the legacy behavior, the old `mailto:` links send a new email to the mailing list with a title of the PEP's number, which is neither terribly useful (the user must be a member of the list, it creates a new, fragmented thread, and doesn't help users looking to join or to view the current thread), nor in keeping with the new PEP 1/PEP 12 guidance.

To fix this issues, I suggest the following two changes:

* Instead of an obfuscated email address as the display text for URLs, we can use the name of the list (preferably prettified a bit with capitalization, at least for common cases). This is just as "obfuscated" for spambots, but clearer to read and more obvious that it points to a `http` rather than `mailto` target, fixing the first two issues
* Instead of creating a `mailto` link for email addresses, at least for the main current/legacy python.org lists (Python-Dev, Python-Ideas, PEPs, Python-List, Typing-SIG and the legacy Distutils-SIG should cover almost everything), following the current guidance we could link instead to the main list page (e.g. `https://mail.python.org/mailman3/lists/python-dev.python.org/`), which solves all the issues above and has information about the list and direct links to the archive, subscribe, etc.

I'd be happy to contribute and test them for your review, if you'd prefer. Just lmk.

AA-Turner:
> email obfuscation... is getting ... applied for the thread link display text

Sorry I don't get this one -- https://python.github.io/peps/pep-0681/ seems to be fine?

> use the name of the list 

Sure

> link instead to the main list page

Less sure on this, but I'll try it out and see.

A

CAM-Gerlach:
> Sorry I don't get this one -- https://python.github.io/peps/pep-0681/ seems to be fine?

Its obfuscated, at least for me:

image

Meanwhile, [PEP 683](https://python.github.io/peps/pep-0683/) is obfuscated...

image

But [PEP 677](https://python.github.io/peps/pep-0677/) is not, despite having the same list address on the allow-list, but a mailto rather than a thread link

image

> Less sure on this, but I'll try it out and see.

As an alternative, for old-style direct emails, we could mirror the very common (as much if not more so than just the raw email address) `List-Name <example@example.com>` form, linking the list name to the list page, and the email with a mailto. Or, perhaps simpler and more practical, have the code convert this form to `List-Title <https://list.url>`_ to display consistently with the bare addresses.

Additionally, it would be idea if the code could code is not parsing the very common List-Name <

PR #2359, opened 2022-02-23
PEP 655: Motivation: Enables mixed key requiredness in the TypedDict alternative syntax

davidfstr:
Thanks for the related feedback @JelleZijlstra and Mehdi Drissi!

### DIFF
--- a/pep-0655.rst
+++ b/pep-0655.rst
@@ -74,7 +74,7 @@ Rationale
 =========
 
 One might think it unusual to propose syntax that prioritizes marking
-*required* keys rather than syntax for marking *potentially-missing* keys, as is
+*required* keys rather than *potentially-missing* keys, as is
 customary in other languages like TypeScript:
 
 .. code-block:: typescript
--- a/pep-0655.rst
+++ b/pep-0655.rst
@@ -55,12 +55,25 @@ a mix of both required and potentially-missing keys:
        title: str
        year: NotRequired[int]
 
+This PEP also makes it possible to define TypedDicts in the
+:pep:`alternative syntax <589#alternative-syntax>`
+with a mix of required and potentially-missing keys,
+which is not currently possible at all because the alternative syntax does
+not support inheritance:
+
+::
+
+   Actor = TypedDict('Actor', {
+       'name': str,
+       'in': NotRequired[List[str]],  # "in" is a keyword, so needs alternative syntax
+   })
+
 
 Rationale
 =========
 
 One might think it unusual to propose syntax that prioritizes marking
-*required* keys rather than syntax for *potentially-missing* keys, as is
+*required* keys rather than syntax for marking *potentially-missing* keys, as is
 customary in other languages like TypeScript:
 
 .. code-block:: typescript
--- a/pep-0655.rst
+++ b/pep-0655.rst
@@ -56,7 +56,7 @@ a mix of both required and potentially-missing keys:
        year: NotRequired[int]
 
 This PEP also makes it possible to define TypedDicts in the
-:pep:`alternative syntax <589#alternative-syntax>`
+:pep:`alternative functional syntax <589#alternative-syntax>`
 with a mix of required and potentially-missing keys,
 which is not currently possible at all because the alternative syntax does
 not support inheritance:
@@ -152,7 +152,7 @@ same time:
        year: NotRequired[Required[int]]  # ERROR
 
 
-The :pep:`alternative syntax <589#alternative-syntax>`
+The :pep:`alternative functional syntax <589#alternative-syntax>`
 for TypedDict also supports
 ``Required[]`` and ``NotRequired[]``:
 
--- a/pep-0655.rst
+++ b/pep-0655.rst
@@ -65,7 +65,8 @@ not support inheritance:
 
    Actor = TypedDict('Actor', {
        'name': str,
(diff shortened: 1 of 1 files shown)


pep-0655.rst
```diff
@@ -55,12 +55,25 @@ a mix of both required and potentially-missing keys:
...
+:pep:`alternative syntax <589#alternative-syntax>`
+with a mix of required and potentially-missing keys,
+which is not currently possible at all because the alternative syntax does
+not support inheritance:
+
+::
+
+   Actor = TypedDict('Actor', {
+       'name': str,
+       'in': NotRequired[List[str]],  # "in" is a keyword, so needs alternative syntax
```
AlexWaygood, line 68: I think "functional syntax" is clearer than "alternative syntax"

```suggestion
:pep:`alternative functional syntax <589#alternative-syntax>`
with a mix of required and potentially-missing keys,
which is not currently possible at all because the functional syntax does
not support inheritance:

::

   Actor = TypedDict('Actor', {
       'name': str,
       'in': NotRequired[List[str]],  # "in" is a keyword, so the functional syntax is necessary
```

davidfstr: > I think "functional syntax" is clearer than "alternative syntax"

Agreed. However the term of art, as put forth in the earlier PEP 589, is ["alternative syntax"](https://www.python.org/dev/peps/pep-0589/#alternative-syntax), so I'm trying to remain consistent with it.

Thus I would advocate leaving the original wording for the most part. However I wouldn't mind changing the first mention in this paragraph from "alternative syntax" to "alternative functional syntax", leaving the other usages as-is. What do you think @AlexWaygood ?

AlexWaygood: Hah, I think I've been hanging out in the [CPython test suite](https://github.com/python/cpython/blob/375a56bd4015596c0cf44129c8842a1fe7199785/Lib/test/test_typing.py#L4534) too much recently. (I've also seen it described [elsewhere](https://github.com/python/mypy/issues/2492#issuecomment-309157987) as the "functional syntax").

> I wouldn't mind changing the first mention in this paragraph from "alternative syntax" to "alternative functional syntax", leaving the other usages as-is.

Consistency with the original PEP is good — let's go with this 👍

pep-0655.rst
```diff
@@ -55,12 +55,25 @@ a mix of both required and potentially-missing keys:
        title: str
        year: NotRequired[int]
 
+This PEP also makes it possible to define TypedDicts in the
+:pep:`alternative functional syntax <589#alternative-syntax>`
+with a mix of required and potentially-missing keys,
+which is not currently possible at all because the alternative syntax does
+not support inheritance:
+
+::
```
CAM-Gerlach, line 64:
```suggestion
not support inheritance::
```

You can just do this, FYI

pep-0655.rst
```diff
@@ -55,12 +55,25 @@ a mix of both required and potentially-missing keys:
...
+
+   Actor = TypedDict('Actor', {
+       'name': str,
+       'in': NotRequired[List[str]],  # "in" is a keyword, so needs alternative syntax
```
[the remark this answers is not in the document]
CAM-Gerlach:
```suggestion
       # "in" is a keyword, so the functional syntax is necessary
       'in': NotRequired[List[str]],
```

To avoid >79 character line length per PEP 8, you could do this...

pep-0655.rst
```diff
@@ -55,12 +55,25 @@ a mix of both required and potentially-missing keys:
...
+   })
+
 
 Rationale
 =========
 
 One might think it unusual to propose syntax that prioritizes marking
-*required* keys rather than syntax for *potentially-missing* keys, as is
+*required* keys rather than syntax for marking *potentially-missing* keys, as is
```
CAM-Gerlach, line 76: Grammar/semantic issue—currently, due to the use of "prioritizes" between the "than" and the intended target, "syntax", this sentence is naively read as follows:

> One might think it unusual to propose syntax that prioritizes (marking *required* keys) rather than (syntax for marking *potentially-missing* keys), as is

when it appears, from the use of the otherwise semantically out of place "syntax for marking", it is intended to be read as (which may require a few readings to parse correctly):

> One might think it unusual to propose (syntax that prioritizes marking *required* keys) rather than (syntax for marking *potentially-missing* keys), as is

To fix this, I suggest going with the "prioritize" target and eliminating the redundant/repetitive "syntax for marking", which is simpler and more concise,

```suggestion
*required* keys rather than *potentially-missing* keys, as is
```

CAM-Gerlach (commented):
A few minor editorial nits

pep-0655.rst
davidfstr: Aye. I prefer keeping the code block element independent of the preceding paragraph element.

Such separation is also consistent with Pandoc's output, which I used to generate most of the RST syntax originally (from Markdown).

JelleZijlstra (approved):

davidfstr:
All pending feedback is now integrated.

(I'm really liking this [new?] GitHub feature were specific diff changes can be proposed.)

CAM-Gerlach:
> (I'm really liking this [new?] GitHub feature where specific diff changes can be proposed.)

It has actually been around for a few years now, but yes, it is extremely useful for both authors and reviewers, particularly for applications such as this.

PR #2363, opened 2022-02-25
PEP 676: Use `docutils.nodes.Node.findall`

AA-Turner:
For forwards compatability.

A

### DIFF
--- a/generate_rss.py
+++ b/generate_rss.py
@@ -6,6 +6,7 @@ import datetime
 import email.utils
 from pathlib import Path
 
+import docutils
 from docutils import frontend
 from docutils import nodes
 from docutils import utils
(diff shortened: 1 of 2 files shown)


generate_rss.py
```diff
@@ -19,6 +20,14 @@ def _format_rfc_2822(dt: datetime.datetime) -> str:
     return email.utils.format_datetime(dt, usegmt=True)
 
 
+# Monkeypatch nodes.Node.findall for forwards compatability
+if docutils.__version_info__ < (0, 18):
```
CAM-Gerlach, line 24: Just to confirm, @AA-Turner — the TODO comment mentions docutils >= 0.18.1. Should this be

```python
 if docutils.__version_info__ < (0, 18, 1):
```

or is there a reason for the discrepancy?

AA-Turner: Ahh, it should've been `<=` but sure the triple is more explicit. Fixed in https://github.com/python/peps/commit/8b31a9365f59125fe05285d244470a22a8294ff3

A

CAM-Gerlach: Thanks

PR #2366, opened 2022-02-26
Add PEP titles and sections as reference title text

AA-Turner:
This adds the PEP number, title, and section target title to the title text of the link element.

A

### DIFF
--- a/pep_sphinx_extensions/pep_processor/parsing/pep_role.py
+++ b/pep_sphinx_extensions/pep_processor/parsing/pep_role.py
@@ -8,7 +8,12 @@ class PEPRole(roles.ReferenceRole):
     def run(self) -> tuple[list[nodes.Node], list[nodes.system_message]]:
         # Get PEP URI from role text.
         pep_str, _, fragment = self.target.partition("#")
-        pep_num = int(pep_str)
+        try:
+            pep_num = int(pep_str)
+        except ValueError:
+            msg = self.inliner.reporter.error(f'invalid PEP number {self.target}', line=self.lineno)
+            prb = self.inliner.problematic(self.rawtext, self.rawtext, msg)
+            return [prb], [msg]
         pep_base = self.inliner.document.settings.pep_url.format(pep_num)
         if fragment:
             ref_uri = f"{pep_base}#{fragment}"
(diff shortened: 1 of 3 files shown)


pep_sphinx_extensions/pep_processor/transforms/pep_references.py
```diff
@@ -0,0 +1,36 @@
...
+            # Cache target_ids
+            env = self.document.settings.env
+            try:
+                target_ids = env.document_ids[filename]
+            except KeyError:
+                env.document_ids[filename] = target_ids = env.get_doctree(filename).ids
+
+            # Create title text string. We hijack the 'reftitle' attribute so
+            # that we don't have to change things in the HTML translator
+            node["reftitle"] = env.titles[filename].astext().replace(" – ", ", ")
```
CAM-Gerlach, line 32:
```suggestion
            node["reftitle"] = env.titles[filename].astext()
```

Is there a reason ` - ` isn't just retained here, to be consistent with actual displayed PEP titles, and a comma is substituted instead? I'm not sure how this is motivated by the semantics here; if we do need to pick a different character, a colon would be more appropriate.

pep_sphinx_extensions/pep_processor/parsing/pep_role.py
```diff
@@ -1,14 +1,28 @@
...
+        else:
+            ref_uri = pep_base
+        if self.has_explicit_title:
+            title = self.title
+        else:
+            title = f"PEP {pep_num}"
+        try:
+            return [nodes.reference("", title,
+                                    internal=True, refuri=ref_uri, classes=["pep"],
+                                    _title_tuple=(pep_num, fragment))], []
```
CAM-Gerlach, line 24: Style nit—IMO, the current structure is potentially confusing, hard to read and follow, and I find it visually ugly. I suggest:

```suggestion
            reference_node = nodes.reference(
                    "", title, internal=True, refuri=ref_uri, classes=["pep"],
                    _title_tuple=(pep_num, fragment))
            return [reference_node], []
```

or even better,

```suggestion
            reference_node = nodes.reference(
                "",
                title,
                internal=True,
                refuri=ref_uri,
                classes=["pep"],
                _title_tuple=(pep_num, fragment),
            )
            return [reference_node], []
```

CAM-Gerlach (commented):
I spot tested this on a few different PEPs and PEP link types, and everything worked great—I particularly liked the inclusion of the section name; that was a very thoughtful and useful touch. I only had one small comment on the formatting of the titletext, and one nitpick on formatting the code.

pep_sphinx_extensions/pep_processor/transforms/pep_references.py
CAM-Gerlach: Yes, but context is important here. The footnotes contained commas because they were formatted as _citations_, whereas this is used as _title-text_ of a link, which is intended to display the _title_ of the linked document—which uses a dash. So I think we should just follow precedent and be consistent with the actual title, leaving it as-is.

I don't really see the issue with appending the section, since you were thoughtful to clearly separate it with the unique and unambiguous section symbol (`§`) and considering the standard format is `title § section`, it would be odd to not use the actual title.

CAM-Gerlach (approved):
LGTM, @AA-Turner , this is a terrific improvement. This should be good to merge at your convenience.

CAM-Gerlach:
:point_up: Just FYI, @warsaw , since IIRC you requested this.

Also, @AA-Turner would it be straightforward to add similar titletext popups in a future PR for footnotes with the footnote content (like Wikipedia has), to avoid users having to click around to find it?

PR #2375, opened 2022-03-02
PRS: Update Post-History processing for inline links

AA-Turner:
For Post-History fields of the form proposed in https://github.com/python/peps/pull/2358#issuecomment-1055622370

A


AA-Turner:
> Also, when I attempt to run this locally with a PEP containing a field formatted as specified, I get a TypeError on the union between the two types in the return type annotation.

`conda install "python==3.10"` would fix this, or I could add the future import.

> this processing isn't actually applied until if and when PEP 676 is accepted and fully implemented

True, although we could backport it as we have a few other changes, it would be fairly easy.

A

CAM-Gerlach:
> conda install "python==3.10" would fix this, or I could add the future import.

I mean, we currently declare 3.9 compat. I'm not sure we should require bleeding edge Python, especially when the cost of a future import is fairly trivial.

> True, although we could backport it as we have a few other changes, it would be fairly easy.

Yeah, though it would be effort that will have little long-term value.

CAM-Gerlach:
Also, what does "PRS" stand for? I've never heard that abbreviation before...

AA-Turner:
> what does "PRS" stand for

"PEP Rendering System"

> effort that will have little long-term value

Agree. Hopefully 676 will have a decision soon.

> cost of a future import is fairly trivial

Yes, I'll add one. 3.10's "bleeding edge" status can be argued later, I'm not requiring 3.11 quite yet ;-)

A

JelleZijlstra:
PEP Rendering System? Programming Reusable Sphinxes? Python-Related Services?

pep_sphinx_extensions/pep_processor/transforms/pep_headers.py
```diff
@@ -115,3 +117,24 @@ def _pretty_thread(text: nodes.Text) -> nodes.Text:
...
+        try:
+            # if Post-History has no links, ``pair.split(maxsplit=1)``
+            # will raise ValueError
+            date, uri = pair.split(maxsplit=1)
+            node = nodes.reference("",
+               date.strip(),
+               refuri=uri.strip(" \f\n\r\t><"),
+               internal=False
+           )
+        except ValueError:
```
JelleZijlstra, line 135: Should we instead fail if the field doesn't parse properly? It would be good to give that feedback to authors.

JelleZijlstra: True, I guess we need some way to avoid erroring on anything that's already there, while alerting on things that look like the desired syntax but aren't quite the same.

Also, wouldn't this get a false positive on PEP 3135:
```
pep-3135.txt:Post-History: 28-Apr-2007, 29-Apr-2007 (1), 29-Apr-2007 (2), 14-May-2007, 12-Mar-2009
```

JelleZijlstra: Also PEP 551:
```
pep-0551.rst:Post-History: 24-Aug-2017 (security-sig), 28-Aug-2017 (python-dev)
```

AA-Turner: Perhaps a normalisation pass is needed here. (Or ignoring anything that doesn't start with "http")

A

JelleZijlstra: I wouldn't mind us changing the handful of PEPs with very eccentric Post-History formats (648, 538, 467, 367, 311 also stand out; and 424 and 428 have bare URLs).

CAM-Gerlach:
> "PEP Rendering System"

If we're going to use prefixes for infra PRs, can we not use cryptic, bespoke, made-up initialisms instead of widely-used plain language abbreviations like "Infra" here? Particularly ones that are easy to confuse with "PRs" (which is how I read it). [Acronyms Seriously Suck](https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764) (and I work for NASA, I should know).

AA-Turner:
I broke up the Post-History changes into four commits with rough categories, and also implemented the change in `pep2html.py`

A

CAM-Gerlach (changes requested):
At least per my testing, the custom processing is still not triggering for the legacy script, but is for the same PEP for the Sphinx system, after pulling this PR branch and re-running both—the links are still rendered in plain text, just as before.

Per the discussion on #2358, to move forward I propose we:

* Merge #2358 specifying a simplified, structured subset of the reST syntax for links
* Drop the bespoke processing part of this PR
* Retain the normalizations in this PR, converting the few that were linkified to the reST-based format, and merge that
* I follow up with a PR (already prototyped) adding automated linting checks for it, like for the other headers

Alternatively, if we do decide to go with this approach instead, we should be more strict about how we handle errors so they don't pass silently, and I'll need to do more testing/checking of various scenarios.

CAM-Gerlach:
Hey @AA-Turner , given our discussion on #2358 , looks like this needs dropping the first three commits for now, and updating the links in the "extra information" and "bare hyperlinks" commits to use the slightly different format. Would you prefer to take care of that, or would you like me to here? I could also open another PR, but I'd rather avoid pinging everybody again.

Once this is merged, I have a PR tested and ready to go revising the existing header validation checks to ensure both this header and the others are correct and programmatically parsable, as well as clarifying a few inconsistencies/ambiguities in this regard in PEP 1/12/the templates and updating a handful of remaining PEPs that don't follow the rules. This will allow both us and others to do any number of useful things programmatically in the future, both here and externally, without further messing with the existing PEPs or requiring workarounds or hardcoding.

AA-Turner:
This is waiting for #2358 to be resolved (as it now uses the reST link format proposed by that PR).

A

pep-0428.txt
```diff
@@ -8,7 +8,7 @@ Type: Standards Track
 Content-Type: text/x-rst
 Created: 30-Jul-2012
 Python-Version: 3.4
-Post-History: https://mail.python.org/pipermail/python-ideas/2012-October/016338.html
+Post-History: `05-Oct-2012 <https://mail.python.org/pipermail/python-ideas/2012-October/016338.html>`__`
```
CAM-Gerlach, line 11:
```suggestion
Post-History: `05-Oct-2012 <https://mail.python.org/pipermail/python-ideas/2012-October/016338.html>`__
```

CAM-Gerlach (commented):
One typo, otherwise :heavy_check_mark:

PR #2377, opened 2022-03-02
PEP 678: Clarify how the notes tuple is updated and when it is copied

Zac-HD:
Following https://github.com/python/steering-council/issues/105#issuecomment-1056023519.

### DIFF
--- a/pep-0678.rst
+++ b/pep-0678.rst
@@ -140,10 +140,10 @@ are collecting multiple exception objects to handle together. [1]_
 Specification
 =============
 
-``BaseException`` gains a new read-only attribute ``__notes__``, an initially
-empty tuple, and a new method ``.add_note(note: str)``.  ``note`` is added to
-the exception's notes, which appear in the standard traceback after the
-exception string. A ``TypeError`` is raised if ``note`` is not a string.
+``BaseException`` gains a new new method ``.add_note(note: str)``. Notes are
+exposed as a tuple via the read-only attribute ``__notes__``, which is backed
+by an internal list, and appear in the standard traceback after the exception
+string.  ``.add_note()`` raises ``TypeError`` if ``note`` is not a string.
 
 ``del err.__notes__`` clears the contents of the ``__notes__`` attribute,
 leaving it an empty tuple as if ``.add_note()`` had never been called.  This
@@ -319,8 +319,8 @@ proposed ``__notes__`` semantics, but this would be rarely and inconsistently
 applicable.
 
 
-Store notes in ``ExceptionGroup``\ s
-------------------------------------
+Store notes in ``ExceptionGroup``\ s, but not other ``Exception``\ s
+--------------------------------------------------------------------
 Initial discussions proposed making a more focussed change by thinking about
 how to associate messages with the nested exceptions in ``ExceptionGroup`` s,
 such as a list of notes or mapping of exceptions to notes.  However, this would
--- a/pep-0678.rst
+++ b/pep-0678.rst
@@ -143,9 +143,9 @@ Specification
 
 ``BaseException`` gains a new new method ``.add_note(note: str)``. Notes are
 exposed as a tuple via the read-only attribute ``__notes__``, and appear in
-the standard traceback after the exception string.  ``.add_note()`` replaces
+the standard traceback after the exception string.  ``.add_note(note)`` replaces
 ``__notes__`` with a new tuple equal to ``__notes__ + (note,)``, if ``note``
-is a string, raising ``TypeError`` otherwise.
+is a string, and raises ``TypeError`` otherwise.
 
 ``del err.__notes__`` clears the contents of the ``__notes__`` attribute,
 leaving it an empty tuple as if ``.add_note()`` had never been called.  This
--- a/pep-0678.rst
+++ b/pep-0678.rst
@@ -319,15 +319,16 @@ proposed ``__notes__`` semantics, but this would be rarely and inconsistently
 applicable.
 
 
-Store notes in ``ExceptionGroup``\ s, but not other ``Exception``\ s
---------------------------------------------------------------------
-Initial discussions proposed making a more focussed change by thinking about
-how to associate messages with the nested exceptions in ``ExceptionGroup`` s,
-such as a list of notes or mapping of exceptions to notes.  However, this would
-force a remarkably awkward API and retains a lesser form of the
-cross-referencing problem discussed under "use ``print()``" above; if this PEP
-is rejected we prefer the status quo. Finally, of course, ``__notes__`` are
-not only useful with ``ExceptionGroup``\ s!
+Don't attach notes to ``Exception``\ s, just store them in ``ExceptionGroup``\ s
+--------------------------------------------------------------------------------
+The initial motivation for this PEP was to associate a note with each error
+in an ``ExceptionGroup``.  At the cost of a remarkably awkward API and the
+cross-referencing problem discussed under "use ``print()``" above, this
+use-case could be supported by storing notes on the ``ExceptionGroup``
+instance instead of on each exception it contains.
+
+We believe that the cleaner interface, and other use-cases described above,
+are sufficient to justify the more general feature proposed by this PEP.
 
 
 
--- a/pep-0678.rst
+++ b/pep-0678.rst
@@ -141,9 +141,9 @@ Specification
 =============
 
 ``BaseException`` gains a new new method ``.add_note(note: str)``. Notes are
-exposed as a tuple via the read-only attribute ``__notes__``, which is backed
-by an internal list, and appear in the standard traceback after the exception
-string.  ``.add_note()`` raises ``TypeError`` if ``note`` is not a string.
+exposed as a tuple via the read-only attribute ``__notes__``, and appear in
+the standard traceback after the exception string.  ``.add_note()`` raises
+``TypeError`` if ``note`` is not a string.
 
 ``del err.__notes__`` clears the contents of the ``__notes__`` attribute,
 leaving it an empty tuple as if ``.add_note()`` had never been called.  This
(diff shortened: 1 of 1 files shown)


pep-0678.rst
```diff
@@ -140,10 +140,10 @@ are collecting multiple exception objects to handle together. [1]_
 Specification
 =============
 
-``BaseException`` gains a new read-only attribute ``__notes__``, an initially
-empty tuple, and a new method ``.add_note(note: str)``.  ``note`` is added to
-the exception's notes, which appear in the standard traceback after the
-exception string. A ``TypeError`` is raised if ``note`` is not a string.
+``BaseException`` gains a new new method ``.add_note(note: str)``. Notes are
+exposed as a tuple via the read-only attribute ``__notes__``, which is backed
+by an internal list, and appear in the standard traceback after the exception
```
JelleZijlstra, line 145: I don't think the PEP needs to say what the internal implementation is.

Zac-HD: Fair enough, taken that back out - I've answered Brett on the SC issue anyway.

pep-0678.rst
```diff
@@ -319,8 +319,8 @@ proposed ``__notes__`` semantics, but this would be rarely and inconsistently
 applicable.
 
 
-Store notes in ``ExceptionGroup``\ s
-------------------------------------
+Store notes in ``ExceptionGroup``\ s, but not other ``Exception``\ s
```
CAM-Gerlach, line 322: Besides being long and unwieldly, this heading is rather confusing to me who's been passively following this PEP but has not delved into the nitty gritty details of its content. as a non-expert on its content. Specifically, is it saying that that other ``Exceptions``s do not have ``notes`` stored in them (which I *think* is the case, from having read the SC issue)? Or that ``Exceptions`` are not stored in ``ExceptionGroup``s? I'm assuming the former, but the present syntax is ambiguous and could confuse someone not reading it carefully multiple times, or not particularly familiar with the new ``ExceptionGroups`` and ``notes`` constructs.

CAM-Gerlach (commented):
One comment on clarity.

pep-0678.rst
Zac-HD: ...yeah, that section hasn't aged well; I'll rewrite entirely to clarify.

pep-0678.rst
```diff
@@ -319,15 +319,16 @@ proposed ``__notes__`` semantics, but this would be rarely and inconsistently
...
-such as a list of notes or mapping of exceptions to notes.  However, this would
-force a remarkably awkward API and retains a lesser form of the
-cross-referencing problem discussed under "use ``print()``" above; if this PEP
-is rejected we prefer the status quo. Finally, of course, ``__notes__`` are
-not only useful with ``ExceptionGroup``\ s!
+Don't attach notes to ``Exception``\ s, just store them in ``ExceptionGroup``\ s
+--------------------------------------------------------------------------------
+The initial motivation for this PEP was to associate a note with each error
+in an ``ExceptionGroup``.  At the cost of a remarkably awkward API and the
+cross-referencing problem discussed under "use ``print()``" above, this
```
CAM-Gerlach, line 326: Especially since it is a number of sections above, I suggest actually linking the appropriate section, e.g. by adding an explicit reference right above the print() section heading, i.e.

```rst
.. _print_idea:

Use print() ...
-------------
```

then adding the appropriate link here, i.e.

```suggestion
cross-referencing problem discussed `above <print_idea_>`_, this
```

pep-0678.rst
```diff
@@ -140,10 +140,10 @@ are collecting multiple exception objects to handle together. [1]_
...
 =============
 
-``BaseException`` gains a new read-only attribute ``__notes__``, an initially
-empty tuple, and a new method ``.add_note(note: str)``.  ``note`` is added to
-the exception's notes, which appear in the standard traceback after the
-exception string. A ``TypeError`` is raised if ``note`` is not a string.
+``BaseException`` gains a new new method ``.add_note(note: str)``. Notes are
+exposed as a tuple via the read-only attribute ``__notes__``, and appear in
+the standard traceback after the exception string.  ``.add_note()`` raises
+``TypeError`` if ``note`` is not a string.
```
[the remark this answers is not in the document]
Zac-HD: Hmm, that's a little surprising because it makes the obvious test for `ExceptionGroup.derive()` (`a.__notes__ is b.__notes__`) fail.  Would it be much more annoying to have `add_note()` do the moral equivalent of `e.__dict__["__notes__"] = e.__notes__ + (note,)`?

(I assume you mean *access*, not *call*)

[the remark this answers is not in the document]
iritkatriel: I can do that. Then to answer Brett’s question the pep needs to say that a new tuple is created when we add/clear notes.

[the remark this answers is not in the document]
Zac-HD: OK, let's go with that approach.

(I know it's *possible* to mutate tuples, but it's still pretty weird, and would prevent a pure-Python backport)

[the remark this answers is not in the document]
iritkatriel: Sure, I’m not suggesting we do that, just that Brett’s point was that the pep should be clear that we’re not doing it.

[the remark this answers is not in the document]
iritkatriel: I've updated the PR now.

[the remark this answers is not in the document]
JelleZijlstra: This is a tangent, but I think even in C you wouldn't be able to grow a tuple in place; https://docs.python.org/3.10/c-api/tuple.html#c._PyTuple_Resize exists but to grow the tuple it has to reallocate it in a different place.

JelleZijlstra (approved):

Zac-HD:
OK, I've pushed a (I sincerely hope it's) final half-sentence specifying that `add_note(note)` behaves like `__notes__ += (note,)` (allocates a new tuple on change, not access, never mutates in place).  Should be ready to merge.

pep-0678.rst
```diff
@@ -140,10 +141,11 @@ are collecting multiple exception objects to handle together. [1]_
 Specification
 =============
 
-``BaseException`` gains a new read-only attribute ``__notes__``, an initially
-empty tuple, and a new method ``.add_note(note: str)``.  ``note`` is added to
-the exception's notes, which appear in the standard traceback after the
-exception string. A ``TypeError`` is raised if ``note`` is not a string.
+``BaseException`` gains a new new method ``.add_note(note: str)``. Notes are
+exposed as a tuple via the read-only attribute ``__notes__``, and appear in
+the standard traceback after the exception string.  ``.add_note()`` replaces
```
iritkatriel, line 146:
```suggestion
the standard traceback after the exception string.  ``.add_note(note)`` replaces
```

pep-0678.rst
```diff
@@ -140,10 +141,11 @@ are collecting multiple exception objects to handle together. [1]_
...
+``__notes__`` with a new tuple equal to ``__notes__ + (note,)``, if ``note``
+is a string, raising ``TypeError`` otherwise.
```
iritkatriel, line 148:
```suggestion
is a string, and raises ``TypeError`` otherwise.
```

CAM-Gerlach (approved):
Mean to do so earlier, but LGTM from a PEP editor perspective, pending approval by @iritkatriel .

iritkatriel (approved):
LGTM

gvanrossum:
I'm moving some discussion here since I don't think it belongs in https://github.com/python/steering-council/issues/105 (I don't like having a substantive discussion in the SC tracker).

@iritkatriel wrote:
> This was the actual change in the pep (https://github.com/python/peps/pull/2377/files).

(Note that that is this PR.)

> Do you think it’s over specifying how notes is updated?

I really want this to grow the size of exception objects by only one pointer.

The current formulation (as of this PR) then pretty much forces us into this implementation:

```py
class BaseException:
    ...
    __notes: tuple[str, ...] | None = None  # Private variable

    def add_note(self, note: str):
        if self.__notes is None:
            self.__notes = (note,)
        else:
            self.__notes = (*self.__notes__, note)

    @property
    def __notes__(self) -> tuple[str, ...] | None:
        return self.__notes

    @__notes__.deleter
    def __notes__(self):
        self.__notes = None
```

I'd wish to *allow* `add_note()` to use an internal data structure that's a list or a linked list or whatever is deemed most efficient given usage patterns, and have `__notes__` construct a fresh tuple on the fly *without* having to cache it (since the cache would extend the size of the object by another pointer, i.e. 8 bytes). Never mind that's unlikely what the initial implementation will be, changing the implementation shouldn't violate the PEP. But currently the PEP text seems to imply that consecutive calls to the property (without intervening `add_notes()` call) are required to return the same tuple object (which my implementation sketch does, but which I think the PEP needn't require).

JelleZijlstra:
@Zac-HD wrote above:

> Hmm, that's a little surprising because it makes the obvious test for `ExceptionGroup.derive()` (`a.__notes__ is b.__notes__`) fail.

But I'm not sure what this test is or why it's obvious.

Zac-HD:
If you split an `ExceptionGroup`, it should have the same notes; so people will check that and some will use `is` instead of `==`.  I'd left that unspecified until I started getting asked about it - maybe we should instead say explicitly that it's not specified?

gvanrossum:
People shouldn’t compare tuple any identity any more than they should integers.

iritkatriel:
In that case we can remove the last commit from https://github.com/python/cpython/pull/31317  and then we have the notes in a list and we create a new tuple each time it’s accessed.

The pep doesn’t need to say all of this, but to Brett’s point it needs to say that the tuple you got from ``.__notes__ `` is not mutated by a subsequent .add_note().  Is this correct?

JelleZijlstra:
> but to Brett’s point it needs to say that the tuple you got from .__notes__ is not mutated by a subsequent .add_note().

That feels unnecessary. Tuples can't be mutated (unless you do something really crazy in C), so there's no need to spell that out.

gvanrossum:
> That feels unnecessary. Tuples can't be mutated (unless you do something really crazy in C), so there's no need to spell that out.

That was my starting point too, but Brett pushed back specifically because "in C you can do anything." So it seems we need to somehow make that clear without forcing an implementation.

I personally think it's fine to put notes in a list and tuplify on demand without caching. (But I also think it's fine to put them in a tuple, creating a fresh tuple in add_note().)

PR #2382, opened 2022-03-05
PEP 681: Fix minor formatting and typographical issues

davidfstr:
Please see individual commit messages for explanation of each change made.

### DIFF
--- a/pep-0681.rst
+++ b/pep-0681.rst
@@ -180,7 +180,8 @@ these parameters accepts a bool argument, and it must be possible for
 the bool value (``True`` or ``False``) to be statically evaluated.
 
 * ``eq``.  ``order``, ``frozen``, ``init`` and ``unsafe_hash`` are parameters
-  supported in the stdlib dataclass, with meanings defined in :pep:`557 <557#id7>`.
+  supported in the stdlib dataclass, with meanings defined in 
+  :pep:`PEP 557 <557#id7>`.
 * ``hash`` is an alias for the ``unsafe_hash`` parameter.
 * ``kw_only`` and ``slots`` are parameters supported in the stdlib dataclass,
   first introduced in Python 3.10.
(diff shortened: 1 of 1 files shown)


davidfstr:
CC @debonte (PEP Author)

hugovk:
https://github.com/python/peps/pull/2382/commits/2e8212bb57d4c76c97eac6513e1524a2bb432a71:

> * I believe '''-style headers are also recommended by the PEP rST template
  for this level of header, but I haven't confirmed.

Correct! 👍 

```rst
First-Level Title
=================

Second-Level Title
------------------

Third-Level Title
''''''''''''''''' 
```

https://python.github.io/peps/pep-0012/#section-headings

JelleZijlstra (approved):

CAM-Gerlach (commented):
LGTM, thanks! Some clarifications:

> [Nit: Adjust PEP reference to include the word "PEP"](https://github.com/python/peps/pull/2382/commits/d0ab16a4e455ccb629d3db2f6b74dab73c98688c)

Just to make sure we're clear, this is only needed for PEPs that have custom inline link text, which by extension includes those linking to sections (as here). For "normal" `` :pep:`1` `` links, "PEP" is prepended automatically as part of the generated inline link text.

> [Remove gremlin character <0x0c>](https://github.com/python/peps/pull/2382/commits/83b4d2de9bec29603691359f850603da7924d804)

The `FF` (form feed) character is needed for Emacs to recognize the Emacs editor configuration block that follows it at the end of the file. Previously, this was required and included in all PEPs, but as it is effectively obsolete (given only a few PEP editors/authors still use Emacs, and those that do pretty much all use `rst-mode` which renders this unnecessary) this requirement was removed in #2374 .

Without the `FF` character the following block is (AFAIK) useless, so if you eliminate that, the Emacs config comment following it should be removed as well, or otherwise both should be left alone (and the commit dropped).

debonte:
@davidfstr, based on https://github.com/python/peps/pull/2374 and @CAM-Gerlach's comment above, I'd say we should remove the Emacs configuration block at the end of the file.

Thanks for these changes!

AA-Turner (approved):

davidfstr:
I believe all feedback has been integrated and this PR is ready to merge. 👍

PR #2386, opened 2022-03-07
PEP 681: Remove discussion of alternate form

debonte:
Removed the "Alternate form" section and the mention of the alternate form in the "Reference Implementation" section.

Now that the `dataclass_transform` decorator is available in `typing_extensions`, the fact that Pyright supports the dundered alternate form (`__dataclass_transform__`) is not particularly interesting and has caused some reader confusion.

This needed to be removed before requesting PEP approval anyway.

cc: @davidfstr

### DIFF
--- a/pep-0681.rst
+++ b/pep-0681.rst
@@ -532,36 +532,6 @@ decorated with ``dataclass_transform`` is in use.
   `ignored by dataclass mechanisms <#class-var_>`_.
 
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
Agree, this was useful as an experimental measure but the discussion in the PEP is now confusing.

PR #2394, opened 2022-03-09
PEP 683: Drop Eddie from CODEOWNERS

ericsnowcurrently:


CAM-Gerlach:
For reference, a fix for the error introduced in #2387 .

CAM-Gerlach:
Thanks for the quick fix

PR #2398, opened 2022-03-09
legally dropped the middle name and suffix.

ewdurbin:
various government agencies all agree my name doesn't have those bits now :)


JelleZijlstra (approved):

encukou:
> various government agencies all agree my name doesn't have those bits now :)

Congratulations!

PR #2413, opened 2022-03-11
PRS: Fix RSS generation and resolve all warnings

AA-Turner:
closes #2407, closes #2406

This builds on Hugo's fix to also fix everything else.

A

### DIFF
--- a/generate_rss.py
+++ b/generate_rss.py
@@ -79,10 +79,17 @@ def pep_abstract(full_path: Path) -> str:
 
 def main():
     # get the directory with the PEP sources
+    in_dir = Path(__file__).parent
+
+    # get the output directory for target HTML files
     out_dir = Path(__file__).parent / "build"
 
-    # get list of peps with creation time (from "Created:" string in pep source)
-    peps_with_dt = sorted((pep_creation(path), path) for path in out_dir.glob("pep-????.*"))
+    # get list of source peps
+    peps = list(in_dir.glob('pep-*.txt'))
+    peps.extend(in_dir.glob('pep-*.rst'))
+
+    # sort peps by creation time (from "Created:" string in pep source)
+    peps_with_dt = sorted((pep_creation(path), path) for path in peps)
 
     # generate rss items for 10 most recent peps
     items = []
--- a/generate_rss.py
+++ b/generate_rss.py
@@ -38,6 +38,8 @@ line_cache: dict[Path, dict[str, str]] = {}
 
 # Monkeypatch PEP and RFC reference roles to match Sphinx behaviour
 EXPLICIT_TITLE_RE = re.compile(r'^(.+?)\s*(?<!\x00)<(.*?)>$', re.DOTALL)
+
+
 def _pep_reference_role(role, rawtext, text, lineno, inliner,
                         options={}, content=[]):
     matched = EXPLICIT_TITLE_RE.match(text)
@@ -65,6 +67,8 @@ def _pep_reference_role(role, rawtext, text, lineno, inliner,
         ref += "#" + fragment
     roles.set_classes(options)
     return [nodes.reference(rawtext, title, refuri=ref, **options)], []
+
+
 def _rfc_reference_role(role, rawtext, text, lineno, inliner,
                         options={}, content=[]):
     matched = EXPLICIT_TITLE_RE.match(text)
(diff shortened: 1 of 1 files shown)


generate_rss.py
```diff
@@ -31,6 +36,63 @@ def findall(self, *args, **kwargs):
 entry.formatRFC2822 = feed.formatRFC2822 = _format_rfc_2822
 line_cache: dict[Path, dict[str, str]] = {}
 
+# Monkeypatch PEP and RFC reference roles to match Sphinx behaviour
+EXPLICIT_TITLE_RE = re.compile(r'^(.+?)\s*(?<!\x00)<(.*?)>$', re.DOTALL)
```
CAM-Gerlach, line 40: As a script-level CONSTANT, shouldn't this be at the top? If not, then per PEP 8

```suggestion
EXPLICIT_TITLE_RE = re.compile(r'^(.+?)\s*(?<!\x00)<(.*?)>$', re.DOTALL)


```

generate_rss.py
```diff
@@ -31,6 +36,63 @@ def findall(self, *args, **kwargs):
...
+        prb = inliner.problematic(rawtext, rawtext, msg)
+        return [prb], [msg]
+    # Base URL mainly used by inliner.pep_reference; so this is correct:
+    ref = (inliner.document.settings.pep_base_url
+           + inliner.document.settings.pep_file_url_template % pepnum)
+    if fragment:
+        ref += "#" + fragment
+    roles.set_classes(options)
+    return [nodes.reference(rawtext, title, refuri=ref, **options)], []
+def _rfc_reference_role(role, rawtext, text, lineno, inliner,
```
CAM-Gerlach, line 68:
```suggestion


def _rfc_reference_role(role, rawtext, text, lineno, inliner,
```

PEP 8 (and much easier to read)

generate_rss.py
```diff
@@ -31,6 +36,63 @@ def findall(self, *args, **kwargs):
...
+            line=lineno)
+        prb = inliner.problematic(rawtext, rawtext, msg)
+        return [prb], [msg]
+    ref = (inliner.document.settings.rfc_base_url + inliner.rfc_url % rfcnum)
+    if fragment:
+        ref += "#" + fragment
+    roles.set_classes(options)
+    return [nodes.reference(rawtext, title, refuri=ref, **options)], []
+roles.register_canonical_role("pep-reference", _pep_reference_role)
+roles.register_canonical_role("rfc-reference", _rfc_reference_role)
```
CAM-Gerlach, line 94:
```suggestion


roles.register_canonical_role("pep-reference", _pep_reference_role)
roles.register_canonical_role("rfc-reference", _rfc_reference_role)
```

Again, per PEP 8 can we have some space here? Otherwise it reads as part of the function body.

generate_rss.py
```diff
@@ -31,6 +36,63 @@ def findall(self, *args, **kwargs):
 entry.formatRFC2822 = feed.formatRFC2822 = _format_rfc_2822
 line_cache: dict[Path, dict[str, str]] = {}
 
+# Monkeypatch PEP and RFC reference roles to match Sphinx behaviour
+EXPLICIT_TITLE_RE = re.compile(r'^(.+?)\s*(?<!\x00)<(.*?)>$', re.DOTALL)
+def _pep_reference_role(role, rawtext, text, lineno, inliner,
```
CAM-Gerlach, line 41: Do we need to copy/paste this here (at least, I'm assuming that's what we're doing), or can we import it from its current location or a common `utils` module?

CAM-Gerlach (commented):
Thanks. One high-level comment and a couple nits.

generate_rss.py
AA-Turner: > As a script-level CONSTANT, shouldn't this be at the top? 

I'm working to add support for this upstream (in Docutils) -- this is intended as a temporary monkeypatch. All the code was copied from the same backport I used in `pep2html`. So it is at the top of the block, allowing easier deletion all at once when we can.

A

generate_rss.py
AA-Turner: From my response above (didn't see this question):

> All the code was copied from the same backport I used in `pep2html`. 

A

generate_rss.py
CAM-Gerlach: Okay, that makes sense.

generate_rss.py
CAM-Gerlach: Hmm, I guess that makes sense.

CAM-Gerlach (approved):
LGTM at least from my end; @hugovk might have more substantive comments since he's more familiar with this script.

AA-Turner:
I'll merge this now in the interests of pragmatism, as RSS is currently broken and this is a fix. We can change in follow-ups if need be.

A

PR #2421, opened 2022-03-12
Support redirects from non-padded PEP numbers

AA-Turner:
fixes #2420 

A


pep_sphinx_extensions/pep_theme/templates/not_found.html
```diff
@@ -0,0 +1,41 @@
...
+    <title>PEP Not Found | peps.python.org</title>
+    <link rel="shortcut icon" href="{{ pathto('_static/py.png', resource=True) }}"/>
+    <link rel="stylesheet" href="{{ pathto('_static/style.css', resource=True) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto('_static/mq.css', resource=True) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto('_static/dark.css', resource=True) }}" type="text/css" media="(prefers-color-scheme: dark)" id="css-dark"/>
+    <link href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
+    <meta name="description" content="Python Enhancement Proposals (PEPs)"/>
+    <script>
+    const prefixNum = window.location.pathname.slice(1).replace(/^(\d*)$/, "$1")
+    if (!isNaN(+prefixNum)) window.location.pathname = "/pep-" + prefixNum.padStart(4, "0")
```
ezio-melotti, line 17:
```suggestion
    const match = window.location.pathname.match(/pep-(\d+)/)
    if (match != null) {
        const pepNum = match[1]
        window.location.pathname = "/pep-" + pepNum.padStart(4, "0")
    }
```
This should achieve the same thing and it's IMHO a bit clearer and flexible.

pep_sphinx_extensions/pep_theme/templates/not_found.html
```diff
@@ -0,0 +1,41 @@
...
+    </script>
+</head>
+<body>
+    <section id="pep-page-section">
+        <header>
+            <h1>Python Enhancement Proposals</h1>
+            <ul class="breadcrumbs">
+                <li><a href="https://www.python.org/" title="The Python Programming Language">Python</a> &raquo; </li>
+                <li><a href="{{ pathto("pep-0000") }}">PEP Index</a></li>
```
ezio-melotti, line 26: Why not just use `<a href="/">` like at line 34 below?

ezio-melotti (changes requested):
I left a few comments but otherwise looks good.  Is there a way to see it rendered online before merging it?

AA-Turner:
@ezio-melotti -- JS works for now. `https://pep-previews--2421.org.readthedocs.build/42` or `https://pep-previews--2421.org.readthedocs.build/042/` should both redirect to the proper PEP, and there is now a custom 404. 

The auto-linker will need to be set up with no prefix to the domain & then just passing the number.

A

pep_sphinx_extensions/pep_theme/templates/not_found.html
AA-Turner: I don't want to include the `pep-` prefix for the auto-redirects, partly for laziness as it means I can now just type `/X` instead of `/pep-000X` etc.

Happy to use your approach although I'll tweak a little.

A

pep_sphinx_extensions/pep_theme/templates/not_found.html
AA-Turner: `/` is not canonical -- really I should do `pathto("pep-0000")` below.

A

ezio-melotti:
https://pep-previews--2421.org.readthedocs.build/pep-8/ seems to work fine.
https://pep-previews--2421.org.readthedocs.build/pep-foo/ also works.
https://pep-previews--2421.org.readthedocs.build/pep-0123 creates a redirect loop, since it's a valid number, but refers to a pep that doesn't exist.

I this we should add a check on the length of the number.  I'll leave a suggestion in the code.

pep_sphinx_extensions/pep_theme/templates/not_found.html
```diff
@@ -0,0 +1,44 @@
...
+    <title>PEP Not Found | peps.python.org</title>
+    <link rel="shortcut icon" href="{{ pathto('_static/py.png', resource=True) }}"/>
+    <link rel="stylesheet" href="{{ pathto('_static/style.css', resource=True) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto('_static/mq.css', resource=True) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto('_static/dark.css', resource=True) }}" type="text/css" media="(prefers-color-scheme: dark)" id="css-dark"/>
+    <link href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
+    <meta name="description" content="Python Enhancement Proposals (PEPs)"/>
+    <script>
+    const pepNumMatch = window.location.pathname.match(/^\d+$/)
+    if (pepNumMatch !== null) {
```
ezio-melotti, line 17:
```suggestion
    if (pepNumMatch !== null && match[0].length < 4) {
```

pep_sphinx_extensions/pep_theme/templates/not_found.html
```diff
@@ -0,0 +1,44 @@
...
+    <meta name="color-scheme" content="light dark" />
+    <title>PEP Not Found | peps.python.org</title>
+    <link rel="shortcut icon" href="{{ pathto('_static/py.png', resource=True) }}"/>
+    <link rel="stylesheet" href="{{ pathto('_static/style.css', resource=True) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto('_static/mq.css', resource=True) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto('_static/dark.css', resource=True) }}" type="text/css" media="(prefers-color-scheme: dark)" id="css-dark"/>
+    <link href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
+    <meta name="description" content="Python Enhancement Proposals (PEPs)"/>
+    <script>
+    const pepNumMatch = window.location.pathname.match(/^\d+$/)
```
ezio-melotti, line 16: With `^` and `$` in the regex, it won't match `pep-NNN`.  Did you just change this temporarily to test?
Note that in the suggestion below I used `match[0]` because of this change, but if my previous suggestion is restored it should be `match[1]`.

ezio-melotti (changes requested):

AA-Turner:
> redirect loop

The latest change matches `^\d+$` so there's no redirect anymore. 

We could check the number but if invalid the only harm of not checking would be an extra redirect from one invalid URL to another.

A

pep_sphinx_extensions/pep_theme/templates/not_found.html
AA-Turner: >it won't match `pep-NNN`

From above, "I don't want to include the `pep-` prefix for the auto-redirects"

A

pep_sphinx_extensions/pep_theme/templates/not_found.html
AA-Turner: Our comments crossed over, sorry!

> We could check the number but if invalid the only harm of not checking would be an extra redirect from one invalid URL to another.

A

pep_sphinx_extensions/pep_theme/templates/not_found.html
ezio-melotti: Why not?
If we set the autolinking to `https://peps.python.org/pep-<num>/` and num is a valid 4-digits number (e.g. pep-0008), then it will go straight to the right page.  If we omit the `pep-` prefix, then it will have to go through an extra redirect if JS is enabled, or end up at the 404 page if it's disabled.

In case of a valid PEP with a <4 digits number (e.g. pep-8) it will go to `/pep-8/` and get redirected to `/pep-0008`.
In case of an invalid PEP with a <4 digits number (e.g. pep-123) it will go to `/pep-123/`, get redirected to `/pep-0123` and then stop at the 404 page since the PEP doesn't exist and the id is already 4 digits long.

Am I missing something?

ezio-melotti: The autolinking on this repo is currently set to link to `https://python.org/dev/peps/pep-<num>`.  My suggestion that simply matched `pep-(\d+)` works even with this type of URL, but fails with `/^\d+$/`.

CAM-Gerlach:
As mentioned on #2420 , you could do this all in HTML with no JS needed by just using`<meta http-equiv="refresh" content="0; url={redirect_url}">` and auto-generating the redirect pages via a build script. We use [such a script](https://github.com/spyder-ide/spyder-docs/blob/master/scripts/generateredirects.py) that I wrote in production for the [Spyder Docs](https://docs.spyder-ide.org).

Instead of this approach, that could be used here as a post-build step in the CI/makefile (I would be willing to relicense it CC0), as we do, with only slight modifications (since almost everything is done via CLI arguments rather than hardcoded), and would avoid the need for JS as the redirects are statically generated on site build, like the rest of the site.

AA-Turner:
I don't know if the preview site has a part to play in this, but the `http-equiv` approach seems to be markedly slower than JS -- I see the redirect page linger on screen for ~0.5 seconds, versus not seeing it at all for the JS approach. If others could test I'd appreciate it.

A

CAM-Gerlach:
@AA-Turner Good catch; I can replicate the behavior you described. However, the [redirects](https://github.com/spyder-ide/spyder-docs/blob/gh-pages/faq.html) generated by my script I linked are essentially instant (≈100 ms; see, for example, [this page](docs.spyder-ide.org/faq.html)) and nothing renders in the viewport except the final page, with no visible artifacts (and the host is GitHub Pages, just like the production site here).

I suspect the issue is due to your redirect pages including a `<body>` which gets rendered and displayed, whereas mine are just a head, and also omit the unnecessary `shortcut icon` and `viewport` elements, which require an additional HTTP resource fetch and and may also introduce a little more rendering delay, (plus as extra elements in the source and in the DOM, increase page size and parsing time slightly).

I suggest simply eliminating all those elements, which should result in identical behavior (at least in production) to the examples I shared.

pep_sphinx_extensions/pep_processor/html/pep_html_redirects.py
```diff
@@ -0,0 +1,19 @@
...
+from pathlib import Path
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from sphinx.application import Sphinx
+
+
+def prepare_redirect_pages(app: Sphinx) -> Iterator[tuple[str, dict, str]]:
+    for path in Path(app.srcdir).glob("pep-????.???"):
```
CAM-Gerlach, line 14:
```suggestion
    for path in Path(app.srcdir).glob("pep-[0-9][0-9][0-9][0-9].???"):
```

Use the correct glob to ensure this is actually a PEP with a valid number. Otherwise, it can match any arbitrary `txt` or `rst` file with a four-character suffix, which will cause an error on line 18.

pep_sphinx_extensions/pep_theme/templates/redirect.html
```diff
@@ -0,0 +1,16 @@
+{# Redirect template #}
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta http-equiv="refresh" content="0; url={{ pathto(pep_file) }}">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
CAM-Gerlach, line 7:
```suggestion
```

Elide unnecessary element for page with no content that we want to load, parse, render and redirect as fast as possible.

pep_sphinx_extensions/pep_theme/templates/redirect.html
```diff
@@ -0,0 +1,16 @@
...
+    <title>PEP Redirector | peps.python.org</title>
+    <link rel="shortcut icon" href="{{ pathto('_static/py.png', resource=True) }}"/>
```
CAM-Gerlach, line 9:
```suggestion
```

As above. In particular, this element can trigger another HTTP request, which could add further delay to the redirect.

pep_sphinx_extensions/pep_theme/templates/redirect.html
```diff
@@ -0,0 +1,16 @@
...
+    <link rel="canonical" href="{{ pathto(pep_file) }}" />
+    <meta name="description" content="Python Enhancement Proposals (PEPs)"/>
+</head>
+<body>
+    <p><a href="{{ pathto(pep_file) }}">PEP {{ pep_num }}</a></p>
+</body>
```
CAM-Gerlach, line 15:
```suggestion
```

If we don't want the page to render, then we shouldn't include a `<body>`, which will flash onscreen (as you noticed) as well as cause a non-trivial delay to load, parsing and rendering.

pep_sphinx_extensions/pep_processor/html/pep_html_redirects.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from pathlib import Path
```
CAM-Gerlach, line 5:
```suggestion
from pathlib import Path
from typing import TYPE_CHECKING
```

Both `pathlib` and `typing` are stdlib modules, so should go in the same block per PEP 8, and be sorted lexicographically.

CAM-Gerlach (commented):
Various fixes and nits. Otherwise, this looks good and is simpler to implement internally within the Sphinx extension than I thought it would be.

pep_sphinx_extensions/pep_processor/html/pep_html_redirects.py
```diff
@@ -0,0 +1,19 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from sphinx.application import Sphinx
+
+
+def prepare_redirect_pages(app: Sphinx) -> Iterator[tuple[str, dict, str]]:
```
ezio-melotti, line 13: Wouldn't this fail if `TYPE_CHECKING` above is false?

pep_sphinx_extensions/pep_theme/templates/redirect.html
```diff
@@ -0,0 +1,16 @@
+{# Redirect template #}
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
```
ezio-melotti, line 5:
```suggestion
    <meta charset="utf-8">
```
Just a minor nit, but the `/` are unnecessary (also below).

ezio-melotti (changes requested):
I agree with CAM comments, and added a couple more.

pep_sphinx_extensions/pep_theme/templates/redirect.html
CAM-Gerlach: I noticed that as well, though it was sufficiently practically unimportant that I didn't mention it (though it did bother me that it was somewhat inconsistently used even on adjacent `void` elements, presumably because some of them were copied from my script while some were copied from the template).

pep_sphinx_extensions/pep_processor/html/pep_html_redirects.py
CAM-Gerlach: No, because the file uses `` from __future__ import annotations`` which does not evaluate annotations at runtime.

pep_sphinx_extensions/pep_processor/html/pep_html_redirects.py
AA-Turner: > if we ever do

This is I suppose my reluctance to put `[0-9]` rather than `?` -- I seriously think this is very unlikely to happen, and `[0-9]` is five times as many characters as `?`. (The thought applies in all cases that this has been suggested).

A

pep_sphinx_extensions/pep_theme/templates/redirect.html
AA-Turner: > inconsistently used

The `http-eqiv` was the only one missing the closing `/` it seems -- I've now added this. (PS for clarity I didn't copy anything from the script you linked as I thought it could be done in Sphinx internally -- I was reading the MDN documentation and most likely just forgot! Sorry.)

A

pep_sphinx_extensions/pep_processor/html/pep_html_redirects.py
CAM-Gerlach: I hope you didn't stay up all night...or just up really early?

A few more characters perhaps, but in terms of what actually matters, does that make it that much less clear to the reader? And is it really worth sacrificing precision and correctness, and risking a potential future hard build failure just to skimp on a few characters? I realize there is _some_ tradeoff to be made, but it seems the rest of us who have weighed in so far have considered it a worthwhile one.

AA-Turner:
It does seem quicker now, although there may be bias in that I want it to be quicker!

A

pep_sphinx_extensions/pep_processor/html/pep_html_redirects.py
AA-Turner: Done. I still disagree though!

A

pep_sphinx_extensions/pep_theme/templates/redirect.html
ezio-melotti: Why not removing them all then?

Technically it's allowed for HTML5 void elements like `meta` and `link`, but has no effect and it's ignored.
See https://www.w3.org/TR/2011/WD-html-markup-20110113/syntax.html#syntax-elements

AA-Turner: My HTML auto-formatter is XML based and expects self-closing element. Maybe a bad reason but it is the reason!

A

hugovk:
I'd like to hear about Nginx redirects https://github.com/python/peps/issues/2420#issuecomment-1065843624 before reviewing this. Thanks!

CAM-Gerlach:
> I'd like to hear about Nginx redirects https://github.com/python/peps/issues/2420#issuecomment-1065843624 before reviewing this. Thanks!

Unless I'm misunderstanding something, that approach only works for people who type in the old no longer canonical URL, i.e. `www.python.org/dev/peps/pep-N`; what is being proposed here works for the new PEP site without relying on the python.org infra, i.e. `peps.python.org/pep-N` or even `peps.python.org/N`, which is self-contained, makes the URL much shorter, avoids a redirect between different subdomains and infras that triggers an extra DNS query, and establishes a clearer justification for doing this now at the inception of the new site. Meanwhile the other downsides you mentioned,

> Rather than running JavaScript client-side (not everyone has JS enabled, it's relatively slow, more code to maintain)

aren't so much of a factor here with the optimized HTML-based approach.

I certainly think the redirect rules could be adjusted to _also_ do this, but that would supplement, not replace the approach in this PR.

hugovk:
Ah right, yes, Nginx redirects don't apply to this: https://github.com/python/peps/issues/2420#issuecomment-1066160432. Will review this, thanks!

hugovk (changes requested):
Testing:

- https://pep-previews--2421.org.readthedocs.build/pep-8
- https://pep-previews--2421.org.readthedocs.build/pep-12
- https://pep-previews--2421.org.readthedocs.build/pep-101
- https://pep-previews--2421.org.readthedocs.build/pep-foo
- https://pep-previews--2421.org.readthedocs.build/pep-0123

These all redirect to https://pep-previews--2421.org.readthedocs.build/pep-0404/

For example, https://pep-previews--2421.org.readthedocs.build/pep-8 has status code 404

AA-Turner:
Although none of those URLs are valid (the supported scheme is `/\d+`), it is very surprising behaviour.

https://peps.python.org/pep-8 redirects to the standard 404 page -- I wonder if the JS experiment's 404 has been cached or something?

A

CAM-Gerlach:
> These all redirect to https://pep-previews--2421.org.readthedocs.build/pep-0404/

The intended redirect pattern that @AA-Turner implemented is `/N` -> `/pep-NNNN`. However, I can confirm in my testing that _any_ link, not just those beginning with `pep-`, that does not correspond to a PEP number or full pep name goes to PEP 404.

The reason this happens is that this attempts to load `/404`, which lands on the redirect page to PEP 404 and thus redirects to that PEP—i.e., we've made the `404` page be a redirect to PEP 404, which is not what we want. We either need to return to the `pep-N` schema or figure out another way to avoid this happening (the simplest being to add a special case to not generate a redirect for PEP 404).

AA-Turner:
> The reason this happens is that this attempts to load `/404`, which lands on the redirect page to PEP 404 and thus redirects to that PEP

The law of unintended consequences... I suppose we could do `window.history` introspection, but that brings back the drawbacks of JavaScript.

A

hugovk:
Would https://pypi.org/project/sphinx-reredirects/ help?

CAM-Gerlach:
We looked into it, but I don't think 0.0.1 was released yet when we implemented it, and we weren't sure if the wildcard support would fully handle our current and future use cases, so we ended up implementing our own solution. For this case, that doesn't look like that would be a huge issue, but it doesn't support the sort of rewriting we need here, so unfortunately it doesn't look like it'd work (at least at present).

AA-Turner:
I won't have time to resolve this before mid April -- feel free to push to this branch on my repo or close this PR as you see fit. My views are strongly in favour of the simple `/N` prefix -- if we can sort out the redirects to PEP 404 with JS history introspection, I think that's fine -- the JS experiment remains in this branch if it's to be resurrected.

A

CAM-Gerlach:
> if we can sort out the redirects to PEP 404 with JS history introspection, I think that's fine -- the JS experiment remains in this branch if it's to be resurrected.

I really would like to not rely on such hacks, and would prefer not to go with the JS approach if possible for the reasons @hugovk mentioned, so I'd strongly lean toward the two less complicated and costly alternatives here—either reverting to the `/pep-N` scheme, which did already have some support by others, or just excluding PEP 404 from the redirect generation.

hugovk:
Closing, this has been implemented in https://github.com/python/peps/issues/2420. Thanks!

PR #2425, opened 2022-03-13
Styling: Remove field block highlights and restore definition list indent

AA-Turner:
A

### DIFF
--- a/pep_sphinx_extensions/pep_theme/static/style.css
+++ b/pep_sphinx_extensions/pep_theme/static/style.css
@@ -244,16 +244,12 @@ dl.footnote {
     grid-template-columns: fit-content(30%) auto;
     line-height: 1.875;
     width: 100%;
-    border-top: 1px solid var(--colour-background-accent);
-    border-bottom: 1px solid var(--colour-background-accent);
+    border-top: 1px solid var(--colour-rule-strong);
 }
 dl.rfc2822 > dt, dl.rfc2822 > dd,
 dl.footnote > dt, dl.footnote > dd {
     padding: .25rem .5rem .2rem;
-}
-dl.rfc2822 > dt:nth-of-type(even), dl.rfc2822 > dd:nth-of-type(even),
-dl.footnote > dt:nth-of-type(even), dl.footnote > dd:nth-of-type(even) {
-    background-color: var(--colour-background-accent);
+    border-bottom: 1px solid var(--colour-rule-strong);
 }
 dl.footnote > dt {
     font-weight: normal;
--- a/pep_sphinx_extensions/pep_theme/static/style.css
+++ b/pep_sphinx_extensions/pep_theme/static/style.css
@@ -122,9 +122,6 @@ pre {
 dl dt {
     font-weight: bold;
 }
-dl dd {
-    margin: 0;
-}
 
 /* Horizontal rule rule */
 hr {
@@ -261,6 +258,10 @@ dl.footnote > dt:nth-of-type(even), dl.footnote > dd:nth-of-type(even) {
 dl.footnote > dt {
     font-weight: normal;
     border-right: 1px solid var(--colour-background);
+}
+dl.rfc2822 dd,
+dl.footnote dd {
+    margin: 0;
 }
 
 /* Sidebar formatting */


AA-Turner:
https://pep-previews--2425.org.readthedocs.build/pep-0609/#terminology for a page with `<dl>` items.

A

pradyunsg (approved):

merwok (approved):

pep_sphinx_extensions/pep_theme/static/style.css
```diff
@@ -248,20 +245,20 @@ dl.footnote {
...
-    border-bottom: 1px solid var(--colour-background-accent);
 }
 dl.rfc2822 > dt, dl.rfc2822 > dd,
 dl.footnote > dt, dl.footnote > dd {
     padding: .25rem .5rem .2rem;
-}
-dl.rfc2822 > dt:nth-of-type(even), dl.rfc2822 > dd:nth-of-type(even),
-dl.footnote > dt:nth-of-type(even), dl.footnote > dd:nth-of-type(even) {
-    background-color: var(--colour-background-accent);
+    border-bottom: 1px solid var(--colour-background-accent);
```
CAM-Gerlach, line 252:
```suggestion
    border-bottom: 1px solid var(--colour-accent-strong);
```

The horizontal rules substituted for the alternating colors for the PEP headers and footnotes are very low contrast, which combined with being very thin is such that even I can barely see them, well below even the lowest minimum WCAG accessibility standards.

image


This is because they use `--colour-background-accent`, rather than the (much more semantically appropriate) ``--colour-rule-strong``, which is specifically intended for rules like this. Switching to that instead is much more easily visible at 4.3:! and passes WCAG AA:

image

pep_sphinx_extensions/pep_theme/static/style.css
```diff
@@ -248,20 +245,20 @@ dl.footnote {
     line-height: 1.875;
     width: 100%;
     border-top: 1px solid var(--colour-background-accent);
```
CAM-Gerlach, line 247:
```suggestion
    border-top: 1px solid var(--colour-rule-strong);
```

To match other comment.

CAM-Gerlach (changes requested):
Unfortunately, this doesn't yet actually fix the problem reported in #2300 , and vertical margins are still pretty tight overall (as @pradyunsg reported), and there's also a contrast, consistency and accessibility issue with the changes to the header and footnotes styling. See comments for more on both and review suggestions to fix each of them.

Also, you might want to consider vertical rules between the definition list elements as well (perhaps `--colour-rule-light` ones) to help them and each element further stand out, and be a closer match to the original theme.

pep_sphinx_extensions/pep_theme/static/style.css
pradyunsg: I don't care, but I do find it weird to be applying rules of text readability contrast to... borders? :)

pep_sphinx_extensions/pep_theme/static/style.css
```diff
@@ -122,9 +122,6 @@ pre {
 dl dt {
     font-weight: bold;
 }
-dl dd {
-    margin: 0;
-}
```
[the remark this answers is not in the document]
pradyunsg: Let's fix the line spacing in a follow up.

pep_sphinx_extensions/pep_theme/static/style.css
AA-Turner: I didn't mind enough to argue! 

A

AA-Turner:
I thought the issue with `<dl>` was the left margin (indent) -- if there are broader vertical whitespace proposed changes, I'd suggest them for a follow-up (per @pradyunsg's comment)

A

pep_sphinx_extensions/pep_theme/static/style.css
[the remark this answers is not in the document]
CAM-Gerlach: There are definitely other issues with line spacing, but this was one of the most immediately noticeable, and this particular issue is not just that the line spacing is too narrow but that is visibly inconsistent between definitions with multiple paragraphs and those with only a single one, so I'm not sure we should let the fact that there are also other issues with line spacing block fixing this one.

However, if we do this, this PR should be unmarked as resolving #2300 , because this was the initial problem that prompted and was cited in said issue.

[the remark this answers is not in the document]
CAM-Gerlach: Opened as #2427

PR #2452, opened 2022-03-21
PEP 675: Mark as accepted

JelleZijlstra:
cc @pradeep90 (and congratulations!)


JelleZijlstra:
and also @gbleaney

JelleZijlstra:
Also, I opened https://bugs.python.org/issue47088 to track the runtime implementation; I couldn't find either of you on BPO so letting you know here. The implementation should be simple so I'm planning to just do it myself.

CAM-Gerlach (approved):
LGTM, thanks

PR #2458, opened 2022-03-22
PEP 659: Reformat to obey 80 column limit

markshannon:
Also, change "deoptimize" to "de-optimize".

No changes to text or meaning, otherwise.

### DIFF
--- a/pep-0659.rst
+++ b/pep-0659.rst
@@ -11,21 +11,26 @@ Post-History: 11-May-2021
 Abstract
 ========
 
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
FWIW, if you're going to auto-wrap this, the nominal limit as specified in PEP 12 is 79 columns, rather than 80 (although that itself is not strictly enforced, particularly small or occasional deviations)

markshannon:
I've mostly stuck to 79 columns, except in a couple of places that are 80. So this should be fine.

PR #2460, opened 2022-03-23
PEP 669: Update specification to return current callback object

markshannon:


hugovk:
The RTD build preview failed because this branch is a long way behind `main` so it doesn't have the RTD config version to switch the build from default Python 3.7 to required 3.9.

A rebase would fix it, but I wouldn't worry for such a small PR. Recommended for future PRs though :)

JelleZijlstra (approved):

markshannon:
@JelleZijlstra Does the `docs/readthedocs.org:pep-previews` failure matter?

hugovk:
We now have these buttons:

image

See https://github.com/python/peps/issues/2495 for more.

I'll do an update with rebase.

hugovk:
Or rather I won't as I see you just found it :)

markshannon:
I don't like merging things with broken tests. Merging main is easy enough.

PR #2461, opened 2022-03-23
PEP 670: Rearrange and reword the PEP

encukou:
@vstinner & @erlend-aasland, this should make the PEP more understandable & complete

### DIFF
--- a/pep-0670.rst
+++ b/pep-0670.rst
@@ -12,19 +12,15 @@ Python-Version: 3.11
 Abstract
 ========
 
-Convert macros to static inline functions or regular functions to avoid
-macro pitfalls.
+Macros in the C API will be converted to static inline functions or
+regular functions. This will help avoid macro pitfalls in C/C++, and
+make the functions usable from other programming languages.
 
-Convert macros and static inline functions to regular functions to make
-them usable by Python extensions which cannot use macros or static
-inline functions, like extensions written in a programming languages
-other than C or C++.
+To avoid compiler warnings, function arguments of pointer types
+will be cast to appropriate types using additional macros.
 
-Function arguments of pointer types are still cast and return types are
-not changed to prevent emitting new compiler warnings.
-
-Macros which can be used as l-value in an assignment are not converted
-to functions to avoid introducing incompatible changes.
+To avoid introducing incompatible changes, macros which can be used as
+l-value in an assignment will not be converted.
 
 
 Rationale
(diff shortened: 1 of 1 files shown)


pep-0670.rst
```diff
@@ -38,202 +34,132 @@ to maintain.
...
-The following macros should not be converted:
+The following macros will not be converted:
+
+* Object-like macros (i.e. those which don't need parentheses and
+  arguments). For example:
+
+  * Empty macros. Example: ``#define Py_HAVE_CONDVAR``.
+  * Macros only defining a value, even if a constant with a well defined
+    type would be better. Example: ``#define METH_VARARGS 0x0001``.
+  * Py_PYTHON_H
```
vstinner, line 85: Just add Py_PYTHON_H as another example on the first item "Empty macros. Example: #define Py_HAVE_CONDVAR.", no?

pep-0670.rst
```diff
@@ -527,18 +455,19 @@ using them as l-value in assignment. It is an incompatible change made
 on purpose: see `bpo-39573 <https://bugs.python.org/issue39573>`_ for
 the rationale.
 
-This PEP does not convert macros which can be used as l-value to avoid
-introducing incompatible changes.
+This PEP does propose converting macros which can be used as l-value
```
vstinner, line 458:
```suggestion
This PEP does not propose converting macros which can be used as l-value
```

pep-0670.rst
encukou:
```suggestion
```
“used for definitions rather than behavior” is a better reason to exclude this one, IMO. It does fit both :)

vstinner:
Please wait for Erlend's (@erlend-aasland) approval, since he is one of two authors of this PEP :-) cc @erlend-aasland

pep-0670.rst
```diff
@@ -12,19 +12,15 @@ Python-Version: 3.11
...
+Macros in the C API will be converted to static inline functions or
+regular functions. This will help avoid macro pitfalls in C/C++, and
+make the functions usable from other programming languages.
 
-Convert macros and static inline functions to regular functions to make
-them usable by Python extensions which cannot use macros or static
-inline functions, like extensions written in a programming languages
-other than C or C++.
+To avoid compiler warnings, function arguments of pointer types
+will be cast to appropriate types using additional macros.
```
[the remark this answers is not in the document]
vstinner: Yep, but the parenthesis is not needed. It's an important information.

erlend-aasland (approved):
AFAICS, the major part of this PR is a rearrangement of some of the sections. I'm fine with that. The rewordings are ok to me as well.

Thanks Petr and Victor.

PR #2462, opened 2022-03-23
PEP 659: Update to describe inline caches

markshannon:


pep-0659.rst
```diff
@@ -149,14 +145,14 @@ Each instruction that would benefit from specialization is replaced by an
 adaptive version during quickening. For example,
 the ``LOAD_ATTR`` instruction would be replaced with ``LOAD_ATTR_ADAPTIVE``.
 
-Each adaptive instruction maintains a counter,
-and periodically attempts to specialize itself.
+Each adaptive instruction maintains a counter, and periodically attempts
+to specialize itself.
```
brandtbucher, line 149: There is an almost-identical sentence two paragraphs down. Maybe remove one of them?

pep-0659.rst
```diff
@@ -178,85 +174,41 @@ Ancillary data
...
-  3.10 requires two reads, which are dependent.
-* uses much less memory, as the data can be different sizes for different
-  instruction families, and doesn't need an additional array of offsets.
-  can support much larger functions, up to about 5000 instructions
-  per function. 3.10 can support about 1000.
+This is an obvious candidate for specialization. Attributes might belong to
+a normal instance, a class, a module, or one of many other special cases.
 
+``LOAD_ATTR`` would initially be quickened to ``LOAD_ATTR_ADAPTIVE`` which
+would track how often it is executed, and call the ``load_attr_optimize``
```
brandtbucher, line 195:
```suggestion
would track how often it is executed, and call the ``_Py_Specialize_LoadAttr``
```

pep-0659.rst
```diff
@@ -312,32 +264,49 @@ Memory use
...
+reached ~2000 when another byte per instruction and 32 bytes per
+instruction with a cache (``LOAD_GLOBAL`` and ``LOAD_ATTR``).
 
 Comparing memory use to 3.10
 ''''''''''''''''''''''''''''
+
+CPython 3.10 used 2 bytes per instruction, until the execution count
+reached ~2000 when it allocates another byte per instruction and
+32 bytes per instruction with a cache (``LOAD_GLOBAL`` and ``LOAD_ATTR``).
+
```
brandtbucher, line 282: This paragraph is repeated (and the first copy looks incomplete).

```suggestion
Comparing memory use to 3.10
''''''''''''''''''''''''''''

CPython 3.10 used 2 bytes per instruction, until the execution count
reached ~2000 when it allocates another byte per instruction and
32 bytes per instruction with a cache (``LOAD_GLOBAL`` and ``LOAD_ATTR``).

```

pep-0659.rst
```diff
@@ -312,32 +264,49 @@ Memory use
...
-then the memory use is still less than that of 3.10.
-With a more realistic 25%, then memory use is basically the same as the
-hypothetical improved version of 3.10.
+The relative memory use depends on how much code is "hot" enough to trigger
+creation of the cache in 3.10. The break even point, where the memory used
+by 3.10 is the same as for 3.11 is ~70%.
+
+It is also worth noting that the actual bytecode is only part of a code
+object. Code objects also include names, constants and quite a lot of
+debugging information.
```
brandtbucher, line 306: Maybe touch on the line/column-table issue when adding additional cache entries? Could be worded as "the format of these (cold) tables may need to be updated to avoid additional memory use" or something.

pep-0659.rst
```diff
@@ -312,32 +264,49 @@ Memory use
...
+
+In summary, for most applications where many of the functions are relatively
+relatively unused, 3.11 will consume more memory than 3.10, but not by much.
```
brandtbucher, line 309:
```suggestion
In summary, for most applications where many of the functions are relatively
unused, 3.11 will consume more memory than 3.10, but not by much.
```

pep-0659.rst
```diff
@@ -349,7 +318,44 @@ None
 Rejected Ideas
 ==============
 
-Too many to list.
+By implementing a specializing adaptive interpreter with inline data caches,
+we are implicitly rejecting many alternative ways to optimize CPython.
+However, it is worth emphasizing that some ideas, such as just-in-time
+compilation, have not been rejected, merely deferred.
+
+Storing data caches before the bytecode.
```
brandtbucher, line 326: Do you think it's worth enumerating a few reasons why we decided not to go this route?

pep-0659.rst
```diff
@@ -99,47 +99,43 @@ more robust in cases where specialization fails or is not stable.
...
 Overview
 --------
 
-Once any instruction in a code object has executed a few times,
-that code object will be "quickened" by allocating a new array for the
-bytecode that can be modified at runtime, and is not constrained as the
-``code.co_code`` object is. From that point onwards, whenever any
-instruction in that code object is executed, it will use the quickened form.
-
 Any instruction that would benefit from specialization will be replaced by an
```
brandtbucher, line 116: Any reason why you don't discuss the warmup period anymore? Perhaps just because we're moving towards quickening everything now?

brandtbucher (approved):
Thanks, looks good. I left a few comments as I read through:

pep-0659.rst
markshannon: I'm not sure that is relevant. We would probably want to revisit how we store the debug information even if we kept the original bytecode format.

pep-0659.rst
markshannon: Added a brief comment.

pep-0659.rst
markshannon: Yes, quickening everything, means that "warm up" is handled by the adaptive instructions.

PR #2474, opened 2022-03-27
Infra: Include trailing slash in RSS URLs to avoid 301 redirect

hugovk:
https://peps.python.org/peps.rss includes links like:

```rss
      <guid isPermaLink="true">https://peps.python.org/pep-0686</guid>
```

Without the trailing `/`, there's an extra HTTP 301 redirect to add the slash, which adds a small but unnecessary delay which can vary between 20ms and 335ms.

See https://github.com/sphinx-doc/sphinx/pull/10288 for some network waterfall diagrams with and without the extra request.

So let's include it:

```rss
      <guid isPermaLink="true">https://peps.python.org/pep-0686/</guid>
```

Also update pip and include wheel in virtualenv creation to remove a couple of warnings, noticeable during `make rss`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ venv:
 		echo "To recreate it, remove it first with \`make clean-venv'."; \
 	else \
 		$(PYTHON) -m venv $(VENVDIR); \
+		$(VENVDIR)/bin/python3 -m pip install -U pip wheel; \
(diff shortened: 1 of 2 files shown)


CAM-Gerlach (approved):
LGTM @hugovk

JelleZijlstra (approved):

PR #2481, opened 2022-03-29
PEP 570: Mark as Final

CAM-Gerlach:
As discussed in #2479 , PEP 570 (positional-only parameters) was implemented and documented in Python 3.8 but not marked Final, likely as an oversight. This fixes that.

Fixes #2479


JelleZijlstra (approved):

pablogsal (approved):

pablogsal:
Thanks!

PR #2499, opened 2022-04-04
PEP 687: Isolating modules in the standard library

encukou:
From recent discussions around “what should have a PEP”, it's clear that this should have been a PEP long ago. Better late than never, I guess!

We submit this PEP to explain the changes, seek consensus on whether they are good,  propose the remaining changes, and set best practices for new modules.

(We wrote this over several months, sorry if it doesn't follow all the new PEP guidelines.)


JelleZijlstra (approved):
Looks great, a very readable proposal

PR #2507, opened 2022-04-07
PEP 681: Fix header links

JelleZijlstra:
#2481 didn't like the hashes in the URLs

### DIFF
--- a/pep-0681.rst
+++ b/pep-0681.rst
@@ -9,9 +9,9 @@ Type: Standards Track
 Content-Type: text/x-rst
 Created: 02-Dec-2021
 Python-Version: 3.11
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
I've instead updated #2484 to ensure URL fragments are allowed (e.g. skipping to specific posts, as these are); there's no real harm in doing so and it avoids extra work/frustration from authors having to manually trim them. Therefore, this PR isn't really needed any more, unless you/the author don't want to explicitly jump to the specific messages in the thread.

JelleZijlstra:
Since it's the top message in the thread, I feel like it's nicer to link to the whole thread.

CAM-Gerlach:
Yeah, I agree, and upon closer inspection and visiting the links I see the second one actually appears to link to the wrong message (for me in FF on Windows, my browser doesn't actually jump to it when I first visit the page, only when I press enter again on the link with the page loaded—perhaps Hyperkitty has an issue with its JS triggers.

CAM-Gerlach (approved):
LGTM, thanks @JelleZijlstra

PR #2515, opened 2022-04-11
PEP 687: Fix reference to PEP 489

mdboom:
As far as I can tell, PEP 484 is a typo and unrelated to this PEP.


JelleZijlstra:
cc @encukou

encukou:
Indeed. Thanks for the catch!

PR #2516, opened 2022-04-11
PEP 687: Add to CODEOWNERS

JelleZijlstra:
cc @encukou


encukou:
Thanks for the catch!

PR #2517, opened 2022-04-12
PEP 0: Author override for my last name

debonte:
`pep_zero_generator` currently considers my last name to be "Bonte" instead of "De Bonte". I have added an entry in `AUTHOR_OVERRIDES.csv` to fix this.

The [algorithm in `_parse_name`](https://github.com/python/peps/blob/main/pep_sphinx_extensions/pep_zero_generator/author.py#L78) that only includes the next to last word (ex. "De") in the surname if that word is all lowercase seems reasonable. I'm hesitant to change that.


CAM-Gerlach:
We _could_ probably special case "De"...but might not be worth it versus just doing an author override

PR #2518, opened 2022-04-12
PEP 678: update status to Accepted

Zac-HD:
Per the `Resolution` link, the SC has accepted PEP-678.  As I understand it the status is therefore updated to `Accepted`, and will reach `Final` after https://github.com/python/cpython/pull/31317 is merged.

(🥳)


JelleZijlstra (approved):
Congratulations!

PR #2526, opened 2022-04-14
purge CDN cache on deploy

ewdurbin:
issues a purge all to the Fastly service that fronts peps.python.org to make new changes appear fast..... lier.

otherwise this would have led to up to 1 hour delay for changes.


ewdurbin:
I'm not actually certain if I am being "correct" in how the curl command is initiated via GH actions/secrets. As the key is scoped to only allow this one single action on this one single service, I am going to merge, ensure that it doesn't leak things and revoke the token if it does.

ewdurbin:
Sweet. Looks good!

PR #2535, opened 2022-04-18
PEP 517: add missing def for the mandatory hook build_wheel signature

wimglenn:
All the other example hooks have a `def` so this looks like a typo.


CAM-Gerlach (approved):
LGTM, thanks.

brettcannon (approved):

PR #2550, opened 2022-04-23
PEP-0594: Fix tiny typo

lemonez:
fix tiny typo


AA-Turner (approved):
The [current rendered version](https://peps.python.org/pep-0594/#creating-maintaining-a-separate-repo-for-the-deprecated-modules) does have a space between "repository" and the full stop, so this is the correct fix.

A

PR #2553, opened 2022-04-25
PEP 688: Link to current docs

encukou:
Link to current documentation of the protocol first and make the
PEP secondary (it might still be useful for the original motivation/rationale).


CAM-Gerlach:
Great catch, thanks!

PR #2558, opened 2022-04-26
PEP 594: fix typos

JelleZijlstra:


pep-0594.rst
```diff
@@ -476,7 +476,7 @@ mailcap
 
 The `mailcap <https://docs.python.org/3/library/mailcap.html>`__ package
 reads *mail capability* files to assist in handling a file attachment in
-an email. In most modern OSs the email client itself handles reacting to
+an email. In most modern OSes the email client itself handles reacting to
```
hugovk, line 479: Personally I would write "OSs" too, or better yet, rewrite to "operating systems" to avoid it.

Style guides are split.

OSes:

* https://developers.google.com/style/abbreviations?hl=en 
* https://github.com/raspberrypilearning/style-guide/blob/master/style-guide.md

OSs

* https://docs.microsoft.com/en-us/style-guide/acronyms
* https://blog.apastyle.org/apastyle/2013/11/pluralize-numbers-and-abbreviations-without-apostrophes.html
* https://help.apple.com/applestyleguide/#/apsg3acde405

Always write in full, even when singular

* https://stylepedia.net/style/5.1/#o

CAM-Gerlach: I personally have pretty much always seem OSes; syntactically I'm not aware of any instance where its explicitly allowed to put two "s"es together without something ("e", "''", etc.) separating them, which English has gone to great lengths to work around. But I agree it would probably be better to spell it out, or use an appropriate synonym.

JelleZijlstra, line 479: Thanks! I had never seen `OSs` and assumed it was just a typo. Let's just write it out so we don't have to worry about the spelling:
```suggestion
an email. In most modern operating systems the email client itself handles reacting to
```

AA-Turner (approved):

brettcannon (approved):

PR #2559, opened 2022-04-26
PEP 689: Semi-stable C API tier

encukou:
A PEP for https://github.com/python/cpython/issues/91744

### DIFF
--- a/pep-0689.rst
+++ b/pep-0689.rst
@@ -1,0 +1,231 @@
+PEP: 689
+Title: Semi-stable C API tier
+Author: Petr Viktorin <encukou@gmail.com>
+Status: Draft
+Type: Standards Track
+Content-Type: text/x-rst
+Requires: 523
+Created: 22-Apr-2022
+Python-Version: 3.11
+
+
+Abstract
+========
+
+Some functions and types of the C-API are designated *semi-stable*,
+meaning that they will not change in patch (bugfix/security) releases,
+but may change between minor releases (e.g. between 3.11 and 3.12) without
+deprecation warnings.
+
+The exact set of APIs exposed this way may change in Beta releases.
+
+
+Motivation & Rationale
+======================
+
+The Python C-API is currently divided in three tiers (see
+`devguide <https://devguide.python.org/c-api/>`__ for details):
+
+- Limited API, with high compatibility expectations;
+- Public API, which follows the backwards compatibility policy
+  (PEP 387) which calls for deprecation warnings before changes;
+- Private (internal) API, which can change at any time.
+
+We need a tier between Public and Private to accomodate tools like
+advanced debuggers and JIT compilers, which need access to CPython
+internals but assume the C-API they use does not change in patch releases.
+
+
+Setting Stability Expectations
+------------------------------
+
+Currently, there are no guarantees for internal API -- that is, anything
+that requires ``Py_BUILD_CORE`` or is named with a leading underscore.
+This API can change without warning at any time, and code that uses it
+is pinned to a specific build of Python.
+
+However, in practice, even internal API usually happens to be stable
+in patch releases:
+
+- Some CPython core developers take this as an an unwritten rule.
+- Patch releases only contain bugfixes, which are not too likely to
+  change API.
+
+Semi-stable API will make the stability expectations more explicit.
+
+It will also hopefully encourage existing users of the private API to
+reach out to python-dev, so we can expose, standardize and test API
+for some of their use cases.
+
+
+Reserving underscores for Private API
+-------------------------------------
+
+PEP 523 introduced functions for use by debuggers and JIT compilers,
+which are to be only stable across minor releases.
+The functions names have leading underscores to suggest their limited
+stability.
+
+However, leading underscores usually mark *fully private* API.
+CPython developers familiar with the “underscore means internal”
+convention are unlikely to check if underscored functions they are
+changing are documented and used outside CPython itself.
+
+This proposal brings us a bit closer to reserving underscores
+only for truly private, unstable, hands-off API.
+
+
+Warning about API that is changed often
+---------------------------------------
+
+The ``PyCode_New()`` family is an example of functions that are
+documented as unstable, and also often change in practice.
+
+Moving it to the semi-stable tier will make its status obvious even
+to people who don't read the docs carefully enough, and will make it
+hard to use accidentally.
+
+
+Changes during the Beta period
+------------------------------
+
+Since the API itself can change continuously up until Beta 1 (feature freeze)
+of a minor version, major users of this API are unlikely to test
+Alpha releases and provide feedback.
+It is very difficult to determine what needs to be exposed as semi-stable.
+
+Addidions to the semi-stable tier will count as *stabilization*,
+and will be allowed up to Release Candidate 1.
+
+
+Specification
+=============
+
+Several functions and types (“APIs”) will be moved to a new *semi-stable* tier.
+
+They will be expected to stay stable across patch releases,
+but may change or be removed without warning in minor releases (3.x.0),
+including Alpha and Beta releases of 3.x.0.
+
+When they change significantly, code that uses them should no longer compile
+(e.g. arguments should be added/removed, or a function should be renamed,
+but the semantic meaning of an argument should not change).
+
+Their definitions will be moved to a new directory, ``Include/semistable/``,
+and will be included from ``Python.h``.
+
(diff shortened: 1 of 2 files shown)


pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+compilers that support ``Py_DEPRECATED``.
+
+A similar deprecation period will be used when making more APIs semi-stable
+in the future:
+
+- When moving from public API, the deprecation period should follow Python's
+  backwards compatibility policy (currently, it should last at least
+  two releases).
+- When moving from API that is documented as unstable,
+  the deprecation period can only last one release.
```
AA-Turner, line 134: I believe there's an implicit "public" here.
```suggestion
- When moving from public API that is documented as unstable,
  the deprecation period can only last one release.
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+
+The C API backwards compatibility story will be made clearer.
+
+
+Security Implications
+=====================
+
+None known.
+
+
```
AA-Turner, line 196:
```suggestion
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+It might be good to add a similar tier in the Python (not C) API,
+e.g. for ``types.CodeType``.
+However, the opt-in mechanism would need to be different (if any).
+This is outside the scope of the PEP.
+
+
+Open Issues
+===========
+
+- “Semi-stable” is not a perfect name.
```
AA-Turner, line 222: Sphinx will auto-convert the quotes.
```suggestion
- "Semi-stable" is not a perfect name.
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+It is very difficult to determine what needs to be exposed as semi-stable.
+
+Addidions to the semi-stable tier will count as *stabilization*,
+and will be allowed up to Release Candidate 1.
+
+
+Specification
+=============
+
+Several functions and types (“APIs”) will be moved to a new *semi-stable* tier.
```
AA-Turner, line 104:
```suggestion
Several functions and types ("APIs") will be moved to a new *semi-stable* tier.
```

AA-Turner (commented):
More comments than I intended, sorry! Some are very trivial, some are larger points (although I'm not an expert in the C-API, so please ignore things that I've gotten wrong).

A

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+
+The Python C-API is currently divided in three tiers (see
+`devguide <https://devguide.python.org/c-api/>`__ for details):
+
+- Limited API, with high compatibility expectations;
+- Public API, which follows the backwards compatibility policy
+  (PEP 387) which calls for deprecation warnings before changes;
+- Private (internal) API, which can change at any time.
+
+We need a tier between Public and Private to accomodate tools like
```
JelleZijlstra, line 34:
```suggestion
We need a tier between Public and Private to accommodate tools like
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+advanced debuggers and JIT compilers, which need access to CPython
+internals but assume the C-API they use does not change in patch releases.
+
+
+Setting Stability Expectations
+------------------------------
+
+Currently, there are no guarantees for internal API -- that is, anything
```
JelleZijlstra, line 42:
```suggestion
Currently, there are no guarantees for the internal API -- that is, anything
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+that requires ``Py_BUILD_CORE`` or is named with a leading underscore.
+This API can change without warning at any time, and code that uses it
+is pinned to a specific build of Python.
+
+However, in practice, even internal API usually happens to be stable
```
JelleZijlstra, line 47:
```suggestion
However, in practice, even the internal API usually happens to be stable
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+in patch releases:
+
+- Some CPython core developers take this as an an unwritten rule.
+- Patch releases only contain bugfixes, which are not too likely to
+  change API.
```
JelleZijlstra, line 52:
```suggestion
  change the API.
```
And probably in a lot more places

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+
+Semi-stable API will make the stability expectations more explicit.
+
+It will also hopefully encourage existing users of the private API to
+reach out to python-dev, so we can expose, standardize and test API
```
JelleZijlstra, line 57:
```suggestion
reach out to python-dev, so we can expose, standardize and test an API
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+for some of their use cases.
+
+
+Reserving underscores for Private API
+-------------------------------------
+
+PEP 523 introduced functions for use by debuggers and JIT compilers,
```
JelleZijlstra, line 64:
```suggestion
:pep:`523` introduced functions for use by debuggers and JIT compilers,
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+which are to be only stable across minor releases.
```
JelleZijlstra, line 65:
```suggestion
which are stable only across minor releases.
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+
+Changes during the Beta period
+------------------------------
+
+Since the API itself can change continuously up until Beta 1 (feature freeze)
+of a minor version, major users of this API are unlikely to test
+Alpha releases and provide feedback.
+It is very difficult to determine what needs to be exposed as semi-stable.
+
+Addidions to the semi-stable tier will count as *stabilization*,
```
JelleZijlstra, line 97:
```suggestion
Additions to the semi-stable tier will count as *stabilization*,
```

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+The semi-stable C-API tier and ``Py_USING_SEMI_STABLE_API`` will be documented,
+and documentation of each semi-stable API will be updated.
+
+
+Adjustments during Beta periods
+-------------------------------
+
+New APIs can be added to the semi-stable tier, and private APIs can be moved
+to it, up to the first release candidate of a new minor version.
+Consensus on the ``capi-sig`` or ``python-dev`` is needed it the beta period.
```
JelleZijlstra, line 151:
```suggestion
Consensus on the ``capi-sig`` or ``python-dev`` is needed in the Beta period.
```
No strong opinion on whether to capitalize "Beta", but you capitalize it on the next line too.

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+Reference Implementation
+========================
+
+https://github.com/python/cpython/issues/91744
+
+
+Rejected Ideas
+==============
+
+It might be good to add a similar tier in the Python (not C) API,
```
JelleZijlstra, line 213: I feel like in practice this is already true for e.g. the `ast` module; I think Guido brought that up in the past too.

pep-0689.rst
JelleZijlstra, line 222: I don't really like the name either, but haven't been able to think of anything better.

encukou: But it accepts proper quotes as well, as far as I know.

encukou:
Thank you for the editing!
I'll merge so the PEP can be reviewed on python-dev; I don't mean to ignore the unresolved conversations.

pep-0689.rst
```diff
@@ -0,0 +1,231 @@
...
+`devguide <https://devguide.python.org/c-api/>`__ for details):
+
+- Limited API, with high compatibility expectations;
+- Public API, which follows the backwards compatibility policy
+  (PEP 387) which calls for deprecation warnings before changes;
+- Private (internal) API, which can change at any time.
+
+We need a tier between Public and Private to accomodate tools like
+advanced debuggers and JIT compilers, which need access to CPython
+internals but assume the C-API they use does not change in patch releases.
```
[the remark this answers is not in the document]
AA-Turner: The current wording is in the rationale section, which is setting out the justification for the proposal. To me, it seems odd to phrase it "We need <feature> because <thing> requires <requirement>, but we don't guarantee requirement.

My suggested rephrasing more explicity spells out that we want to support <thing> properly, there is a reason the current state doesn't work (no guarantees at an API level of stability), and hence <feature> is needed that provides those guarantees.

So this was more a "readability" suggestion, though as author you very much have the right to reject it!

A

PR #2568, opened 2022-05-02
PEP 1: Update "issue tracker" to link to GitHub Issues

itamaro:
with all the issues migrated to GitHub Issues, it makes sense now for PEP-1 to point to GitHub Issues over BPO

### DIFF
--- a/pep-0001.txt
+++ b/pep-0001.txt
@@ -124,7 +124,7 @@ Start with an idea for Python
 The PEP process begins with a new idea for Python.  It is highly
 recommended that a single PEP contain a single key proposal or new
 idea; the more focused the PEP, the more successful it tends to be.
-Small enhancements and bug fixes often don't need a PEP and
+Most enhancements and bug fixes don't need a PEP and
 can be submitted directly to the `Python issue tracker`_.
 The PEP editors reserve the
 right to reject PEP proposals if they appear too unfocused or too
--- a/pep-0001.txt
+++ b/pep-0001.txt
@@ -123,13 +123,12 @@ Start with an idea for Python
 
 The PEP process begins with a new idea for Python.  It is highly
 recommended that a single PEP contain a single key proposal or new
-idea. Small enhancements or changes often don't need a PEP and can
-be integrated into the Python development workflow by reporting an
-issue to the Python `issue tracker`_ and/or submitting a `pull request`_.
-The more focused the PEP, the more successful it tends to be.
-The PEP editors reserve the right to reject PEP proposals if they
-appear too unfocused or too broad.  If in doubt,
-split your PEP into several well-focused ones.
+idea; the more focused the PEP, the more successful it tends to be.
+Small enhancements and bug fixes often don't need a PEP and
+can be submitted directly to the `Python issue tracker`_.
+The PEP editors reserve the
+right to reject PEP proposals if they appear too unfocused or too
+broad.  If in doubt, split your PEP into several well-focused ones.
 
 Each PEP must have a champion -- someone who writes the PEP using the style
 and format described below, shepherds the discussions in the appropriate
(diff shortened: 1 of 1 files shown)


Rosuav:
Yep, seems good. :+1:  Might also be worth saying "pull request" rather than "patch submission".

itamaro:
> Might also be worth saying "pull request" rather than "patch submission".

oh, good call!

this could use a little wordsmithing. 

the current sentence is:

> Small enhancements or patches often don't need
> a PEP and can be injected into the Python development workflow with a
> patch submission to the Python issue tracker.

I think changing "patch submission" to "pull request" would cause more confusion, since pull requests are not created via the issue tracker (as perhaps patches were attached to BPO issue in the past).

to make this accurate, we need to say something along these lines:

> Small enhancements or changes often don't need a PEP and
> can be integrated into the Python development workflow by reporting
> an issue to the Python issue tracker and submitting a pull request.

if this looks good, I'll update the PR with the changes.

JelleZijlstra:
@itamaro that wording sounds good to me.

Rosuav:
On GitHub, the issue tracker manages pull requests too, so it's not terribly wrong. What about this slight change?

> Small enhancements or changes often don't need a PEP and
> can be integrated into the Python development workflow by reporting
> an issue to the Python issue tracker and/or submitting a pull request.

itamaro:
thanks for the feedback and suggestions - update the PR and added a link to GitHub pull requests too

JelleZijlstra (approved):

CAM-Gerlach:
That's better, though IMO that sentence can be made a lot simpler, and I don't know if its wise to push users directly to pull requests since any non-trivial change requires an issue. Also, due to the automatic re-wrapping to fit the hard breaks, it ends up touching a fairly substantial number of lines and is hard to see what was actually modified,

What about [this change](https://github.com/python/peps/compare/main...CAM-Gerlach:pep-1-12-issue-tracker-links?expand=1) as implemented on [this branch](https://github.com/CAM-Gerlach/peps/tree/pep-1-12-issue-tracker-links) instead?

itamaro:
> What about [this change](https://github.com/python/peps/compare/main...CAM-Gerlach:pep-1-12-issue-tracker-links?expand=1) as implemented on [this branch](https://github.com/CAM-Gerlach/peps/tree/pep-1-12-issue-tracker-links) instead?

that looks good to me

AA-Turner:
s/Small enhancements/Most enhancements/ in either version -- de facto only substantive changes go through the PEP process.

A

AA-Turner:
+0 to Cam's wording, +1 to the change in general.

A

CAM-Gerlach:
After in-person discussion with @itamaro , he asked me to go ahead and push a commit with a slightly modified form of my proposed revision.

pep-0001.txt
```diff
@@ -123,10 +123,10 @@ Start with an idea for Python
 
 The PEP process begins with a new idea for Python.  It is highly
 recommended that a single PEP contain a single key proposal or new
-idea. Small enhancements or patches often don't need
-a PEP and can be injected into the Python development workflow with a
-patch submission to the Python `issue tracker`_. The more focused the
-PEP, the more successful it tends to be.  The PEP editors reserve the
+idea; the more focused the PEP, the more successful it tends to be.
+Small enhancements and bug fixes often don't need a PEP and
```
itamaro, line 127: @AA-Turner suggested changing "small" to "most"

AA-Turner:
```suggestion
Most enhancements and bug fixes often don't need a PEP and
```

AA-Turner (approved):

pep-0001.txt
CAM-Gerlach: Sorry, I made that change but it got lost when I patched my version on top of the further changes on this PR.

AA-Turner:
Thanks @itamaro!

A

PR #2578, opened 2022-05-05
PEP 691: Draft PEP for Simple JSON API

di:


pep-9999.rst
```diff
@@ -0,0 +1,452 @@
+PEP: 9999
```
JelleZijlstra, line 1: You can pick PEP 691

pep-0691.rst
```diff
@@ -0,0 +1,452 @@
...
+* While JSON doesn't natively support an URL type, any value that represents an
+  URL in this API may be either absolute or relative as long as they point to
+  the correct location. If relative, they are relative to the current URL as if
+  it were HTML.
+
+* Additional keys may be added to any dictionary objects in the API responses
+  and clients **MUST** ignore keys that they don't understand.
+
+* By default, any hash algorithm available via `hashlib
+  <https://docs.python.org/3/library/hashlib.html>` (specifically any that can
```
JelleZijlstra, line 265:
```suggestion
  <https://docs.python.org/3/library/hashlib.html>`_ (specifically any that can
```

pep-0691.rst
```diff
@@ -0,0 +1,452 @@
...
+All being said, it is the opinion of this PEP that those three issues combined make
+using separate API routes a less desirable solution than relying on content
+negotiation to select the most ideal representation of the data.
+
+
+Appendix 1: Survey of use cases to cover
+========================================
+
+This was done through a discussion between ``pip`` and ``bandersnarch``
+maintainers, where the two first potential use cases for the new API. This is
```
JelleZijlstra, line 407: Some missing word here

pep-0691.rst
```diff
@@ -0,0 +1,452 @@
+PEP: 691
+Title: JSON-based Simple API for Python Package Indexes
+Author: Donald Stufft <donald@stufft.io>,
+        Pradyun Gedam <pradyunsg@gmail.com>,
+        Cooper Lees <me@cooperlees.com>,
+        Dustin Ingram <di@python.org>
+Status: Draft
+Type: Informational
```
JelleZijlstra, line 8: Are you sure it should be Informational? It reads like a standards-track PEP.

JelleZijlstra (commented):
Congratulations on the PEP submission!

Can you also add a CODEOWNERS entry with @dstufft as the owner for this PEP?

pep-0691.rst
di: This is based on https://peps.python.org/pep-0503/ being informational, but I'm fine to change it if this doesn't make sense anymore.

pep-0691.rst
di:
```suggestion
maintainers, who were the two first potential user for the new API. This is
```

pep-0691.rst
JelleZijlstra: Interesting, I feel like it should not be Informational because you are proposing a new standard that multiple tools will have to implement. An Informational PEP would be appropriate if you're just documenting how this API works in Warehouse.

But ultimately that's a decision for the packaging community, so I won't hold up merging the PEP over it. I can merge it now if you like, or wait for your co-authors or others to chime in.

pep-0691.rst
```diff
@@ -0,0 +1,452 @@
...
+
+  - List of all files for a particular release.
+
+    - Workout URL for release files to download
+
+  - Metadata of each individual artifact.
+
+    - Write out the JSON to mirror storage today (disk/S3)
+
+        - Required metadata used (via Package class - https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch/package.py):
```
JelleZijlstra, line 435: This renders a bit weirdly: https://pep-previews--2578.org.readthedocs.build/pep-0691/#appendix-1-survey-of-use-cases-to-cover

pep-0691.rst
cooperlees: PEP 503 was written post the Simple HTML already existing so I second @JelleZijlstra here - We should probably change.

di: I agree, but let's see what the wider community thinks in discussion. Fine to merge now!

pep-0691.rst
di: I'm not sure what's causing this. Is this a limitation of rST, an expected style for PEPs, or something syntactically wrong?

pep-0691.rst
di:
```suggestion
Type: Standards Track
```

pep-0691.rst
JelleZijlstra: I think it's indented by one space too many.

pep-0691.rst
```diff
@@ -0,0 +1,452 @@
...
+
+          - metadata["info"]
+          - metadata["last_serial"]
+          - metadata["releases"]
+
+            - digests
+            - URL
```
JelleZijlstra, line 442:
```suggestion
      - Required metadata used (via Package class - https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch/package.py):

        - metadata["info"]
        - metadata["last_serial"]
        - metadata["releases"]

          - digests
            - URL
```

pep-0691.rst
JelleZijlstra: I applied the change, will confirm it looks OK in the preview and then merge.

JelleZijlstra (approved):

pep-0691.rst
dstufft: There's like a 95% chance I picked informational on PEP 503 because I didn't care very much and that seemed fine and nobody called me out on it.

CAM-Gerlach:
I'm very happy to hear that this new API is moving forward, and looking forward to seeing it implemented.

There are a handful of minor technical/rendering issues that should be corrected (inconsistent header order, obsolete BDFL-Delegate header, inconstant code-block syntax, incorrect syntax highlighting, dummy Discussions-To/Post-History, etc.), which I'm happy to help take care of.

However, there are two more critical defects that should be resolved promptly (and should have been spotted and fixed prior to merge), and block me making such changes:

* The Copyright section [required by PEP 1](https://peps.python.org/pep-0001/#what-belongs-in-a-successful-pep) is missing and should be added by an author of the PEP (such as yourself) with the concurrence of the others.
* While @dstufft is the SC-designated `PEP-Delegate` for PyPI infra, while he _is_ allowed to also simultaneously be the PEP's `Sponsor` (given none of the others are core devs or PEP editors, IIRC), [per PEP 1](https://peps.python.org/pep-0001/#submitting-a-pep) he _cannot_ simultaneously be PEP-Delegate _and_ an author of the PEP (as that would be an obvious conflict of interest). Therefore, to resolve this, I would suggest he remove himself as an author of the PEP and serve as the PEP's sponsor instead (alternatively, he could recuse himself as the PEP-Delegate, or apply to the SC for a special exception if no one objected).

Also, while they are not strictly required and not as much a priority immediately, for the benefit of both reviewers and the community at large, I'd advise at some point soon adding the relevant Suggested Sections from [PEP 1](https://peps.python.org/pep-0001/#what-belongs-in-a-successful-pep), [PEP 12](https://peps.python.org/pep-0012/#suggested-sections) and the template, particularly "Security Implications", "Backwards Compatibility" and/or "Reference Implementation". Also, given the "answers" in the "FAQ" each closely fit the purpose of the standard "Rejected Ideas" section, I'd suggest retilting them accordingly.

As a final sidenote, there's no need to include a non-working dummy link for `Discussions-To`; as [PEP 12 states](https://peps.python.org/pep-0012/#how-to-use-this-template), you should instead simply elide it when first submitting the PEP, and then just add it (and `Post-History`) immediately after.

PR #2589, opened 2022-05-08
PEP 664: Fix indentation error

pablogsal:


CAM-Gerlach:
FYI for next time, there's a RTD preview, which really helps with catching things like this before merging to production :)

PR #2601, opened 2022-05-12
PEP 623: Change status to Final

methane:
Implemented in https://github.com/python/cpython/pull/92537

### DIFF
--- a/pep-0623.rst
+++ b/pep-0623.rst
@@ -2,7 +2,7 @@ PEP: 623
 Title: Remove wstr from Unicode
 Author: Inada Naoki <songofacandy@gmail.com>
 BDFL-Delegate: Victor Stinner <vstinner@python.org>
-Discussions-To: https://mail.python.org/archives/list/python-dev@python.org/thread/BO2TQHSXWL2RJMINWQQRBF5LANDDJNHH/#BO2TQHSXWL2RJMINWQQRBF5LANDDJNHH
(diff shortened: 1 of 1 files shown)


AA-Turner (approved):
I noticed that there is no Resolution header for the PEP.
We should add that either in this PR or a follow-up.

A

hugovk:
And a `Discussions-To` with https://mail.python.org/archives/list/python-dev@python.org/thread/S7KW2U6IGXZFBMGS6WSJB26NZIBW4OLE/ ?

CAM-Gerlach:
[This thread](https://mail.python.org/archives/list/python-dev@python.org/thread/VQKDIZLZ6HF2MLTNCUFURK2IFTXVQEYA/#VQKDIZLZ6HF2MLTNCUFURK2IFTXVQEYA) is the resolution.

Its hard to be sure since the text is somewhat ambiguous and highly technical, but it sounds like that thread is not actually about PEP 623 itself, which concerns only the APIs using the `wstr` Unicode struct member, while the other APIs are stated to be out of scope by the PEP, whereas the thread is titled "Plan to remove Py_UNICODE APIs _except_ PEP 623" and lists a different set of objects to be removed than the PEP does. However, I can't find any thread about PEP 623 either searching Python-Dev or Discourse. There's [this thread](https://mail.python.org/archives/list/python-dev@python.org/thread/THXVM7FZVT56B7CPEDIYKJG6VMAYIEK5/) on the related PEP 624, but that's a separate PEP.

Inada-san, could you help us out with a link to the thread? Or @vstinner ?

vstinner (approved):
LGTM.

AA-Turner (approved):

PR #2607, opened 2022-05-15
PEP 517: Add missing tarfile import (used in _make_sdist hook)

wimglenn:


CAM-Gerlach:
Seems like this is acceptable, since it fixes a straightforward issue with an example and as of yet, there's no packaging.python.org build system spec?

CAM-Gerlach:
(Also, it seems like I can no longer remove code owners from the reviewer requests, even as a maintainer, as removing them appears to do nothing—unless it is the weird UI/caching issue I was experiencing before, where attributes set in the sidebar randomly didn't update or even reverted themselves)

AA-Turner:
> I can no longer remove code owners from the reviewer requests

Seemed to work for me just now.

A

AA-Turner (approved):

brettcannon (approved):

PR #2616, opened 2022-05-21
Infra: Use Docutils' `list-table` syntax for PEP 0

AA-Turner:
Change in support of subindices.

A

### DIFF
--- a/pep_sphinx_extensions/pep_zero_generator/writer.py
+++ b/pep_sphinx_extensions/pep_zero_generator/writer.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import datetime
-import functools
 from typing import TYPE_CHECKING
 import unicodedata
 
@@ -25,16 +24,6 @@ from pep_sphinx_extensions.pep_zero_generator.errors import PEPError
 
 if TYPE_CHECKING:
     from pep_sphinx_extensions.pep_zero_generator.parser import PEP
-
-title_length = 55
-author_length = 40
-table_separator = "== ====  " + "="*title_length + " " + "="*author_length
-
-# column format is called as a function with a mapping containing field values
-column_format = functools.partial(
-    "{type}{status}{number: >5}  {title: <{title_length}} {authors}".format,
-    title_length=title_length
-)
 
 header = f"""\
 PEP: 0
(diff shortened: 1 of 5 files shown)


AA-Turner:
Save for the CLA check, this is now done -- identical output in PEP 0, and a nice reduction in complexity as (almost) all of the custom PEP 0 transform is now gone (only the email address masking logic remains).

A

AA-Turner:
\*Not quite identical output to PEP 0, PEP titles are no longer truncated, and PEP 604's title now links correctly.

A

pep_sphinx_extensions/pep_zero_generator/writer.py
```diff
@@ -80,21 +69,27 @@ def emit_text(self, content: str) -> None:
...
-        self.emit_table_separator()
+        self.emit_text(".. list-table::")
+        self.emit_text("   :header-rows: 1")
+        self.emit_text("   :widths: auto")
+        self.emit_text("   :class: pep-zero-table")
+        self.emit_newline()
+        self.emit_text("   * - ")
+        self.emit_text("     - PEP")
+        self.emit_text("     - PEP Title")
+        self.emit_text("     - PEP Author(s)")
```
hugovk, line 92: "Author(s)" is a bit ugly, for a table heading can we just put "Authors" or "Author"?

Also I think it's clear that the title is the _PEP_ title, and the authors are the _PEP_ authors, in a list of PEPs on the PEP index page on https://peps.python.org/ :)

```suggestion
        self.emit_text("     - PEP")
        self.emit_text("     - Title")
        self.emit_text("     - Authors")
```

hugovk (commented):
It would be nice to have some tests for the changed `emit_*` functions in `writer.py`. 

(Testing just `emit_pep_category` would also cover `emit_column_headers` and `emit_pep_row`.)

https://codecov.io/gh/python/peps/compare/278070ac9765efbd34e325d6080d8162bd529e99...b3a603d2fede03f207a26eaec584d13e6f8ebb13/src/pep_sphinx_extensions/pep_zero_generator/writer.py?before=pep_sphinx_extensions/pep_zero_generator/writer.py

AA-Turner:
> It would be nice to have some tests

I'm not sure what these would look like -- there's no real point to unit tests of just the function here, as what matters is that the entire list is a valid input to the `list-table` directive (a two-level list where the length of every second level list is uniform).

I'd then argue that the integration test part is handled by if it is parsed correctly as reST!

A

AA-Turner:
(Also, the CLA is still not working -- just tried again)

hugovk:
Yeah, tests not so valuable here. If nothing else, a test would show the code is at least running without exceptions, although we should _probably_ notice it (but if not a super simple test would have helped ;)

CLA: I've pinged @ambv again.

PR #2627, opened 2022-06-03
Update pep-0469.txt

JanGoeteyn:


AA-Turner:
This is a Withdrawn PEP and won't be updated (I agree that the code sample is incorrect, but there is little value in updating it, and elsewhere it is correct, so the meaning is not lost on the reader.)

A

PR #2636, opened 2022-06-08
Several PEPs: Add `Topic: Packaging`

AA-Turner:
Depends on #2579

Opened as draft to not ping for review.

A

PR #2656, opened 2022-06-18
PEP 632: Remove `Topic: Packaging` header

pradyunsg:
This removes the header, as requested by the author in https://github.com/python/peps/pull/2636#discussion_r898173875.


CAM-Gerlach:
Rather unfortunate, considering it is of obvious interest and considerable importance to the packaging ecosystem as a whole, and the Topics are not (at least as originally construed) mutually-exclusive.

AA-Turner:
I merged as it seems @pradyunsg & others only wanted the packaging topic to be PyPA topics -- other PEPs that were to do with python packages, but not the business of packaging, were earlier removed too.

A

PR #2665, opened 2022-06-23
PEP 650: Withdraw and move to Standards Track

pradyunsg:
The discussion and interest in this PEP has died down.

Moving to Standards Track in line with the packaging specification process.

This should probably wait for approval from at least one of the authors.


brettcannon:
@di I'm okay with this if you are.

di (approved):

PR #2673, opened 2022-06-25
PEP 668: Address feedback and mark as accepted

pradyunsg:


pradyunsg:
I'm not sure I understand why the linter does not like this URL.

AA-Turner:
@CAM-Gerlach the link is valid, do we need to update the regex pattern? From the lint job:

```text
'Resolution' must be a direct thread/message URL..................................Failed
- hook id: validate-resolution
- exit code: 1

pep-0668.rst:14:Resolution: https://discuss.python.org/t/10302/44
```

A

pradyunsg:
Figured it out -- the `Resolution` header needs to be at then end of the header list, to pass the lint check.

pradyunsg:
This is ready for a merge from an editor, whenever someone has a moment to click the button. :)

pradyunsg:
Thanks @JelleZijlstra! ^>^

PR #2678, opened 2022-06-27
Remove the requirement for Resolution to be the final header

pradyunsg:
See #2673 for an example of where this existing description was insufficient.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -220,9 +220,9 @@ repos:
         types: [text]
 
       - id: validate-resolution
-        name: "'Resolution' must be a direct thread/message URL and the final header"
+        name: "'Resolution' must be a direct thread/message URL"
         language: pygrep
-        entry: '(?<!\n\n)(?<=\n)Resolution: (?:(?!https://((discuss\.python\.org/t/([\w\-]+/)?\d+(/\d+)?/?)|(mail\.python\.org/pipermail/[\w\-]+/\d{4}-[A-Za-z]+/[A-Za-z0-9]+\.html)|(mail\.python\.org/archives/list/[\w\-]+@python\.org/(message|thread)/[A-Za-z0-9]+/?(#[A-Za-z0-9]+)?))\n(?=\n)))'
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -220,7 +220,7 @@ repos:
         types: [text]
 
       - id: validate-resolution
-        name: "'Resolution' must be a direct thread/message URL"
+        name: "'Resolution' must be a direct thread/message URL and the final header"
         language: pygrep
         entry: '(?<!\n\n)(?<=\n)Resolution: (?:(?!https://((discuss\.python\.org/t/([\w\-]+/)?\d+(/\d+)?/?)|(mail\.python\.org/pipermail/[\w\-]+/\d{4}-[A-Za-z]+/[A-Za-z0-9]+\.html)|(mail\.python\.org/archives/list/[\w\-]+@python\.org/(message|thread)/[A-Za-z0-9]+/?(#[A-Za-z0-9]+)?))\n(?=\n)))'
```
CAM-Gerlach, line 225: I would have liked to include more detailed descriptions of the checks in the CLI, but we also don't want to overflow the terminal width, and this one is a fair bit longer than all the other checks. As such, since (AFAIK) it isn't really any more critical that Resolution be at the end compared to any other check being in its nominal order (at least, AFAIK), IMO the better solution for PEP authors is to just drop the check for it being at the end, which avoids any extra nitpickyness to begin with.

```suggestion
        name: "'Resolution' must be a direct thread/message URL"
        language: pygrep
        entry: '(?<!\n\n)(?<=\n)Resolution: (?:(?!https://((discuss\.python\.org/t/([\w\-]+/)?\d+(/\d+)?/?)|(mail\.python\.org/pipermail/[\w\-]+/\d{4}-[A-Za-z]+/[A-Za-z0-9]+\.html)|(mail\.python\.org/archives/list/[\w\-]+@python\.org/(message|thread)/[A-Za-z0-9]+/?(#[A-Za-z0-9]+)?))\n))'
```

CAM-Gerlach (changes requested):
IMO, its better to just drop the check for it being at the end, since it isn't really that necessary, as opposed to making the description more complicated and increasing the min terminal width substantially.

.pre-commit-config.yaml
pradyunsg: Sure, why not. 🤷🏽

pradyunsg:
Please squash-merge this.

JelleZijlstra (approved):

CAM-Gerlach:
> Please squash-merge this, if this gets merged.

That's the only option we have enabled :smile:

CAM-Gerlach (approved):
LGTM, thanks @pradyunsg !

PR #2683, opened 2022-06-28
Infra: gitignore generated `/topic` and wipe via `make clean`

hugovk:
Fixes #2659.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,5 @@ __pycache__
 *.swp
 /build
 /package
+/topic
 /venv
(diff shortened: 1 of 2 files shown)


CAM-Gerlach (approved):
For the record, IMO the general approach @pradyunsg proposed, checking in actual topic index pages to the source and using a custom `pep-table` directive to generate all the tabs would be a more sustainable long-term solution than adding our own custom reST source file generation and somewhat hackily getting Sphinx to recognize them, but may as well do this for now.

PR #2685, opened 2022-06-28
PEP 694: Grammar changes + reason phrase deprecation mention

Wouterkoorn:
Some small grammar changes, and added mention of HTTP2 having deprecated the HTTP status reason phrase.

### DIFF
--- a/pep-0694.rst
+++ b/pep-0694.rst
@@ -443,7 +443,7 @@ Session Status
 
 Similarly to file upload, the session URL is provided in the response to
 creating the upload session, and clients **MUST NOT** assume that there is any
-communality to what those URLs look like from one session to the next.
+commonality to what those URLs look like from one session to the next.
 
 To check the status of a session, clients issue a ``GET`` request to the
 session URL, to which the server will respond with the same response that
--- a/pep-0694.rst
+++ b/pep-0694.rst
@@ -38,10 +38,10 @@ Beyond the above, there are a number of major issues with the current API:
   not go last, possibly some hard to build packages are attempting to be built
   from source.
 
-- It has very limited support for communicating back to the user, limited entirely
-  to the HTTP status code and reason phrase, which has been deprecated 
-  since HTTP/2 (:rfc:`7540` #section-8.1.2.4). It has no 
-  support for multiple errors, warnings, deprecations, etc.
+- It has very limited support for communicating back to the user, with no support 
+  for multiple errors, warnings, deprecations, etc. It is limited entirely to the 
+  HTTP status code and reason phrase, of which the reason phrase has been 
+  deprecated since HTTP/2 (:rfc:`RFC 7540 <7540#section-8.1.2.4>`). 
 
 - The metadata for a release/file is submitted alongside the file, however this
   metadata is famously unreliable, and most installers instead choose to download
--- a/pep-0694.rst
+++ b/pep-0694.rst
@@ -263,7 +263,7 @@ to actually upload each of those files.
 
 There is no set endpoint for actually uploading the file, that is given to the
 client by the server as part of the creation of the upload session, and clients
-**MUST NOT** assume that there is any communality to what those URLs look like from
+**MUST NOT** assume that there is any commonality to what those URLs look like from
 one session to the next.
 
 To initiate a file upload, a client sends a ``POST`` request to the upload URL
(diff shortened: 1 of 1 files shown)


Wouterkoorn:
I could not find a way to link to a specific section on the referenced RFC...

pep-0694.rst
```diff
@@ -263,7 +263,7 @@ to actually upload each of those files.
 
 There is no set endpoint for actually uploading the file, that is given to the
 client by the server as part of the creation of the upload session, and clients
-**MUST NOT** assume that there is any stability to what those URLs look like from
+**MUST NOT** assume that there is any communality to what those URLs look like from
```
JelleZijlstra, line 266:
```suggestion
**MUST NOT** assume that there is any commonality to what those URLs look like from
```

pep-0694.rst
```diff
@@ -443,7 +443,7 @@ Session Status
 
 Similarly to file upload, the session URL is provided in the response to
 creating the upload session, and clients **MUST NOT** assume that there is any
-stability to what those URLs look like from one session to the next.
+communality to what those URLs look like from one session to the next.
```
JelleZijlstra, line 446:
```suggestion
commonality to what those URLs look like from one session to the next.
```

JelleZijlstra (commented):
cc @dstufft

dstufft (approved):

pep-0694.rst
```diff
@@ -39,9 +39,9 @@ Beyond the above, there are a number of major issues with the current API:
   from source.
 
 - It has very limited support for communicating back to the user, limited entirely
-  to the HTTP status code, and status message (something which I'm not sure is
-  even technically valid HTTP?). It has no support for multiple errors, warnings,
-  deprecations, etc.
+  to the HTTP status code and reason phrase, which has been deprecated
```
gvanrossum, line 42: This is ambiguous -- is the status code also deprecated, or only the reason phrase?

dstufft: Just the reason phrase, which the status quo file upload API on PyPI uses custom reason phrases to communicate errors back to the user. It would be good to make it less ambiguous though.

pep-0694.rst
```diff
@@ -39,9 +39,9 @@ Beyond the above, there are a number of major issues with the current API:
   from source.
 
 - It has very limited support for communicating back to the user, limited entirely
-  to the HTTP status code, and status message (something which I'm not sure is
-  even technically valid HTTP?). It has no support for multiple errors, warnings,
-  deprecations, etc.
+  to the HTTP status code and reason phrase, which has been deprecated 
+  since HTTP/2 (:rfc:`7540` #section-8.1.2.4). It has no
```
hugovk, line 43: > I could not find a way to link to a specific section on the referenced RFC...

Here's the magic words. You can change "RFC 7540" to whatever you'd like as the link display text:

```suggestion
  since HTTP/2 (:rfc:`RFC 7540 <7540#section-8.1.2.4>`). It has no 
```

pep-0694.rst
CAM-Gerlach: Yeah, it definitely should be less ambiguous. As a starting point, what about

```suggestion
  to the HTTP status code and the (deprecated) reason phrase,
```

pep-0694.rst
Wouterkoorn: In the preview this does not seem to be working. Are you confident this will work on peps.python.org?

CAM-Gerlach: You have to actually commit the suggestion to regenerate the [preview](https://pep-previews--2685.org.readthedocs.build/pep-0694/), unless you mean you tested it locally? If so, could you provide a link/screenshot, and/or be more specific as to what actual behavior you are seeing?

Unless there's something subtle I'm missing, the syntax is indeed correct.

Wouterkoorn: Apologies for the vague phrasing of 'preview'. I was talking about the github preview you can see when editing, which does work with just :rfc:`7540`, but not with this change.
I'll commit this suggested syntax, thanks!

Wouterkoorn: It works :heavy_check_mark:

CAM-Gerlach: > I was talking about the github preview you can see when editing, which does work with just :rfc:`7540`, but not with this change.

Oh, the GitHub preview should not be trusted in the slightest, as its reST parser is rather limited to begin with, and that's without all our custom extensions to process and render the PEPs.

PR #2713, opened 2022-07-14
PEP 695: Incorporated recent feedback from python-dev.

erictraut:
* Someone pointed out that rust does include a way to specify default type arguments, something I missed when compiling the survey in Appendix A.
* Updated the "Rejected Ideas" section to address a couple of questions that were raised, including the use of angle brackets and the use of decorator-style syntax.


gvanrossum (approved):
“Unsightly and cumbersome”, indeed. :-)

PR #2748, opened 2022-08-05
PEP-669: Fix version number

markshannon:


CAM-Gerlach:
If the PEP is primarily targeting and describing Python 3.12, can we also add a

```email
Python-Version: 3.12
```

header between `Created` and `Post-History`?

PR #2757, opened 2022-08-10
Lint: Add missing Python-Version header in PEPs

CAM-Gerlach:
As originally discussed in #2755 , a number of older non-Packaging Accepted/Final Standards Track PEPs are missing Python-Version headers even though they were implemented. I've added those missing them via cross-referencing the What's New in Python documents, as well as the CPython commit history and the PEPs themselves.


ericsnowcurrently:
The PEP 421 change is fine.

PR #2760, opened 2022-08-11
PEP 669: Clarify and restrict interaction with PEP 523

markshannon:
@encukou is this better?


encukou:
Definitely. This is much clearer, thank you!


(looking at it on a phone as long weekend starts; I'll re-read on a proper screen next week)

On August 11, 2022 6:59:43 PM GMT+02:00, Mark Shannon ***@***.***> wrote:
***@***.*** is this better?

PR #2766, opened 2022-08-16
PEP 683: Updates for Round 4 of Discussions

ericsnowcurrently:
Note: I've left some TODO comments as placeholders in response to some of the feedback.

CC @eduardo-elizondo

### DIFF
--- a/pep-0683.rst
+++ b/pep-0683.rst
@@ -81,7 +81,10 @@ possible negative impact to consider.  The threshold for an "acceptable"
 performance penalty for immortal objects is 2% (the consensus at the
 2022 Language Summit).  A naive implementation of the approach described
 below makes CPython roughly 4% slower.  However, the implementation
-is performance-neutral once known mitigations are applied.
+is ~performance-neutral~ once known mitigations are applied.
+
+TODO: Update the performance impact for the latest branch
+(both for GCC and for clang).
 
 
 Motivation
@@ -211,7 +214,7 @@ Performance
 -----------
 
 A naive implementation shows `a 4% slowdown`_.  We have demonstrated
-a return to performance-neutral with a handful of basic mitigations
+a return to ~performance-neutral~ with a handful of basic mitigations
 applied.  See the `mitigations`_ section below.
 
 On the positive side, immortal objects save a significant amount of
@@ -220,6 +223,9 @@ objects provide opportunities for specialization in the eval loop that
 would improve performance.
 
 .. _a 4% slowdown: https://github.com/python/cpython/pull/19474#issuecomment-1032944709
+
+TODO: Update the performance impact for the latest branch
+(both for GCC and for clang).
 
 Backward Compatibility
 ----------------------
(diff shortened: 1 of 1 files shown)


ericsnowcurrently:
@eduardo-elizondo, am I missing anything from the discussion (aside from the TODO items)?

pep-0683.rst
```diff
@@ -72,14 +72,19 @@ Here's a high-level look at the implementation:
 If an object's refcount matches a very specific value (defined below)
 then that object is treated as immortal.  The CPython C-API and runtime
 will not modify the refcount (or other runtime state) of an immortal
-object.
+object.  The runtime will now be explicitly responsible for deallocating
+all immortal objects during finalization, unless statically allocated.
+(See `Object Cleanup`_ below.)
```
CAM-Gerlach, line 77: FYI, you might want to consider using [the `:ref:` role and ref target labels](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref) (e.g. `` :ref:`683-object-cleanup` linking to `` .. _683-object-cleanup:``) instead of relying on implicit internal links, as they don't break if you change the section headings, will raise warnings if they do break or are incorrect, and can be used across files, builders and even in other docs through Intersphinx.

pep-0683.rst
```diff
@@ -503,9 +514,10 @@ mutable object won't actually be modified.
 On the other hand, some mutable objects will never be shared between
 threads (at least not without a lock like the GIL).  In some cases it
 may be practical to make some of those immortal too.  For example,
-``sys.modules`` is a per-interpreter dict that we do not expect to ever
-get freed until the corresponding interpreter is finalized.  By making
-it immortal, we no longer incur the extra overhead during incref/decref.
+``sys.modules`` is a per-interpreter dict that we do not expect to
```
CAM-Gerlach, line 517: FYI, if you like, you can use Intersphinx to actually cross-reference this in the CPython docs, i.e. 

```suggestion
:data:`python:sys.modules` is a per-interpreter dict that we do not expect to
```

ericsnowcurrently: That's good to know.  Thanks.  It isn't particularly beneficial in this case but I can think of other places where it might be.

pep-0683.rst
ericsnowcurrently: Thanks for pointing this out.  I'll take a look.

pep-0683.rst
```diff
@@ -72,14 +72,19 @@ Here's a high-level look at the implementation:
...
+all immortal objects during finalization, unless statically allocated.
+(See `Object Cleanup`_ below.)
 
 Aside from the change to refcounting semantics, there is one other
 possible negative impact to consider.  The threshold for an "acceptable"
 performance penalty for immortal objects is 2% (the consensus at the
 2022 Language Summit).  A naive implementation of the approach described
-below makes CPython roughly 6% slower.  However, the implementation
-is performance-neutral once known mitigations are applied.
+below makes CPython roughly 4% slower.  However, the implementation
```
eduardo-elizondo, line 83: This is probably much much worse today, maybe we should re-measure? Or just remove the 4% completely and mention that it's "significantly slower"?

eduardo-elizondo:
Changes loko solid!

pep-0683.rst
ericsnowcurrently: I think I'll leave "roughly 4%".  An exactly accurate value doesn't communicate anything extra and "significantly slower" doesn't offer any good scale.

PR #2767, opened 2022-08-18
PEP 669: nitpicks & clarifications

encukou:
A few wording fixes & clarifications.
It's entirely possible the clarifications are wrong, please clarify the other way if so :)


markshannon:
Thanks, look good.

PR #2772, opened 2022-08-25
PEP 697 (new): C-API for Extending Opaque Types

encukou:
I've juggled with this long enough, time for more eyes to see it.

### DIFF
--- a/pep-0697.rst
+++ b/pep-0697.rst
@@ -1,0 +1,357 @@
+PEP: 697
+Title: C API for Extending Opaque Types
+Author: Petr Viktorin <encukou@gmail.com>
+Status: Draft
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 2022-08-23
+Python-Version: 3.12
+
+
+Abstract
+========
+
+Add limited C API for extending types whose ``struct`` is opaque,
+by allowing code to only deal with data specific to a particular (sub)class.
+
+Make the mechanism usable with ``PyHeapType``.
+
+
+Motivation
+==========
+
+Extending opaque types
+----------------------
+
+In order to allow changing/optimizing CPython, and allow freedom for alternate
+implementations of the C API, best practice is to not expose memory layout
+(C structs) in public API, and instead rely on accessor functions.
+(When this hurts performance, direct struct access can be allowed in a
+less stable API tier, at the expense of compatibility with diferent
+versions/implementations of the interpreter.)
+
+However, when a particular type's instance struct is hidden, it becomes
+difficult to subclass it.
+The usual subclassing pattern, explained `in the tutorial <https://docs.python.org/3/extending/newtypes_tutorial.html#subclassing-other-types>`_,
+is to put the base class ``struct`` as the first member of the subclass ``struct``.
+The tutorial shows this on a ``list`` subtype with extra state; adapted to
+a heap type (``PyType_Spec``) the example reads:
+
+.. code-block:: c
+
+    typedef struct {
+        PyListObject list;
+        int state;
+    } SubListObject;
+
+    static PyType_Spec Sublist_spec = {
+        .name = "sublist.SubList",
+        .basicsize = sizeof(SubListObject),
+        .itemsize = 0,
+        .flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
+        .slots = SubList_slots
+    };
+
+Since the superclass struct (``PyListObject``) is part of the subclass struct
+(``SubListObject``):
+
+- ``PyListObject`` size must be known at compile time, and
+- the size must be the same across all interpreters/versions the compiled
+  extension is ABI-compatible with.
+
+But in limited API/stable ABI, we do not expose the size of ``PyListObject``,
+so that it can vary between CPython versions (and even between possible
+alternate ABI-compatible C API implementations).
+
+With the size not available, limited API users must resort to workarounds such
+as querying ``__basicsize__`` and plugging it into ``PyType_Spec`` at runtime,
+and divining the correct offset for their extra data.
+This requires making assumptions about memory layout, which the limited API
+is supposed to hide.
+
+
+Extending variable-size objects
+-------------------------------
+
+Another scenario where the traditional way to extend an object does not work
+is variable-sized objects, i.e. ones with non-zero ``tp_itemsize``.
+If the instance struct ends with a variable-length array (such as
+in ``tuple`` or ``int``), subclasses cannot add their own extra data without
+detailed knowledge about how the superclass allocates and uses its memory.
+
+Some types, such as CPython's ``PyHeapType``, handle this by storing
+variable-sized data after the fixed-size struct.
+This means that any subclass can add its own fixed-size data.
+(Only one class in the inheritance hierarchy can use variable-sized data, though.)
+This PEP proposes API that makes this practice easier, and ensures the
+variable-sized data is properly aligned.
+
+Note that many variable-size types, like ``int`` or ``tuple``, do not use
+this mechanism. 
+This PEP does not propose any changes to existing variable-size types (like
+``int`` or ``tuple``) except ``PyHeapType``.
+
+
+Extending ``PyHeapType`` specifically
+-------------------------------------
+
+The motivating problem this PEP solves is creating metaclasses, that is,
+subclasses of ``type``.
+The underlying ``PyHeapTypeObject`` struct is both and variable-sized, and
+opaque in the limited API.
+
+Projects such as language bindings and frameworks that need to attach custom
+data to metaclasses currently resort to questionable workarounds.
+The situation is worse in projects that target the Limited API.
+
+For an example of the currently necessary workarounds, see:
+`nb_type_data_static <https://github.com/wjakob/nanobind/blob/f3044cf44763e105428e4e0cf8f42d951b9cc997/src/nb_type.cpp#L1085>`_
+in the not-yet-released limited-API branch of ``nanobind``
+(a spiritual successor of the popular C++ binding generator ``pybind11``).
+
+
+Rationale
+=========
+
+This PEP proposes a different model: instead of the superclass data being
+part of the subclass data, the extra space a subclass needs is specified
+and accessed separately.
+(How base class data is accessed is left to whomever implements the base class:
+they can for example  provide accessor functions, expose a part of its
+``struct`` for better performance, or do both.)
+
+The proposed mechanism allows using static, read-only ``PyType_Spec``
+even if the superclass struct is opaque, like ``PyTypeObject`` in
+the Limited API.
+
+Combined with a way to create class from ``PyType_Spec`` and a custom metaclass,
+this will allow libraries like nanobind or JPype to create metaclasses
+without making assumptions about ``PyTypeObject``'s memory layout.
+The approach generalizes to non-metaclass types as well.
+
+
+Specification
+=============
+
+In the code blocks below, only function headers are part of the specification.
+Other code (the size/offset calculations) are details of the initial CPython
+implementation, and subject to change.
+
+Relative ``basicsize``
+----------------------
+
+The ``basicsize`` member of ``PyType_Spec`` will be allowed to be zero or
+negative.
+In that case, it will specify the inverse of *extra* storage space instances of
+the new class require, in addition to the basicsize of the base class.
+That is, the basicsize of the resulting class will be:
+
+.. code-block:: c
+
+   type->tp_basicsize = _align(base->tp_basicsize) + _align(-spec->basicsize);
+
+where ``_align`` rounds up to a multiple of ``alignof(max_align_t)``.
+When ``spec->basicsize`` is zero, ``base->tp_basicsize`` will be inherited
+directly instead (i.e. set to ``base->tp_basicsize`` without aligning).
+
+On an instance, the memory area specific to a subclass -- that is, the
+“extra space” that subclass reserves in addition its base -- will be available
+using a new function, ``PyObject_GetTypeData``.
+In CPython, this function will be defined as:
+
+.. code-block:: c
+
+   void *
+   PyObject_GetTypeData(PyObject *obj, PyTypeObject *cls) {
+       return (char *)obj + _align(cls->tp_base->tp_basicsize);
+   }
+
+Another function will be added to retreive the size of this memory area:
+
+.. code-block:: c
+
+   Py_ssize_t
+   PyObject_GetTypeDataSize(PyTypeObject *cls) {
+       return cls->tp_basicsize - _align(cls->tp_base->tp_basicsize);
+   }
+
+The functionality comes with two important caveats, which will be pointed out
+in documentation:
+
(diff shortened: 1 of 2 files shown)


pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+This PEP does not propose any changes to existing variable-size types (like
+``int`` or ``tuple``) except ``PyHeapType``.
+
+
+Extending ``PyHeapType`` specifically
+-------------------------------------
+
+The motivating problem this PEP solves is creating metaclasses, that is,
+subclasses of ``type``.
+The underlying ``PyHeapTypeObject`` struct is both and variable-sized, and
```
gvanrossum, line 100: both _what_ and variable-sized? Missing word?

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+opaque in the limited API.
+
+Projects such as language bindings and frameworks that need to attach custom
+data to metaclasses currently resort to questionable workarounds.
+The situation is worse in projects that target the Limited API.
+
+For an example of the currently necessary workarounds, see:
+`nb_type_data_static <https://github.com/wjakob/nanobind/blob/f3044cf44763e105428e4e0cf8f42d951b9cc997/src/nb_type.cpp#L1085>`_
```
gvanrossum, line 108: Can't say I understand what's going on there, since it's modern C++ and I don't read that.

Could you perhaps explain what it is doing?

Are you aware of pure-C examples of similar workarounds? Or is C++ the main target here?

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+in the not-yet-released limited-API branch of ``nanobind``
+(a spiritual successor of the popular C++ binding generator ``pybind11``).
+
+
+Rationale
```
gvanrossum, line 113: I don't think this is the right section heading here. Maybe "proposal"?

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+=========
+
+This PEP proposes a different model: instead of the superclass data being
+part of the subclass data, the extra space a subclass needs is specified
+and accessed separately.
+(How base class data is accessed is left to whomever implements the base class:
```
gvanrossum, line 119: whoever

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+in the not-yet-released limited-API branch of ``nanobind``
+(a spiritual successor of the popular C++ binding generator ``pybind11``).
+
+
+Rationale
+=========
+
+This PEP proposes a different model: instead of the superclass data being
+part of the subclass data, the extra space a subclass needs is specified
+and accessed separately.
```
gvanrossum, line 118: I don't understand what this is describing. Are you proposing to use a pointer?

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+(How base class data is accessed is left to whomever implements the base class:
+they can for example  provide accessor functions, expose a part of its
+``struct`` for better performance, or do both.)
+
+The proposed mechanism allows using static, read-only ``PyType_Spec``
```
gvanrossum, line 123: IIUC `PyType_Spec` is specific to `PyHeapType`. So this makes me wonder -- perhaps the PEP is just about making `PyHeapType` more usable? Especially from C++? And you should just say so in the PEP title, rather than claiming more generality?

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+even if the superclass struct is opaque, like ``PyTypeObject`` in
+the Limited API.
+
+Combined with a way to create class from ``PyType_Spec`` and a custom metaclass,
+this will allow libraries like nanobind or JPype to create metaclasses
+without making assumptions about ``PyTypeObject``'s memory layout.
+The approach generalizes to non-metaclass types as well.
```
gvanrossum, line 130: But do we care?

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+In the code blocks below, only function headers are part of the specification.
+Other code (the size/offset calculations) are details of the initial CPython
+implementation, and subject to change.
+
+Relative ``basicsize``
+----------------------
+
+The ``basicsize`` member of ``PyType_Spec`` will be allowed to be zero or
+negative.
+In that case, it will specify the inverse of *extra* storage space instances of
```
gvanrossum, line 145: "inverse" makes me think of 1/x, not -x. How about this?
```suggestion
In that case, its absolute value will specify the amount of *extra* storage space instances of
```

gvanrossum (commented):
This is not a complete review, I realize I am not qualified to understand the specification.

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+Title: C API for Extending Opaque Types
+Author: Petr Viktorin <encukou@gmail.com>
+Status: Draft
+Type: Standards Track
+Content-Type: text/x-rst
+Created: 23-Aug-2022
+Python-Version: 3.12
+
+
+Abstract
```
CAM-Gerlach, line 11: If you prefer, instead of manually specifying the highlighting language of each of your code blocks, since this PEP is exclusively C-related, you could just do this:

```suggestion
Python-Version: 3.12


.. highlight:: c


Abstract
```

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+In order to allow changing/optimizing CPython, and allow freedom for alternate
+implementations of the C API, best practice is to not expose memory layout
+(C structs) in public API, and instead rely on accessor functions.
+(When this hurts performance, direct struct access can be allowed in a
+less stable API tier, at the expense of compatibility with diferent
+versions/implementations of the interpreter.)
+
+However, when a particular type's instance struct is hidden, it becomes
+difficult to subclass it.
+The usual subclassing pattern, explained `in the tutorial <https://docs.python.org/3/extending/newtypes_tutorial.html#subclassing-other-types>`_,
```
CAM-Gerlach, line 35: Minor sidenote: You could consider using an intersphinx reference for these cases rather than a hardcoded URL—though you'll need to add proper ref target labels to the source tutorial headings.

You can also use it to easily and reliably link existing functions, structs, etc., if you think it prudent, instead of hardcoding links or making the reader search for them themselves if they want more details.

CAM-Gerlach (commented):
I didn't review this (yet) in any detail, but it overall looks fairly solid at least from a technical and rendering perspective. I did have a couple sidenotes about things you _could_ take advantage of, if you wanted.

pep-0697.rst
encukou:
```suggestion
The underlying ``PyHeapTypeObject`` struct is both variable-sized and
```
Word *should* be missing. Thanks for the catch!

pep-0697.rst
encukou: Really? It would be “left to her”, not “left to she”.

pep-0697.rst
encukou: Yes. It enables the [example in the tutorial](https://docs.python.org/3.10/extending/newtypes_tutorial.html#subclassing-other-types) to work with stable ABI. (And in case you're following Mark Shannon, I should add that *any* stable ABI will need to solve the issue somehow.)

The tutorial's example – extending `list` with an extra C variable – is particularly impractical/simplified, sure, but it shows that currently, subclassing ties you to the ABI of the superclass.

pep-0697.rst
encukou: Thanks for the clarification!
(English math terminology is confusing. I looked up that [inverse can mean both](https://en.wikipedia.org/wiki/Inverse), and hoped the meaning's clear when dealing with integers and negatives.)

pep-0697.rst
encukou: It's about defining types. There are two ways to do that, ``PyType_Spec`` and static types. 
Maybe I should be clearer that it doesn't apply to static types?

It's not specific to C++.

pep-0697.rst
encukou: I'll rearrange this to make it clearer.

pep-0697.rst
encukou: A high-level overview. Pointer-level details are in Specification.
(I'll clarify this.)

pep-0697.rst
encukou: It's querying the Python attribute ``__basicsize__`` and plugging it into ``PyType_Spec`` used to create a subclass. I'll move the English description here.

I'm using nanobind (C++) for examples & initial design discussions, because it's small and free of historical baggage. But Java or Rust bindings have similar use cases, and I'll consider the PEP a failure if JPype & Pyo3 don't find it useful.

(My own motivation is improving subclassing API in general, but that's not what people behind binding generators are asking for...)

encukou:
@gvanrossum Thank you very much for your questions!
I'm obviously too deep in the woods, and your comments should be very helpful in marking the way here.

Would you mind publishing the PEP now, so I can share the technical details with the people who are affected (or even have been asking for such a feature), while I work on the introductory bits?

pep-0697.rst
CAM-Gerlach: I do as well, FWIW—just wanted to offer that as an option.

pep-0697.rst
CAM-Gerlach: BTW, see https://github.com/python/peps/pull/2773#issuecomment-1229104900

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+
+But in limited API/stable ABI, we do not expose the size of ``PyListObject``,
+so that it can vary between CPython versions (and even between possible
+alternate ABI-compatible C API implementations).
+
+With the size not available, limited API users must resort to workarounds such
+as querying ``__basicsize__`` and plugging it into ``PyType_Spec`` at runtime,
+and divining the correct offset for their extra data.
+This requires making assumptions about memory layout, which the limited API
+is supposed to hide.
```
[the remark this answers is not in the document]
gvanrossum: Ah, I realize you buried the lede here. The key motivation is "to attach C-level state to custom types". You might explain this use case more clearly up front.

pep-0697.rst
```diff
@@ -0,0 +1,357 @@
...
+
+Another scenario where the traditional way to extend an object does not work
+is variable-sized objects, i.e. ones with non-zero ``tp_itemsize``.
+If the instance struct ends with a variable-length array (such as
+in ``tuple`` or ``int``), subclasses cannot add their own extra data without
+detailed knowledge about how the superclass allocates and uses its memory.
+
+Some types, such as CPython's ``PyHeapType``, handle this by storing
+variable-sized data after the fixed-size struct.
+This means that any subclass can add its own fixed-size data.
```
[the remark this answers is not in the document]
gvanrossum: Okay, more detail would be helpful.

pep-0697.rst
gvanrossum: Just don't expect your readers to be familiar with nanobind. It's fine to explain that nanobind provides a use case, but you still need to explain what problem it is solving and how, and how its solution is sub-optimial and how your PEP provides a better solution. I'm guessing from what you explained in comments above that when a subclass of a variable-sized type (like tuple) needs to add extra state it must define a struct to hold the extra state and write a custom function that calculates the offset of that extra state relative to the nominal (`PyObject *`) start of the object, and it also needs to override the new-instance allocation (`tp_new`? `tp_alloc`?) to allocate space for the extra state.

pep-0697.rst
gvanrossum: OK, it looks like maybe it's so high level as to be meaningless (at least to someone who basically only learns from examples like myself).

pep-0697.rst
gvanrossum: But it's "left to (_she_ who implements the base class)". See https://www.thoughtco.com/whoever-and-whomever-1689622#toc-how-to-use-whomever, the example "The prize should be given to **whoever** wins the race".

pep-0697.rst
gvanrossum: I figured it doesn't apply to static types (3rd party static types are outmoded anyway :-). I understand it's not unique to C++ but your only example uses C++ and for a non-C++-user that example was more baffling than helpful. The excess generality that I'm objecting most to is mentioned below, "The approach generalizes to non-metaclass types as well."

pep-0697.rst
gvanrossum: Hm. I see no particular problem to extending list, since the list instance itself is not variable-size -- it is a plain struct whose `sizeof` doesn't lie, and the example even works with a static type. (Perhaps that's why list was chosen in the tutorial? :-) Admittedly it doesn't work with the stable ABI (because the typedef `PyListObject` isn't in the ABI) but that's not the same problem you explained to me in the comments above (or I still misunderstood your explanation). The problem with defining types with extra state seems to be caused by the basic type object being variable-sized, not because it's not in the ABI. Or have I got that wrong? The equivalent at the non-metaclass level would seem to me to define a subclass of `tuple` that adds extra state to each tuple instance.

Although now that I read this back I'm not sure whether I've fully understood the problem that you are solving.

pep-0697.rst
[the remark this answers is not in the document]
kumaraditya303: Please use the correct name `PyHeapTypeObject` here to avoid confusion.

pep-0697.rst
kumaraditya303: The C equivalent snippet of the code would be helpful. The code is small and easy is to write in C.

pep-0697.rst
kumaraditya303: Additive inverse is the correct terminology here but absolute value works too :)

pep-0697.rst
encukou: So, I'm trying to link to [Python 3.11's `PyListObject`](https://docs.python.org/3.11/c-api/list.html?highlight=pylistobject#c.PyListObject). I added a `python3.11` mapping and tried:
```
:c:type:`python3.11:PyListObject`
```
but I'm getting:
```
WARNING: Unparseable C cross-reference: 'python3.11:PyListObject'
Invalid C declaration: Expected identifier in nested name. [error at 8]
  python3.11:PyListObject
  --------^
```
So I read Intersphinx docs, and got this to work:
```
:external+python3.11:c:type:`PyListObject`
```
Is that the way? If so, it doesn't sound like something to burden PEP authors with, tbh

gvanrossum:
I'm happy for you to publish the PEP, major clarifications often happen after the first draft is published.

I'm also with you on the intersphinx stuff being too complex, I just use a URL that points to the current dev-level docs (or occasionally a specific version if that makes sense). I expect we'll be getting pushback from the sphinx nerds among the PEP editors team though. :-)

AA-Turner:
> I'm also with you on the intersphinx stuff being too complex, I just use a URL that points to the current dev-level docs (or occasionally a specific version if that makes sense). I expect we'll be getting pushback from the sphinx nerds among the PEP editors team though. :-)

I think I qualify on the second point! I'd back using plain links in PEPs though for simplicity.

A

encukou:
OK, merging to reserve the number, but I'll rewrite the explanations before opening the discussion.
Thanks for the reviews!

PR #2774, opened 2022-08-26
PEP 668: Change to Standards Track

pradyunsg:
This PEP describes a new feature for Python Packaging interoperability,
which is better classified under this type.


CAM-Gerlach (approved):
If it has a PEP delegate and an Accepted state, as it does, then yes it should be Standards Track (or Process, which this is not), per PEP 1 and the PEP process. Informational PEPs are essentially just whitepapers rather than standards, and per PEP 1,

> do not necessarily represent a Python community consensus or recommendation

PR #2777, opened 2022-08-28
PEP 696: Improve section "Using another TypeVarLike as the default"

Gobot1234:
Follows further discussion on typing-sig

cc @erictraut

### DIFF
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -199,14 +199,19 @@ for the ``TypeVarLike`` (``T2``) the following rules apply.
 Scoping Rules
 ~~~~~~~~~~~~~
 
-``T1`` must be used before ``T2`` in the parameter list of the generic.
+``T1`` must be used before ``T2`` in the parameter list of the generic
+or bound in an outer class or function scope.
 
 .. code:: py
 
    DefaultT = TypeVar("DefaultT", default=T)
 
-   class Foo(Generic[T, DefaultT]): ...  # Valid
-   def bar(x: T, y: DefaultT): ...       # Valid
+   class Foo(Generic[T, DefaultT]): ...   # Valid
+   def bar(x: T, y: DefaultT): ...        # Valid
+   class Foo(Generic[T]):
+       class Bar(Generic[DefaultT]): ...  # Valid
+   def outer(x: T):
+       def inner(y: DefaultT): ...        # Valid
 
    StartT = TypeVar("StartT", default="StopT")  # Swapped defaults around from previous example
    StopT = TypeVar("StopT", default=int)
@@ -214,6 +219,7 @@ Scoping Rules
                      # ^^^^^^ Invalid: ordering does not allow StopT to be bound
    def baz(x: DefaultT, y: T): ...
             # ^^^^^^^^ Invalid: ordering does not allow DefaultT to be bound
+
 
 Bound Rules
 ~~~~~~~~~~~
@@ -248,7 +254,8 @@ constraints of ``T2`` must be a superset of the constraints of ``T1``.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 ``TypeVarLike``\ s are valid as parameters to generics inside of a
-``default`` when the first parameter is in scope.
+``default`` when the first parameter is in scope as determined by the
+previous section.
 
 .. code:: py
 
@@ -257,7 +264,6 @@ constraints of ``T2`` must be a superset of the constraints of ``T1``.
 
    def foo(x: ListDefaultT): ...
    def bar(y: T, x: ListDefaultT): ...
-
 
 
 ``Generic`` ``TypeAlias``\ es
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -193,14 +193,14 @@ default to the type of ``start`` and step default to ``int | None``.
    reveal_type(slice[str]())                   # type is slice[str, str, int | None]
    reveal_type(slice[str, bool, timedelta]())  # type is slice[str, bool, timedelta]
 
-When using another ``TypeVarLike`` (``T1``) as the default, the default
-for the ``TypeVarLike`` (``T2``) the following rules apply.
+When using a ``TypeVarLike`` as the default to another ``TypeVarLike``.
+Where ``T1`` is the default for ``T2`` the following rules apply.
 
 Scoping Rules
 ~~~~~~~~~~~~~
 
 ``T1`` must be used before ``T2`` in the parameter list of the generic
-or bound in an outer class or function scope.
+or be bound in an outer class or function scope.
 
 .. code:: py
 
@@ -219,7 +219,6 @@ or bound in an outer class or function scope.
                      # ^^^^^^ Invalid: ordering does not allow StopT to be bound
    def baz(x: DefaultT, y: T): ...
             # ^^^^^^^^ Invalid: ordering does not allow DefaultT to be bound
-
 
 Bound Rules
 ~~~~~~~~~~~
@@ -242,7 +241,7 @@ constraints of ``T2`` must be a superset of the constraints of ``T1``.
 .. code:: py
 
    T1 = TypeVar("T1", bound=int)
-   TypeVar("Ok", int, str default=T1)                 # Valid
+   TypeVar("Ok", int, str, default=T1)                # Valid
    TypeVar("Invalid", float, str, default=T1)         # Invalid: upper bound int is incompatible with constraints float or str
 
    T1 = TypeVar("T1", int, str)
@@ -262,8 +261,18 @@ previous section.
    T = TypeVar("T")
    ListDefaultT = TypeVar("ListDefaultT", default=list[T])
 
-   def foo(x: ListDefaultT): ...
-   def bar(y: T, x: ListDefaultT): ...
+   class Bar(Generic[T, ListDefaultT]):
+      def __init__(self, x: T, y: ListDefaultT): ...
+
+   reveal_type(Bar[int])             # type is Bar[int, list[int]]
+   reveal_type(Bar[int, list[str]])  # type is Bar[int, list[str]]
+   reveal_type(Bar[int, str])        # type is Bar[int, str]
+
+Specialisation Rules
+~~~~~~~~~~~~~~~~~~~~
+
+``TypeVarLike``\ s currently cannot be further subscripted. This might
+change if Higher Kinded TypeVars are implemented.
 
 
 ``Generic`` ``TypeAlias``\ es
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -15,7 +15,7 @@ Abstract
 
 This PEP introduces the concept of type defaults for
 ``TypeVarLike``\ s (``TypeVar``, ``ParamSpec`` and ``TypeVarTuple``),
-which act as defaults for a type parameter when none is specified or
+which act as defaults for a type parameter when one is not specified or
 the constraint solver isn't able to solve a type parameter to anything.
 
 Default type argument support is available in some popular languages
@@ -28,7 +28,7 @@ and can be found in its
 Motivation
 ----------
 
-.. code:: py
+.. code-block:: py
 
    T = TypeVar("T", default=int)  # This means that if no type is specified T = int
 
@@ -43,7 +43,7 @@ One place this `regularly comes
 up <https://github.com/python/typing/issues/975>`__ is ``Generator``. I
 propose changing the *stub definition* to something like:
 
-.. code:: py
+.. code-block:: py
 
    YieldT = TypeVar("YieldT")
    SendT = TypeVar("SendT", default=None)
@@ -55,7 +55,7 @@ propose changing the *stub definition* to something like:
 
 This is also useful for a ``Generic`` that is commonly over one type.
 
-.. code:: py
+.. code-block:: py
 
    class Bot: ...
 
@@ -74,12 +74,12 @@ also helps non-typing users who rely on auto-complete to speed up their
 development.
 
 This design pattern is common in projects like:
- - `discord.py <https://github.com/Rapptz/discord.py>`__ - where the
+ - `discord.py <https://github.com/Rapptz/discord.py>`__ — where the
    example above was taken from.
- - `NumPy <https://github.com/numpy/numpy>`__ - the default for types
+ - `NumPy <https://github.com/numpy/numpy>`__ — the default for types
    like ``ndarray``'s ``dtype`` would be ``float64``. Currently it's
    ``Unknown`` or ``Any``.
- - `TensorFlow <https://github.com/tensorflow/tensorflow>`__ - this
+ - `TensorFlow <https://github.com/tensorflow/tensorflow>`__ — this
    could be used for Tensor similarly to ``numpy.ndarray`` and would be
    useful to simplify the definition of ``Layer``.
 
@@ -96,7 +96,7 @@ a ``default`` value. Doing so should ideally raise a ``TypeError`` in
 ``typing._GenericAlias``/``types.GenericAlias``, and a type checker
 should flag this an error.
 
-.. code:: py
+.. code-block:: py
 
    DefaultStrT = TypeVar("DefaultStrT", default=str)
    DefaultIntT = TypeVar("DefaultIntT", default=int)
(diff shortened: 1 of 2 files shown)


erictraut (commented):
Overall, I'm fine with this round of changes. I added a few questions/suggestions.

pep-0696.rst
```diff
@@ -184,16 +187,84 @@ default to the type of ``start`` and step default to ``int | None``.
...
+Constraint Rules
+~~~~~~~~~~~~~~~~
+
+``T2``'s constraint must either contain the bound of ``T1`` or the
+constraints of ``T2`` must be a superset of the constraints of ``T1``.
+
+.. code:: py
+
+   T1 = TypeVar("T1", bound=int)
+   TypeVar("Ok", int, str default=T1)                 # Valid
```
JelleZijlstra, line 245:
```suggestion
   TypeVar("Ok", int, str, default=T1)                 # Valid
```

pep-0696.rst
```diff
@@ -184,16 +187,84 @@ default to the type of ``start`` and step default to ``int | None``.
...
+``TypeVarLike``\ s are valid as parameters to generics inside of a
+``default`` when the first parameter is in scope as determined by the
+previous section.
+
+.. code:: py
+
+   T = TypeVar("T")
+   ListDefaultT = TypeVar("ListDefaultT", default=list[T])
+
+   def foo(x: ListDefaultT): ...
```
JelleZijlstra, line 265: I'm having a hard time understanding what this would mean.

Gobot1234: You're right this doesn't make sense and should be an error, whoops

pep-0696.rst
```diff
@@ -78,15 +79,15 @@ This design pattern is common in projects like:
  - `NumPy <https://github.com/numpy/numpy>`__ - the default for types
    like ``ndarray``'s ``dtype`` would be ``float64``. Currently it's
    ``Unknown`` or ``Any``.
- - `TensorFlow <https://github.com/tensorflow/tensorflow>`__ (this
+ - `TensorFlow <https://github.com/tensorflow/tensorflow>`__ - this
```
CAM-Gerlach, line 82:
```suggestion
 - `TensorFlow <https://github.com/tensorflow/tensorflow>`__ — this
```

Typographical nit, but this should be an em dash (you can also do `---` and Sphinx will replace it, if you don't remember the ALT code)

pep-0696.rst
```diff
@@ -184,16 +187,93 @@ default to the type of ``start`` and step default to ``int | None``.
...
+                     # ^^^^^^ Invalid: ordering does not allow StopT to be bound
+   def baz(x: DefaultT, y: T): ...
+            # ^^^^^^^^ Invalid: ordering does not allow DefaultT to be bound
+
+Bound Rules
+~~~~~~~~~~~
+
+``T2``'s bound must be a subtype of ``T1``'s bound.
+
+.. code:: py
```
CAM-Gerlach, line 228: FYI, ``code-block`` is the modern Sphinx directive and should be preferred; ``code`` is the legacy docutils version with different behavior in some areas.

pep-0696.rst
```diff
@@ -184,16 +187,93 @@ default to the type of ``start`` and step default to ``int | None``.
...
+   TypeVar("Ok", int, str, default=T1)                # Valid
+   TypeVar("Invalid", float, str, default=T1)         # Invalid: upper bound int is incompatible with constraints float or str
+
+   T1 = TypeVar("T1", int, str)
+   TypeVar("AlsoOk", int, str, bool, default=T1)      # Valid
+   TypeVar("AlsoInvalid", bool, complex, default=T1)  # Invalid: {bool, complex} is not a superset of {int, str}
+
+
+``TypeVarLike``\ s As Parameters To Generics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
CAM-Gerlach, line 253: Another typographical nit, but since this was specifically changed to use title case for h3+ headings (most PEPs I've seen tend to use sentence case for those headings, but up to you—just be consistent), you don't capitalize articles, short conjunctions, etc. in title case. Also, FYI, for this an the rest, you can just escape the `s` directly rather than the space before the `s`.

```suggestion
``TypeVarLike``\s as Parameters to Generics
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

pep-0696.rst
```diff
@@ -345,7 +432,7 @@ with non-defaults can be checked at compile time.
 Rejected Alternatives
 ---------------------
 
-Allowing the ``TypeVarLike``\ s defaults to be passed to ``type.__new__``'s ``**kwargs``
+Allowing The ``TypeVarLike``\ s Defaults To Be Passed To ``type.__new__``'s ``**kwargs``
 ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
```
CAM-Gerlach, line 436: Whether the first "to" is capitalized depends on whether you follow AP that says yes for to in infinitives, or most of the other style guides (APA, MLA, Chicago, etc) that say no. I generally follow AP by default, but it is somewhat of an outlier here so its really up to you

```suggestion
Allowing the ``TypeVarLike``\s Defaults to Be Passed to ``type.__new__``'s ``**kwargs``
'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
```

or

```suggestion
Allowing the ``TypeVarLike``\s Defaults To Be Passed to ``type.__new__``'s ``**kwargs``
'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
```

pep-0696.rst
```diff
@@ -184,16 +187,93 @@ default to the type of ``start`` and step default to ``int | None``.
...
+   TypeVar("AlsoOk", int, str, bool, default=T1)      # Valid
+   TypeVar("AlsoInvalid", bool, complex, default=T1)  # Invalid: {bool, complex} is not a superset of {int, str}
+
+
+``TypeVarLike``\ s As Parameters To Generics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``TypeVarLike``\ s are valid as parameters to generics inside of a
+``default`` when the first parameter is in scope as determined by the
+previous section.
```
CAM-Gerlach, line 257: Can you `:ref:` the "previous section" you're talking about here, for clarity?

pep-0696.rst
```diff
@@ -373,7 +460,7 @@ Ideally, if :pep:`637` wasn't rejected, the following would be acceptable:
    class Box(Generic[T = int]):
        value: T | None = None
 
-Allowing non-defaults to follow defaults
+Allowing Non-defaults To Follow Defaults
```
CAM-Gerlach, line 463: AP is as written, otherwise

```suggestion
Allowing Non-Defaults to Follow Defaults
```

pep-0696.rst
```diff
@@ -184,16 +187,93 @@ default to the type of ``start`` and step default to ``int | None``.
...
+
+   T = TypeVar("T", bound=float)
+   TypeVar("Ok", default=T, bound=int)        # Valid
+   TypeVar("AlsoOk", default=T, bound=float)  # Valid
+   TypeVar("Invalid", default=T, bound=str)   # Invalid: str is not a subtype of float
+
+Constraint Rules
+~~~~~~~~~~~~~~~~
+
+``T2``'s constraint must either contain the bound of ``T1`` or the
```
CAM-Gerlach, line 238:
```suggestion
Either ``T2``'s constraint must contain the bound of ``T1``, or the
```

A comma here helps disambiguate the intended grouping (assuming I've parsed the intent correctly as "(T2's constraint must contain the bound of T2) or (the constrains of T2 must be a superset of the constraints of T2)"

pep-0696.rst
```diff
@@ -184,16 +187,93 @@ default to the type of ``start`` and step default to ``int | None``.
...
+   reveal_type(slice[str]())                   # type is slice[str, str, int | None]
+   reveal_type(slice[str, bool, timedelta]())  # type is slice[str, bool, timedelta]
+
+When using a ``TypeVarLike`` as the default to another ``TypeVarLike``.
+Where ``T1`` is the default for ``T2`` the following rules apply.
+
+Scoping Rules
+~~~~~~~~~~~~~
+
+``T1`` must be used before ``T2`` in the parameter list of the generic
```
CAM-Gerlach, line 202:
```suggestion
``T1`` must be used before ``T2`` in the parameter list of the generic,
```

Help make the intended grouping more clear, given the double or

pep-0696.rst
```diff
@@ -160,19 +161,21 @@ literal "``...``".
...
 ''''''''''''''''''''''''''''''''''''''''''''
 
-To use another ``TypeVarLike``\ s as the default they have to be of the
-same type. When using another ``TypeVarLike`` (T1) as the default, the default
-for the ``TypeVarLike`` (T2), T1 must be used before in the signature
-of the class it appears in before T2. T2's bound must be a subtype of
-T1's bound.
+This allows for a value to be used again when the constraints solver
+fails to solve a constraint for a type or the a type parameter to a
+generic is missed out but another is specified.
```
CAM-Gerlach, line 174:
```suggestion
fails to solve a constraint for a type, or the type parameter to a
generic is missing but another is specified.
```

There appears to be a typo here, and a missing comma. Also, I couldn't figure out precisely what is meant by "another" here—another _what_ is specified?

pep-0696.rst
```diff
@@ -15,7 +15,8 @@ Abstract
 
 This PEP introduces the concept of type defaults for
 ``TypeVarLike``\ s (``TypeVar``, ``ParamSpec`` and ``TypeVarTuple``),
-which act as defaults for a type parameter when none is specified.
+which act as defaults for a type parameter when none is specified or
```
CAM-Gerlach, line 18:
```suggestion
which act as defaults for a type parameter when one is not specified or
```

Avoid potential confusion with `None`

pep-0696.rst
```diff
@@ -184,16 +187,93 @@ default to the type of ``start`` and step default to ``int | None``.
...
+
+   reveal_type(Bar[int])             # type is Bar[int, list[int]]
+   reveal_type(Bar[int, list[str]])  # type is Bar[int, list[str]]
+   reveal_type(Bar[int, str])        # type is Bar[int, str]
+
+Specialisation Rules
+~~~~~~~~~~~~~~~~~~~~
+
+``TypeVarLike``\ s currently cannot be further subscripted. This might
+change if Higher Kinded TypeVars are implemented.
```
CAM-Gerlach, line 275: Assuming "Higher Kinded TypeVars" is indeed a proper name of a concept or proposal, as implied by the case, could you link what that is here?

CAM-Gerlach (commented):
Had a handful of textual and clarity suggestions and coments

pep-0696.rst
Gobot1234: Good to know

pep-0696.rst
Gobot1234: Thanks, that reads better, I missed out type parameter so this now reads
> fails to solve a constraint for a type, or the type parameter to a
> generic is missing but a different type parameter is specified.

which I think is clearer

pep-0696.rst
```diff
@@ -160,19 +161,21 @@ literal "``...``".
...
    DefaultTs = TypeVarTuple("DefaultTs", default=Unpack[tuple[str, int]])
 
-   class Foo(Generic[DefaultTs]): ...
+   class Foo(Generic[*DefaultTs]): ...
 
    reveal_type(Foo())             # type is Foo[str, int]
    reveal_type(Foo[int, bool]())  # type is Foo[int, bool]
 
-Using another ``TypeVarLike`` as the default
+Using Another ``TypeVarLike`` as ``default``
```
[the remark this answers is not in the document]
Gobot1234: Ok so coming back to this, I'm pretty confident that this is possible to implement (this sure isn't perfect and I had to disable some mypy sanity checks for things like default=T where T didn't use to be valid being passed) https://github.com/python/mypy/compare/master...Gobot1234:mypy:TypeVar-defaults.

```py
from typing import Callable, Generic, TypeVar
from typing_extensions import ParamSpec, TypeAlias, TypeVarTuple, Unpack, reveal_type

T = TypeVar("T", default=int)
T2 = TypeVar("T2", default=Callable[[T], int])


class Foo(Generic[T, T2]): ...


reveal_type(Foo)  # revealed type is type[__main__.Foo[T = builtins.int, T2 = (T = builtins.int) -> builtins.int]]
reveal_type(Foo[str])  # revealed type is type[__main__.Foo[builtins.str, T2 = (builtins.str) -> builtins.int]]
reveal_type(Foo[str, int])  # revealed type is type[__main__.Foo[builtins.str, int]]

PreSpecialised: TypeAlias = Foo[str]
reveal_type(PreSpecialised)    # revealed type is type[__main__.Foo[builtins.str, T2 = (builtins.str) -> builtins.int]]
reveal_type(PreSpecialised[int])  # This currently doesn't work but I don't think it will be hard to fix


P = ParamSpec("P", default=[int, str])
P2 = ParamSpec("P2", default=P)

class Bar(Generic[P, P2]): ...
def foo(fn: Callable[P, int]) -> bool: ...
reveal_type(foo)  # revealed type is def [P = [builtins.int, builtins.str]] (fn: (*P.args, **P.kwargs) -> builtins.int) -> builtins.bool
reveal_type(Bar)  # revealed type is type[__main__.Bar[P = [builtins.int, builtins.str], P2 = P = [builtins.int, builtins.str]]]
```
(I've slightly altered the representations mypy gives to make things a tiny bit clearer)
FYI this doesn't support TypeVarTuple defaults just cause mypy doesn't support PEP 646 yet, but enabling them should be about 20 to 30 lines of code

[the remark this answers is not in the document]
Gobot1234: I've just found out mypy does sort of support TypeVarTuples so I'll add support for that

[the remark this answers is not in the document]
Gobot1234: Ok this is actually a lot harder to support than I initially thought, I think I'm best waiting before trying to add this to mypy, there are a lot of gaps in the TypeVarTuple support. I can PR this feature in further down the road.

pep-0696.rst
```diff
@@ -156,44 +157,126 @@ literal "``...``".
...
+Constraint Rules
+~~~~~~~~~~~~~~~~
+
+Either ``T2``'s constraint must contain the bound of ``T1``, or the
+constraints of ``T2`` must be a superset of the constraints of ``T1``.
+
+.. code-block:: py
+
+   T1 = TypeVar("T1", bound=int)
+   TypeVar("Ok", int, str, default=T1)                # Valid
```
JelleZijlstra, line 246: This seems sketchy because `T1` may be solved to a subtype of `int`, but `Ok` can only be `int` or `str` exactly.

pep-0696.rst
```diff
@@ -156,44 +157,126 @@ literal "``...``".
...
+``default`` when the first parameter is in scope as determined by the
+:ref:`previous section <scoping-rules>`.
+
+.. code-block:: py
+
+   T = TypeVar("T")
+   ListDefaultT = TypeVar("ListDefaultT", default=list[T])
+
+   class Bar(Generic[T, ListDefaultT]):
+      def __init__(self, x: T, y: ListDefaultT): ...
```
JelleZijlstra, line 267:
```suggestion
       def __init__(self, x: T, y: ListDefaultT): ...
```

pep-0696.rst
```diff
@@ -220,17 +303,17 @@ Subclassing
...
 
    class Bar(SubclassMe[int, DefaultStrT]): ...
    reveal_type(Bar())        # type is Bar[str]
    reveal_type(Bar[bool]())  # type is Bar[bool]
 
-   class Foo(SubclassMe[int]): ...
+   class Foo(SubclassMe[float]): ...
 
-   reveal_type(Foo())  # type is <subclass of SubclassMe[int, int]>
+   reveal_type(Foo())  # type is <subclass of SubclassMe[int, float]>
```
JelleZijlstra, line 316: Shouldn't this be `SubclassMe[float, str]`? Also, the reveal_type() output would just be "type is Foo".

Gobot1234: Yeah you're correct on both points. The `<subclass of SubclassMe[int, str]>` is just to show `DefaultStrT` is constrained to `str` in `SubclassMe`. To avoid this I could actually use `DefaultStrT` as an attribute but I think this might be a little harder to follow. Do you have any other ideas as to how else this could be shown?

JelleZijlstra (approved):

PR #2781, opened 2022-09-02
PEP 249: Replace StandardError with Exception

### LINKED ISSUE #2776: PEP 249 refers to StandardError (gone in Python 3)
PEP 249 / DB API 2.0 refers to the StandardError Python exception as the starting point for its exception hierarchy, but this exception is gone in Python 3.

malemburg:
To avoid confusion when using the DB-API 2.0 in the context of Python 3.

Add a note about a future upgrade to the Warning base class.

Fixes #2776.

### DIFF
--- a/pep-0249.txt
+++ b/pep-0249.txt
@@ -1295,17 +1295,16 @@ Footnotes
     ``WHERE`` clause, or clearly document a different interpretation
     of the ``.rowcount`` attribute.
 
-.. [10] In Python 2 and earlier versions of this PEP,
-    ``exceptions.StandardError`` was used as the base class for all
-    DB-API exceptions. Since ``StandardError`` was removed in Python 3,
-    database modules targeting Python 3 should use
-    ``exception.Exception`` as base class instead. Both
-    ``StandardError`` and ``Exception`` are available as builtin
-    objects, so it is not necessary to import the ``exceptions`` module
-    for this. The PEP was updated to use ``Exception`` throughout the
-    text, to avoid confusion. The change should not affect existing
-    modules or uses of those modules, since all DB-API error exception
-    classes are still rooted at the ``Error`` or ``Warning`` classes.
+.. [10] In Python 2 and earlier versions of this PEP, ``StandardError``
+    was used as the base class for all DB-API exceptions. Since
+    ``StandardError`` was removed in Python 3, database modules
+    targeting Python 3 should use ``Exception`` as base class instead.
+    Both ``StandardError`` and ``Exception`` are available as builtin
+    exception objects. The PEP was updated to use ``Exception``
+    throughout the text, to avoid confusion. The change should not
+    affect existing modules or uses of those modules, since all DB-API
+    error exception classes are still rooted at the ``Error`` or
+    ``Warning`` classes.
 
 .. [11] In a future revision of the DB-API, the base class for
     ``Warning`` will likely change to the builtin ``Warning`` class. At
--- a/pep-0249.txt
+++ b/pep-0249.txt
@@ -118,7 +118,7 @@ exceptions or subclasses thereof:
 `Warning`_
     Exception raised for important warnings like data truncations
     while inserting, etc. It must be a subclass of the Python
-    ``StandardError`` (defined in the module exceptions).
+    ``Exception`` [10]_ [11]_.
 
 
 .. _Error:
@@ -128,7 +128,7 @@ exceptions or subclasses thereof:
     exceptions. You can use this to catch all errors with one single
     ``except`` statement. Warnings are not considered errors and thus
     should not use this class as base. It must be a subclass of the
-    Python ``StandardError`` (defined in the module exceptions).
+    Python ``Exception`` [10]_.
 
 
 .. _InterfaceError:
@@ -201,8 +201,8 @@ exceptions or subclasses thereof:
 
 This is the exception inheritance layout::
 
-    StandardError
-    |__Warning
+    Exception [10]_
+    |__Warning [11]_
     |__Error
        |__InterfaceError
        |__DatabaseError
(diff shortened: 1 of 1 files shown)


vstinner:
cc @erlend-aasland

vstinner:
cc @methane

Rosuav:
LGTM. Not sure the Py2 historical note is going to matter much but it's good to have it in case anyone's wondering as they make upgrades.

malemburg:
Posted RFC to DB-SIG: https://mail.python.org/pipermail/db-sig/2022-September/006329.html

pep-0249.txt
```diff
@@ -1297,6 +1295,23 @@ Footnotes
     ``WHERE`` clause, or clearly document a different interpretation
     of the ``.rowcount`` attribute.
 
+.. [10] In Python 2 and earlier versions of this PEP,
+    ``exceptions.StandardError`` was used as the base class for all
+    DB-API exceptions. Since ``StandardError`` was removed in Python 3,
+    database modules targeting Python 3 should use
+    ``exception.Exception`` as base class instead. Both
```
anthony-tuininga, line 1302: Python 3 doesn't have the ``exceptions`` module either!

```
import exceptions
ModuleNotFoundError: No module named 'exceptions'
```

malemburg: Good point :-)

pep-0249.txt
```diff
@@ -1297,6 +1295,22 @@ Footnotes
     ``WHERE`` clause, or clearly document a different interpretation
     of the ``.rowcount`` attribute.
 
+.. [10] In Python 2 and earlier versions of this PEP, ``StandardError``
+    was used as the base class for all DB-API exceptions. Since
+    ``StandardError`` was removed in Python 3, database modules
+    targeting Python 3 should use ``Exception`` as base class instead.
+    Both ``StandardError`` and ``Exception`` are available as builtin
```
zware, line 1302: `StandardError` was removed as mentioned in the previous sentence, it isn't a builtin :)

malemburg: I guess it's better to remove that sentence altogether. I mentioned it, since I had removed the import of the exceptions module (for Python 2).

erlend-aasland (approved):

pep-0249.txt
```diff
@@ -1297,6 +1295,20 @@ Footnotes
...
+    targeting Python 3 should use ``Exception`` as base class instead.
+    The PEP was updated to use ``Exception`` throughout the text, to
+    avoid confusion. The change should not affect existing modules or
+    uses of those modules, since all DB-API error exception classes are
+    still rooted at the ``Error`` or ``Warning`` classes.
+
+.. [11] In a future revision of the DB-API, the base class for
+    ``Warning`` will likely change to the builtin ``Warning`` class. At
+    the time of writing of the DB-API 2.0 in 1999, the warning framework
+    in Python had not yet existed.
```
AA-Turner, line 1310: Minor on grammar:
```suggestion
    in Python did not yet exist.
```

pep-0249.txt
```diff
@@ -201,8 +201,8 @@ exceptions or subclasses thereof:
 
 This is the exception inheritance layout::
```
AA-Turner, line 203: This currently renders as:
image

I don't imagine you want that, so I'd suggest changing to "text" (no highlighting) and removing the underscores from the footnotes, as they won't be interpreted as reStructuredText.

```suggestion
This is the exception inheritance layout:

.. code-block:: text
```

pep-0249.txt
```diff
@@ -201,8 +201,8 @@ exceptions or subclasses thereof:
...
 
-    StandardError
-    |__Warning
+    Exception [10]_
+    |__Warning [11]_
```
AA-Turner, line 205: (per above)
```suggestion
    Exception [10]
    |__Warning [11]
```

AA-Turner (approved):

CAM-Gerlach:
Late to the party, was busy with the Artemis 1 launch, but just FYI as of #2702 you can now use intersphinx links to be able to do e.g. `` :exc:`Exception` ``, etc, which will not only format it as an exception, but also link to the relevant exception, if you want.

malemburg:
On 14.09.2022 05:27, C.A.M. Gerlach wrote:
> Late to the party, was busy with the Artemis 1 launch, but just FYI as 
> of #2702 <https://github.com/python/peps/pull/2702> you can now use 
> intersphinx links to be able to do e.g. |:exc:`Exception`|, etc, which 
> will not only format it as an exception, but also link to the relevant 
> exception, if you want.

Thanks for the tip.

Is there a reference somewhere of what formatting can be used for PEPs ?

-- 
Marc-Andre Lemburg
http://www.malemburg.com/

CAM-Gerlach:
Yup, its standard Sphinx reST;  PEP-12 both gives an intro to reST as well as PEP-specific guidance, but per #2337 I've been meaning to update it to focus just on the latter and for the rest, refer to the more complete and up to date [Sphinx reST primer](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) (which itself is just a mildly updated version of the [reST primer in the CPython devguide](https://devguide.python.org/documentation/markup/), not to be confused with the earlier and more limited [docutils reST primer](https://docutils.sourceforge.io/docs/user/rst/quickstart.html)). There's also the docutils [Quick Reference](https://docutils.sourceforge.io/docs/user/rst/quickref.html) and [full reference](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html) to basic reST, and the Sphinx [Directives](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html), [Domains](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html) and [Roles](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html) pages for the constructs Sphinx adds.

PR #2790, opened 2022-09-12
PEP 664: Add 3.11.0rc2 release date and correct date for 3.11.0 final

pablogsal:


hugovk (approved):
FYI for others about the final release date:

> Due to the fact that we needed to delay the last release candidate by a week and because of personal scheduling problems I am delaying the final release to 2022-10-24 (three weeks from the original date).

https://blog.python.org/2022/09/python-3110rc2-is-now-available.html

PR #2797, opened 2022-09-17
PEP 625: Approval

pfmoore:


CAM-Gerlach (approved):
LGTM

jpl-jengelke:
What an unfortunate decision this was to cram source distributions into the box of gzipped tarballs. It's very unfriendly to the [73% of machines worldwide that run versions of Windows OS](https://gs.statcounter.com/os-market-share/desktop/worldwide). It requires the installation of a third-party app on many, if not all, of these devices just to view source.

webknjaz:
@jpl-jengelke how is it limiting? Wheels are zip files and tarballs are readable with CPython's stdlib out of the box. Plus there's `python -Im tarfile`. If you have Python, you can unpack it w/o needing any extra software.

hugovk:
Please continue the discussion at https://discuss.python.org

jpl-jengelke:
[Here is the info and discussion thread on PEP 625...](https://discuss.python.org/t/pep-625-file-name-of-a-source-distribution/)

PR #2800, opened 2022-09-22
PEP 684: Updates for Round 1 of Discussions

ericsnowcurrently:
Note: I've left some TODO (actually "XXX") comments as placeholders in response to some of the feedback.

### DIFF
--- a/pep-0684.rst
+++ b/pep-0684.rst
@@ -238,25 +238,35 @@ The following runtime state will be moved to ``PyInterpreterState``:
 
 * all global objects that are not safely shareable (fully immutable)
 * the GIL
-* mutable, currently protected by the GIL
-* mutable, currently protected by some other per-interpreter lock
-* mutable, may be used independently in different interpreters
-* all other mutable (or effectively mutable) state
-  not otherwise excluded below
+* most mutable data that's currently protected by the GIL
+* mutable data that's currently protected by some other per-interpreter lock
+* mutable data that may be used independently in different interpreters
+  (also applies to extension modules, including those with multi-phase init)
+* all other mutable data not otherwise excluded below
 
-Furthermore, a number of parts of the global state have already been
-moved to the interpreter, such as GC, warnings, and atexit hooks.
+Furthermore, a portion of the full global state has already been
+moved to the interpreter, including GC, warnings, and atexit hooks.
 
-The following state will not be moved:
+The following runtime state will not be moved:
 
 * global objects that are safely shareable, if any
-* immutable, often ``const``
-* treated as immutable
-* related to CPython's ``main()`` execution
-* related to the REPL
-* set during runtime init, then treated as immutable
-* mutable, protected by some global lock
-* mutable, atomic
+* immutable data, often ``const``
+* effectively immutable data (treated as immutable), for example:
+
+  * some state is initialized early and never modified again
+  * hashes for strings (``PyUnicodeObject``) are idempotently calculated
+    when first needed and then cached
+
+* all data that is guaranteed to be modified exclusively in the main thread,
+  including:
+
+  * state used only in CPython's ``main()``
+  * the REPL's state
+  * data modified only during runtime init (effectively immutable afterward)
+
+* mutable data that's protected by some global lock (other than the GIL)
+* global state in atomic variables
+* mutable global state that can be changed (sensibly) to atomic variables
 
 Note that currently the allocators (see ``Objects/obmalloc.c``) are shared
 between all interpreters, protected by the GIL.  They will need to move
--- a/pep-0684.rst
+++ b/pep-0684.rst
@@ -388,7 +388,13 @@ Extension modules have many of the same problems as the runtime when
 state is stored in global variables.  :pep:`630` covers all the details
 of what extensions must do to support isolation, and thus safely run in
 multiple interpreters at once.  This includes dealing with their globals.
-All other extensions are considered incompatible with multiple interpreters.
+
+If an extension implements multi-phase init (see :pep:`489`) it is
+considered compatible with multiple interpreters.  All other extensions
+are considered incompatible.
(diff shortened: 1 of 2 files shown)


pep-0684.rst
```diff
@@ -273,11 +389,24 @@ state is stored in global variables.  :pep:`630` covers all the details
...
 
-Extension modules that do not implement isolation will only run in
-the main interpreter.  In all other interpreters, the import will
-raise ``ImportError``.  This will be done through
+If an extension implements multi-phase init (see :pep:`489`) it is
+considered compatible with multiple interpreters.  All other extensions
+are considered incompatible.
+
+.. XXX Do we need an extra indicator (e.g. moduledef slot) for
+   I-said-I-was-isolated-but-now-I-really-mean-it (per-interpreter GIL)?
```
ericsnowcurrently, line 397: FYI, this is a simple placeholder for a long, still unresolved exchange between Petr and me:

<details>
<summary>expand</summary>

https://mail.python.org/archives/list/python-dev@python.org/message/VH74ERJYZ4VDGQQN52LF5Q56EABACHX3/

```
> > Now we get to the break in compatibility mentioned above.  Some
> > extensions are safe under multiple interpreters, even though they
> > haven't indicated that.  Unfortunately, there is no reliable way for
> > the import system to infer that such an extension is safe, so
> > importing them will still fail.  This case is addressed in
> > `Extension Module Compatibility`_ below.
>
> Extensions that use multi-phase init should already be compatible with
> multiple interpreters. Multi-phase init itself is the flag that
> indicates this.

Correct and ExtensionFileLoader will use that if
PyInterpreterConfig.strict_extensions is set, regardless of whether or
not we need a second extension module indicator for
I-said-I-was-isolated-but-now-I-really-mean-it (per-interpreter GIL).

> But they might not be compatible with *per-interpreter GIL*. I don't
> like how that's conflated with multiple interpreters here.

Hmm, I suppose in my mind they *have* been the same thing. :)

> For example, extension modules can currently support multiple
> interpreters, but rely on the GIL to protect calls to a non-threadsafe
> library, access shared memory, etc. As an example, the PEP 630 “opt-out”
> is not thread-safe.

So, you are saying that some mutli-phase init extensions may still be
relying on the GIL as a lock for some shared state.  In the case of
your "opt-out" example, there is a possible (albeit super unlikely)
race on "loaded".  So such an extension needs to be able to separately
opt in to being used without a GIL between interpreters.  Is all that
correct?

Out of curiosity, do you have any examples of extensions that
implement multi-phase init but need to opt out (like in your example)?
 Is it only the case where the maintainer is in the process of
isolating the module, so the opt-out is temporary?

Aside from the unsafe-flag-to-indicate-not-isolated case, do you know
of any other examples where a module is safe for use between
interpreters but still relies on the shared GIL?  I'm struggling to
imagine such a scenario, but where they don't also opt out of
multi-interpreter support already.

FWIW, my assumption is that, if an extension has been made isolated
enough for use between multiple interpreters, then it is extremely
likely that it is also isolated enough to use without a GIL shared
between the interpreters.

> It seems to me that there should be a separate flag (slot) to indicate
> support for per-interpreter GIL, and the `strict_extensions` bit should
> work with that.

I think I see what you are saying.  My concern is that anything beyond
the default settings is an obstacle for extension maintainers, so
opt-in is especially painful if it ends up being the common case.  Of
course, it may be unavoidable in the end.

While we may end up needing a flag to indicate
yes-I'm-isolated-but-not-that-isolated, the following alternative came
to mind:

* multi-phase init extensions should be expected to be isolated, thus
compatible with use in multiple interpreters (already true)
* multi-phase init extensions should be expected to be fully isolated,
thus compatible with per-interpreter GIL
* there would be a new module def slot that a multi-phase init
extension can use to explicitly opt out
* the ExtensionFileLoader would enforce loading the module only once
in the process (and use a dedicated granular lock to prevent races)

Instead of using its own static "loaded" variable, the module in your
opt-out example would use this new slot.

To me, really-truly-fully-isolated is the sensible long-term default
for multi-phase init.  Most maintainers will implement multi-phase
init, using PEP 630 to get isolated enough.  We can avoid the extra
step for the common case.  So our future selves would be much happier
if we go with an explicit opt-out now. :)  This follows my earlier
assumption that few extensions will be safe in multi-interpreter but
not per-interpreter GIL.

FWIW, I was going to say perhaps we could get away with treating the
vast majority of extensions as already safe in multiple interpreters,
to avoid requiring extensions to implement multi-phase init.  However,
I can already think of a number of relatively common cases where that
isn't true (e.g. static types). :)

Plus, multi-phase init is such a good thing and doesn't require that
much effort (especially if we provide an opt-out slot for
multi-interpreter support).  Per-interpreter GIL would be a pretty
good carrot. :)
```
</details>

encukou:
I sense a misunderstanding here :(
Could you please post in the thread again, and point out what's still unresolved? I don't want to repeat too much of what's in the thread, but in short, I think that:

* Currently, there is no "I-said-I-was-isolated" flag, so it doesn't make much sense to me to talk about adding an "I-said-I-was-isolated-but-now-I-really-mean-it" one. Multi-phase init modules are assumed to support multiple interpreters. Today, it is possible to support multiple interpreters well without making modules *fully* isolated.
* Using the GIL as a general lock for calling non-thread-safe API is currently OK even with really-truly-fully-isolated modules. (My example of non-threadsafe API was perhaps too simple, but while holding the GIL for I/O is a bad idea, it does work as far as locking is concerned. Jim said that on that thread. Substitute any non-threadsafe API – say, [`inet_ntoa`](https://github.com/python/cpython/blob/main/Modules/socketmodule.c#L6522))

ericsnowcurrently:
I was planning on starting a new round of discussion, this time on discuss.python.org.  I'd including a summary of what is still unresolved.

PR #2807, opened 2022-09-30
PEP 684: Updates for Round 2 of Discussions

ericsnowcurrently:
I'll be updating this branch in response to discussion on https://discuss.python.org/t/pep-684-a-per-interpreter-gil/19583.

Significant changes:

* settled on keeping the allocators global but requiring that they *all* be thread-safe
* the state of the existing "small block" will be moved to `PyInterpreterState`
* dropped references to mimalloc
* simplified the C-API changes
* clarified the situation with incompatible extension modules
* proposed that extensions always opt in to per-interpreter GIL support with a new `PyModuleDef` slot (at least until we have enough evidence that multi-phase init is sufficient)
* expanded "How to Teach This"

### DIFF
--- a/pep-0684.rst
+++ b/pep-0684.rst
@@ -224,8 +224,8 @@ following changes, in the order they must happen:
 3. finally, move the GIL down into ``PyInterpreterState``
 4. everything else
 
-   * add to the public C-API
-   * implement restrictions in ``ExtensionFileLoader``
+   * update the C-API
+   * implement extension module restrictions
    * work with popular extension maintainers to help
      with multi-interpreter support
 
--- a/pep-0684.rst
+++ b/pep-0684.rst
@@ -50,11 +50,11 @@ CPython Runtime State
 
 Properly isolating interpreters requires that most of CPython's
 runtime state be stored in the ``PyInterpreterState`` struct.  Currently,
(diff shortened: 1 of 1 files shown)


pep-0684.rst
```diff
@@ -516,24 +524,33 @@ there is no change in behavior under multiple interpreters created
...
+Such modules will have to start wrapping any direct or indirect use
+of that state in a lock.
+
 Now we get to the break in compatibility mentioned above.  Some
 extensions are safe under multiple interpreters, even though they
 haven't indicated that.  Unfortunately, there is no reliable way for
 the import system to infer that such an extension is safe, so
 importing them will still fail.  This case is addressed in
-`Extension Module Compatibility`_ below.
+`Extension Module Compatibility`_ above.
```
CAM-Gerlach, line 553: BTW, being lazy and forgetful, I usually try to avoid having to update these references if I re-arrange sections by omitting a direction, and just stating e.g. "This case is address in :ref:`XXX`" or "the :ref:`XXX` section", since users can just click the link that points directly to the section in question. Also, I think I mentioned it before so I won't repeat that part, but another benefit of `:ref:` is it will automatically insert the section name and update in case you change its name, rather than having to hardcode it and manually update it everywhere.

PR #2810, opened 2022-10-04
Remove ``include_patterns`` monkeypatching

AA-Turner:
This is included in Sphinx core from version 5.1.

A


JelleZijlstra (approved):

CAM-Gerlach:
A bit annoying, since it means that the Sphinx versions supported for building the PEPs and the CPython docs don't intersect at all, so I'll need to have two envs and an extra few hundred MB on disk, but oh well...

PR #2816, opened 2022-10-05
Infra + many PEPs: Add Sphinx Lint and fix errors

hugovk:
Lint PEPs using [Sphinx Lint](https://github.com/sphinx-contrib/sphinx-lint) via pre-commit, already used for the devguide.

Let's run it with two checks disabled:

* `trailing-whitespace`: there's 300+ of these, probably not worth fixing
* `role-with-double-backticks`: temporarily disabled due to a false positive, reported at https://github.com/sphinx-contrib/sphinx-lint/issues/34

This PR also fixes the errors it finds.

Currently Sphinx Lint only runs on .rst files; I also fixed the errors it would find if the old .txt PEPs were renamed to .rst (which may happen as part of https://github.com/python/peps/issues/10).

Whilst we're editing these PEPs:

* Remove old `Version` and `Last-Modified` headers
* Remove old emacs metadata
* Redirect old BPO links to GH issues


AA-Turner (approved):
Looks OK

A

JelleZijlstra (approved):

PR #2818, opened 2022-10-05
PEP 688: Second version

JelleZijlstra:


pep-0688.rst
```diff
@@ -120,57 +189,58 @@ subclassed at runtime, but supports the ``__instancecheck__`` and
...
 For runtime uses, the ``ABC.register`` API can be used to register
 buffer classes with ``typing_extensions.Buffer``.
-
-When ``types.Buffer`` is available, ``typing_extensions`` should simply
-re-export it. Thus, users who register their buffer class manually
-with ``typing_extensions.Buffer.register`` should use a guard to make
-sure their code continues to work once ``types.Buffer`` is in the
-standard library. 
+When ``collections.abc.Buffer`` is available, ``typing_extensions`` should simply
+re-export it. ``MutableBuffer`` will work similarly.
```
AlexWaygood, line 243: Do we need this? It feels a little bit like an implementation detail

```suggestion
```

pep-0688.rst
```diff
@@ -209,41 +279,45 @@ more consistent, so we believe the migration cost is worth it.
...
+An earlier version of this PEP proposed adding a new ``types.Buffer`` type with
+an ``__instancecheck__`` implemented in C so that ``isinstance()`` checks can be
+used to check whether a type implements the buffer protocol. This avoids the
+complexity of exposing the full buffer protocol to Python code, while still
+allowing the type system to check for the buffer protocol.
 
-Nevertheless, the ABC proposal has the advantage that it does not require C changes.
-This PEP proposes to adopt a version of it in the third-party ``typing_extensions``
-package for the benefit of users of older Python versions.
+However, this approach
```
AlexWaygood, line 311:
```suggestion
However, that approach
```

pep-0688.rst
```diff
@@ -209,41 +279,45 @@ more consistent, so we believe the migration cost is worth it.
...
+does not compose well with the rest of the type system, because ``types.Buffer``
+would be a nominal type, not a structural one. For example, there would be no way
+to represent "an object that supports both the buffer protocol and ``__len__``". With
+the current proposal, the ``__buffer__`` class is like any other method, so a
```
AlexWaygood, line 315: This is the first time you've referred to the class as "the ``__buffer__`` class" in this document :)

pep-0688.rst
```diff
@@ -209,41 +279,45 @@ more consistent, so we believe the migration cost is worth it.
...
+``Protocol`` can be defined combining it with any other method.
+
+More generally, no other part of Python works like the proposed ``types.Buffer``.
+The current proposal is more consistent with the rest of the language, where
+C-level slots usually have corresponding Python-level magic methods.
```
AlexWaygood, line 320:
```suggestion
C-level slots usually have corresponding Python-level special methods.
```

pep-0688.rst
```diff
@@ -78,37 +85,99 @@ that this shortcut has caused significant problems for the ``psycopg`` project.
...
+
+To help implementations of ``__buffer__``, we add ``inspect.BufferFlags``,
+a subclass of ``enum.IntFlag``. This enum contains all flags defined in the
+C buffer protocol; for example, ``inspect.BufferFlags.SIMPLE`` has the same
+value as the ``PyBUF_SIMPLE`` constant.
+
+collections.abc.Buffer and MutableBuffer
+----------------------------------------
+
+We add two new abstract base classes, ``collections.abc.Buffer`` and
```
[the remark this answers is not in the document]
JelleZijlstra: Yes, let me add that.

pep-0688.rst
```diff
@@ -120,57 +189,58 @@ subclassed at runtime, but supports the ``__instancecheck__`` and
...
-that do not have ``types.Buffer`` available.
+workaround, two abstract base classes, ``typing_extensions.Buffer``
+and ``MutableBuffer``, will be provided for Python versions
+that do not have ``collections.abc.Buffer`` available.
 
-For the benefit of
-static type checkers, ``typing_extensions.Buffer`` can be used as
-a base class in stubs to mark types as supporting the buffer protocol.
+For the benefit of static type checkers, ``typing_extensions.Buffer``
+can be used as a base class in stubs to mark types as supporting the buffer protocol.
```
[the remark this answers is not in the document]
JelleZijlstra: Yes, because the protocols won't apply before 3.12 where there's no `__buffer__` method.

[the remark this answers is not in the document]
AlexWaygood: Ahh, I see. Could you make this more explicit? Something like `In Python versions where objects supporting the buffer protocol have a __buffer__ method, inheriting from collections.abc.Buffer will not be necessary to indicate that an object supports the buffer protocol. However, in older Python versions, it will be necessary to explicitly inherit from typing_extensions.Buffer to indicate to type checkers that a class supports the buffer protocol, since objects supporting the buffer protocol will not have a __buffer__ method.`

pep-0688.rst
```diff
@@ -190,6 +190,8 @@ They can also be used in ``isinstance`` and ``issubclass`` checks:
    >>> issubclass(str, Buffer)
    False
 
+In the typeshed stub files, these classes should be defined as ``Protoocol``s.
```
AlexWaygood, line 193:
```suggestion
In the typeshed stub files, these classes should be defined as ``Protoocol``s,
following the precedent of other simple ABCs in ``collections.abc`` such as
``collections.abc.Iterable`` or ``collections.abc.Sized``.
```

pep-0688.rst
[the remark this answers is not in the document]
JelleZijlstra: Thanks, pushed a variation of that wording

AlexWaygood (approved):
Looks great!

PR #2822, opened 2022-10-06
Be more tolerant of PEP 523 plugins.

markshannon:
@Yhg1s @pablogsal 
Is this acceptable wording?


pep-0669.rst
```diff
@@ -289,26 +289,18 @@ Backwards Compatibility
...
-Arguably, they already were as the author does not know of any PEP 523 plugin
-that supports ``sys.settrace`` or ``sys.setprofile`` correctly.
-This PEP merely formalizes that.
+There are some compatibility issues with :pep:`523`, as the behavior
+of :pep:`523` plugins is outside of the VM's control.
+It is up to :pep:`523` plugins to ensure that they respect the semantics
+of this PEP. Simple plugins that do not change the state of the VM, and
+defer execution to ``_PyEval_EvalFrameDefault()`` should continue to work.
+
+``sys.settrace`` and ``sys.setprofile`` will act as if they were tools
```
CAM-Gerlach, line 298: Just FYI, you can actually link these with `` :func:`sys.settrace` ``, etc., if you want, just like in the CPython docs, since we added intersphinx support

PR #2844, opened 2022-10-23
PEP 696: Changes from Pyright implementation

Gobot1234:
list vs tuple for default will be changed when https://github.com/python/typing/issues/1274 is resolved


pep-0696.rst
```diff
@@ -345,34 +346,17 @@ subtype of one of the constraints.
...
-- a new class ``_DefaultMixin`` needs to be added which is a superclass
-  of ``TypeVar``, ``ParamSpec``, and ``TypeVarTuple``.
-
-   -  the type passed to default would be available as a ``__default__``
-      attribute and a sentinel would need to be added that can be
-      checked against to see if it was supplied.
+- The classes ``TypeVar``, ``ParamSpec``, and ``TypeVarTuple`` would
+  all expose the type passed to default which would be available as a
+  ``__default__`` attribute with the default being ``None`` if no
+  argument is passed and ``NoneType`` if ``default=None``.
```
JelleZijlstra, line 359: This sentence is rather long, I'd split it up. Also, should quote `default`.

And can you remind me where this `NoneType` conversion comes from?

Gobot1234: The NoneType conversion comes from here, converting None to NoneType is how current typing seems to handle similar cases for more see https://github.com/python/typing_extensions/pull/85.
```py
In [1]: from typing import Optional, get_args

In [2]: get_args(Optional[int])
Out[2]: (int, NoneType)
```

JelleZijlstra: Thanks, makes sense

PR #2847, opened 2022-10-26
PEP 689: Rewrite, with PyUnstable_ prefix

encukou:
I'm getting back to this one!
The `PyUnstable_` prefix is an unwieldy compromise, but, better than nothing.

### DIFF
--- a/pep-0689.rst
+++ b/pep-0689.rst
@@ -2,22 +2,14 @@ PEP: 689
 Title: Unstable C API tier
 Author: Petr Viktorin <encukou@gmail.com>
 Discussions-To: https://mail.python.org/archives/list/python-dev@python.org/thread/PQXSP7E2B6KNXTJ2AERWMKKX42YP5D6O/
-Status: Deferred
+Status: Draft
 Type: Standards Track
 Content-Type: text/x-rst
 Requires: 523
 Created: 22-Apr-2022
 Python-Version: 3.12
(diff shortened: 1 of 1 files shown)


pep-0689.rst
```diff
@@ -28,183 +20,228 @@ meaning that they will not change in patch (bugfix/security) releases,
 but may change between minor releases (e.g. between 3.11 and 3.12) without
 deprecation warnings.
 
+C-API with an leading underscore is designated *internal*, meaning that it
```
JelleZijlstra, line 23:
```suggestion
Any C-API with a leading underscore is designated *internal*, meaning that it
```

pep-0689.rst
```diff
@@ -213,19 +250,43 @@ Reference Implementation
 https://github.com/python/cpython/issues/91744
 
 
+Open Issues
+===========
+
+With the ``PyUnstable_`` ptefix, is the opt-in macro necessary?
```
JelleZijlstra, line 256:
```suggestion
With the ``PyUnstable_`` prefix, is the opt-in macro necessary?
```

pep-0689.rst
CAM-Gerlach:
```suggestion
Any C API with a leading underscore is designated *internal*, meaning that it
```

No dash here, per syntax or convention

PR #2852, opened 2022-10-27
PEP 689: Link to new discussion thread

encukou:


JelleZijlstra (approved):
Confirmed the link is right

PR #2854, opened 2022-10-27
PEP 675: Clarify Motivation query examples are abridged

Eclips4:
Fixes https://github.com/python/peps/issues/2853

### DIFF
--- a/pep-0675.rst
+++ b/pep-0675.rst
@@ -42,7 +42,8 @@ insert it into a predefined SQL query:
     def query_user(conn: Connection, user_id: str) -> User:
         query = f"SELECT * FROM data WHERE user_id = {user_id}"
         conn.execute(query)
-        ...  # some actions to transform data to abstract class User, and return it
+        ...  # Transform data to a User object and return it
+
     query_user(conn, "user123")  # OK.
 
 However, the user-controlled data ``user_id`` is being mixed with the
--- a/pep-0675.rst
+++ b/pep-0675.rst
@@ -141,9 +141,8 @@ from a format string using ``user_id``, and cannot be passed to
 
     def query_user(conn: Connection, user_id: str) -> User:
         query = f"SELECT * FROM data WHERE user_id = {user_id}"
-        conn.execute(query)
+        conn.execute(query)  # Error: Expected LiteralString, got str.
         ...
-        # Error: Expected LiteralString, got str.
 
 The method remains flexible enough to allow our more complicated
 example:
--- a/pep-0675.rst
+++ b/pep-0675.rst
@@ -39,10 +39,10 @@ insert it into a predefined SQL query:
 
 ::
 
-    def query_user(conn: Connection, user_id: str) -> None:
+    def query_user(conn: Connection, user_id: str) -> User:
         query = f"SELECT * FROM data WHERE user_id = {user_id}"
         conn.execute(query)
-
+        ...  # some actions to transform data to abstract class User, and return it
     query_user(conn, "user123")  # OK.
 
 However, the user-controlled data ``user_id`` is being mixed with the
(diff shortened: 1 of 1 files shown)


pep-0675.rst
```diff
@@ -39,7 +39,7 @@ insert it into a predefined SQL query:
 
 ::
 
-    def query_user(conn: Connection, user_id: str) -> User:
+    def query_user(conn: Connection, user_id: str) -> None:
         query = f"SELECT * FROM data WHERE user_id = {user_id}"
         conn.execute(query)
```
CAM-Gerlach, line 44:
```suggestion
    def query_user(conn: Connection, user_id: str) -> User:
        query = f"SELECT * FROM data WHERE user_id = {user_id}"
        conn.execute(query)
        ...
```

I.e. like this

CAM-Gerlach (changes requested):
It seems to me that these function definition bodies are likely deliberately abridged for the sake of brevity in the example, to omit logic (i.e. constructing and returning the User objects) not directly relevant to the motivation of the problem. It doesn't appear to make sense for these functions, whose purpose is to look up a user's data in the db, to not return said data and merely throw it away.

This could be made more clear and explicit, however, by adding an ellipses at the end of the body, per standard (typing and otherwise) convention.

@JelleZijlstra ?

CAM-Gerlach:
Also @pradeep90

pep-0675.rst
```diff
@@ -42,7 +42,7 @@ insert it into a predefined SQL query:
     def query_user(conn: Connection, user_id: str) -> User:
         query = f"SELECT * FROM data WHERE user_id = {user_id}"
         conn.execute(query)
-
+        ...  # some actions to transform data to abstract class User, and return it
     query_user(conn, "user123")  # OK.
```
CAM-Gerlach, line 46:
```suggestion
        ...  # Transform data to a User object and return it

    query_user(conn, "user123")  # OK.
```

IMO, if we want a comment, this can be made more concise, and clearer that the data is being transformed to a User _instance object_ rather than a _class object_. Also, add missing line break.

pep-0675.rst
```diff
@@ -141,6 +141,7 @@ from a format string using ``user_id``, and cannot be passed to
     def query_user(conn: Connection, user_id: str) -> User:
         query = f"SELECT * FROM data WHERE user_id = {user_id}"
         conn.execute(query)
+        ...
         # Error: Expected LiteralString, got str.
```
CAM-Gerlach, line 145:
```suggestion
        conn.execute(query)  # Error: Expected LiteralString, got str.
        ...
```

The typecheck error occurs on this line, not after the `...`, so let's put it there

CAM-Gerlach (commented):
Some further refinements

Eclips4:
Thanks for your advices!

CAM-Gerlach:
Just FYI, it looks like you mistakenly made the changes in your own `main` branch rather than in a feature branch; so far it seems GitHub has managed not to screw things up on this PR yet, but to avoid further trouble on this PR and future ones, I strongly suggest:

**Until this PR is merged (or closed)**: _Don't_ push to, pull from, merge to/from, update, or create new branches from your `main` branch.

**Once it is merged**, _either_:
  * Delete your fork and local clone (and if necessary, re-fork/clone it), _or_
  * Run the following commands to re-sync your `main` branch to `upstream` (assuming you have an `upstream` remote set to this repo):
    
    ```shell
    git switch main
    git fetch --all
    git reset --hard upstream/main
    git push -f
    ```
  
**In the future**: Make sure to always create a new branch (`git switch -c new-feature-branch-name`) before committing any changes (ideally, as soon as you start work, so you don't forget)

Eclips4:
I'm sorry, it's really mistake, 
Should I do something now?

CAM-Gerlach:
> I'm sorry, it's really mistake,

No worries; it happens sometimes and as far as I can tell its not a problem for us right now; its just that it can cause a problem for you in the future so I wanted to let you know about it.

>  Should I do something now?

No, **_you shouldn't do anything now_** other than _not_ doing anything :joy: ...

> Don't push to, pull from, merge to/from, update, or create new branches from your main branch.

...to avoid causing any issues for you with this PR. _**After**_ this PR is merged (or closed), you should do one of the two things I suggested. Thanks!

pradeep90 (approved):
No strong opinions about this. The ellipses seem fine.

CAM-Gerlach:
(FYI @Eclips4 , now you can do one of the two options I mention under "After it is merged/closed" above)

PR #2856, opened 2022-10-29
PEP 649: Fix nested list formatting

PeterJCLaw:
This fixes the formatting of a couple of nested lists so that they render correctly (i.e: as lists rather than paragraphs).


hugovk (approved):
Thank you!

Issue #2862, opened 2022-11-02
PEP 0383: Broken attribution link


textshell:
The link to "Markus Kuhn’s idea that he called UTF-8b" got broken again when gmane shut down.
From looking at the history i think a working replacement link is
https://web.archive.org/web/20090830064219/http://mail.nl.linux.org/linux-utf8/2000-07/msg00040.html

(submitting as issue because CLA requirements for pull requests)

hugovk:
Thanks, please see PR https://github.com/python/peps/pull/2863.

PR #2863, opened 2022-11-02
PEP 383: Fix broken link

hugovk:
Fixes https://github.com/python/peps/issues/2862, and removes some redundant metadata.


gvanrossum (approved):
So much for "perma". :-)

PR #2869, opened 2022-11-06
PEP 616: Mark as Final

gustavgransbo:
This resolves #2868.

### DIFF
--- a/pep-0616.rst
+++ b/pep-0616.rst
@@ -2,7 +2,7 @@ PEP: 616
 Title: String methods to remove prefixes and suffixes
 Author: Dennis Sweeney <sweeney.dennis650@gmail.com>
 Sponsor: Eric V. Smith <eric@trueblade.com>
-Status: Accepted
+Status: Final
 Type: Standards Track
 Content-Type: text/x-rst
 Created: 19-Mar-2020


JelleZijlstra (approved):

AA-Turner (approved):

AA-Turner:
Impressively, merging this managed to create an empty commit -- some sort of race condition in GitHub's logic perhaps? 

The offending commit: https://github.com/python/peps/commit/5e6cef777696de43fdb33cd34557f8fb02d9a8a6

A

CAM-Gerlach:
Never seen that happen before—I've reported it. Seems this is also missing a Resolution header, as many PEPs are.

JelleZijlstra:
@AA-Turner it's probably fine but I'd have waited for a few days to give the PEP owner a chance to say that we should wait. Some PEPs may have rough corners that aren't immediately apparent and that still need to be finalized. No need to revert though, unless the PEP owner actually asks for it.

AA-Turner:
Ahh, I got slightly mixed up and thought Eric was the author of 616, rather than the sponsor -- thank you for the note!

A

AA-Turner:
> Seems this is also missing a Resolution header

https://mail.python.org/archives/list/python-dev@python.org/thread/2VGR3BMACYLARB7SKGX3SDPPXIXZSJE2/

A

PR #2882, opened 2022-11-12
PEP 3141: Fixes typo Number.Complex.imag() #2880

trcmohitmandokhot:
Resubmit typo error fix.


CAM-Gerlach:
It seems the other one still works, so we can close this one I think

PR #2889, opened 2022-11-16
PEP 248: Fix Author and Created date headers

malemburg:
The headers did not include the original author (Greg Stein) and listed a wrong create date.

### DIFF
--- a/pep-0248.txt
+++ b/pep-0248.txt
@@ -1,6 +1,6 @@
 PEP: 248
 Title: Python Database API Specification v1.0
-Author: Greg Stein <gstein@lyra.org>, Marc-André Lemburg (editor) <mal@lemburg.com>
+Author: Greg Stein <gstein@lyra.org>, Marc-André Lemburg as Editor <mal@lemburg.com>
 Discussions-To: db-sig@python.org
 Status: Final
 Type: Informational
--- a/pep-0248.txt
+++ b/pep-0248.txt
@@ -5,7 +5,7 @@ Discussions-To: db-sig@python.org
 Status: Final
 Type: Informational
 Content-Type: text/x-rst
-Created: 8-May-1996
+Created: 08-May-1996
 Post-History:
 Superseded-By: 249
 
--- a/pep-0248.txt
+++ b/pep-0248.txt
@@ -1,6 +1,6 @@
 PEP: 248
 Title: Python Database API Specification v1.0
-Author: Greg Stein <gstein@lyra.org>, Editor: Marc-André Lemburg <mal@lemburg.com>
+Author: Greg Stein <gstein@lyra.org>, Marc-André Lemburg (editor) <mal@lemburg.com>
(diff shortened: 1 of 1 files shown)


pep-0248.txt
```diff
@@ -1,11 +1,11 @@
 PEP: 248
 Title: Python Database API Specification v1.0
-Author: Marc-André Lemburg <mal@lemburg.com>
+Author: Greg Stein <gstein@lyra.org>, Editor: Marc-André Lemburg <mal@lemburg.com>
```
AA-Turner, line 3: "Editor" isn't a currently allowed header, and we've been tightening the validation here -- the simplest solution would be to remove "Editor:" here I'd suggest.

```suggestion
Author: Greg Stein <gstein@lyra.org>, Marc-André Lemburg <mal@lemburg.com>
```

A

malemburg: The problem is that Greg authored the DB-API v1, but I was editor for it after he left. Perhaps I can write "Marc-André Lemburg (editor) <...". Would that work ?

AA-Turner: I believe it would pass the automatic validation, though it might cause an issue in PEP 0 -- I can't remember if the indexing is on email or name.

I'd suggest try it and check PEP 0 to ensure you're not duplicated!

A

malemburg: The linter doesn't like the parens.

malemburg: All good now.

AA-Turner: I checked locally with `python -m pre_commit.languages.pygrep --multiline '(?<=\n)Author:(?:(?!((( +|\n {1,8})[^!#$%&()*+,/:;<=>?@\[\\\]\^_`{|}~]+( <[\w!#$%&''*+\-/=?^_{|}~.]+(@| at )[\w\-.]+\.[A-Za-z0-9]+>)?)(,|(?=\n[^ ])))+
\n(?=[A-Z])))' pep-0249.txt` and it is a fairly restrictive pattern -- even square brackets aren't allowed.

Your latest commit (`as Editor`) duplicates you and sorts oddly: https://pep-previews--2889.org.readthedocs.build/ 

`Marc-André Lemburg, Editor` passes and sorts correctly, though still duplicates you.

A

malemburg:
Hmm, looking at the preview, the index now lists "Stein, as Editor". Grumble :-)

AA-Turner:
Another option, though less nice from a metadata perspective, is to add a note to the top of the PEP, e.g.

```
.. note:: This PEP was originally written by Greg Stein for DB-API version 1, with Marc-André Lemburg becoming editor for the PEP [spam]
```

Would that resolve the attribution problem?

A

malemburg:
I've now added a short notice to the acknowledgements to the same effect.

AA-Turner (approved):

malemburg:
Thanks for helping, @AA-Turner

PR #2897, opened 2022-11-20
PEP 698: reword "value add" -> "benefits"

stroxler:
As suggested in after-merge comments on
https://github.com/python/peps/pull/2893

Maybe I've spent too much time in big-company roadmapping meetings, I didn't used to accidentally say things like this :D


CAM-Gerlach (approved):
This PR is a quick win with high return on investment that aligns holistic synergy between the content and the customer journey of those reading it.

PR #2902, opened 2022-11-23
PEP 696: Add further clarification and edge cases

Gobot1234:
cc @erictraut

### DIFF
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -217,8 +217,9 @@ Note
 ^^^^
 ``TypeVarTuple``\s as defaults are not supported due to
 :ref:`scoping-rules` not allowing usage of ``TypeVarLikes`` from outer
-scopes and multiple ``TypeVarTuple``\s cannot appear in the same
-type parameter lists.
+scopes, multiple ``TypeVarTuple``\s cannot appear in the same
+type parameter lists to a class and function defaults not being
+supported.
 
 .. _scoping-rules:
 
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -213,11 +213,11 @@ default to the type of ``start`` and step default to ``int | None``.
 When using a ``TypeVarLike`` as the default to another ``TypeVarLike``.
 Where ``T1`` is the default for ``T2`` the following rules apply.
 
-Note
-^^^^
-``TypeVarTuple``\s as defaults are not supported due to
+.. note::
+
+   ``TypeVarTuple``\s as defaults are not supported due to
 :ref:`scoping-rules` not allowing usage of ``TypeVarLikes`` from outer
-scopes, multiple ``TypeVarTuple``\s cannot appear in the same
+scopes, multiple ``TypeVarTuple``\s not being able appear in the same
 type parameter lists to a class and function defaults not being
 supported.
 
@@ -529,19 +529,19 @@ convenient, could have a ``TypeVarLike`` with no default follow a
 This would have also been a breaking change for a small number of cases
 where the code relied on ``Any`` being the implicit default.
 
-Allowing ``TypeVarLike``\ s with defaults to be used in function signatures
-'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+Allowing ``TypeVarLike``\s with defaults to be used in function signatures
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
-A previous version of this PEP allowed ``TypeVarLike``\ s to be used in
-function signatures. This was removed for the rationale described In
+A previous version of this PEP allowed ``TypeVarLike``\s to be used in
+function signatures. This was removed for the rationale described in
 :ref:`function-defaults`. Hopefully, this can be added in the future if
 a way to get the runtime value of a type parameter is added.
 
 Allowing ``TypeVarLikes`` from outer scopes in ``default``
 ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
-This was deemed too niche a feature to be worth the added complexity, if
-any cases arise where this is needed, it can be added in a future PEP.
+This was deemed too niche a feature to be worth the added complexity.
+If any cases arise where this is needed, it can be added in a future PEP.
 
 Acknowledgements
 ----------------
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -144,7 +144,7 @@ future, this might be possible (see `Interaction with PEP
 
 ``ParamSpec`` defaults are defined using the same syntax as
 ``TypeVar`` \ s but use a ``list`` or ``tuple`` of types or an ellipsis
-literal "``...``" or another in-scope ``ParamSpec`` (see :ref:`scoping-rules`).
+literal "``...``" or another in-scope ``ParamSpec`` (see :ref:`696-scoping-rules`).
 
 .. code-block:: py
 
@@ -161,7 +161,7 @@ literal "``...``" or another in-scope ``ParamSpec`` (see :ref:`scoping-rules`).
 
 ``TypeVarTuple`` defaults are defined using the same syntax as
 ``TypeVar`` \ s but use an unpacked tuple of types instead of a single type
-or another in-scope ``TypeVarTuple`` (see :ref:`scoping-rules`).
+or another in-scope ``TypeVarTuple`` (see :ref:`696-scoping-rules`).
 
 .. code-block:: py
 
@@ -216,12 +216,12 @@ Where ``T1`` is the default for ``T2`` the following rules apply.
 .. note::
 
    ``TypeVarTuple``\s as defaults are not supported due to
-:ref:`scoping-rules` not allowing usage of ``TypeVarLikes`` from outer
-scopes, multiple ``TypeVarTuple``\s not being able appear in the same
-type parameter lists to a class and function defaults not being
-supported.
+   :ref:`696-scoping-rules` not allowing usage of ``TypeVarLikes`` from
+   outer scopes, multiple ``TypeVarTuple``\s not being able appear in
+   the same type parameter lists to a class and function defaults not
+   being supported.
 
-.. _scoping-rules:
+.. _696-scoping-rules:
 
 Scoping Rules
 ~~~~~~~~~~~~~
@@ -275,7 +275,7 @@ The constraints of ``T2`` must be a superset of the constraints of ``T1``.
 
 ``TypeVarLike``\ s are valid as parameters to generics inside of a
 ``default`` when the first parameter is in scope as determined by the
-:ref:`previous section <scoping-rules>`.
+:ref:`previous section <696-scoping-rules>`.
 
 .. code-block:: py
 
@@ -369,7 +369,7 @@ subtype of one of the constraints.
    TypeVar("Ok", float, str, default=float)     # Valid
    TypeVar("Invalid", float, str, default=int)  # Invalid: expected one of float or str got int
 
-.. _function-defaults:
+.. _696-function-defaults:
 
 Function Defaults
 '''''''''''''''''
@@ -534,7 +534,7 @@ Allowing ``TypeVarLike``\s with defaults to be used in function signatures
 
 A previous version of this PEP allowed ``TypeVarLike``\s to be used in
 function signatures. This was removed for the rationale described in
-:ref:`function-defaults`. Hopefully, this can be added in the future if
+:ref:`696-function-defaults`. Hopefully, this can be added in the future if
 a way to get the runtime value of a type parameter is added.
 
 Allowing ``TypeVarLikes`` from outer scopes in ``default``
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -215,17 +215,17 @@ Where ``T1`` is the default for ``T2`` the following rules apply.
 
 .. note::
 
-   ``TypeVarTuple``\s are not supported as defaults due to:
+   ``TypeVarTuple``\s are not supported because:
 
-   - :ref:`696-scoping-rules` not allowing usage of ``TypeVarLikes``
+   - :ref:`696-scoping-rules` does not allow usage of ``TypeVarLikes``
      from outer scopes.
-   - Multiple ``TypeVarTuple``\s not being able appear in the same type
-     parameter lists to a class as defined in
+   - Multiple ``TypeVarTuple``\s cannot appear in the type
+     parameter list for a single class, as specified in
      :pep:`646#multiple-type-variable-tuples-not-allowed`.
-   - Function defaults not being supported.
+   - ``TypeVarLike`` defaults in functions are not supported.
 
-   These reasons leave no current valid location for a
-   ``TypeVarTuple``; hence, they are not supported.
+   These reasons leave no current valid location where a
+   ``TypeVarTuple`` could have a default.
 
 .. _696-scoping-rules:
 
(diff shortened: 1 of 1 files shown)


pep-0696.rst
```diff
@@ -265,9 +276,11 @@ The constraints of ``T2`` must be a superset of the constraints of ``T1``.
    class Bar(Generic[T, ListDefaultT]):
        def __init__(self, x: T, y: ListDefaultT): ...
 
-   reveal_type(Bar[int])             # type is Bar[int, list[int]]
-   reveal_type(Bar[int, list[str]])  # type is Bar[int, list[str]]
-   reveal_type(Bar[int, str])        # type is Bar[int, str]
+   reveal_type(Bar)                    # type is type[Bar[T, list[ListDefaultT = T]]]
```
erictraut, line 279: I presume this should be `type[Bar[T, ListDefaultT = list[T]]`?

pep-0696.rst
```diff
@@ -265,9 +276,11 @@ The constraints of ``T2`` must be a superset of the constraints of ``T1``.
...
+   reveal_type(Bar[int])               # type is type[Bar[int, ListDefaultT = list[int]]]
```
erictraut, line 280: I think you're saying that partial specialization is applied in this case? This is not what I'd expect. I think that when the use indicates an intent to specialize a generic type by using an index expression form, it should be fully specialized, with defaults applied as necessary. In other words, I think the revealed type here should be `type[Bar[int, list[int]]`.

Allowing partial specialization here would be inconsistent with existing rules for specialization where all TypeVar's have implied default types of "Any". Consider, for example, `reveal_type(dict[str])`. This should reveal `dict[str, Any]`, not `dict[str, _VT = Any]`.

Gobot1234: Yeah that makes sense, I've fixed the examples to remove examples of this

pep-0696.rst
```diff
@@ -206,20 +206,26 @@ default to the type of ``start`` and step default to ``int | None``.
...
+   reveal_type(Foo(1, ""))  # type is Foo[int, str]
+   Foo[int](1, "")          # Invalid: Foo[int, str] cannot be assigned to self: Foo[int, int] in Foo.__init__
+   Foo[int]("", 1)          # Invalid: Foo[str, int] cannot be assigned to self: Foo[int, int] in Foo.__init__
 
 When using a ``TypeVarLike`` as the default to another ``TypeVarLike``.
 Where ``T1`` is the default for ``T2`` the following rules apply.
 
+Note
+''''
+Currently ``TypeVarTuple``\s as defaults are not supported due to
```
erictraut, line 218: I recommend removing the word "Currently" because it's confusing. A PEP is meant to be a proposed design, so it should be unequivocal about what's supported and what's not. Of course, future changes can always change or extend a design. It may be worth adding a note in the "Rejected Ideas" section that explains why this was rejected. You can state there that this could be added in the future.

pep-0696.rst
```diff
@@ -206,20 +206,26 @@ default to the type of ``start`` and step default to ``int | None``.
...
+:ref:`scoping-rules` not allowing usage of ``TypeVarLikes`` from outer
+scopes and multiple ``TypeVarTuple``\s cannot appear in the same
```
erictraut, line 220: I think you mean to say that "TypeVarTuple's from outer scopes are not supported in default types". The scoping rules don't dictate this, but there are other good reasons to disallow this.

Also, I'm not sure what you mean by "multiple TypeVarTuples cannot appear in the same signature". Normally, "signature" refers to the parameter list of a function. Are you referring here to the type expression used for the default type? Why is it illegal for multiple TypeVarTuples to appear here? Is it legal for the same TypeVarTuple to appear multiple times, as in `tuple[int, *Ts] | tuple[str, *Ts]`?

pep-0696.rst
```diff
@@ -234,6 +240,8 @@ or be bound in an outer class or function scope.
    class slice(Generic[StartT, StopT, StepT]): ...
                      # ^^^^^^ Invalid: ordering does not allow StopT to be bound
 
+Currently using a ``TypeVarLike`` from an outer scope as a default is not supported.
```
erictraut, line 243: Same as comment above. I recommend removing the word "Currently" because it's confusing.

pep-0696.rst
Gobot1234: The scoping rules don't solely mean that they aren't allowed in defaults, it is also the combination that there aren't any cases where you can have a TypeVarTuple default to another and have be it resolved because you can't have
```py
class Foo[*Ts, *Us = *Ts]: ...
```
or because you can't have defaults in functions.

Unless I'm missing something there aren't any cases where you can use a TypeVarTuple with a TypeVarTuple default so they should just be disallowed.

pep-0696.rst
```diff
@@ -190,38 +196,52 @@ default to the type of ``start`` and step default to ``int | None``.
...
    class slice(Generic[StartT, StopT, StepT]): ...
 
+   reveal_type(slice)  # type is type[slice[StartT = int, StopT = StartT, StepT = int | None]]
    reveal_type(slice())                        # type is slice[int, int, int | None]
    reveal_type(slice[str]())                   # type is slice[str, str, int | None]
    reveal_type(slice[str, bool, timedelta]())  # type is slice[str, bool, timedelta]
 
+   T2 = TypeVar("T2", default=DefaultStrT)
+
+   class Foo(Generic[DefaultStrT, T2]):
```
erictraut, line 206: This references DefaultStrT, which isn't defined in this sample.

pep-0696.rst
```diff
@@ -507,6 +529,19 @@ convenient, could have a ``TypeVarLike`` with no default follow a
 This would have also been a breaking change for a small number of cases
 where the code relied on ``Any`` being the implicit default.
 
+Allowing ``TypeVarLike``\ s to be used in function signatures
```
erictraut, line 532: Do you mean "TypeVarLike" with a default type? Obviously, "TypeVarLike"s are already supported in function signatures today.

erictraut (changes requested):
I have a bunch of additional feedback on the draft PEP that I'll send you by email — mostly corrections to the sample code.

pep-0696.rst
Gobot1234: I was just using those from before as you might have guessed, I wasn't particularly aware of any stipulations that the snippets were meant to all contain the context to run them in an isolated way

pep-0696.rst
```diff
@@ -507,6 +529,19 @@ convenient, could have a ``TypeVarLike`` with no default follow a
 This would have also been a breaking change for a small number of cases
 where the code relied on ``Any`` being the implicit default.
 
+Allowing ``TypeVarLike``\ s with defaults to be used in function signatures
+'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+A previous version of this PEP allowed ``TypeVarLike``\ s to be used in
+function signatures. This was removed for the rationale described In
```
CAM-Gerlach, line 536:
```suggestion
function signatures. This was removed for the rationale described in
```

Fix typo

pep-0696.rst
```diff
@@ -507,6 +529,19 @@ convenient, could have a ``TypeVarLike`` with no default follow a
...
+:ref:`function-defaults`. Hopefully, this can be added in the future if
+a way to get the runtime value of a type parameter is added.
+
+Allowing ``TypeVarLikes`` from outer scopes in ``default``
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+This was deemed too niche a feature to be worth the added complexity, if
+any cases arise where this is needed, it can be added in a future PEP.
```
CAM-Gerlach, line 544:
```suggestion
This was deemed too niche a feature to be worth the added complexity.
If any cases arise where this is needed, it can be added in a future PEP.
```

Fix grammar error (major comma splice)

pep-0696.rst
```diff
@@ -346,6 +369,8 @@ subtype of one of the constraints.
    TypeVar("Ok", float, str, default=float)     # Valid
    TypeVar("Invalid", float, str, default=int)  # Invalid: expected one of float or str got int
 
+.. _function-defaults:
```
CAM-Gerlach, line 372: Since every ref target across every PEP shares a namespace, to avoid collisions please prefix all ref target labels with the PEP number, i.e.

```rst
.. _696-function-defaults:
```

You'll also need to update their usage accordingly in `:ref:` roles.

If you only add the number to the existing ref target labels, as suggested above, it will be automatically stripped when generating the HTML fragment anchors, so it won't break any existing external links (and passing `-n` will generate a warning on any broken internal ones).

pep-0696.rst
```diff
@@ -190,38 +196,52 @@ default to the type of ``start`` and step default to ``int | None``.
...
+   reveal_type(Foo(1, ""))  # type is Foo[int, str]
+   Foo[int](1, "")          # Invalid: Foo[int, str] cannot be assigned to self: Foo[int, int] in Foo.__init__
+   Foo[int]("", 1)          # Invalid: Foo[str, int] cannot be assigned to self: Foo[int, int] in Foo.__init__
+
 When using a ``TypeVarLike`` as the default to another ``TypeVarLike``.
 Where ``T1`` is the default for ``T2`` the following rules apply.
 
+Note
+^^^^
+``TypeVarTuple``\s as defaults are not supported due to
```
CAM-Gerlach, line 218:
```suggestion
.. note::

   ``TypeVarTuple``\s as defaults are not supported due to
```

May as well make this an actual Note admonition, no?

pep-0696.rst
```diff
@@ -190,38 +196,52 @@ default to the type of ``start`` and step default to ``int | None``.
...
+:ref:`scoping-rules` not allowing usage of ``TypeVarLikes`` from outer
+scopes, multiple ``TypeVarTuple``\s cannot appear in the same
```
CAM-Gerlach, line 220:
```suggestion
scopes, multiple ``TypeVarTuple``\s not being able appear in the same
```

I'm rather confused parsing this syntax, but I _think_ this _might_ be what you intend to say?

Gobot1234:
Thank you for the review @CAM-Gerlach, I'll fix the namespacing issues in the ref tag

pep-0696.rst
```diff
@@ -190,38 +196,52 @@ default to the type of ``start`` and step default to ``int | None``.
...
 When using a ``TypeVarLike`` as the default to another ``TypeVarLike``.
 Where ``T1`` is the default for ``T2`` the following rules apply.
 
+.. note::
+
+   ``TypeVarTuple``\s as defaults are not supported due to
+:ref:`scoping-rules` not allowing usage of ``TypeVarLikes`` from outer
+scopes, multiple ``TypeVarTuple``\s not being able appear in the same
+type parameter lists to a class and function defaults not being
+supported.
```
CAM-Gerlach, line 222:
```suggestion
   :ref:`scoping-rules` not allowing usage of ``TypeVarLikes`` from outer
   scopes, multiple ``TypeVarTuple``\s not being able appear in the same
   type parameter lists to a class and function defaults not being supported.
```

I forgot to explicitly include indenting the rest of the lines in my actual suggestion, sorry—that's why, predictably, you're seeing a build failure.

CAM-Gerlach (approved):
LGTM from a PEP editorial perspective, thanks

pep-0696.rst
```diff
@@ -190,38 +196,52 @@ default to the type of ``start`` and step default to ``int | None``.
...
+   Foo[int](1, "")          # Invalid: Foo[int, str] cannot be assigned to self: Foo[int, int] in Foo.__init__
+   Foo[int]("", 1)          # Invalid: Foo[str, int] cannot be assigned to self: Foo[int, int] in Foo.__init__
+
 When using a ``TypeVarLike`` as the default to another ``TypeVarLike``.
 Where ``T1`` is the default for ``T2`` the following rules apply.
 
-.. _scoping-rules:
+.. note::
+
+   ``TypeVarTuple``\s as defaults are not supported due to
```
JelleZijlstra, line 218: This sentence is very hard to understand, could you rewrite it?

pep-0696.rst
```diff
@@ -507,6 +529,19 @@ convenient, could have a ``TypeVarLike`` with no default follow a
 This would have also been a breaking change for a small number of cases
 where the code relied on ``Any`` being the implicit default.
 
+Allowing ``TypeVarLike``\s with defaults to be used in function signatures
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+A previous version of this PEP allowed ``TypeVarLike``\s to be used in
```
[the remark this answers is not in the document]
JelleZijlstra:
```suggestion
A previous version of this PEP allowed ``TypeVarLike``\s with defaults to be used in
```

pep-0696.rst
```diff
@@ -507,6 +529,19 @@ convenient, could have a ``TypeVarLike`` with no default follow a
...
+function signatures. This was removed for the rationale described in
```
JelleZijlstra, line 536:
```suggestion
function signatures. This was removed for the reasons described in
```
(wording nit)

JelleZijlstra (changes requested):

pep-0696.rst
CAM-Gerlach: For starters on this line, you could do

```suggestion
   ``TypeVarTuple``\s are not supported as defaults due to
```

JelleZijlstra: It's more that I got lost on the next few lines trying to figure out the structure of the sentence :)

Gobot1234: Does 
```rst
.. note::

   ``TypeVarTuple``\s are not supported as defaults due to:

   - :ref:`696-scoping-rules` not allowing usage of ``TypeVarLikes``
     from outer scopes
   - Multiple ``TypeVarTuple``\s not being able appear in the same type
     parameter lists to a class as defined in
     :pep:`646#multiple-type-variable-tuples-not-allowed`
   - Function defaults not being supported

   These reasons leave no current valid location for a
   ``TypeVarTuple``, hence, they are not supported.
```
read better?

CAM-Gerlach: (Fixed a few minor punctuation issues)

JelleZijlstra: I would rewrite each bullet point as a sentence, something like
Does 
```rst
.. note::

   ``TypeVarTuple``\s are not supported because:

   - :ref:`696-scoping-rules` does not allow usage of ``TypeVarLikes``
     from outer scopes.
   - Multiple ``TypeVarTuple``\s cannot appear in the type
     parameter list for a single class, as specified in
     :pep:`646#multiple-type-variable-tuples-not-allowed`.
   - ``TypeVar`` defaults in functions are not supported.

   These reasons leave no current valid location where a
   ``TypeVarTuple`` could have a default.
```

Gobot1234: That reads better now, thanks

JelleZijlstra (approved):

pep-0696.rst
JelleZijlstra: I just looked at the preview (https://pep-previews--2902.org.readthedocs.build/pep-0696/#using-another-typevarlike-as-default) and the "Note" formatting here looks way too prominent for my taste. Maybe remove the ".. note" and just make this a normal paragraph.

CAM-Gerlach: Yeah, as the person who suggested that in the first place (because the text was in a manual `Note:` block, especially with the reformat, seems like it makes more sense and flows better as prose now.

JelleZijlstra (approved):

EMMWIS:
code yet

PR #2875, opened 2022-11-08
PEP 8104: 2023 Term Steering Council election

ewdurbin:
Initial draft with pertinent dates and details.

PR #2912, opened 2022-12-06
enable netlify

ewdurbin:


ewdurbin:
Not sure what's up... seems it might be an incompatibility with python3.8 (best available on netlify)

AA-Turner:
I don't think PEPs need Netlify, as we already have render previews. Is there a reference to the discussion for this at all?

A

ewdurbin:
wires crossed here, I read that as CAM asking for netlify here https://github.com/python/peps/pull/2875#pullrequestreview-1172967034

AA-Turner:
No worries, thank you--good that the CPython previews are now enabled.

A

CAM-Gerlach:
Sorry for the confusion, and thanks for taking care of that—it will be a huge benefit there to both PR authors and reviewers alike!

PR #2922, opened 2022-12-13
PEP 620: Rejected

vstinner:


encukou:
The proper status is Withdrawn: #2923

PR #2924, opened 2022-12-14
PEP 8: Make example conform to PEP 8

johan-bolmsjo:


gvanrossum:
This is an example of "A foolish consistency is the hobgoblin of little mind", quoted in the PEP as the first subheading.

PR #2938, opened 2022-12-20
Revert "PEP 517: add missing def for the mandatory hook build_wheel signature"

apidwalinsubport:
Reverts python/peps#2535

### DIFF
--- a/pep-0517.txt
+++ b/pep-0517.txt
@@ -211,7 +211,7 @@ build_wheel
 
 ::
 
-    def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
+    build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
         ...
 
 Must build a .whl file, and place it in the specified ``wheel_directory``. It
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
This appears to another spammer or bot, FYI—this PR has no plausible logical reason to revert the commit (which is easy for a bot or unskilled human to mass-spam), its a nearly brand new account, very high rate of contributions, contributions seemingly random or purely for inflating its numbers, with no logic to them, and mostly on its own repositories, just like the others we've seen.

@python/organization-owners :point_up:

Rosuav:
Huh. The bot says that all commit authors have signed the CLA. So either the spammer is registered as having signed, or a reversion is listed differently from a regular commit.

ambv:
Well, spammers are also able to sign CLAs. This is what happened in this case.

CAM-Gerlach:
Yeah, I anecdotally suspect based on the fact they've shown just enough minimal interaction and variety elsewhere that it is probably a human spammer, rather than just a bot. Or if it is, they've gotten a good deal better...

PR #2942, opened 2022-12-31
PEP 702: Type system support for deprecations

JelleZijlstra:
This is the first time anyone other than me sees this document, so I welcome any feedback.

### DIFF
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -247,7 +247,7 @@ I found:
 * 9 deprecations of function parameters (supported by this PEP through
   decorating overloads)
 * 1 deprecation of a constant
-* 38 deprecations that are not easily detected in the type system (for
+* 38 deprecations that are not easily detectable in the type system (for
   example, for calling :func:`asyncio.get_event_loop` without an active
   event loop)
 
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -24,8 +24,8 @@ giving their users time to migrate to new APIs. Python provides a mechanism for
 achieving these goals: the :exc:`DeprecationWarning` warning class, which is
 used to show warnings when deprecated functionality is used. This mechanism is
 widely used: as of the writing of this PEP, the CPython main branch contains
-about 150 distinct deprecations. Many third-party libraries also use
-:exc:`!DeprecationWarning` to mark deprecations.
+about 150 distinct code paths that raise :exc:`!DeprecationWarning`. Many
+third-party libraries also use :exc:`!DeprecationWarning` to mark deprecations.
 
 However, the current mechanism is often insufficient to ensure that users
 of deprecated functionality update their code in time. For example, the
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -1,4 +1,4 @@
-PEP: 9999
+PEP: 702
 Title: Typing support for deprecations
 Author: Jelle Zijlstra <jelle.zijlstra@gmail.com>
 Status: Draft
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -285,7 +285,7 @@ show that this feature is not commonly needed.
 Acknowledgments
 ===============
 
-A discussion in the typing-sig meetup group led to useful feedback on this
+A call with the typing-sig meetup group led to useful feedback on this
 proposal.
 
 Copyright
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -171,6 +171,8 @@ while others (e.g., library developers who want their library to remain
 compatible with future versions of Python) would want to catch any use of
 deprecated functionality in their CI pipeline. Therefore, it is recommended
 that type checkers provide configuration options that cover both use cases.
+As with any other type checker error, it is also possible to ignore deprecations
+using ``# type: ignore`` comments.
 
 Deprecation policy
 ------------------
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -25,11 +25,11 @@ achieving these goals: the :exc:`DeprecationWarning` warning class, which is
 used to show warnings when deprecated functionality is used. This mechanism is
 widely used: as of the writing of this PEP, the CPython main branch contains
 about 150 distinct deprecations. Many third-party libraries also use
-:exc:`DeprecationWarning` to mark deprecations.
+:exc:`!DeprecationWarning` to mark deprecations.
 
 However, the current mechanism is often insufficient to ensure that users
 of deprecated functionality update their code in time. For example, the
-removal of various long-deprecated ``unittest`` features had to be
+removal of various long-deprecated :mod:`unittest` features had to be
 `reverted <https://github.com/python/cpython/commit/b50322d20337ca468f2070eedb051a16ee1eba94>`__
 from Python 3.11 to give users more time to update their code.
 Users may run their test suite with warnings disabled for practical reasons,
@@ -107,7 +107,7 @@ For deprecated classes and functions, this includes:
 
 There are some additional scenarios where deprecations could come into play:
 
-* An object may implement a :data:`typing.Protocol`, but only if deprecated
+* An object may implement a :class:`typing.Protocol`, but only if deprecated
   methods are considered.
 * A class may use the ``@override`` decorator from :pep:`698` to assert that
   its method overrides a base class method, but the base class method is
@@ -154,10 +154,10 @@ Here is how type checkers should handle usage of this library:
 Runtime behavior
 ----------------
 
-At runtime, the decorator sets an attribute ``.__deprecated__`` on the decorated
+At runtime, the decorator sets an attribute ``__deprecated__`` on the decorated
 object. The value of the attribute is the message passed to the decorator.
 The decorator returns the original object. Notably, it does not issue a runtime
-``DeprecationWarning``.
+:exc:`DeprecationWarning`.
 
 Type checker behavior
 ---------------------
@@ -200,7 +200,7 @@ Runtime warnings
 ----------------
 
 Users might expect usage of the ``@deprecated`` decorator to issue a
-``DeprecationWarning`` at runtime. However, this would raise a number of
+:exc:`DeprecationWarning` at runtime. However, this would raise a number of
 thorny issues:
 
 * When the decorator is applied to a class or an overload, the warning
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -1,0 +1,275 @@
+PEP: 9999
+Title: Typing support for deprecations
+Author: Jelle Zijlstra <jelle.zijlstra@gmail.com>
+Discussions-To: 
+Status: Draft
+Type: Standards Track
+Topic: Typing
+Content-Type: text/x-rst
+Created: 
+Python-Version: 3.12
+Post-History: 
+
+
+Abstract
+========
+
+This PEP adds an ``@typing.deprecated()`` decorator that marks a class or function
+as deprecated, enabling static checkers to warn when it is used.
+
+Motivation
+==========
+
+As software evolves, new functionality is added and old functionality becomes
+obsolete. Library developers want to work towards removing obsolete code while
+giving their users time to migrate to new APIs. Python provides a mechanism for
+achieving these goals: the :exc:`DeprecationWarning` warning class, which is
+used to show warnings when deprecated functionality is used. This mechanism is
+widely used: as of the writing of this PEP, the CPython main branch contains
+about 150 distinct deprecations. Many third-party libraries also use
+:exc:`DeprecationWarning` to mark deprecations.
+
+However, the current mechanism is often insufficient to ensure that users
+of deprecated functionality update their code in time. For example, the
+removal of various long-deprecated ``unittest`` features had to be
+`reverted <https://github.com/python/cpython/commit/b50322d20337ca468f2070eedb051a16ee1eba94>`__
+from Python 3.11 to give users more time to update their code.
+Users may run their test suite with warnings disabled for practical reasons,
+or deprecations may be triggered in code paths that are not covered by tests.
+
+Providing more ways for users to find out about deprecated functionality
+can speed up the migration process. This PEP proposes to leverage static type
+checkers to communicate deprecations to users. Such checkers have a thorough
+semantic understanding of user code, enabling them to detect and report
+deprecations that a single ``grep`` invocation could not find. In addition, many type
+checkers integrate with IDEs, enabling users to see deprecation warnings
+right in their editors.
+
+Rationale
+=========
+
+At first glance, deprecations may not seem like a topic that type checkers should
+touch. After all, type checkers are concerned with checking whether code will
+work as is, not with potential future changes. However, the analysis that type
+checkers perform on code to find type errors is very similar to the analysis
+that would be needed to detect usage of many deprecations. Therefore, type
+checkers are well placed to find and report deprecations.
+
+Other languages already have similar functionality:
+
+* GCC supports a ``deprecated`` `attribute <https://gcc.gnu.org/onlinedocs/gcc-3.1.1/gcc/Type-Attributes.html>`__
+  on function declarations. This powers CPython's ``Py_DEPRECATED`` macro.
+* GraphQL `supports <https://spec.graphql.org/June2018/#sec-Field-Deprecation>`__
+  marking fields as ``@deprecated``.
+* Kotlin `supports <https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/>`__
+  a ``Deprecated`` annotation.
+* Scala `supports <https://www.scala-lang.org/api/2.12.5/scala/deprecated.html>`__
+  an ``@deprecated`` annotation.
+* Swift `supports <https://docs.swift.org/swift-book/ReferenceManual/Attributes.html>`__
+  using the ``@available`` attribute to mark APIs as deprecated.
+* TypeScript `uses <https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#deprecated>`__
+  the ``@deprecated`` JSDoc tag to issue a hint marking use of
+  deprecated functionality.
+
+Several users have requested support for such a feature:
+
+* `typing-sig thread <https://mail.python.org/archives/list/typing-sig@python.org/thread/E24WTMQUTGKPFKEXVCGGEFFMG7LDF3WT/>`__
+* `Pyright feature request <https://github.com/microsoft/pyright/discussions/2300>`__
+* `mypy feature request <https://github.com/python/mypy/issues/11439>`__
+
+
+Specification
+=============
+
+The new ``@typing.deprecated()`` decorator can be used on a class, function or method.
+With overloaded functions, the decorator may be applied to individual overloads,
+indicating that the particular overload is deprecated. The decorator may also be
+applied to the overload implementation function, indicating that the entire function
+is deprecated.
+
+The decorator takes a single argument of type ``str``, which is a message that should
+be shown by the type checker when it encounters a usage of the decorated object.
+The content of deprecation messages is up to the user, but it may include the version
+in which the deprecated object is to be removed, and information about suggested
+replacement APIs.
+
+Type checkers should produce a diagnostic whenever they encounter a usage of an
+object marked as deprecated. For deprecated overloads, this includes all calls
+that resolve to the deprecated overload.
+For deprecated classes and functions, this includes:
+* References through module, class, or instance attributes (``module.deprecated_object``,
+  ``module.SomeClass.deprecated_method``, ``module.SomeClass().deprecated_method()``)
+* Any usage of deprecated objects in their defining module
+  (``x = deprecated_object()`` in ``module.py``)
+* If ``import *`` is used, usage of deprecated objects from the
+  module (``from module import *; x = deprecated_object()``)
+* ``from`` imports (``from module import deprecated_object``)
+
+There are some additional scenarios where deprecations could come into play:
+* An object may implement a :data:`typing.Protocol`, but only if deprecated
+  methods are considered.
+* A class may use the ``@override`` decorator from :pep:`698` to assert that
+  its method overrides a base class method, but the base class method is
+  deprecated.
+As these scenarios appear complex and relatively unlikely to come up in practice,
+this PEP does not mandate that type checkers detect them.
+
+Example
+-------
+
+As an example, consider this library stub named ``library.pyi``:
+
+.. code-block:: python
+
+   from typing import deprecated
+
+   @deprecated("Use Spam instead")
+   class Ham: ...
+
+   @deprecated("It is pining for the fiords")
+   def norwegian_gray(x: int) -> int: ...
+
+   @deprecated("Only str will be allowed")
+   @overload
+   def foo(x: int) -> str: ...
+   @overload
+   def foo(x: str) -> str: ...
+
+Here is how type checkers should handle usage of this library:
+
+.. code-block:: python
+
+   from library import Ham  # error: Use of deprecated class Ham. Use Spam instead.
+
+   import library
+
+   library.norwegian_gray(1)  # error: Use of deprecated function norwegian_gray. It is pining for the fiords.
+   map(library.norwegian_gray, [1, 2, 3])  # error: Use of deprecated function norwegian_gray. It is pining for the fiords.
+
+   library.foo(1)  # error: Use of deprecated overload for foo. Only str will be allowed.
+   library.foo("x")  # no error
+
+Runtime behavior
+----------------
+
+At runtime, the decorator sets an attribute ``.__deprecated__`` on the decorated
+object. The value of the attribute is the message passed to the decorator.
+The decorator returns the original object. Notably, it does not issue a runtime
+``DeprecationWarning``.
+
+Type checker behavior
+---------------------
+
+This PEP does not specify exactly how type checkers should present deprecation
+diagnostics to their users. However, some users (e.g., application developers
+targeting only a specific version of Python) may not care about deprecations,
+while others (e.g., library developers who want their library to remain
+compatible with future versions of Python) would want to catch any use of
+deprecated functionality in their CI pipeline. Therefore, it is recommended
+that type checker provide configuration options that cover both use cases.
+
+Backwards compatibility
+=======================
+
+Creating a new decorator poses no backwards compatibility concerns.
+As with all new typing functionality, the ``@deprecated`` decorator
+will be added to the ``typing_extensions`` module, enabling its use
+in older versions of Python.
+
+How to teach this
+=================
+
+For users who encounter deprecation warnings in their IDE or type
+checker output, the messages they receive should be clear and self-explanatory.
+Usage of the ``@deprecated`` decorator will be an advanced feature
+mostly relevant to library authors. The decorator should be mentioned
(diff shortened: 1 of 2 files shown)


hauntsaninja (commented):
I think it's somewhat unfortunate to not issue a runtime warning (I imagine most users would ideally want one). The SC has also indicated they'd like static typing to feel more cohesive with the runtime, so we should try harder to find something that could work. 

Should we have an opt-in kwarg a la Protocol's runtime_checkable? Type checkers could warn if we decorate an overload; we could even detect it at runtime if the decorators are in the right order. Decorating a class could cause `__init__` to raise the warning? (I think the popular deprecated library does this; can just use similar semantics)

hauntsaninja:
(Let me know if I should move this comment to discuss or typing-sig; I just didn't see a discussion link in there)

pep-0702.rst
```diff
@@ -0,0 +1,277 @@
...
+Content-Type: text/x-rst
+Created: 30-Dec-2022
+Python-Version: 3.12
+
+
+Abstract
+========
+
+This PEP adds an ``@typing.deprecated()`` decorator that marks a class or function
+as deprecated, enabling static checkers to warn when it is used.
```
ItsDrike, line 16:
```suggestion
as deprecated, enabling static checkers to produce a warning when it is used.
```

pep-0702.rst
```diff
@@ -0,0 +1,277 @@
...
+==========
+
+As software evolves, new functionality is added and old functionality becomes
+obsolete. Library developers want to work towards removing obsolete code while
+giving their users time to migrate to new APIs. Python provides a mechanism for
+achieving these goals: the :exc:`DeprecationWarning` warning class, which is
+used to show warnings when deprecated functionality is used. This mechanism is
+widely used: as of the writing of this PEP, the CPython main branch contains
+about 150 distinct deprecations. Many third-party libraries also use
+:exc:`DeprecationWarning` to mark deprecations.
```
ItsDrike, line 28:
```suggestion
:exc:`DeprecationWarning` to mark deprecations too.
```

pep-0702.rst
```diff
@@ -0,0 +1,277 @@
...
+Specification
+=============
+
+The new ``@typing.deprecated()`` decorator can be used on a class, function or method.
+With overloaded functions, the decorator may be applied to individual overloads,
+indicating that the particular overload is deprecated. The decorator may also be
+applied to the overload implementation function, indicating that the entire function
+is deprecated.
+
+The decorator takes a single argument of type ``str``, which is a message that should
```
ItsDrike, line 88: This will probably have to be a literal string, not just any string, as type checkers won't be able to obtain a string that's dynamically generated, and hence won't be able to use it in the warning.

JelleZijlstra:
Thanks for the feedback, but yes I think this is better directed to the discuss/typing-sig thread.

My plan is:
- Wait a few days to give people the chance to provide feedback on typos/organization/copyedits
- Merge the PEP
- Open discussions on both Discuss and typing-sig (since it's a typing PEP but also implicates the core development workflow

ItsDrike (changes requested):
Looks pretty good. 

Though I actually think that introducing `typing.Deprecated` would be something worth while, due to some additional uses it provides.

Mainly when deprecating type-aliases, like: `FOO_TYPE: TypeAlias = Union[A, B, C]`, it would be quite nice to be able to mark this type alias as deprecated with: `FOO_TYPE: Deprecated[TypeAlias, "Replaced by BAR_TYPE"] = Union[A, B, C]`. As this is currently completely impossible with the type system, and it's not even possible for a library to produce a runtime warning for these cases, which is why something like `typing.Deprecated` would be very nice to see.

pep-0702.rst
JelleZijlstra: Good point, will add that

pep-0702.rst
JelleZijlstra: The sentence already has "also"

pep-0702.rst
JelleZijlstra: I like the more concise current wording better

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+==========
+
+As software evolves, new functionality is added and old functionality becomes
+obsolete. Library developers want to work towards removing obsolete code while
+giving their users time to migrate to new APIs. Python provides a mechanism for
+achieving these goals: the :exc:`DeprecationWarning` warning class, which is
+used to show warnings when deprecated functionality is used. This mechanism is
+widely used: as of the writing of this PEP, the CPython main branch contains
+about 150 distinct deprecations. Many third-party libraries also use
+:exc:`DeprecationWarning` to mark deprecations.
```
gvanrossum, line 28: Is it easy to get some numbers for e.g. the 1000 most popular PyPI packages?

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+
+The new ``@typing.deprecated()`` decorator can be used on a class, function or method.
+With overloaded functions, the decorator may be applied to individual overloads,
+indicating that the particular overload is deprecated. The decorator may also be
+applied to the overload implementation function, indicating that the entire function
+is deprecated.
+
+The decorator takes a single argument of type ``str``, which is a message that should
+be shown by the type checker when it encounters a usage of the decorated object.
+The message must be a string literal or a string with a ``Literal`` type.
```
gvanrossum, line 90: If it's a string with a `Literal` type, shouldn't it be constrained to a single value? I wouldn't want to be a type checker having to implement something like
```py
message: typing.Literal["error", "warning"]

@typing.deprecated(message)
def f(x):
    ...
```

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+in which the deprecated object is to be removed, and information about suggested
+replacement APIs.
+
+Type checkers should produce a diagnostic whenever they encounter a usage of an
+object marked as deprecated. For deprecated overloads, this includes all calls
+that resolve to the deprecated overload.
+For deprecated classes and functions, this includes:
+
+* References through module, class, or instance attributes (``module.deprecated_object``,
+  ``module.SomeClass.deprecated_method``, ``module.SomeClass().deprecated_method()``)
```
gvanrossum, line 101: Surely also `module.SomeClass().deprecated_method` (i.e. no parens).

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+* Any usage of deprecated objects in their defining module
+  (``x = deprecated_object()`` in ``module.py``)
```
gvanrossum, line 103: This one could be problematic? *Internal* uses might need to be kept around, if the deprecated object returns something whose API is not deprecated. OTOH after
```py
deprecated_alias = deprecated_object
```
using `module.deprecated_alias` elsewhere should be considered deprecated.

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+Type checker behavior
+---------------------
+
+This PEP does not specify exactly how type checkers should present deprecation
+diagnostics to their users. However, some users (e.g., application developers
+targeting only a specific version of Python) may not care about deprecations,
+while others (e.g., library developers who want their library to remain
+compatible with future versions of Python) would want to catch any use of
+deprecated functionality in their CI pipeline. Therefore, it is recommended
+that type checker provide configuration options that cover both use cases.
```
gvanrossum, line 171: Hopefully a deprecated use will not issue a warning/error when `# type: ignore` is added.

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+checker output, the messages they receive should be clear and self-explanatory.
+Usage of the ``@deprecated`` decorator will be an advanced feature
+mostly relevant to library authors. The decorator should be mentioned
+in relevant documentation (e.g., :pep:`387` and the :exc:`DeprecationWarning`
+documentation) as an additional way to mark deprecated functionality.
+
+Reference Implementation
+========================
+
+TODO
```
gvanrossum, line 194: Just suggest it to Eric Traut as a holiday project, he'll do it overnight. :-)

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+thorny issues:
+
+* When the decorator is applied to a class or an overload, the warning
+  would not be raised as expected.
+* Users may want to control the ``warn`` call in more detail (e.g.,
+  changing the warning class).
+* ``typing.py`` generally aims to avoid affecting runtime behavior.
+
+Users who want to use ``@deprecated`` while also issuing a runtime warning
+can use the ``if TYPE_CHECKING:`` idiom, for example:
```
gvanrossum, line 213: I hope this doesn't become too popular, or we'd be better off doing this ourselves (it would be tricky to add later -- if people add other mechanisms they would get double warnings). And we _do_ have `reveal_type()`, which prints to stderr at runtime.

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+all current deprecations in the CPython standard library.
+
+I found:
+
+* 74 deprecations of functions, methods and classes (supported by this PEP)
+* 28 deprecations of whole modules (largely due to :pep:`594`)
+* 9 deprecations of function parameters (supported by this PEP through
+  decorating overloads)
+* 1 deprecation of a constant
+* 38 deprecations that are not easily detected in the type system (for
```
gvanrossum, line 250:
```suggestion
* 38 deprecations that are not easily detectable in the type system (for
```

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+This PEP covers deprecations of classes, functions and overloads. This
+allows type checkers to detect many but not all possible deprecations.
+To evaluate whether additional functionality would be worthwhile, I
+`examined <https://gist.github.com/JelleZijlstra/ff459edc5ff0918e22b56740bb28eb8b>`__
+all current deprecations in the CPython standard library.
+
+I found:
+
+* 74 deprecations of functions, methods and classes (supported by this PEP)
+* 28 deprecations of whole modules (largely due to :pep:`594`)
```
gvanrossum, line 246: Hm. This does make me think that maybe we should try a bit harder to come up with something for modules. (Maybe at least when they are submodules of a package? Surely that must be the most common case for 3rd party packages.)

pep-0702.rst
JelleZijlstra: Working on it now for pyanalyze, might also do mypy if I find time.

pep-0702.rst
JelleZijlstra: I feel like the suggested implementation for whole-module deprecations a few paragraphs down is sound, so we can add it if we think it's sufficiently useful. I'll bring this up when I open discussion threads to get more feedback on whether it is useful.

This number might be a bit of a distortion because CPython happens to be in the middle of lots of PEP 594 deprecations, so the number of whole-module deprecations is higher than it would usually be.

pep-0702.rst
JelleZijlstra:
```suggestion
  ``module.SomeClass.deprecated_method``, ``module.SomeClass().deprecated_method``)
```

pep-0702.rst
JelleZijlstra: Yes, the error should be `# type: ignore`able like any other type checker error. I can add an explicit mention of that if you think it's useful.

Also, unrelated typo I noticed:
```suggestion
that type checkers provide configuration options that cover both use cases.
```

pep-0702.rst
JelleZijlstra: @hauntsaninja also brought this up above. I'm now thinking of adding a `runtime=True` kwarg to the decorator that makes it also throw a runtime warning.

pep-0702.rst
JelleZijlstra: Yes, it should be a single literal, I'll rephrase. I wanted to leave the door open for something like:

```python
MESSAGE: Final = "This is deprecated for reasons x, y, z. You should use Spam instead."

@deprecated(MESSAGE)
def func1(): ...

@deprecated(MESSAGE)
def func2(): ...
```

pep-0702.rst
AlexWaygood: On the one hand, users will probably expect some runtime behaviour, and will probably be surprised if there isn't a runtime effect.

On the other hand, adding runtime behaviour feels like we might be stepping on the toes of third-party packages that have spent a lot of time thinking about this kind of thing, like https://github.com/tantale/deprecated.

(I don't really have a solution here or a strong preference, just noting that it might be an issue.)

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+
+   library.foo(1)  # error: Use of deprecated overload for foo. Only str will be allowed.
+   library.foo("x")  # no error
+
+Runtime behavior
+----------------
+
+At runtime, the decorator sets an attribute ``.__deprecated__`` on the decorated
+object. The value of the attribute is the message passed to the decorator.
+The decorator returns the original object. Notably, it does not issue a runtime
```
CAM-Gerlach, line 159:
```suggestion
The decorator returns the modified original object. Notably, it does not issue a runtime
```

Doesn't it return the object with the above modification? To me, it reads rather confusing to say that it modifies it and then returns the original.

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+in older versions of Python.
+
+How to teach this
+=================
+
+For users who encounter deprecation warnings in their IDE or type
+checker output, the messages they receive should be clear and self-explanatory.
+Usage of the ``@deprecated`` decorator will be an advanced feature
+mostly relevant to library authors. The decorator should be mentioned
+in relevant documentation (e.g., :pep:`387` and the :exc:`DeprecationWarning`
```
CAM-Gerlach, line 188: PEP 387 (PEP-387) specifies a normative process for Python's own development, rather than being itself (user) documentation. Maybe it would be worth being explicit here what change this PEP is recommending to that one (e.g. to require `@deprecated` when possible as part of the Python deprecation process, alongside `DeprecationWarning`/`warnings._deprecated` and `.. deprecated::`/`.. deprecated-removed::` in the docs?)

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+TODO
+
+Rejected ideas
+==============
+
+Runtime warnings
+----------------
+
+Users might expect usage of the ``@deprecated`` decorator to issue a
+``DeprecationWarning`` at runtime. However, this would raise a number of
```
CAM-Gerlach, line 203:
```suggestion
:exc:`DeprecationWarning` at runtime. However, this would raise a number of
```

IMO, worth being consistent here

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+thorny issues:
+
+* When the decorator is applied to a class or an overload, the warning
+  would not be raised as expected.
```
CAM-Gerlach, line 207: For a class, couldn't it be raised on `__init__`, though?

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+* Users may want to control the ``warn`` call in more detail (e.g.,
+  changing the warning class).
```
CAM-Gerlach, line 209: That's true—in general, this would not really be useful for most CPython deprecation warnings, which use the recommended `warnings._deprecated` helper to do several useful thinks.

pep-0702.rst
CAM-Gerlach: Yeah, for what it worth, it does `raise` some potential considerations, but if people are otherwise going to be doing ugly, boilerplately and verbose things, it's better to just provide the functionality built in now, IMO.

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+   library.norwegian_gray(1)  # error: Use of deprecated function norwegian_gray. It is pining for the fiords.
+   map(library.norwegian_gray, [1, 2, 3])  # error: Use of deprecated function norwegian_gray. It is pining for the fiords.
+
+   library.foo(1)  # error: Use of deprecated overload for foo. Only str will be allowed.
+   library.foo("x")  # no error
+
+Runtime behavior
+----------------
+
+At runtime, the decorator sets an attribute ``.__deprecated__`` on the decorated
```
CAM-Gerlach, line 157:
```suggestion
At runtime, the decorator sets an attribute ``__deprecated__`` on the decorated
```

Minor nit, but to be consistent with the usage elsewhere here and in the CPython docs

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+``Annotated`` could be added. However, this would create a new place
+in the type system where strings are just strings, not forward references,
+complicating the implementation of type checkers. In addition, my data
+show that this feature is not commonly needed.
+
+Acknowledgments
+===============
+
+A discussion in the typing-sig meetup group led to useful feedback on this
+proposal.
```
CAM-Gerlach, line 272: Could you link this?

CAM-Gerlach (commented):
Various minor suggestions and comments.

Trying out the new PEP checklist:

## New PEP Checklist

* [x] File created from the [latest PEP template](https://github.com/python/peps/blob/main/pep-0012/pep-NNNN.rst?plain=1)

### Content requirements

* [x] PEP topic [discussed in a suitable venue](https://peps.python.org/pep-0001/#start-with-an-idea-for-python) and general agreement that a PEP is appropriate
* [x] Content is reasonably sound, complete and makes at least basic technical sense
* [x] Abstract and Copyright sections completed, with the required wording in the latter
* [x] [Suggested sections](https://peps.python.org/pep-0012/#suggested-sections) included (unless not applicable)
* [ ] Title clearly, accurately and concisely describes the content [See comment]
* [x] Prose is intelligible and generally free of basic grammar and syntax errors, at least to the degree they significantly detract from understanding [Except as listed in comments]
* [x] Code is in code blocks, well-formatted (PEP 7/PEP 8) and has the right [lexer name](https://pygments.org/docs/lexers/) if non-Python (or ``text``)
* [x] ~~Any project stated in the PEP as supporting/endorsing benefiting from it confirms such~~ None found


### Technical requirements

* [x] Filename in the correct format (``pep-NNNN.rst``)
* [x] ``PEP``, ``Title``, ``Author``, ``Status`` (``Draft``), ``Type`` and ``Created`` headers filled out correctly
* [x] ``PEP-Delegate``, ``Requires``, ``Python-Version`` and ``Replaces`` filled out if appropriate
* [x] ``Sponsor`` listed and approves if not authored by a core dev or PEP editor
* [x] PEP builds with no warnings, CI checks pass and content displays as intended in the rendered preview [Except for as listed in comments]
* [x] PEP number assigned by the PEP editors and filename & ``PEP`` header updated accordingly
* [x] Sponsor/author(s) that are core devs/PEP editors added to ``.github/CODEOWNERS`` for the PEP

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+   @deprecated("Use Spam instead")
+   class Ham: ...
+
+   @deprecated("It is pining for the fiords")
+   def norwegian_gray(x: int) -> int: ...
+
+   @deprecated("Only str will be allowed")
+   @overload
+   def foo(x: int) -> str: ...
+   @overload
```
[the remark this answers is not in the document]
JelleZijlstra: Yes, this is the convention in stubs. We can reconsider this in Black if the PEP is accepted but for now I'd prefer to stick with the compact syntax.

pep-0702.rst
JelleZijlstra: It returns the original object, since it is modified in place.

pep-0702.rst
JelleZijlstra: No, it was a Zoom call.

pep-0702.rst
CAM-Gerlach: Right, I understand that, but it is still somewhat strange for me to read, since the object is mutated and thus not exactly the same as the original.

pep-0702.rst
CAM-Gerlach: Okay, then perhaps that could be clarified? I read this as a discussion on the Typing-SIG mailing list. E.g.

```suggestion
A Zoom call in the typing-sig meetup group led to useful feedback on this
proposal.
```

or

```suggestion
A Zoom discussion in the typing-sig meetup group led to useful feedback on this
proposal.
```

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
+PEP: 702
+Title: Typing support for deprecations
```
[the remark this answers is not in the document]
JelleZijlstra: I am going with "Marking deprecations using the type system" for now.

(And happy soon-to-be-New Year!)

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+
+* `typing-sig thread <https://mail.python.org/archives/list/typing-sig@python.org/thread/E24WTMQUTGKPFKEXVCGGEFFMG7LDF3WT/>`__
+* `Pyright feature request <https://github.com/microsoft/pyright/discussions/2300>`__
+* `mypy feature request <https://github.com/python/mypy/issues/11439>`__
+
+
+Specification
+=============
+
+The new ``@typing.deprecated()`` decorator can be used on a class, function or method.
```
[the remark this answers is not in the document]
JelleZijlstra: Pushing a slightly different wording

pep-0702.rst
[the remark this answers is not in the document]
JelleZijlstra: Rewriting this line to link `Final` instead.

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+* Any usage of deprecated objects in their defining module
+  (``x = deprecated_object()`` in ``module.py``)
+* If ``import *`` is used, usage of deprecated objects from the
+  module (``from module import *; x = deprecated_object()``)
+* ``from`` imports (``from module import deprecated_object``)
+
+There are some additional scenarios where deprecations could come into play:
+
+* An object may implement a :data:`typing.Protocol`, but only if deprecated
+  methods are considered.
```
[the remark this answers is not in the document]
JelleZijlstra: Pushing a rewrite

pep-0702.rst
JelleZijlstra: I think we should, but that hardly fits under "How to teach this" any more; I'll add a new section to the spec instead.

pep-0702.rst
JelleZijlstra: Or perhaps `__new__`. Correct support for all such variations feels like a huge can of worms.

pep-0702.rst
JelleZijlstra: Pushed a change

pep-0702.rst
JelleZijlstra: https://dev.to/hugovk/how-to-search-5000-python-projects-31gk specifically, I'm currently at 472/5000 downloaded and will report back when it is finished :)

pep-0702.rst
JelleZijlstra: I'm not sure why internal uses are special; e.g. when the `assertEquals` method in unittest is deprecated, unittest itself shouldn't be using it either.

pep-0702.rst
```diff
@@ -0,0 +1,278 @@
...
+Motivation
+==========
+
+As software evolves, new functionality is added and old functionality becomes
+obsolete. Library developers want to work towards removing obsolete code while
+giving their users time to migrate to new APIs. Python provides a mechanism for
+achieving these goals: the :exc:`DeprecationWarning` warning class, which is
+used to show warnings when deprecated functionality is used. This mechanism is
+widely used: as of the writing of this PEP, the CPython main branch contains
+about 150 distinct deprecations. Many third-party libraries also use
```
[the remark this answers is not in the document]
JelleZijlstra: It's code paths that lead to raising DeprecationWarning (e.g. warnings.warn, `warnings._deprecated`, whatever the equivalent in C is).

pep-0702.rst
JelleZijlstra: Added some numbers (at least 3.5k deprecations in the top 5k packages).

pep-0702.rst
```diff
@@ -79,15 +79,17 @@ Several users have requested support for such a feature:
...
 With overloaded functions, the decorator may be applied to individual overloads,
 indicating that the particular overload is deprecated. The decorator may also be
 applied to the overload implementation function, indicating that the entire function
 is deprecated.
 
 The decorator takes a single argument of type ``str``, which is a message that should
 be shown by the type checker when it encounters a usage of the decorated object.
-The message must be a string literal or a string with a ``Literal`` type.
+The message must be a string literal or be statically resolvable to a single string
+(e.g., a :data:`typing.Final` variable of type ``str``).
```
gvanrossum, line 100: Can't say I believe this is the right constraint either. It's totally possible to have something of type `Final[str]` whose value cannot be determined by the type checker -- that type just means that it cannot be reassigned, but the initial value could still be dynamically computed.

pep-0702.rst
```diff
@@ -170,6 +172,17 @@ compatible with future versions of Python) would want to catch any use of
 deprecated functionality in their CI pipeline. Therefore, it is recommended
 that type checkers provide configuration options that cover both use cases.
 
+Deprecation policy
+------------------
+
+CPython's deprecation policy (:pep:`387`) will require that new deprecations
```
gvanrossum, line 188: This is what's being _proposed_, right?

JelleZijlstra: Yes, should that be made more obvious somehow?

gvanrossum: I had to read it twice before I was sure of it. Maybe:
```suggestion
We propose that CPython's deprecation policy (:pep:`387`) is updated to require that new deprecations
```

pep-0702.rst
thomkeh: Pyright also really doesn't like code in the `else` branch of `if TYPE_CHECKING`: https://github.com/microsoft/pyright/discussions/2330

pep-0702.rst
gvanrossum: > This number might be a bit of a distortion because CPython happens to be in the middle of lots of PEP 594 deprecations, so the number of whole-module deprecations is higher than it would usually be.

I actually expect that more ancient, poorly-maintained modules in the stdlib will eventually get the axe.

pep-0702.rst
CAM-Gerlach: What about

```suggestion
(e.g., a :data:`~typing.Final` variable of type ``str`` assigned a string literal).
```

CAM-Gerlach (approved):
LGTM from my side, pending resolution of Guido's final comments. Thanks @JelleZijlstra !

pep-0702.rst
gvanrossum: Presumably for pyright that wouldn't be a problem, since it already tracks literal values. In mypy this will probably require something new, since it doesn't do this. No idea about Pyre, pytype and others. Maybe we should leave this discussion open for the more public debate of the PEP's merits and details. A base position to retreat to might be that it has to be a string literal? PEP 675 (`LiteralString`) comes to mind.

pep-0702.rst
JelleZijlstra: Will bring this up as a point of discussion when starting threads.

pep-0702.rst
JelleZijlstra: Will bring this up (too) as a point of discussion when starting threads.

pep-0702.rst
JelleZijlstra: I'll just go back to only string literals; type checkers can relax the requirements if they wish.

JelleZijlstra:
Thanks everyone for the feedback!

PR #2946, opened 2023-01-02
PEP 702: More discussion of precedent and runtime behavior

JelleZijlstra:

### DIFF
--- a/pep-0702.rst
+++ b/pep-0702.rst
@@ -246,9 +246,9 @@ thorny issues:
 * When the decorator is applied to a class or an overload, the warning
   would not be raised as expected. For classes, the warning could be
   raised on instantiation, but this would not cover usage in type
-  annotations or ``isinstance`` checks.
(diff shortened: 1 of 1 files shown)


pep-0702.rst
```diff
@@ -236,7 +244,9 @@ Users might expect usage of the ``@deprecated`` decorator to issue a
 thorny issues:
 
 * When the decorator is applied to a class or an overload, the warning
-  would not be raised as expected.
+  would not be raised as expected. For classes, the warning could be
+  raised on instantiation, but this would not cover usage in type
+  annotations or ``isinstance`` checks.
 * Users may want to control the ``warn`` call in more detail (e.g.,
   changing the warning class).
```
CAM-Gerlach, line 251:
```suggestion
  annotations or :func:`isinstance` checks.
* Users may want to control the :func:`~warnings.warn` call in more detail
  (e.g., changing the warning class).
```

You could cross-reference these for readers, if you wanted.

PR #2948, opened 2023-01-04
PEP 698: discuss.python.org thread to post-history

stroxler:
As far as I know this is the last update needed ahead of SC review.


JelleZijlstra (approved):

CAM-Gerlach:
If that is the most current/canonical thread, and where the resolution is going to be posted, you might want to update the `Discussions-To` to point to it as well.

PR #2952, opened 2023-01-07
Uncap the Sphinx requirement

AA-Turner:
This updates the Sphinx requirement to allow 6.1.2, but blacklists 6.1.0 and 6.1.1 (broken versions) from being installed.

A

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # Requirements for building PEPs with Sphinx
 Pygments >= 2.9.0
-Sphinx >= 5.1.1,<=6.0.0
+Sphinx >= 5.1.1,!=6.1.0,!=6.1.1,<=6.1.2
 docutils >= 0.19.0
 
 # For RSS
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,8 @@
 # Requirements for building PEPs with Sphinx
 Pygments >= 2.9.0
-Sphinx >= 5.1.1,!=6.1.0,!=6.1.1,<=6.1.2
+# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
+# See https://github.com/sphinx-doc/sphinx/pull/11100
+Sphinx >= 5.1.1, != 6.1.0, != 6.1.1
 docutils >= 0.19.0
 
 # For RSS


requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Requirements for building PEPs with Sphinx
 Pygments >= 2.9.0
-Sphinx >= 5.1.1,<=6.0.0
+Sphinx >= 5.1.1,!=6.1.0,!=6.1.1,<=6.1.2
```
JelleZijlstra, line 3: Did you mean `>=` at the end? Do we think it'll be safe to auto-upgrade to 6.1.3 or 6.2.0 when the time comes?

We could consider pinning to a specific version.

AA-Turner: I went with continuing to cap but very happy to remove the cap (if so we should just do the below):
```suggestion
Sphinx >= 5.1.1,!=6.1.0,!=6.1.1
```

CAM-Gerlach, line 3: I previously suggesting keeping the upper bound to avoid the possibility of the preview builds and the live site suddenly breaking on unrelated PRs due to a Sphinx version upgrade, and persisting until manually fixed and PRs rebased, as happened in #2949 here. This still maintains a relatively forgiving range of compatible versions, while ensuring Sphinx version upgrades are done in a controlled and atomic manner.

Of course, it means we need to bump that periodically, so perhaps not worth it, at least down to the patch version? Upper-bounding to the minor version (`<6.2`), which would still pick up small bugfixes but not larger changes (like those here that broke images), and avoid some of the churn, might be a more reasonable compromise.

Either way, we should document why this constraint is imposed, and use consistent (and more readable) formatting. Here are each of the options with that implemented:

Upper-bound to patch version (current):

```suggestion
# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
# See https://github.com/sphinx-doc/sphinx/pull/11100
Sphinx >= 5.1.1, != 6.1.0, != 6.1.1, <= 6.1.2
```

Upper bound to minor version (compromise):

```suggestion
# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
# See https://github.com/sphinx-doc/sphinx/pull/11100
Sphinx >= 5.1.1, != 6.1.0, != 6.1.1, < 6.2
```

No upper bound (previous):

```suggestion
# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
# See https://github.com/sphinx-doc/sphinx/pull/11100
Sphinx >= 5.1.1, != 6.1.0, != 6.1.1
```

CAM-Gerlach: I previously suggesting keeping the upper bound to avoid the possibility of the preview builds and the live site suddenly breaking on unrelated PRs due to a Sphinx version upgrade, and persisting until manually fixed and PRs rebased, as happened in #2949 here. This still maintains a relatively forgiving range of compatible versions, while ensuring Sphinx version upgrades are done in a controlled and atomic manner.

Of course, it means we need to bump that periodically, so perhaps not worth it, at least down to the patch version? Upper-bounding to the minor version (`<6.2`), which would still pick up small bugfixes but not larger changes (like those here that broke images), and avoid some of the churn, might be a more reasonable compromise.

Either way, we should document why this constraint is imposed, and use consistent (and more readable) formatting. Here are each of the options with that implemented:

Upper-bound to patch version (current):

```suggestion
# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
# See https://github.com/sphinx-doc/sphinx/pull/11100
Sphinx >= 5.1.1, != 6.1.0, != 6.1.1, <= 6.1.2
```

Upper bound to minor version (compromise):

```suggestion
# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
# See https://github.com/sphinx-doc/sphinx/pull/11100
Sphinx >= 5.1.1, != 6.1.0, != 6.1.1, < 6.2
```

No upper bound (previous):

```suggestion
# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
# See https://github.com/sphinx-doc/sphinx/pull/11100
Sphinx >= 5.1.1, != 6.1.0, != 6.1.1
```

PR #2960, opened 2023-01-13
PEP 561: Clarify 4th step in Module Resolution Order

not-my-profile:
Fixes https://github.com/python/typing/issues/1334.

\cc @ethanhs as PEP author.

### DIFF
--- a/pep-0561.rst
+++ b/pep-0561.rst
@@ -182,6 +182,7 @@ laid out as follows::
         └── hexagon
             └── __init__.pyi
 
+.. _mro:
 
 Type Checker Module Resolution Order
 ------------------------------------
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

pep-0561.rst
```diff
@@ -200,8 +201,10 @@ resolve modules containing type information:
 3. Stub packages - these packages SHOULD supersede any installed inline
    package. They can be found at ``foopkg-stubs`` for package ``foopkg``.
 
-4. Inline packages - if there is nothing overriding the installed
-   package, *and* it opts into type checking, inline types SHOULD be used.
+4. Packages with a ``py.typed`` marker file - if there is nothing overriding
+   the installed package, *and* it opts into type checking, the types
+   bundled with the package should be used (be they in ``.pyi`` type
```
hauntsaninja, line 206: nit: preserve the RFC like wording
```suggestion
   bundled with the package SHOULD be used (be they in ``.pyi`` type
```

not-my-profile: Nice catch! Fixed.

ethanhs (approved):
Thanks for improving the wording!

PR #2964, opened 2023-01-16
PEP 704: Require virtual environments by default for installers

pradyunsg:
Enforce the use of virtual environments and establish a conventional name for the virtual environment directory.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -584,6 +584,7 @@ pep-0700.rst  @pfmoore
 pep-0701.rst  @pablogsal @isidentical @lysnikolaou
 pep-0702.rst  @jellezijlstra
 pep-0703.rst  @ambv
+pep-0703.rst  @brettcannon @pradyunsg
 # ...
 # pep-0754.txt
 # ...
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -584,7 +584,7 @@ pep-0700.rst  @pfmoore
 pep-0701.rst  @pablogsal @isidentical @lysnikolaou
 pep-0702.rst  @jellezijlstra
 pep-0703.rst  @ambv
-pep-0703.rst  @brettcannon @pradyunsg
+pep-0704.rst  @brettcannon @pradyunsg
 # ...
 # pep-0754.txt
 # ...
--- a/pep-0704.rst
+++ b/pep-0704.rst
@@ -15,35 +15,42 @@ Post-History: `16-Jan-2023 <https://discuss.python.org/t/22846>`__
 Abstract
 ========
 
-This PEP proposes that package installers like ``pip`` MUST require a virtual environment by default, on Python 3.13+.
+This PEP recommends that package installers like ``pip``  require a virtual environment by default, on Python 3.13+.
+
 
 Motivation
 ==========
 
-Python virtual environments are an essential part of the development workflow for Python. However, it requires extra effort since it is an opt-in feature, and requires users to either:
+Python virtual environments are an essential part of the development workflow for Python. However, they require extra effort since they are an opt-in feature, and requires users to either:
 
 - take explicit steps to activate/deactivate a virtual environment
-- use ``<path-to-venv>/<bin-path>/<executable>`` to run files
+- use :samp:`{<path-to-venv>}/{<bin-path>}/{<executable>}` to run files
 
-This creates a barrier to entry for many new users, since virtual environments serve as a pre-requisite for typical Python development workflows and how virtual environments work is a lot of information for anyone new. It can take a lot of extra time and effort to explain them.
+For new users, things will seemingly work correctly when virtual environments are not used until they don't. Further, activating a virtual environment uses slightly different syntax and mechanisms on different platforms. This complicates the introduction to virtual environments since now information and context about how/why they are useful needs to explained to justify adding additional steps to the workflow.
 
-Further, activation of virtual environment uses a slightly different syntax/mechanism to activate on different platforms.
+It also creates a scope for mistakes, since users need to remember to activate the virtual environment before running an installer like ``pip`` or configure those installers to error out. On certain Linux distributions, forgetting to do so can result in the installer modifying files that are owned by the operating system (which is partially mitigated by :pep:`668` for distributions that opt-in to marking their environments accordingly).
 
-It also creates a scope for mistakes, since users must remember to activate the virtual environment before running an installer like ``pip``. On certain Linux distributions, this can result in the installer modifying files that are owned by the operating system (this is partially mitigated by :pep:`668` for distributions that opt-in to marking their environments accordingly).
 
-This PEP proposes that the default behaviour of installers like ``pip`` should be to require a virtual environment to be active. This would make it easier for new users to get started with Python, and would reduce the scope for mistakes.
+Rationale
+=========
 
+Changing the default behaviour of installers like ``pip`` to require a virtual environment to be active would:
+
+- make it easier for new users to get started with Python (since there's a consistent experience and virtual environments are understood as a thing you're required to use)
+- reduce the scope for accidental installation issues for all users by default (by explicitly flagging when you're not using a virtual environment).
+
+Setting up a convention of placing the virtual environment in-tree in a directory named ``.venv`` removes an decision point for common workflows and creates a clear convention within the ecosystem.
 
 Specification
 =============
 
-When a user runs an installer without an active virtual environment, the installer MUST print an error message and exit with a non-zero exit code. The error message MUST inform the user that a virtual environment is required, SHOULD provide shell-specific instructions on how to create and activate a virtual environment named ``.venv``, and SHOULD provide a link to a documentation page that explains how to create and activate a virtual environment.
+When a user runs an installer without an active virtual environment, the installer SHOULD print an error message and exit with a non-zero exit code. The error message SHOULD inform the user that a virtual environment is required, SHOULD provide shell-specific instructions on how to create and activate a virtual environment named ``.venv``, and SHOULD provide a link to a documentation page that explains how to create and activate a virtual environment.
 
-The installer SHOULD also provide a mechanism to disable this behaviour, allowing a user to opt-in to running the tool outside of a virtual environment. The installer MAY include this in the error message that it prints, and MUST include it in the documentation page it links to.
+The installer SHOULD also provide a mechanism to disable this behaviour, allowing a user to opt-in to running the tool outside of a virtual environment. The installer MAY include this in the error message that it prints, and SHOULD include it in the documentation page it links to.
 
 If the installer includes a link to a documentation page in the error message described above, the documentation page SHOULD be written with a focus on new users. It should explain what virtual environments are, why they are required, and how to create and activate a virtual environment. It SHOULD include instructions for the most common shells and platforms.
 
-Installers MAY choose to implement this default behaviour on any Python versions, but MUST implement it on Python 3.13 or newer.
+Installers MAY choose to implement this default behaviour on any Python versions, but SHOULD implement it on Python 3.13 or newer.
 
 
 Backwards Compatibility
--- a/pep-0704.rst
+++ b/pep-0704.rst
@@ -9,8 +9,7 @@ Type: Standards Track
 Topic: Packaging
 Content-Type: text/x-rst
 Created: 16-Jan-2023
-Python-Version: 3.13
-Post-History: 16-Jan-2023
+Post-History: `16-Jan-2023 <https://discuss.python.org/t/22846>`__
 
 
 Abstract
--- a/pep-0704.rst
+++ b/pep-0704.rst
@@ -44,11 +44,22 @@ Setting up a convention of placing the virtual environment in-tree in a director
 Specification
 =============
 
-When a user runs an installer without an active virtual environment, the installer SHOULD print an error message and exit with a non-zero exit code. The error message SHOULD inform the user that a virtual environment is required, SHOULD provide shell-specific instructions on how to create and activate a virtual environment named ``.venv``, and SHOULD provide a link to a documentation page that explains how to create and activate a virtual environment.
+Requiring a virtual environment by default
+------------------------------------------
 
-The installer SHOULD also provide a mechanism to disable this behaviour, allowing a user to opt-in to running the tool outside of a virtual environment. The installer MAY include this in the error message that it prints, and SHOULD include it in the documentation page it links to.
+When a user runs an installer without an active virtual environment, the installer SHOULD print an error message and exit with a non-zero exit code.
 
-If the installer includes a link to a documentation page in the error message described above, the documentation page SHOULD be written with a focus on new users. It should explain what virtual environments are, why they are required, and how to create and activate a virtual environment. It SHOULD include instructions for the most common shells and platforms.
+The error message SHOULD inform the user that a virtual environment is required, SHOULD provide shell-specific instructions on how to create and activate a virtual environment named ``.venv``, and SHOULD provide a link to a documentation page that explains how to create and activate a virtual environment.
+
+See `Implementation notes`_ for more details.
+
+Opting out of virtual environments
+----------------------------------
+
+The installer SHOULD also provide a explicit opt-in before allowing an end user to use it outside of a virtual environment. If the installer does not provide this functionality, it SHOULD mention this in the error message and documentation page.
+
+Consistent timeline for the change
+----------------------------------
 
 Installers MAY choose to implement this default behaviour on any Python versions, but SHOULD implement it on Python 3.13 or newer.
 
@@ -75,14 +86,34 @@ This PEP does not introduce any new security implications.
 How to Teach This
 =================
 
-This PEP does require that new users create and use a virtual environment to get started with using Python packages. This is however a best practice, as demonstrated by the section on "basics of how to install Python packages" in the Python Packaging User Guide, which explains how/what virtual environments are before discussing using ``pip``.[1]_
+This PEP does require that new users create and use a virtual environment to get started with using Python packages. This is however a best practice, as `demonstrated <https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments>`__ by the section on "basics of how to install Python packages" in the Python Packaging User Guide, which explains how/what virtual environments are before discussing using ``pip``.
 
 
 Reference Implementation
 ========================
 
-There is no reference implementation for this PEP, however the proposed behaviour is already implemented in ``pip`` and can be achieved by setting ``PIP_REQUIRE_VENV`` environment variable to ``1`` (not setting it would result in the proposed opt-in behaviour).
+There is no reference implementation for this PEP, however the proposed behaviour is largely implemented in ``pip`` and can be achieved by setting ``PIP_REQUIRE_VENV`` environment variable to ``1`` (not setting it would result in the proposed opt-in behaviour).
 
+
+Implementation Notes
+====================
+
+Detecting an active virtual environment
+---------------------------------------
+
+:pep:`As discussed in PEP 668 <668#backwards-compatibility>`, the logic for robustly detecting a virtual environment is something like::
+
+    def is_virtual_environment():
+        return sys.base_prefix != sys.prefix or hasattr(sys, "real_prefix")
+
+Documentation on using a virtual environment
+--------------------------------------------
+
+Package installers are expected to provide a link to a documentation page in the error message.
+
+Ideally, such a documentation page would explain what virtual environments are, why they are required, and how to create and activate a virtual environment using ``venv``. It should include instructions for the most common shells and platforms.
+
+Such a documentation page should be made available in the `Python Packaging User Guide <https://packaging.python.org>`__ to reduce duplicated effort across installers for covering this topic.
 
 Rejected Ideas
 ==============
@@ -131,11 +162,6 @@ Open Issues
 
 None.
 
-
-Footnotes
-=========
-
-.. [1] https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
 
 Copyright
 =========
--- a/pep-0704.rst
+++ b/pep-0704.rst
@@ -15,7 +15,7 @@ Post-History: `16-Jan-2023 <https://discuss.python.org/t/22846>`__
 Abstract
 ========
 
-This PEP recommends that package installers like ``pip``  require a virtual environment by default, on Python 3.13+.
+This PEP recommends that package installers like ``pip``  require a virtual environment by default on Python 3.13+.
 
 
 Motivation
@@ -26,7 +26,7 @@ Python virtual environments are an essential part of the development workflow fo
 - take explicit steps to activate/deactivate a virtual environment
 - use :samp:`{<path-to-venv>}/{<bin-path>}/{<executable>}` to run files
 
-For new users, things will seemingly work correctly when virtual environments are not used until they don't. Further, activating a virtual environment uses slightly different syntax and mechanisms on different platforms. This complicates the introduction to virtual environments since now information and context about how/why they are useful needs to explained to justify adding additional steps to the workflow.
+For new users, things will seemingly work correctly when virtual environments are not used -— until they don't. Further, activating a virtual environment uses slightly different syntax and mechanisms on different platforms. This complicates the introduction to virtual environments, since now information and context about how/why they are useful needs to explained to justify adding additional steps to the workflow.
 
 It also creates a scope for mistakes, since users need to remember to activate the virtual environment before running an installer like ``pip`` or configure those installers to error out. On certain Linux distributions, forgetting to do so can result in the installer modifying files that are owned by the operating system (which is partially mitigated by :pep:`668` for distributions that opt-in to marking their environments accordingly).
 
@@ -39,7 +39,7 @@ Changing the default behaviour of installers like ``pip`` to require a virtual e
 - make it easier for new users to get started with Python (since there's a consistent experience and virtual environments are understood as a thing you're required to use)
 - reduce the scope for accidental installation issues for all users by default (by explicitly flagging when you're not using a virtual environment).
 
-Setting up a convention of placing the virtual environment in-tree in a directory named ``.venv`` removes an decision point for common workflows and creates a clear convention within the ecosystem.
+Setting up a convention of placing the virtual environment in-tree in a directory named ``.venv`` removes a decision point for common workflows and creates a clear convention within the ecosystem.
 
 Specification
 =============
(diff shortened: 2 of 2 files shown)


CAM-Gerlach:
Looks like PEP 704 is available, so it's all yours.

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+If the installer includes a link to a documentation page in the error message described above, the documentation page SHOULD be written with a focus on new users. It should explain what virtual environments are, why they are required, and how to create and activate a virtual environment. It SHOULD include instructions for the most common shells and platforms.
+
+Installers MAY choose to implement this default behaviour on any Python versions, but MUST implement it on Python 3.13 or newer.
+
+
+Backwards Compatibility
+=======================
+
+This PEP is backwards incompatible with workflows where users who are using installers outside of virtual environments. Such users will be prompted with an error message and will need to either:
```
JelleZijlstra, line 53:
```suggestion
This PEP is backwards incompatible with workflows where users are using installers outside of virtual environments. Such users will be prompted with an error message and will need to either:
```

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Users who are already using virtual environments will not be affected by this change.
+
+Workflow tools (which manage virtual environments for the user, under the hood) should be unaffected, since they should already be using the virtual environment for running the installer.
+
+
+Security Implications
+=====================
+
+This PEP does not introduce any new security implications.
```
h-vetinari, line 66: I'm pretty sure it can be argued that this improves the security situation, as it will guide users to avoid to "just add `sudo`" to get things to work, and potentially installing arbitrary code with those privileges. By defaulting to environments, the user doesn't need an escape hatch for not having enough privileges to install something - things will work fine just with standard rights.

Rosuav: We already have a non-root option available, and it's the default in current versions of pip (user installation). This isn't really a security concern IMO.

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+- take explicit steps to activate/deactivate a virtual environment
+- use ``<path-to-venv>/<bin-path>/<executable>`` to run files
+
+This creates a barrier to entry for many new users, since virtual environments serve as a pre-requisite for typical Python development workflows and how virtual environments work is a lot of information for anyone new. It can take a lot of extra time and effort to explain them.
+
+Further, activation of virtual environment uses a slightly different syntax/mechanism to activate on different platforms.
+
+It also creates a scope for mistakes, since users must remember to activate the virtual environment before running an installer like ``pip``. On certain Linux distributions, this can result in the installer modifying files that are owned by the operating system (this is partially mitigated by :pep:`668` for distributions that opt-in to marking their environments accordingly).
+
+This PEP proposes that the default behaviour of installers like ``pip`` should be to require a virtual environment to be active. This would make it easier for new users to get started with Python, and would reduce the scope for mistakes.
```
CAM-Gerlach, line 35: These two sentences are really Rationale rather than Motivation, as they try to explain why the solution you're proposing should be accepted to solve the problem outlined in the Motivation. Furthermore, I'd strongly suggest expanding that Rationale section at least somewhat so it actually explains how implementing this default behavior will achieve the aim you state here, as currently on the surface it appears to contradict itself (which, to be clear, I believe you have a reasonably plausible case for—you just have to actually make it :slightly_smiling_face: ).

In particular, most of the aforementioned Motivation focuses at some length on the UX challenges that virtual environment present for new users, as it describes the extra steps that using venvs require, states that this "creates a barrier to entry for many new users", and mentions that "how virtual environments work is a lot of information for anyone new. It can take a lot of extra time and effort to explain them."

However, in the Rationale section, it states having installers "require a virtual environment to be active" by default "would make it easier for new users to get started with Python". As the previous lines just established that understanding venvs and knowing the actions to take to create and active one are "a barrier to entry for many new users", and there is no mention here of _how_ this would be made easier for new users, it sounds like a _prima facie_ contradiction that requiring new users to create and activate a venv before installing any package would make it easier, rather than harder, "for new users to get started".

Also, by contrast, the Motivation more briefly touches on the issue of accidentally installing packages without a venv active, mentioning a single concrete example that is at least partially mitigated by another accepted PEP, despite this arguably being the broader and more obvious benefit here.

Therefore, I suggest:

* Make a Rationale section starting with this paragraph.
* Mention how this proposal will actually make thins easier for new users by, e.g.,
  * Mandating that tools give users simple, specific and actionable advice on how to create and activate one, and links to help resources
  * Standardizes the naming of venvs to avoid user confusion and improve tool automation
  * Etc.
* Add at least >=one more example to the Motivation of the real-world problems not activating a venv causes users

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+=======================
+
+This PEP is backwards incompatible with workflows where users are using installers outside of virtual environments. Such users will be prompted with an error message and will need to either:
+
+- explicitly opt-in to running the installer outside of a virtual environment, or
+- create and use a virtual environment
+
+Users who are already using virtual environments will not be affected by this change.
+
+Workflow tools (which manage virtual environments for the user, under the hood) should be unaffected, since they should already be using the virtual environment for running the installer.
```
CAM-Gerlach, line 60: They will be affected if they don't use `venv`/`virtualenv` virtual environments (e.g. Conda envs, I'd assume PDM PEP 582 envs, and possibly others), assuming tools only detect them. Either way, that really should be explicitly addressed here.

pep-0704.rst
CAM-Gerlach, line 66: You could potentially argue that it could improve security slightly, as any malicious package will (at least initially) only be installed into a single isolated virtualenv, as opposed to the system environment used for all projects (perhaps even with administrator rights).

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+Security Implications
+=====================
+
+This PEP does not introduce any new security implications.
+
+
+How to Teach This
+=================
+
+This PEP does require that new users create and use a virtual environment to get started with using Python packages. This is however a best practice, as demonstrated by the section on "basics of how to install Python packages" in the Python Packaging User Guide, which explains how/what virtual environments are before discussing using ``pip``.[1]_
```
CAM-Gerlach, line 72:
```suggestion
This PEP does require that new users create and use a virtual environment to get started with using Python packages. This is, however, a best practice, as demonstrated by the `section on the basics of how to install Python packages <https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments>`__ in the Python Packaging User Guide, which explains what virtual environments are before discussing using ``pip``.
```

* Fixed a couple grammar errors.

* The footnote syntax is broken.

* Further, It doesn't really make sense to require to user to find and click through to a footnote, just to have to click on another (bare) link. Instead, just link the content directly. Per PEP 12:

  > Native reST hyperlink targets or inline hyperlinks in the text should be used in preference to footnotes for including URLs to online resources.

  See also #2130

* In fact, you could normally just do `` :ref:`sections-ref-target-label-on-pypug` `` to link it directly, but unfortunately it doesn't (yet) have one.

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+
+Reference Implementation
+========================
+
+There is no reference implementation for this PEP, however the proposed behaviour is already implemented in ``pip`` and can be achieved by setting ``PIP_REQUIRE_VENV`` environment variable to ``1`` (not setting it would result in the proposed opt-in behaviour).
```
CAM-Gerlach, line 78:
```suggestion
There is no reference implementation for this PEP. However, the proposed behaviour is already implemented in ``pip`` and can be activated by setting the ``PIP_REQUIRE_VENV`` environment variable to ``1``. (Leaving it unset results in the proposed opt-in behaviour of not requiring a virtual environment for installation.)
```

* Clarify what you mean by the last clause (I was very confused by it, and it took several re-reads to parse out)
* Fix several grammar and punctuation errors and issues
* Fix apparent typo ("achieved" -> "activated")

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Do not specify a name for the virtual environment directory
+-----------------------------------------------------------
+
+Using a consistent name for the virtual environment directory is important for a few reasons:
+
+1. It makes it easier for users to find the virtual environment directory, and to activate it.
+2. It removes a decision point for new users, since they do not need to decide on a name for the virtual environment directory.
+3. It creates a clear convention within the ecosystem, which makes it easier for users to find documentation.
+4. It ensures consistency across different tools, so that differences in the error message of different tools do not confuse users.
```
CAM-Gerlach, line 91:
```suggestion
* It makes it easier for users to find the virtual environment directory, and to activate it.
* It removes a decision point for new users, since they do not need to decide on a name for the virtual environment directory.
* It creates a clear convention within the ecosystem, which makes it easier for users to find documentation.
* It ensures consistency across different tools, so that differences in the error messages do not confuse users.
```

* Use unordered lists for items that don't have a clear and unambiguous sequential order (which also avoids it inevitably getting out of sync if items are added, moved or removed)
* Fix grammar error and avoid repetition

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Functionally, the directory name does not matter much as long as there is a single consistent suggestion.
+
+The name ``.venv`` was picked since it:
+
+1. does not conflict with any valid Python import name
+2. does not conflict ``venv`` module in the standard library
+3. has pre-existing usage in the Python community
+4. has support for auto-detection in common text editors
+5. can be typed without modifier keys on common keyboard layouts
```
CAM-Gerlach, line 104:
```suggestion
* does not conflict with any valid Python import name
* does not conflict ``venv`` module in the standard library
* has pre-existing usage in the Python community
* has support for auto-detection in common text editors
* can be typed without modifier keys on common keyboard layouts
```

Same here

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Do not couple tooling behaviour with a Python version
+-----------------------------------------------------
+
+This PEP creates a coupling between the behaviour of installers and the Python version, enforcing a behaviour change in installers based on a Python version.
+
+This is already a rollout mechanism being used for behaviour changes in the installation tooling. For example, ``pip`` on Python 3.11 will use ``importlib.metadata`` instead of ``pkg_resources`` for parsing/fetching package metadata, and ``sysconfig`` instead of ``distutils.sysconfig`` for getting the paths to unpack wheels into.
```
CAM-Gerlach, line 111:
```suggestion
This is already a rollout mechanism being used for behaviour changes in the installation tooling. For example, ``pip`` on Python 3.11 will use :mod:`importlib.metadata` instead of :mod:`!pkg_resources` for fetching/parsing package metadata, and :mod:`sysconfig` instead of :mod:`!distutils.sysconfig` for getting the paths to unpack wheels into.
```

* Use appropriate markup/cross references
* Flip words ("fetching" -> "parsing") to logical order (fetching comes before parsing, no?)

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+The difference with those cases is that they're supposed to be largely transparent to end users. This PEP is proposing a behaviour change that is not transparent to end users, and requires them to take action.
+
+The primary benefit of this is that it allows for redistributors to adapt their tooling in time for the new Python version and provides a clear and consistent point for change across the ecosystem. It also puts a clear deadline on when the default behaviour will consistently require a virtual environment by default (once Python 3.12 goes end-of-life).
+
+The primary issue with this approach is that it enforces a behaviour change on users when they upgrade to a new Python version, which can hamper the adoption of a new Python version. However, this is a migration/upgrade for existing users and it is a common expectation that _some_ changes to be needed for migration/upgrades.
```
CAM-Gerlach, line 117:
```suggestion
The primary issue with this approach is that it enforces a behaviour change on users when they upgrade to a new Python version, which can hamper the adoption of a new Python version. However, this is a migration/upgrade for existing users and it is a common expectation that *some* changes will be needed for migration/upgrades.
```

* Fix syntax mistake
* Fix grammar error

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+None.
+
+
+Footnotes
+=========
+
+.. [1] https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
+
+Copyright
```
CAM-Gerlach, line 133:
```suggestion
None.


Copyright
```

Turned into a proper link; see above.

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+It also creates a scope for mistakes, since users must remember to activate the virtual environment before running an installer like ``pip``. On certain Linux distributions, this can result in the installer modifying files that are owned by the operating system (this is partially mitigated by :pep:`668` for distributions that opt-in to marking their environments accordingly).
+
+This PEP proposes that the default behaviour of installers like ``pip`` should be to require a virtual environment to be active. This would make it easier for new users to get started with Python, and would reduce the scope for mistakes.
+
+
+Specification
+=============
+
+When a user runs an installer without an active virtual environment, the installer MUST print an error message and exit with a non-zero exit code. The error message MUST inform the user that a virtual environment is required, SHOULD provide shell-specific instructions on how to create and activate a virtual environment named ``.venv``, and SHOULD provide a link to a documentation page that explains how to create and activate a virtual environment.
```
CAM-Gerlach, line 41: What counts as a "virtual environment"? Only an environment created with `venv`/`virtualenv`? What about Conda environments? Or PEP 582 environments? Or other types of isolated environments? The PEP should either explicitly define this, or link (`:term:`) to a precise and authoritative definition, e.g. in the PyPA glossary.

CAM-Gerlach, line 41: Also, what counts as "an installer"? Obviously, `pip` counts (as its mentioned by name), and I'm _assuming_ Hatch, PDM and Poetry count as well when used in that capacity, while `apt`, `dnf`, `brew` and `choco` don't, though that isn't explicitly stated.

If you limit it to "Python-specific" installer, what about tools like shiv, pex, etc.? And what about, of course, Conda? The PEP should provide a precise definition of that term and examples of what tools would and would not quality (similar to PEP 668 for what qualifies as an externally-managed environment).

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+
+The installer SHOULD also provide a mechanism to disable this behaviour, allowing a user to opt-in to running the tool outside of a virtual environment. The installer MAY include this in the error message that it prints, and MUST include it in the documentation page it links to.
+
+If the installer includes a link to a documentation page in the error message described above, the documentation page SHOULD be written with a focus on new users. It should explain what virtual environments are, why they are required, and how to create and activate a virtual environment. It SHOULD include instructions for the most common shells and platforms.
+
+Installers MAY choose to implement this default behaviour on any Python versions, but MUST implement it on Python 3.13 or newer.
```
CAM-Gerlach, line 47: IMO, it would be best to be explicit here that you mean the Python environment the installer is being instructed to install into (as I'm _assuming_ you mean, vs. e.g. the runtime environment of the installer, or the default/system Python.

CAM-Gerlach (commented):

* [x] File created from the [latest PEP template](https://github.com/python/peps/blob/main/pep-0012/pep-NNNN.rst?plain=1)
* [x] PEP has next available number, with filename (``pep-NNNN.rst``) and ``PEP`` header set accordingly
* [x] ``PEP``, ``Title``, ``Author``, ``Status`` (``Draft``), ``Type`` and ``Created`` headers filled out correctly
* [ ] ``Sponsor``, ``PEP-Delegate``, ``Topic``, ``Requires`` and ``Replaces`` headers completed if appropriate
* [x] ~~Code is well-formatted (PEP 7/PEP 8) and is in [code blocks, with the right lexer names](https://peps.python.org/pep-0012/#literal-blocks) if non-Python~~ N/A
* [ ] Authors/sponsor added to ``.github/CODEOWNERS`` for the PEP


    * [x] Motivation
    * [x] Rationale
    * [x] Specification
    * [x] Backwards Compatibility
    * [x] Security Implications
    * [x] How to Teach This
    * [x] Reference Implementation
    * [x] Rejected Ideas
    * [x] Open Issues
* [ ] ~~``Python-Version`` set to valid (pre-beta) future Python version~~ Not needed for Packaging PEP that doesn't touch core Python
* [x] Any project stated in the PEP as supporting/endorsing/benefiting from it confirms such
* [x] Right before or after initial merging, [PEP discussion thread](https://peps.python.org/pep-0001/#discussing-a-pep) created and linked to in ``Discussions-To`` and ``Post-History``


## General comments

* When you're just referring to pip, the name of the package installer (as opposed to running the `pip` _command_ on a CLI), I would suggest omitting the literal formatting, to be semantically correct, consistent with other names and pip's own docs, and avoid visual noise in the source and rendered output.
* Just FYI, we've enabled Intersphinx against both the Python.org docs and packaging.python.org, so you can use all the normal directives you'd expect (`:ref:`, `:term:`, `:module:`, etc) to link against things like the PyPA glossary (which is rather outdated though I have some proposed updates in the works; for my working draft of PEP 639 I basically had to redefine most of the relevant terms myself with updated definitions).
* I've included a handful of non-suggestion comments regarding the PEP clearly and precisely expressing what it intends to in terms of terminology and rationale, that may stray a bit into the gray area of content/subject matter expertise. If you feel any of them stray too far, I can defer them here and bring them up on the thread instead.

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
+PEP: 704
+Title: Require virtual environments by default for installers
```
CAM-Gerlach, line 2:
```suggestion
Title: Require virtual environments by default for package installers
```

@ned-deily suggested this to avoid any confusion with the installers for Python itself, and I agree FWIW

pep-0704.rst
dstufft: There's roughly no security win, since the ability to execute arbitrary code means you more or less have all the powers the user does.

h-vetinari: > There's roughly no security win, since the ability to execute arbitrary code means you more or less have all the powers the user does.

Yeah, but it _does_ avoid the "admin escape hatch" that several users have learned to take if they cannot install something with their regular user, and it turns out they're trying to install to the system python. It's great if this is not done by default anymore, but it still helps even if you're accidentally installing something as root user to not mess up your system python, but put them in a separate env (though when you're installing something as root, obviously any potential security improvements aren't relevant anymore).

pradyunsg: > Yeah, but it does avoid the "admin escape hatch" that several users have learned to take if they cannot install something with their regular user, and it turns out they're trying to install to the system python. 

We've closed that with EXTERNALLY-MANAGED (PEP 668).

pep-0704.rst
CAM-Gerlach: (But this could, of course, be deferred to a followup at your discretion, as it is really more of a content issue rather than editorial matter).

pep-0704.rst
```diff
@@ -0,0 +1,137 @@
...
+Author: Pradyun Gedam <pradyunsg@gmail.com>
+Sponsor: Brett Cannon <brett@python.org>
```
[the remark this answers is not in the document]
brettcannon: Yep, I agree.

pep-0704.rst
pradyunsg: I prefer the current style.

pep-0704.rst
pradyunsg: I prefer the numbered list here, since this is an ordered list to me.

pep-0704.rst
pradyunsg: I don't think the markup references are particularly valuable -- is that a blocking concern?


It's primarily used for parsing rather than fetching; hence the order.

pradyunsg:
Are there any blocking concerns here, or can this be merged?

JelleZijlstra:
Did Brett (as the sponsor) approve the text of the PEP?

Also, you should add the PEP to CODEOWNERS (it's in CAM's checklist above; there's also an entry unchecked about headers but I am not sure what the issue is).

pradyunsg:
Yes: https://github.com/python/peps/pull/2964#discussion_r1082037627

CODEOWNERS added.

JelleZijlstra:
OK, then I think we can merge but I'll wait for a bit for @CAM-Gerlach to confirm.

pep-0704.rst
```diff
@@ -0,0 +1,170 @@
...
+1. does not conflict with any valid Python import name
+2. does not conflict ``venv`` module in the standard library
+3. has pre-existing usage in the Python community
+4. has support for auto-detection in common text editors
+5. can be typed without modifier keys on common keyboard layouts
+
+Do not couple tooling behaviour with a Python version
+-----------------------------------------------------
+
+This PEP creates a coupling between the behaviour of installers and the Python version, enforcing a behaviour change in installers based on a Python version.
```
CAM-Gerlach, line 147: This repeats an nearly the same phrase twice (the first of which _also_ repeats the section title on the line above), which is seems thoroughly redundant. I therefore suggest either one or the other:

```suggestion
This PEP creates a coupling between the behaviour of installers and the Python version.
```

or 

```suggestion
This PEP enforces a behaviour change in installers based on a Python version.
```

pep-0704.rst
```diff
@@ -0,0 +1,170 @@
...
+
+The difference with those cases is that they're supposed to be largely transparent to end users. This PEP is proposing a behaviour change that is not transparent to end users, and requires them to take action.
+
+The primary benefit of this is that it allows for redistributors to adapt their tooling in time for the new Python version and provides a clear and consistent point for change across the ecosystem. It also puts a clear deadline on when the default behaviour will consistently require a virtual environment by default (once Python 3.12 goes end-of-life).
+
+The primary issue with this approach is that it enforces a behaviour change on users when they upgrade to a new Python version, which can hamper the adoption of a new Python version. However, this is a migration/upgrade for existing users and it is a common expectation that *some* changes will be needed for migration/upgrades.
```
CAM-Gerlach, line 155:
```suggestion
The primary issue with this approach is that it enforces a behaviour change on users when they upgrade to a new Python version, which can hamper the adoption of a new Python version. However, this is a migration/upgrade for existing users, and it is a common expectation that *some* changes will be needed for migration/upgrades.
```

Somehow missed this missing comma before, sorry

pep-0704.rst
```diff
@@ -0,0 +1,170 @@
...
+
+The author of this PEP believes that enforcing the benefits of applying this consistently throughout the ecosystem with a deadline, outweigh the drawbacks of enforcing a best-practice on users when they upgrade.
```
CAM-Gerlach, line 157:
```suggestion
The author of this PEP believes that the benefits of applying this consistently throughout the ecosystem with a deadline outweigh the drawbacks of enforcing a best-practice on users when they upgrade.
```

Remove spurious word and comma

CAM-Gerlach:
> there's also an entry unchecked about headers but I am not sure what the issue is).

That was for confirming that the sponsor approved*; I saw Brett's comment and marked it as resolved, but forgot to update the checklist accordingly, sorry. Everything's checked now :heavy_check_mark: ; there's just a handful of remaining small typos, textual fixes and syntax tweaks that either got missed from suggestions that were outdated/resolved for other reasons, on newly added material, or that I missed the first time around.

\* In earlier versions of the checklist, I had a separate item for that, but Petr suggested merging it with the others which I went along with, which I'm now having second thoughts about...but I don't want to drag #2956 out even more since I already revised it several times.

CAM-Gerlach (commented):
A handful of remaining small typos, textual fixes and syntax tweaks that either got missed from suggestions that were outdated/resolved for other reasons, on newly added material, or that I missed the first time around. Besides that, and in terms of the checklist, LGTM, thanks.

pep-0704.rst
CAM-Gerlach: Looks like there were all missed when this suggestion was outdated by another change; I meant to update and re-make it, but didn't end up getting to it in time, sorry.

pep-0704.rst
CAM-Gerlach: Fair enough

pep-0704.rst
CAM-Gerlach: Okay, up to you :)

.github/CODEOWNERS
```diff
@@ -584,6 +584,7 @@ pep-0700.rst  @pfmoore
 pep-0701.rst  @pablogsal @isidentical @lysnikolaou
 pep-0702.rst  @jellezijlstra
 pep-0703.rst  @ambv
+pep-0703.rst  @brettcannon @pradyunsg
```
zakj, line 587:
```suggestion
pep-0704.rst  @brettcannon @pradyunsg
```

CAM-Gerlach: Oops, missed that somehow

pep-0704.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Motivation
+==========
+
+Python virtual environments are an essential part of the development workflow for Python. However, they require extra effort since they are an opt-in feature, and requires users to either:
+
+- take explicit steps to activate/deactivate a virtual environment
+- use :samp:`{<path-to-venv>}/{<bin-path>}/{<executable>}` to run files
+
+For new users, things will seemingly work correctly when virtual environments are not used until they don't. Further, activating a virtual environment uses slightly different syntax and mechanisms on different platforms. This complicates the introduction to virtual environments since now information and context about how/why they are useful needs to explained to justify adding additional steps to the workflow.
```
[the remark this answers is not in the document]
pradyunsg:
```suggestion
For new users, things will seemingly work correctly when virtual environments are not used -— until they don't. Further, activating a virtual environment uses slightly different syntax and mechanisms on different platforms. This complicates the introduction to virtual environments, since now information and context about how/why they are useful needs to explained to justify adding additional steps to the workflow.
```

pep-0704.rst
pradyunsg: This feels wrong to me as a reader; so skipping it. :)

pep-0704.rst
```diff
@@ -0,0 +1,170 @@
...
+
+It also creates a scope for mistakes, since users need to remember to activate the virtual environment before running an installer like ``pip`` or configure those installers to error out. On certain Linux distributions, forgetting to do so can result in the installer modifying files that are owned by the operating system (which is partially mitigated by :pep:`668` for distributions that opt-in to marking their environments accordingly).
```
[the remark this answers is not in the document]
pradyunsg: This feels wrong to me as a reader; so skipping it. :)

pep-0704.rst
CAM-Gerlach: As above—it sounds wrong to me as a technical writer and copyeditor, and the same grammar applies—these are two independent clauses separated by a coordinating conjunction, so they need a comma. Particularly so since the second clause repeats a phrase ("migration/upgrades") from the previous clause, and the distance created by the comma also helps make this repetition feel less jarring and more intentional.

Also, fixed a plural/singular agreement error.

That being said, same as above applies—it's your PEP :)

pradyunsg: Let's leave this as-is.

potiuk (commented):
Just a comment about that one. I really like that we have an opt-out now rather than opt-in (but it's good we have an explicit opt-out as well that is useful in many cases - for example in most container use-cases. Those are specific use-cases and having and opt-out possible is more than enough for those cases - especially for legacy use cases.

Also having a default convention for the venv usage is great as part of this PEP. This will make a number of use-cases simpler and less decisions to make and having implicit steps for activation of ~/.venv environment is a good one too.

pradyunsg:
Please don't comment on the PR, there's a discuss.python.org thread for this PEP which is where discussions should happen.

potiuk:
> Please don't comment on the PR, there's a discuss.python.org thread for this PEP which is where discussions should happen.

sure. I was not sure what the etiquette is but if you say so, I am happy to add this comment to the discourse.

PR #2966, opened 2023-01-17
Adds Rationale section in the PEP

kushaldas:
Adds `Rationale` section to the PEP

### DIFF
--- a/pep-0582.rst
+++ b/pep-0582.rst
@@ -45,6 +45,17 @@ being introduced to Python (or programming for the first time).
   as in a previous terminal with virtual environment activated.
 
 
+Rationale
+=========
+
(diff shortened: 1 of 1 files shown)


brettcannon:
@kushaldas FYI we try to make sure the PEP number is in the title so people know what the PR is about. Otherwise this sort of title is hard to understand when you're looking at your GitHub notifications. 😉

CAM-Gerlach:
(Like this)

Also, while it is already mentioned in our existing PR template, our new templates/checklists in #2956 should make it more obvious and harder to miss—while it doesn't currently, we could add it as a checkbox item for all PRs, which would take up less space while being harder to miss then the present template.

PR #3006, opened 2023-02-05
PEP 702: Runtime warnings by default

JelleZijlstra:


hauntsaninja (approved):
These changes look good!

If we continue to be concerned about the fact that overloads won't actually raise the warning, we could include some "type checkers may require that category=None be explicitly passed to deprecated overloads" language.

pep-0702.rst
```diff
@@ -175,10 +176,24 @@ Here is how type checkers should handle usage of this library:
...
-object. The value of the attribute is the message passed to the decorator.
-The decorator returns the original object. Notably, it does not issue a runtime
-:exc:`DeprecationWarning`.
+The ``@deprecated`` parameter takes two keyword-only arguments:
+
+* ``category``: A warning class. Defaults to :exc:`DeprecationWarning`. If this
+  is set to ``None``, no warning is issued at runtime and the decorator returns
+  the original object, except for setting the ``__deprecated__`` attribute (see below).
+* ``stacklevel``: The number of stack frames to skip when emitting the warning.
+  Defaults to 2.
```
Jackenmen, line 185: Maybe it would be better to define this as the number of levels above whatever function wrapping behavior that needs to happen in the implementation? This way it wouldn't tie the implementation to the public API and probably be more obvious to the user. The implementation would just add the appropriate number to `stacklevel` when calling `warnings.warn()` in the wrapper.

I suppose that then, a new default of `1` would always be equivalent to saying that you want the deprecation warning to be shown at the function/constructor's call site.

JelleZijlstra: That's a good point, let me change that.

Relatedly I had to think about when it would even make sense to use this argument, and the one scenario I can think of is when there is another decorator above `@deprecated` that adds an additional wrapper.

Jackenmen: The kind of usage that would benefit from `stacklevel` which I can think of would hurt from it being type checked (i.e. internal functions checking for argument deprecation in multiple functions where you wouldn't want type checker to report the call site as type error). Actually, thinking about it... It feels like it's a given that `stacklevel != 1` *conflicts* with proposed type checker behavior as the type checkers are supposed to report the issue on the call site, regardless of the `stacklevel` (and a different type checker behavior wouldn't make sense).

I can't shake off the feeling that it's wrong *not to* have this argument even though the above argument strongly suggests that it's not needed. Might be worth bringing it up in the discussion thread?

JelleZijlstra: My stacked decorators scenario sounds pretty reasonable to me, what do you think?

Jackenmen: Oh yeah, I guess it does. I somehow forgot you brought that scenario up while thinking about my own :smile: It may not be often that you *need* to use `@deprecated` decorator below some other *wrapping* (so `@overload` excluded) decorator but it probably will come up in some scenarios. I'm convinced :)

JelleZijlstra: Added this, along with the language @hauntsaninja asked for.

Jackenmen: Seems good to me, I think this also made the description of `stacklevel` clearer for potential implementers since it now explains the behavior of the argument.

Somewhat off-topic for the PEP text but in relation to decorator stacking, this might be a realistic scenario for `stacklevel != 1` if you'll later need some test case for the implementation:
```
class Foo:
    @functools.cached_property
    @deprecated("bar will go away soon", stacklevel=2)
    def bar(self):
        return 42
```

PR #3027, opened 2023-02-24
PEP 8: Allow both one and two spaces after period

erlend-aasland:
Authored-by: Hugo van Kemenade <hugovk@users.noreply.github.com>

### DIFF
--- a/pep-0008.txt
+++ b/pep-0008.txt
@@ -705,8 +705,8 @@ letter (never alter the case of identifiers!).
 Block comments generally consist of one or more paragraphs built out of
 complete sentences, with each sentence ending in a period.
 
-You should use one or two spaces after a sentence-ending period in multi-
-sentence comments, except after the final sentence.
+You should use one or two spaces after a sentence-ending period in
+multi-sentence comments, except after the final sentence.
 
 Ensure that your comments are clear and easily understandable to other 
 speakers of the language you are writing in.
--- a/pep-0008.txt
+++ b/pep-0008.txt
@@ -705,7 +705,7 @@ letter (never alter the case of identifiers!).
 Block comments generally consist of one or more paragraphs built out of
 complete sentences, with each sentence ending in a period.
 
-You should use two spaces after a sentence-ending period in multi-
+You should use one or two spaces after a sentence-ending period in multi-
 sentence comments, except after the final sentence.
 
 Ensure that your comments are clear and easily understandable to other 


pradyunsg:
FWIW you're missing a `Co-` for the commit body for GitHub to recognise it. Otherwise, `git commit --amend --author [same text]` would achieve making @hugovk the author and you could drop/change the commit body.

erlend-aasland:
> FWIW you're missing a `Co-` for the commit body for GitHub to recognise it. Otherwise, `git commit --amend --author [same text]` would achieve making @hugovk the author and you could drop/change the commit body.

Right, I deliberately removed the `Co-`; I'll amend it and force-push.

pradyunsg (approved):

gvanrossum (approved):

pep-0008.txt
```diff
@@ -705,7 +705,7 @@ letter (never alter the case of identifiers!).
 Block comments generally consist of one or more paragraphs built out of
 complete sentences, with each sentence ending in a period.
 
-You should use two spaces after a sentence-ending period in multi-
+You should use one or two spaces after a sentence-ending period in multi-
 sentence comments, except after the final sentence.
```
JelleZijlstra, line 709:
```suggestion
You should use one or two spaces after a sentence-ending period in
multi-sentence comments, except after the final sentence.
```
Noticed this doesn't get rendered correctly:
Screen Shot 2023-02-24 at 4 27 37 PM

gvanrossum:
Can we make it self-referential? :-)

hugovk:
Re: https://github.com/python/peps/issues/3024#issuecomment-1443985658

Do we also want to add something to say it "doesn't require consistency (not even within a paragraph)"?

---

Whilst editing this file, let's also remove the redundant Version and Last-Modified headers, and the Local Variables footer, as we've gradually been doing elsewhere.

CAM-Gerlach:
> Do we also want to add something to say it "doesn't require consistency (not even within a paragraph)"?

Has that been enough of a problem (for code comments at least, which is what this mention is about and not the reST document that triggered this) that it's worth an explicit sentence stating so? Consistency is *nice*, just not required or worth changes to existing text that isn't already rewritten.

CAM-Gerlach (approved):
LGTM, thanks @erlend-aasland and @hugovk !

hugovk (approved):

JelleZijlstra (approved):

PR #3030, opened 2023-02-28
PEP 709: Remove obsolete incompatibility subsection

carljm:
This "incompatibility" does not actually exist in the current reference implementation, I just hadn't realized that it had been fixed.


carljm:
The commit https://github.com/python/cpython/pull/101441/commits/463c740b7199650fca15870b4ce583652970dd7f on the reference implementation https://github.com/python/cpython/pull/101441 confirms that this section is no longer correct.

CAM-Gerlach (approved):
LGTM, thanks @carljm

PR #3041, opened 2023-03-07
Cleanup PEP-697 before PSC acceptance.

gpshead:
Clarify in the abstract that this does not apply to all types, things like tuple and int are excluded.  But could be done in the future.

Fix ``basesize`` vs ``basicsize`` field name typos.  oops! 😁 

Remove the Open Issues section and mention that using a `PyType_Spec` flag instead of a negative basicsize was rejected and offer an explanation why.  _(ultimately either way would work with similar impact)_


encukou:
Thank you!

PR #3086, opened 2023-04-03
PEP 101: PRs to the release management PEP shouldn't ping Guido and Barry

ambv:
While they are the original authors, this is a living document that changes whenever the release process is tweaked.

----
:books: Documentation preview :books:: https://pep-previews--3086.org.readthedocs.build/


ned-deily (approved):

Yhg1s:
Should @zooba be on the list as well?

PR #3118, opened 2023-04-22
PEP 649: distinguish annotations requested as strings vs objects

carljm:
Larry suggested to make this a separate PR from https://github.com/python/peps/pull/3113 since it might require more discussion.

----
:books: Documentation preview :books:: https://pep-previews--3118.org.readthedocs.build/


larryhastings:
And again--thanks, Carl!

PR #3133, opened 2023-05-02
PEP 696: Use list for ParamSpec default [Fix]

cdce8p:
Support for tuples for `ParamSpec` default values was removed in #3052.

From the PEP
```rst
``ParamSpec`` defaults are defined using the same syntax as
``TypeVar`` \ s but use a ``list`` of types or an ellipsis
literal "``...``" or another in-scope ``ParamSpec`` (see :ref:`696-scoping-rules`).
```

/CC: @Gobot1234

----
:books: Documentation preview :books:: https://pep-previews--3133.org.readthedocs.build/

### DIFF
--- a/pep-0696.rst
+++ b/pep-0696.rst
@@ -150,7 +150,7 @@ literal "``...``" or another in-scope ``ParamSpec`` (see :ref:`696-scoping-rules
 
 .. code-block:: py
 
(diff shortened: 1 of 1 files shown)


Gobot1234 (approved):
Thanks for spotting this

PR #3207, opened 2023-07-17
PEP 257: update code example to Python 3

francisohara24:
Replace `print` keyword with call to `print()` function in "Handling Docstring Indentation" example.

----
:books: Documentation preview :books:: https://pep-previews--3207.org.readthedocs.build/


AA-Turner:
We don't tend to update PEPs, which are fixed change proposals.

A

francisohara24:
Alright.

PR #3224, opened 2023-07-31
PEP 323: Resolve unreferenced footnotes

AA-Turner:
Footnotes [1] was never referenced, and the reference to [2] was removed in 36e7a69 -- this commit removes the markup syntax.

A

----
:books: Documentation preview :books:: https://pep-previews--3224.org.readthedocs.build/

### DIFF
--- a/pep-0323.txt
+++ b/pep-0323.txt
@@ -475,23 +475,14 @@ specific iterator type at the same time.
 References
 ==========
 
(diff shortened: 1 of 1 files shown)


hugovk (approved):
We could convert them to plain bullet points, and maybe even remove [2] altogether.

PR #3225, opened 2023-07-31
PEP 326: Remore unreferenced footnotes and renumber

AA-Turner:
Footnote [10] does not exist.  Footnotes [5], [8], and [11] are unreferenced. Commit 778f8da removed the reference to [5] (then numbered [6]) and added [8] with no reference. [11] was added in 64497e6 with no reference. Without an apparent need for these three, they are removed. The remaining footnotes and their text references are re-numbered.

----
:books: Documentation preview :books:: https://pep-previews--3225.org.readthedocs.build/


terryjreedy:
Thanks for the cleanup.

PR #3230, opened 2023-08-01
PEP 343: Resolve unreferenced footnotes

AA-Turner:
Footnote [12] was added in f860cb2 with no reference -- I can't find a working archive.org link (all say that the patch is private), so I've removed [12] and re-numbered the remaining footnotes.

A

----
:books: Documentation preview :books:: https://pep-previews--3230.org.readthedocs.build/


gvanrossum (approved):
Not worth shedding tears over.

PR #3235, opened 2023-08-01
PEP 419: Resolve unreferenced footnotes

AA-Turner:
Footnote [4] referenced the initial email to Python-Dev, and [5] referenced https://github.com/python/cpython/issues/58935, both of which seem severable from the main text.

A

----
:books: Documentation preview :books:: https://pep-previews--3235.org.readthedocs.build/


hugovk:
Let's keep them either as bullet points or plain `[x]` things.

PR #3237, opened 2023-08-01
PEP 448: Resolve unreferenced footnotes

AA-Turner:
Footnote [3] has been present since the beginning (originally as [2]) -- whilst unreferened, it provides useful background to the PEP. I've removed the syntax, but I'm on the fence about just removing it.

A


hugovk (approved):
Thanks, I think it's fine to leave it.

Oh, and whilst editing, we can also remove the old `Version` and `Last-Modified` headers as we've done before (e.g. https://github.com/python/peps/pull/2816).

PR #3242, opened 2023-08-02
PEP 502: Resolve unreferenced footnotes

AA-Turner:
The references to footnotes [10] and [11] were removed in 1bd2f20; I've added references in what seems a sensible place.

A

----
:books: Documentation preview :books:: https://pep-previews--3242.org.readthedocs.build/


hugovk (approved):
And let's also remove the old `Version` and `Last-Modified` headers as we've done before (e.g. https://github.com/python/peps/pull/2816).

PR #3254, opened 2023-08-02
PEP 654: Resolve unreferenced footnotes

AA-Turner:
Footnote [13] was added in cef697a, which also added the stanza on "Implementing a better task spawning API..." to which I've added a reference to [13].

A

----
:books: Documentation preview :books:: https://pep-previews--3254.org.readthedocs.build/


iritkatriel (approved):
Thank you.

PR #3270, opened 2023-08-05
Update GitHub Actions workflows

AA-Turner:
Misc GHA workflow updates:

- Standardise formatting
- Add permissions section
- Add concurrency section
- Set ``FORCE_COLOR``
- Ensure ``pip`` is cached

A

----
:books: Documentation preview :books:: https://pep-previews--3270.org.readthedocs.build/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,19 @@
-name: Lint
+name: Lint PEPs
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  FORCE_COLOR: 1
 
 jobs:
   pre-commit:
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -1,20 +1,32 @@
...
-      - name: Check out repo
-        uses: actions/checkout@v3
-
+      - uses: actions/checkout@v3
       - name: Set up Python 3
         uses: actions/setup-python@v4
         with:
-          python-version: '3.x'
+          python-version: "3.x"
+          cache: pip
```
hugovk, line 29: I don't think this is necessary for this workflow, we're not installing anything directly with pip, and the pre-commit action takes care of its own caching. If anything, the extra overhead makes it a little slower :)

AA-Turner: Ahh, fair enough -- I had thought that the ``pre-commit`` action used pip, I didn't realise it had its own cache.

hugovk: Yep, the action does use pip. Hmm, looking closer, even with this pip cache and its own cache it's still downloading files (and they're all wheels):

https://github.com/python/peps/actions/runs/5771517466/job/15645551533#step:4:26

What it's actually caching, is the pre-commit cache: `~/.cache/pre-commit` which is the git repos for the hooks.

Well, the result is the same, `cache: pip` doesn't cache anything...

PR #3278, opened 2023-08-07
PEP 723: Fix swapped references

JelleZijlstra:
See https://github.com/python/peps/pull/3277/files#r1286078922

cc @ofek

----
:books: Documentation preview :books:: https://pep-previews--3278.org.readthedocs.build/


AA-Turner (approved):
Thanks for spotting!

A

PR #3289, opened 2023-08-14
Revert the link for 'Putting Metaclasses to Work' (Forman & Danforth)

AA-Turner:


gvanrossum:
Maybe wait until the link actually stops working? Though I still don’t understand how one is supposed to find that link (other than a Google search, which could be seeded by this very PEP).

*If* we decide to remove this we should remove the other one in PEP 551.

shailshouryya:
I didn't look through the archive.org link properly until after typing this out, at which point I realized I was talking about an entirely different scenario (archiving the link to the Amazon product page of the book, as opposed to the actual book) and what I typed out does not apply. However, I don't want to delete the entire thing, and maybe someone finds something useful out of it for some other purpose.

Also some context for others who might need to do some digging (the lawsuit is mentioned the linked issue #3259 above, but adding references):
- https://www.theregister.com/2023/08/14/internet_archive_sued_by_music_labels/
- https://www.musicbusinessworldwide.com/sony-universal-among-companies-suing-internet-archive-over-alleged-massive-ongoing-violation-of-music-copyrights/

therefore, the archive.org link will likely go away if the publishers have their way.

<br>
<hr>

### The original thing I typed (which is not relevant, see above):

(Adding here in case this is useful)

Since https://www.amazon.com/gp/product/0201433052 is just a link to buy the book, the link should hopefully stay unaffected by the lawsuit (I think?). However, if the entire organization goes down, then the link will probably disappear - at that point, though, the entire site will probably go down at that point, along with all the other links there.

It looks like there are some alternatives to archive.org, and I was going through https://www.searchenginejournal.com/wayback-machine-5-alternatives-to-try/477224/ to see what they're like, but I haven't really tried them yet.

Short summary based off some (very light) research:
- [Memento](https://mementoweb.org/about/) seems to require an extension, but the extension doesn't seem to have that many downloads and figuring out the [database](https://wiki.mementodatabase.com/index.php?title=Support:Getting_started_with_Memento) might take a while
- archive.today (https://archive.ph/ and apparently https://archive.is/) seems to be used to archive pages with dynamic content (that archive.org usually does not capture) - I've seen links to these domains before, but the site seems a bit finnicky at times and apparently people without ad-blockers see ads (too much?)
- [WebCite](https://webcitation.org/) is "currently not accepting archiving requests. The archival state/snapshots of websites that have been archived with WebCite in the past can still be accessed and cited." Searching for https://www.amazon.com/gp/product/0201433052 returns no results (try the URL at https://webcitation.org/query.php)
- GitHub (mentioned in article) is not relevant here
- Country-Specific Web Archives
  - The UK Web Archive (https://www.webarchive.org.uk/ukwa/).
  - The Library of Congress Web Archive (https://www.loc.gov/web-archives/collections/).
  - Web Archive Singapore (https://eresources.nlb.gov.sg/webarchives/landing-page).
  - The Croatian Web Archive (https://haw.nsk.hr/en).


There's also [this list](https://www.softwaretestinghelp.com/top-wayback-machine-alternatives), but going through that will take even more time (but maybe we can get feedback if anyone has tried any of them before).
</strike>

hugovk:
Resolved conflict, and marking as draft until the link stops working.

hugovk:
> Maybe wait until the link actually stops working?

15 months on, and the archive.org link still works. Shall we close this? 

(If we did change the link, rather than Amazon I'd prefer to use the author's, publisher's or a library catalogue page.)

gvanrossum (commented):
Why not keep both links? We can take the archives.org link down when archives.org removes it.

juniorcastro0152-hub:
From d2f1ec19caabbfdc7d9437d18bc4411413342fee Mon Sep 17 00:00:00 2001
From: Adam Turner <9087854+aa-turner@users.noreply.github.com>
Date: Tue, 15 Aug 2023 00:29:26 +0100
Subject: [PATCH] Revert the link for 'Putting Metaclasses to Work' (Forman &
 Danforth)

---
 pep-3119.txt | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/pep-3119.txt b/pep-3119.txt
index fda56b40427..5c39e7666d8 100644
--- a/pep-3119.txt
+++ b/pep-3119.txt
@@ -862,7 +862,7 @@ References
 
 .. [8] Putting Metaclasses to Work: A New Dimension in Object-Oriented
    Programming, by Ira R. Forman and Scott H. Danforth
-   (https://archive.org/details/PuttingMetaclassesToWork)
+   (https://www.amazon.com/gp/product/0201433052)
 
 [9] Partial order, in Wikipedia
 \   (https://en.wikipedia.org/wiki/Partial_order)

gvanrossum:
I'm closing this now. The archive.org link still works, the book is still out of print, Amazon currently lists a few copies for sale from used bookstores, but no e-book*. People who insist on buying the book should be able to track it down easily wherever used books are sold using the title and author. (My own copy is not for sale.)

I asked an LLM about the book publishers' lawsuit, and the conclusion seems to be that archive.org can keep publishing scans of works for which no e-book version exists.

PR #3297, opened 2023-08-21
Finalize PEP 723

ofek:
----
:books: Documentation preview :books:: https://pep-previews--3297.org.readthedocs.build/

### DIFF
--- a/pep-0723.rst
+++ b/pep-0723.rst
@@ -406,9 +406,9 @@ would live as single-file scripts:
   executable or script runner.
 
 This PEP argues that reusing our TOML-based metadata format is the best for
-each category of user and that the block comment is only approachable for
-those who have familiarity with ``requirements.txt``, which represents a
-small subset of users.
+each category of user and that the requirements-like block comment is only
+approachable for those who have familiarity with ``requirements.txt``, which
+represents a small subset of users.
 
 * For the average person automating a task or the data scientist, they are
   already starting with zero context and are unlikely to be familiar with
--- a/pep-0723.rst
+++ b/pep-0723.rst
@@ -63,7 +63,7 @@ __ https://rust-lang.github.io/rfcs/3424-cargo-script.html
 
 The format is intended to bridge the gap between different types of users
 of Python. Users will benefit from seamless interoperability with tools that
-are that already work with TOML.
+already work with TOML.
 
 One of the central themes we discovered from the recent
 `packaging survey <https://discuss.python.org/t/22420>`__ is that users have
--- a/pep-0723.rst
+++ b/pep-0723.rst
@@ -306,25 +306,29 @@ is TOML and resembles a ``pyproject.toml`` file.
 The two allowed tables are ``[run]`` and ``[tool]``. The ``[run]`` table may
 contain the following fields:
 
-+---------------------+------------------------------------------------------+------------------------------------------------------+
-| Fields              | Description                                          | Tool behavior                                        |
-+=====================+======================================================+======================================================+
-| ``dependencies``    | A list of strings that specifies the runtime         | Tools will error if the specified dependencies       |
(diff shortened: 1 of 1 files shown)


ofek:
This is ready for review

pep-0723.rst
```diff
@@ -284,27 +261,20 @@ installed in the user's environment.
...
-Since the format chosen is the same as the official metadata standard, we can
-have a page that describes how to embed the metadata in scripts and to learn
-about metadata itself direct users to the living document that describes
-`project metadata <pyproject metadata_>`_.
-
-We will document that the name and version fields in the ``[project]`` table
-may be elided for simplicity. Additionally, we will have guidance explaining
-that single-file scripts cannot (yet) be built into a wheel via standard means.
+We can have a page that describes how to embed the metadata in scripts. This
+will come naturally to many users because the allowed fields happen to match
```
AA-Turner, line 276: I hestitate to define what will come naturally -- Python's users globally are a heterogeneous set, who we don't know! Many uses in e.g. data science are by people who aren't expereienced programmers, etc.

I think the first line here "We can have..." isn't needed, as any proposal implemented will inevitably have documentation.

PEP-616 and PEP-655 have good how to teach this sections -- I think the major challenges with teaching here are the novel comment syntax (metadata blocks), and then the meaning assigned to those. This PEP should probably reccomend adding a descriptive guide to the PPUG at the least, probably also mentioning that the structed blocks only have meaning when used with tools that understand them, and that the core Python interpreter doesn't use them at all.

pep-0723.rst
```diff
@@ -494,7 +413,8 @@ Additionally, since the original block comment alternative format went against
 the recommendation of :pep:`8` and as a result linters and IDE auto-formatters
 that respected the recommendation would
 `fail by default <https://discuss.python.org/t/29905/247>`__, the final
-proposal uses standard comments starting with a single ``#`` character.
+proposal uses standard comments starting with a single ``#`` character without
+any obvious start nor end sequence.
```
AA-Turner, line 417: Perhaps briefly outline what the original block comment alternative format was?

pep-0723.rst
```diff
@@ -101,50 +94,61 @@ The following are some of the use cases that this PEP wishes to support:
...
+first two characters of each line if the second character is a space, otherwise
+just the first character (which means the line consists of only a single
+``#``).
 
-This is the canonical regular expression that MUST be used to parse the
+Tools reading embedded metadata MAY respect the standard Python encoding
+declaration. If they choose not to do so, they MUST process the file as UTF-8.
+
+The first type of metadata block is named ``pyproject`` which represents a
+subset of content that one would see in a ``pyproject.toml`` file.
```
AA-Turner, line 115: I think "subset" is slightly misleading here, as the "[run]" table isn't valid in pyproject.toml, and as far as I can tell this PEP doesn't seek to make it legal -- so it is *similar* data, but not a strict subset.

AA-Turner (commented):
Thanks for this update!

I'm a little unclear on if the ``TYPE`` blocks are user-definable or not -- e.g. if my script had ``# /// BLAH``, do you expect tools to produce an error, optionally produce errors, ignore it, etc?

A

ofek:
@AA-Turner I think I addressed all of your comments!

ofek:
all finished

ofek:
That table took me so long to format... I didn't know there was an easier way 😅

AA-Turner:
See [tables](https://www.docutils.org/docs/ref/rst/restructuredtext.html#tables), [``.. list-table::``](https://www.docutils.org/docs/ref/rst/directives.html#list-table), and [``.. csv-table::``](https://www.docutils.org/docs/ref/rst/directives.html#csv-table).

AA-Turner (approved):

PR #3484, opened 2023-10-13
PEP 729, 731: Remove Python-Version

JelleZijlstra:
As these PEPs don't propose changes to Python, the Python-Version
header doesn't really make sense. I had assumed the header was required
but it's not.

----
:books: Documentation preview :books:: https://pep-previews--3484.org.readthedocs.build/


gvanrossum (approved):
I had the same thought. :-)

hugovk (approved):

PR #3543, opened 2023-11-23
PEP-8105: update with nominations and voter roll

ewdurbin:
----
:books: Documentation preview :books:: https://pep-previews--3543.org.readthedocs.build/


hugovk (approved):
🗳️

PR #3591, opened 2023-12-15
PEP 709: Mark as Final

hugovk:
* [x] Pull request title in appropriate format (``PEP 123: Mark Final``)
* [ ] Canonical docs/spec linked with a ``canonical-doc`` directive (or ``canonical-pypa-spec``, for packaging PEPs)

PEP 709 was implemented in Python 3.12, we should mark it as Final so it's no longer listed under [Accepted PEPs (accepted; may not be implemented yet)](https://peps.python.org/#accepted-peps-accepted-may-not-be-implemented-yet).

Should this have a `.. canonical-doc::` banner at the top to link to canonical docs?

For example, compare https://peps.python.org/pep-0634/

All I could find in the CPython docs is the [What's New entry](https://docs.python.org/3/whatsnew/3.12.html#pep-709-comprehension-inlining), but perhaps as an internal change, "canonical docs" doesn't apply in this instance?

----
📚 Documentation preview 📚: https://pep-previews--3591.org.readthedocs.build/

### DIFF
--- a/peps/pep-0709.rst
+++ b/peps/pep-0709.rst
@@ -3,9 +3,8 @@ Title: Inlined comprehensions
 Author: Carl Meyer <carl@oddbird.net>
 Sponsor: Guido van Rossum <guido@python.org>
 Discussions-To: https://discuss.python.org/t/pep-709-inlined-comprehensions/24240
-Status: Accepted
+Status: Final
 Type: Standards Track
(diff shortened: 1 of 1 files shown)


carljm:
I don't think canonical docs applies, since it's an internal implementation detail with only a few user-visible edge cases.

Thanks for doing this update!

PR #3617, opened 2024-01-09
PEP 11: Add note about interpreting current Windows release

zooba:
Just proposed text right now for discussion, but as the text of the policy requires a bit of interpretation (since Microsoft changes their lifecycle and set of releases frequently), I think it's worth writing our current intention in here.

----
📚 Documentation preview 📚: https://pep-previews--3617.org.readthedocs.build/


ambv:
This looks sensible to me. We can always modify it forward if somebody does scream.

PR #3634, opened 2024-01-26
PEP 739: don't use raw text in specification titles

FFY00:
Just a minor visual improvement, as with raw text it becomes quite difficult to distinguish the smaller subsections.

----
📚 Documentation preview 📚: https://pep-previews--3634.org.readthedocs.build/


FFY00:
Yes, I will submit a PR adding the new headers, and addressing some of the feedback, soon.

PR #3691, opened 2024-02-25
PEP 394: Fix formatting

hugovk:
And remove redundant fields, and update BPO link.

<table>
<tr>
<th>Before
<th>After

<tr>
<td>
image

image

image

<td>
image

image

image
</table>

----
📚 Documentation preview 📚: https://pep-previews--3691.org.readthedocs.build/


encukou:
Thanks!

PR #3694, opened 2024-02-28
PEP 730: Mark as accepted

JelleZijlstra:
----
📚 Documentation preview 📚: https://pep-previews--3694.org.readthedocs.build/


ned-deily (commented):
Thanks!

PR #3704, opened 2024-03-07
 PEP 737: Remove Python changes

vstinner:
PR related to the Steering Council decision: https://discuss.python.org/t/pep-737-unify-type-name-formatting/39872/50

cc @warsaw @gpshead

----
📚 Documentation preview 📚: https://pep-previews--3704.org.readthedocs.build/


vstinner:
I added quotes of @warsaw and @Yhg1s to better explain the rejection of the Python API changes.

PR #3707, opened 2024-03-08
Infra: Add OG metadata to improve previews of shared links

hugovk:
I think because this is a custom theme, for some reason the sphinxext-opengraph we use for CPython docs and the devguide doesn't work here. 

But it's easy enough to add the Open Graph metadata to the template header, omitting the longer `<meta name="description"  content="...">` that sphinxext-opengraph generates.

OG docs: https://ogp.me/

# Demo

On Discord, with a temporary build using `<meta property="og:image" content="https://docs.python.org/3/_static/og-image.png">`, because `https://peps.python.org/_static/og-image.png` isn't deployed yet.

image

# Currently

image

----
📚 Documentation preview 📚: https://pep-previews--3707.org.readthedocs.build/


JacobCoffee (approved):
this would be very nice!

PR #3753, opened 2024-04-14
PEP 560: Add Typing label

Hnasar:
I was looking for this PEP on 
https://peps.python.org/topic/typing/ but because
it's missing the label, the sphinx pep_zero_generator skipped this one.

----
📚 Documentation preview 📚: https://pep-previews--3753.org.readthedocs.build/


JelleZijlstra:
Thanks, I missed this one when I added the topic because it didn't make type system changes.

PR #3803, opened 2024-05-29
PEP 667: Fix word duplication

kerel-fs:
Fixed an editorial issue in the accepted PEP 667: redundancy in the phrase "will be always be visible.".

----
📚 Documentation preview 📚: https://pep-previews--3803.org.readthedocs.build/


AlexWaygood:
Thanks!

PR #3807, opened 2024-06-01
PEP 581: Mark as Final

hugovk:
* [ ] Final implementation has been merged (including tests and docs)
* [ ] PEP matches the final implementation
* [x] Pull request title in appropriate format (``PEP 123: Mark Final``)

Helps https://github.com/python/peps/issues/2872, which says:

> [PEP-581](https://peps.python.org/581) _Using GitHub Issues for CPython_: Since the PEP's content is tightly focused on the rationale for choosing GitHub rather than the current, active process for using it and is thus now of historical rather than procedural value, it should be marked Final (rather than Active) like other transition-oriented Process PEPs have.

Also remove some redundant fields.

----
📚 Documentation preview 📚: https://pep-previews--3807.org.readthedocs.build/

### DIFF
--- a/peps/pep-0581.rst
+++ b/peps/pep-0581.rst
@@ -1,13 +1,10 @@
 PEP: 581
 Title: Using GitHub Issues for CPython
-Version: $Revision$
-Last-Modified: $Date$
 Author: Mariatta <mariatta@python.org>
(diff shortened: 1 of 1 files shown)


Mariatta (approved):
Thanks

PR #3822, opened 2024-06-05
Add note to PEP-746 about  raising NotImplementedError

adriangb:
----
📚 Documentation preview 📚: https://pep-previews--3822.org.readthedocs.build/

### DIFF
--- a/peps/pep-0746.rst
+++ b/peps/pep-0746.rst
@@ -90,6 +90,8 @@ Implementations may be generic and may use overloads that return ``Literal[True]
 to indicate if the metadata is valid for the given type.
 
 Implementations may raise a NotImplementedError if they cannot determine if the metadata is valid for the given type.
+Tools calling ``__supports_type__`` at runtime should catch this exception and treat it as if ``__supports_type__``
+was not present; they should not take this as an indication that the metadata is invalid for the type.
 
 Tools that use the metadata at runtime may choose to ignore the implementation of ``__supports_type__``; this PEP does not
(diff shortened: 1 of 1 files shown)


peps/pep-0746.rst
```diff
@@ -89,6 +89,13 @@ For example, to support a generic ``Gt`` metadata, one might write::
 Implementations may be generic and may use overloads that return ``Literal[True]`` or ``Literal[False]``
 to indicate if the metadata is valid for the given type.
 
+Implementations may raise a NotImplementedError if they cannot determine if the metadata is valid for the given type.
+Tools calling ``__supports_type__`` at runtime should catch this exception and treat it as if ``__supports_type__``
+was not present; they should not take this as an indication that the metadata is invalid for the type.
+
+Tools that use the metadata at runtime may choose to ignore the implementation of ``__supports_type__``; this PEP does not
+specify how the method should be used at runtime, only that it may be available for use.
```
adriangb, line 97: I don't love this wording. Especially in contrast to the paragraph above which is almost certainly making statements about how `__supports_type__` should be used and interpreted at runtime. The point I am trying to make is that if tools like Pydantic or msgspec encounter `Annotated[int, Gt(0)]` they are allowed to check the constraint in Rust / C and are not obligated to call into Python just to run `__supports_type__`. I'm open to suggestions on better wording.

JelleZijlstra: I think this wording is fine, actually.

JelleZijlstra (approved):

PR #3833, opened 2024-06-09
PEP 749: use `annotationlib` as the proposed module name throughout

qexat:
The name `annotationlib` is mentioned through, but the examples use `annotations` instead. This commit fixes that.

----
📚 Documentation preview 📚: https://pep-previews--3833.org.readthedocs.build/


AlexWaygood (approved):
Thanks!

PR #3883, opened 2024-07-31
PEP 751: address comments

brettcannon:
----
📚 Documentation preview 📚: https://pep-previews--3883.org.readthedocs.build/

### DIFF
--- a/peps/pep-0751.rst
+++ b/peps/pep-0751.rst
@@ -655,6 +655,9 @@ Per-file locking
 Per-package locking
 ===================
 
+Some values for ``packages.files.origin`` left out to make creating this
+example more easily as it was done by hand.
+
 .. code-block:: toml
 
   version = '1.0'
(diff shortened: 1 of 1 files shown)


peps/pep-0751.rst
```diff
@@ -177,6 +177,14 @@ All keys listed below are required unless otherwise noted. If two keys are
 mutually exclusive to one another, then one of the keys is required while the
 other is disallowed.
 
+Key order in tables -- including the top-level table -- SHOULD be emitted by
+lockers in the order they specified in this PEP when applicable and not
+otherwise specified to minimize noise in diffs. If the keys are not explicitly
```
JelleZijlstra, line 182: This sentence doesn't make sense to me and I'm not sure what you're trying to say. I think "they" should be removed, and I'm not sure you need "and not otherwise specified".

PR #3950, opened 2024-09-07
PEP 752: minor touchups

ofek:
----
📚 Documentation preview 📚: https://pep-previews--3950.org.readthedocs.build/

### DIFF
--- a/peps/pep-0752.rst
+++ b/peps/pep-0752.rst
@@ -133,8 +133,9 @@ Parent Namespace
     A namespace's parent refers to the namespace without the trailing
     hyphenated component e.g. the parent of ``foo-bar`` is ``foo``.
 Child Namespace
-    A namespace's child refers to the namespace with an additional trailing
-    hyphenated component e.g. the child of ``foo`` is ``foo-bar``.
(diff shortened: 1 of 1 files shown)


peps/pep-0752.rst
```diff
@@ -133,8 +133,9 @@ Parent Namespace
     A namespace's parent refers to the namespace without the trailing
     hyphenated component e.g. the parent of ``foo-bar`` is ``foo``.
 Child Namespace
-    A namespace's child refers to the namespace with an additional trailing
-    hyphenated component e.g. the child of ``foo`` is ``foo-bar``.
+    A namespace's child refers to the namespace with additional trailing
+    hyphenated components e.g. ``foo-bar`` is a valid child of ``foo`` as is
+    ``foo-bar-baz``.
```
ofek, line 138: To match the API:

```rst
* ``children``: This is an array of any child namespaces. For example, if the
  namespace is ``foo`` and there are active grants for ``foo-bar`` and
  ``foo-bar-baz`` then this would be ``["foo-bar", "foo-bar-baz"]``.
```

warsaw (approved):

PR #3952, opened 2024-09-07
PEP 752, 755: update history

ofek:
----
📚 Documentation preview 📚: https://pep-previews--3952.org.readthedocs.build/

### DIFF
--- a/peps/pep-0752.rst
+++ b/peps/pep-0752.rst
@@ -3,7 +3,7 @@ Title: Implicit namespaces for package repositories
 Author: Ofek Lev <ofekmeister@gmail.com>
 Sponsor: Barry Warsaw <barry@python.org>
 PEP-Delegate: Dustin Ingram <di@python.org>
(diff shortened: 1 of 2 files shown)


peps/pep-0752.rst
```diff
@@ -9,6 +9,7 @@ Type: Standards Track
 Topic: Packaging
 Created: 13-Aug-2024
 Post-History: `18-Aug-2024 <https://discuss.python.org/t/61227>`__,
+              `07-Sep-2024 <https://discuss.python.org/t/63191>`__,
```
hugovk, line 12: That's the PEP 755 thread, here's the new 752 one:
```suggestion
              `07-Sep-2024 <https://discuss.python.org/t/63192>`__,
```

And let's update `Discussions-To` to point to this new one as well.

ofek: Good catch, thank you!

hugovk (approved):
Thanks!

PR #3955, opened 2024-09-12
PEP 752: Address feedback, round 4

ofek:
1. Expressed the use case for open/restricted namespaces in the motivation section.
2. Made it clear that the current automatic protection against typosquatting used by PyPI is insufficient.
3. Added examples of attacks to the appendix.
4. Fixed versioning of the JSON API.
5. Rejected the idea of encouraging projects to maintain their own package repositories.
6. Rejected the idea of fixed top-level prefixes used for reservations.
7. Rejected the idea of using DNS.
8. Expressed in the rationale that using PEP 541 requests in addition to this proposal would minimize risk to users.
9. Added a community buy-in section enumerating feedback from relevant projects and communities.

----
📚 Documentation preview 📚: https://pep-previews--3955.org.readthedocs.build/

### DIFF
--- a/peps/pep-0752.rst
+++ b/peps/pep-0752.rst
@@ -24,29 +24,42 @@ Motivation
 ==========
 
 The current ecosystem lacks a way for projects with many packages to signal a
-verified pattern of ownership. Some examples:
+verified pattern of ownership. Such projects fall into two categories.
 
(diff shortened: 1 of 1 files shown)


ofek:
This is ready for review!

warsaw (approved):

PR #3966, opened 2024-09-16
PEP 757: Make PyLong_DigitArray.obj private

vstinner:
----
📚 Documentation preview 📚: https://pep-previews--3966.org.readthedocs.build/


vstinner:
cc @steve-s

PR #4001, opened 2024-09-27
PEP 740: Update api-version

woodruffw:
Both PEP 740 and [PEP 708](https://peps.python.org/pep-0708/) stipulated API version 1.2 as the next JSON API version, but PEP 708 was finished first! So this just bumps to the next unused API version.

CC @di 

----
📚 Documentation preview 📚: https://pep-previews--4001.org.readthedocs.build/


di (approved):
Nice catch!

PR #4004, opened 2024-09-27
PEP 747: fix rst syntax

carljm:
This causes a mis-rendering on https://peps.python.org/pep-0747/

----
📚 Documentation preview 📚: https://pep-previews--4004.org.readthedocs.build/


Mariatta (approved):
I checked the build and confirmed that it fixed the rendering.

PR #4013, opened 2024-10-01
PEP 758: Small updates to PEG syntax and formatting

pablogsal:


pablogsal:
Of course, you only notice the errors after merging :facepalm:

AA-Turner (approved):

PR #4019, opened 2024-10-05
PEP 746: Rename __supports_type__ to __supports_annotated_base__

adriangb:
----
📚 Documentation preview 📚: https://pep-previews--4019.org.readthedocs.build/


AA-Turner (approved):
In general we should keep [line length](https://peps.python.org/pep-0012/#general) to ~80, I appreciate though that the name is now 28 characters long...

A

PR #4047, opened 2024-10-11
PEP 735: Fix a minor mistake

sirosen:
This line of text was not removed when the relevant behaviors were
removed from the PEP. It is therefore inaccurate.

---

I am not sure what this change qualifies as, based on the list below:

* Change is either:
    * [ ] To a Draft PEP

It's an Accepted PEP, but this is definitely an oversight.
I assume, since it's a packaging PEP, that if we get Paul's approval as the PEP delegate, it becomes the middle case?

----
📚 Documentation preview 📚: https://pep-previews--4047.org.readthedocs.build/

### DIFF
--- a/peps/pep-0735.rst
+++ b/peps/pep-0735.rst
@@ -197,15 +197,13 @@ Dependency Groups are not Hidden Extras
 ---------------------------------------
 
 Dependency Groups are very similar to extras which go unpublished.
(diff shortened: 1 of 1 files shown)


hugovk:
cc @pfmoore

pfmoore (approved):

PR #4058, opened 2024-10-16
Meta: Allow PEP authors to use a name other than their real name

JelleZijlstra:
Fixes #4052.

### DIFF
--- a/peps/pep-0001.rst
+++ b/peps/pep-0001.rst
@@ -606,9 +606,9 @@ optional and are described below.  All other headers are required.
 
     PEP: <pep number>
     Title: <pep title>
-    Author: <list of authors' real names and optionally, email addrs>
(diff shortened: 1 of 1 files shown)


warsaw (commented):
Should there be a requirement to provide handles for the author on DPO, Discord, and/or GitHub?  People should be able to use the name they want but other folks in the community should also be able to unambiguously contact them on the various discussion platforms.

JelleZijlstra:
I feel that would add too much process in a place where we haven't really felt a need for it. My proposed text already says authors should use the name they use in discussions surrounding the PEP. Adding something more complicated feels onerous.

willingc (approved):
Thanks! We can always amend down the road if we see an issue with the ability to contact someone.

savannahostrowski (approved):

PR #4095, opened 2024-10-30
PEP 753: link to PyPA spec

woodruffw:
The equivalent PyPI page is now live at https://packaging.python.org/en/latest/specifications/well-known-project-urls/, so this adds the appropriate banner.

----
📚 Documentation preview 📚: https://pep-previews--4095.org.readthedocs.build/


hugovk:
Thanks!

PR #4111, opened 2024-11-06
PEP 757: Use union

vstinner:
----
📚 Documentation preview 📚: https://pep-previews--4111.org.readthedocs.build/

### DIFF
--- a/peps/pep-0757.rst
+++ b/peps/pep-0757.rst
@@ -107,37 +107,51 @@ Export API
 
    Export of a Python :class:`int` object.
 
-   There are two cases:
+   .. c:member:: int8_t kind
 
-   * If :c:member:`digits` is ``NULL``, only use the :c:member:`value` member.
-     Calling :c:func:`PyLong_FreeExport` is optional in this case.
-   * If :c:member:`digits` is not ``NULL``, use :c:member:`negative`,
-     :c:member:`ndigits` and :c:member:`digits` members.
-     Calling :c:func:`PyLong_FreeExport` is mandatory in this case.
+      There are two cases:
 
-   .. c:member:: int64_t value
+      * If :c:member:`kind` is :c:macro:`PyLongExport_VALUE_KIND`, use
+        :c:member:`data.value`. Calling :c:func:`PyLong_FreeExport` is
+        optional in this case.
+      * If :c:member:`kind` is :c:macro:`PyLongExport_ARRAY_KIND`, use
+        :c:member:`data.array`. Calling :c:func:`PyLong_FreeExport` is
+        mandatory in this case.
 
-      The native integer value of the exported :class:`int` object.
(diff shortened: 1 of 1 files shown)


vstinner:
@skirpichev: What do you think of using an union in the export?

skirpichev (commented):
With current proposal we don't gain anything.  E.g. some contract like "this will not fail for integers".

I'm neutral on keeping the ``kind`` field wrt using the return value of PyLong_Export() to encode the union type (yet the struct, probably, will be bigger), as https://github.com/python/peps/pull/4026/commits/c0a725b64d9e6253309fe41c0afafed0bafff870 does.

vstinner:
I'm not changing any contract. It's just the structure which changes to make it more obvious which members should be used depending on the kind.

peps/pep-0757.rst
```diff
@@ -270,16 +284,18 @@ Code::
...
 
-        if (long_export.digits) {
-            mpz_import(z, long_export.ndigits, int_digits_order, int_digit_size,
-                       int_endianness, int_nails, long_export.digits);
-            if (long_export.negative) {
+        if (long_export.kind == PyLongExport_ARRAY_KIND) {
+            mpz_import(z, long_export.data.array.ndigits,
+                       int_digits_order, int_digit_size, int_endianness,
+                       int_nails, long_export.data.array.digits);
+            if (long_export.data.array.negative) {
```
NameNoQuality, line 291: #4113

NameNoQuality (approved):
#4113

peps/pep-0757.rst
```diff
@@ -107,37 +107,51 @@ Export API
...
-      The native integer value of the exported :class:`int` object.
-      Only valid if :c:member:`digits` is ``NULL``.
+      .. c:namespace:: NULL
 
-   .. c:member:: uint8_t negative
+      ========================================  =====
+      Kind                                      Value
+      ========================================  =====
+      .. c:macro:: PyLongExport_VALUE_KIND      ``1``
+      .. c:macro:: PyLongExport_ARRAY_KIND      ``2``
```
skirpichev, line 127: Should we expose these details?  Lets just keep symbolic constants.

skirpichev (commented):
> I'm not changing any contract.

Bad news.  I would buy such change if one allows us to skip error checking.  Or simplify current implementation (e.g. if only one export type will be used).

> It's just the structure which changes to make it more obvious which members should be used depending on the kind.

I think that current PEP shape is simple enough.  Lets not rewrite it again ;)  I doubt that in future we will need something else than current two versions of export.

With union the structure will be, probably, more compact.  This is only benefit of this change, for me.  Not sure if it worth.

Count me neutral on this.

vstinner:
Ok, I close this PR.

PR #4133, opened 2024-11-21
PEP 672: Mention Unicode TR 55

encukou:
Unicode Technical Report 55 “Unicode Source Code Handling” was published in January and contains lots of relevant information and general recommendations.

---

This is a change to an Informational PEP, which should be OK per [PEP 1](https://peps.python.org/pep-0001/#pep-maintenance) but doesn't fit the template:

* Change is either:
    * [ ] To a Draft PEP

----
📚 Documentation preview 📚: https://pep-previews--4133.org.readthedocs.build/


hugovk (approved):
Thank you! 

We can also delete the redundant `Content-Type: text/x-rst` from the header and the emacs stuff from the end of the file.

PR #4158, opened 2024-12-06
PEP 768: Safe external debugger interface for CPython

pablogsal:

PR #4160, opened 2024-12-09
PEP 768: Add pablogsal to `CODEOWNERS`

hugovk:
Follow on from https://github.com/python/peps/pull/4158.


pablogsal:
Oh sorry I totally missed this :(

PR #4168, opened 2024-12-13
PEP 757: Set Status to Final

vstinner:
----
📚 Documentation preview 📚: https://pep-previews--4168.org.readthedocs.build/


vstinner:
Oh, there is already https://github.com/python/peps/pull/4167

PR #4172, opened 2024-12-15
PEP 695: Fix runtime behavior example of generic classes

Viicos:
At runtime, `Generic` is added as a base class:

```python
class A[T](Base):
    pass

A.__bases__
#> (<class '__main__.Base'>, <class 'typing.Generic'>)
```

Note: I know the PEP is a historical document and this one is marked as final. However, this seems to be the only place where the runtime behavior is documented.

* Change is either:
    * [ ] To a Draft PEP
    * [x] To an Accepted or Final PEP, with Steering Council approval

----
📚 Documentation preview 📚: https://pep-previews--4172.org.readthedocs.build/

### DIFF
--- a/peps/pep-0695.rst
+++ b/peps/pep-0695.rst
@@ -1033,7 +1033,7 @@ Equivalent to::
 
    def695 __generic_parameters_of_C():
        T = typing.TypeVar('T')
-       class C(Base):
+       class C(Base, typing.Generic[T]):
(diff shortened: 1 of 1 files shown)


Viicos:
My bad, seems like this is documented [here](https://docs.python.org/3/reference/compound_stmts.html#generic-classes). Feel free to close if necessary.

PR #4173, opened 2024-12-15
PEP 768: Expand the security considerations section

pablogsal:


pablogsal:
CC: @zooba

pablogsal:
CC: @godlygeek @ivonastojanovic

peps/pep-0768.rst
```diff
@@ -294,6 +294,24 @@ An example usage of the API would look like:
...
 
+Configuration API
+-----------------
+
+To allow redistributors, system administrators, or users to disable this
+mechanism, several methods will be provided to control the behavior of the
+interpreter:
+
+A new environment ``PYTHON_DISABLE_REMOTE_DEBUGGING`` environment variable will
+be provided to control the behaviour at runtime. If set to any value, the
```
godlygeek, line 305: Does the empty string count as a value?

godlygeek, line 305: We should clarify when this variable will be checked - is it at process startup, or is it when the eval breaker bit is checked, right before the code would be executed?

If it's checked at startup, we should expose the flag via the debug offsets, so that debuggers can check this and print out a useful error message explaining why they won't work, instead of believing that they have attached and then never having their script run.

peps/pep-0768.rst
```diff
@@ -294,6 +294,24 @@ An example usage of the API would look like:
...
+interpreter will ignore any attempts to attach a debugger using this mechanism.
+
+This environment variable will be added together with a new ``-X disable-remote-debugging``
+flag to the Python interpreter to allow users to disable this feature at runtime.
+
+Additionally a new  ``--without-remote-debugging`` flag will be added to the
+``configure`` script to allow redistributors to build Python without support for
+remote debugging if they so desire.
```
godlygeek, line 313: We should additionally expose a flag indicating that remote debugging is disabled via the debug offsets, so that debuggers can check this and print out a useful error message explaining why they won't work, instead of believing that they have attached and then never having their script run. That will provide much nicer end user experience. All 3 methods (env var, `-X`, and `--without-remote-debugging`) should set that flag.

peps/pep-0768.rst
```diff
@@ -310,6 +328,10 @@ processes that can already write to arbitrary memory within your process and
 execute arbitrary code on the machine (in order to create the file containing
 the Python code to be executed).
 
+Furtermore, the execution of the code is done is gated by the interpreter's
```
godlygeek, line 331:
```suggestion
Furthermore, the execution of the code is gated by the interpreter's
```

peps/pep-0768.rst
```diff
@@ -310,6 +328,10 @@ processes that can already write to arbitrary memory within your process and
...
+audit hooks, which can be used to monitor and control the execution of the code
```
godlygeek, line 332:
```suggestion
audit hooks, which can be used to monitor or prevent the execution of the code
```

godlygeek (changes requested):

peps/pep-0768.rst
pablogsal: I would say yes. WDYT?

godlygeek: I'm ok either way, but I think the PEP should say "(including an empty string)" explicitly. There's a lot of people who don't realize that `unset FOO` and `FOO=` do different things...

pablogsal, line 305:
```suggestion
be provided to control the behaviour at runtime. If set to any value (including an empty string), the
```

peps/pep-0768.rst
```diff
@@ -294,6 +294,24 @@ An example usage of the API would look like:
...
+A new environment ``PYTHON_DISABLE_REMOTE_DEBUGGING`` environment variable will
+be provided to control the behaviour at runtime. If set to any value (including an empty string), the
+interpreter will ignore any attempts to attach a debugger using this mechanism.
+
+This environment variable will be added together with a new ``-X disable-remote-debugging``
+flag to the Python interpreter to allow users to disable this feature at runtime.
+
+Additionally a new  ``--without-remote-debugging`` flag will be added to the
+``configure`` script to allow redistributors to build Python without support for
+remote debugging if they so desire.
```
pablogsal, line 313:
```suggestion
Additionally a new  ``--without-remote-debugging`` flag will be added to the
``configure`` script to allow redistributors to build Python without support for
remote debugging if they so desire.

A new flag indicating the status of remote debugging will be made available via the debug offsets so tools can query if a remote process has disabled the feature. This way, tools can offer a useful error message explaining why they won't work, instead of believing that they have attached and then never having their script run. T
```

peps/pep-0768.rst
```diff
@@ -294,6 +294,26 @@ An example usage of the API would look like:
...
+
+A new flag indicating the status of remote debugging will be made available via the debug offsets so tools can query if a remote process has disabled the feature. This way, tools can offer a useful error message explaining why they won't work, instead of believing that they have attached and then never having their script run. T
```
godlygeek, line 315:
```suggestion
A new flag indicating the status of remote debugging will be made available via the debug offsets so tools can query if a remote process has disabled the feature. This way, tools can offer a useful error message explaining why they won't work, instead of believing that they have attached and then never having their script run.
```

godlygeek (approved):
LGTM, other than a little typo

pablogsal:
Let's wait for @zooba to take a look before landing

peps/pep-0768.rst
```diff
@@ -294,6 +294,26 @@ An example usage of the API would look like:
...
         print(f"Failed to execute code: {e}")
 
+Configuration API
+-----------------
+
+To allow redistributors, system administrators, or users to disable this
+mechanism, several methods will be provided to control the behavior of the
+interpreter:
+
+A new environment ``PYTHON_DISABLE_REMOTE_DEBUGGING`` environment variable will
```
godlygeek, line 304: Spotted another typo:
```suggestion
A new ``PYTHON_DISABLE_REMOTE_DEBUGGING`` environment variable will
```

peps/pep-0768.rst
```diff
@@ -294,6 +294,26 @@ An example usage of the API would look like:
...
+interpreter will ignore any attempts to attach a debugger using this mechanism.
+
+This environment variable will be added together with a new ``-X disable-remote-debugging``
+flag to the Python interpreter to allow users to disable this feature at runtime.
+
+Additionally a new  ``--without-remote-debugging`` flag will be added to the
+``configure`` script to allow redistributors to build Python without support for
+remote debugging if they so desire.
+
+A new flag indicating the status of remote debugging will be made available via the debug offsets so tools can query if a remote process has disabled the feature. This way, tools can offer a useful error message explaining why they won't work, instead of believing that they have attached and then never having their script run.
```
zooba, line 315:
```suggestion
A new flag indicating the status of remote debugging will be made available via
the debug offsets so tools can query if a remote process has disabled the
feature. This way, tools can offer a useful error message explaining why they
won't work, instead of believing that they have attached and then never having
their script run.
```

peps/pep-0768.rst
zooba, line 304: Depending on how you feel, I _slightly_ prefer `DISABLE_REMOTE_DEBUG` (and likewise for the other uses) over debugGING. Saves a few characters, and I always find ING-words to be weird feature names.

Probably just my own bias though. Might be an American intonation thing as well ("remote DEE-bug") - I mostly speak to Americans about debugging 😆

peps/pep-0768.rst
```diff
@@ -310,6 +330,10 @@ processes that can already write to arbitrary memory within your process and
 execute arbitrary code on the machine (in order to create the file containing
 the Python code to be executed).
```
zooba, line 331: This file is no longer created.

peps/pep-0768.rst
```diff
@@ -368,6 +392,33 @@ proposed interface.
...
+* For an external attacker, the ability to write to arbitrary memory in a
+  process is already a severe security issue. This interface does not introduce
+  any new attack surface, as the attacker would already have the ability to
+  execute arbitrary code in the process. This interface behaves in exactly
+  the same way as existing debuggers, and does not introduce any new additional
+  security risks.
+* For an attacker who has gained arbitrary memory write access to a process but
+  not arbitrary code execution, this interface does not provide a way to escalate
+  to arbitrary code execution as long as the attacker has not gained control over
+  additional machine resources external to the program.
```
zooba, line 407: A few too many negatives in here? I hope I've flipped them around correctly:

```suggestion
  not arbitrary code execution, this interface does not allow them to escalate.
  The ability to calculate and write to specific memory locations is required,
  which is not available without compromising other machine resources that
  are external to the Python process.
```

peps/pep-0768.rst
```diff
@@ -368,6 +392,33 @@ proposed interface.
...
+  to arbitrary code execution as long as the attacker has not gained control over
+  additional machine resources external to the program. 
+  
+Additionally, the fact that the code to be executed is gated by the interpreter's
+audit hooks means that the execution of the code can be monitored and controlled
+by system administrators. This means that even if the attacker has compromised the
+application **and the filesystem**, leveraging this interface for malicious
+purposes provides a very risky proposition for an attacker, as they risk
+exposing their actions to system administrators that could not only detect the
+attack but also take action to prevent it. Finally, is important to note that
```
zooba, line 415:
```suggestion
attack but also take action to prevent it.

Finally, is important to note that
```

Plus rewrapping... not doing all that in a PR comment, sorry 😆

zooba (commented):
LGTM

peps/pep-0768.rst
pablogsal: I suppose you wrote this before checking out the other PR but just to clarify: in the new model the debugger tool must write a file to be executed

PR #4174, opened 2024-12-15
PEP 472: Fix Sphinx reference warnings

hugovk:
For https://github.com/python/peps/issues/4087.

The `keyword-X` footnotes were added in the PEP's first commit and never referenced:

https://github.com/python/peps/commit/84d82f06f22a789b7ed3923e37552daef13eba14

I've turned them into bullet points, and replaced the broken https://article.gmane.org/gmane.comp.python.ideas/ links with corresponding https://mail.python.org/archives/list/python-ideas@python.org/ ones.

I also turned the one remaining footnote into an inline link.

----
📚 Documentation preview 📚: https://pep-previews--4174.org.readthedocs.build/

### DIFF
--- a/peps/pep-0472.rst
+++ b/peps/pep-0472.rst
@@ -1,12 +1,9 @@
 PEP: 472
 Title: Support for indexing with keyword arguments
-Version: $Revision$
-Last-Modified: $Date$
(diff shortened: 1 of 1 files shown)


willingc (approved):
Thanks @hugovk

PR #4267, opened 2025-02-11
texto_descodificado = texto_codificado.decode('windows-1250')     ret…

rflowerss:
…urn texto_descodificado

traductor.translate(texto, src='auto', dest='es')
    return traduccion.text

----
📚 Documentation preview 📚: https://pep-previews--4267.org.readthedocs.build/


Rosuav:
I don't think this was meant to be a pull request. Check what you were pushing to, you may have sent this to the wrong repository.

PR #4313, opened 2025-03-20
Some minor copy edits for PEP 751

pfmoore:
@brettcannon This is just a couple of typo fixes for PEP 751 that I spotted while reviewing.

There are only two that are significant:

1. I added a note that tools could choose to only support a subset of VCS implementations. I think this is basically self-evident (we wouldn't want to require everyone to support Subversion) but it's worth being explicit.
2. I removed an incomplete comment in "How to teach this". I couldn't work out what you wanted to say, but feel free to put back the complete version if you think it's worth having 🙂

The rest are just typos.

Now off to review the Discourse threads. I could be some time 🙂


brettcannon (approved):

brettcannon:
Thanks!

PR #4376, opened 2025-04-15
PEP 785: rename method, and a rejected idea

Zac-HD:
Brett Canon [observed](https://discuss.python.org/t/pep-785-new-methods-for-easier-handling-of-exceptiongroups/88244/2):

> I will say the name `flat_exceptions()` confused me initially as a “flat exception” isn’t a thing.

and on reflection I agreed that this made sense, renaming the method to `.leaf_exceptions()` to match terminology from PEP-654.  If this PEP is accepted, I plan to check for appropriate places to introduce the term to the exception-group tutorial and reference documentation.

Updating [my GitHub search](https://github.com/search?q=%2F%5C.leaf_exceptions%5C%28%2F+language%3APython&type=code) for existing uses of this name still finds zero hits.  Interestingly, https://github.com/python-trio/exceptiongroup/pull/13 proposed a `leaf_exceptions()` helper for a precursor to `ExceptionGroup`; I've added a footnote with some of that history.

### DIFF
--- a/peps/pep-0785.rst
+++ b/peps/pep-0785.rst
@@ -329,6 +329,20 @@ less magical and tempting to use in cases where it would not be appropriate.
 We could be argued around though, if others prefer this form.
 
 
+Preserve additional attributes
+------------------------------
+
+We decided against preserving the ``__cause__`` and ``__suppress_context__``
+attributes, because they are not changed by re-raising the exception, and we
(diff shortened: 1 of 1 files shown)


gpshead (approved):
probably also worth recording the `flat_exceptions` name in a Rejected Ideas section.

PR #4456, opened 2025-06-10
PEP 489: Fix syntax highlighting

hugovk:
Most of the code blocks are C, so set that as the document highlight, and use Python on blocks where needed.

Also add inline code formatting.

----
📚 Documentation preview 📚: https://pep-previews--4456.org.readthedocs.build/

### DIFF
--- a/peps/pep-0489.rst
+++ b/peps/pep-0489.rst
@@ -12,6 +12,10 @@ Python-Version: 3.5
 Post-History: 23-Aug-2013, 20-Feb-2015, 16-Apr-2015, 07-May-2015, 18-May-2015
 Resolution: https://mail.python.org/pipermail/python-dev/2015-May/140108.html
 
+.. canonical-doc:: :ref:`python:initializing-modules`.
+                   For Python 3.14+, see :ref:`py3.14:extension-modules`
+                   and :ref:`py3.14:pymoduledef`
+
 .. highlight:: c
 
 Abstract


encukou:
Umm... Does this historical document really need updating to current publication standards?

hugovk:
No, It's not absolutely needed, but would improve readability: the doc still exists and I was referred to it via https://github.com/python-pillow/Pillow/pull/8983. I guess I should have looked up the real docs instead :)

Anyway, I can ditch the second commit if it makes it easier to review, and/or add the canonical docs banner to the top (to https://docs.python.org/3/c-api/module.html#multi-phase-initialization ?) to make it clearer this is history?

AA-Turner:
I think this is fine to change, it's editorial/style rather than the content, which is the class of changes we tend to take more exception to.

A

encukou:
> add the canonical docs banner to the top

That's the thing to do :)

It would be nice to get https://github.com/python/cpython/pull/135126 in before that, and link to its new [`extension-modules.rst`](https://cpython-previews--135126.org.readthedocs.build/en/135126/c-api/extension-modules.html). Could I ask for a review there?

encukou:
Now merged. Would a canonical docs banner like this be appropriate?
```rst
.. canonical-doc:: :ref:`python:initializing-modules`.
                   For Python 3.14+, see :ref:`py3.14:extension-modules`
                   and :ref:`py3.14:pymoduledef`
```

hugovk:
Thanks, added!

I would like to keep at least the first commit, as while this PEP is indeed historical, it is still read, and we even link to it a dozen times from the CPython docs.

PR #4472, opened 2025-06-24
PEP 793: Two small updates

encukou:
- `PyType_GetModuleByToken` will return a strong reference, unlike `PyType_GetModuleByDef`
- ~~Remove incorrect note that the new API creates modules from a *spec*. (The note could be updated to say *slots*, but, it isn't that important.)~~

---

* Change is either:
    * [x] To a Draft PEP

----
📚 Documentation preview 📚: https://pep-previews--4472.org.readthedocs.build/


encukou:
Ugh, ignore the second change. I got confused by *def* vs. *spec* again.
(If only there was a PEP to phase of one of those out...)

PR #4487, opened 2025-07-08
PEP 628: Fix 'Pi is [(still)] wrong' links

hugovk:
Like https://github.com/python/cpython/issues/136122.

Vi Hart has removed intentionally removed their videos from YouTube: https://www.patreon.com/posts/vihart-channel-129842460

One is available on their Vimeo. The other is not, so use another YouTube copy. Also update other links to https.

----
📚 Documentation preview 📚: https://pep-previews--4487.org.readthedocs.build/


ncoghlan:
Thanks!

PR #4502, opened 2025-07-20
ensure `make venv` exits on error

### LINKED ISSUE #4501: Prevent `make venv` from saying it succeeded when it failed
As reported in https://github.com/python/devguide/issues/1607 and https://github.com/python/cpython/issues/136854, if an error happens when executing `make venv`, it will still be reported as successful ("The venv has been created in the ./venv directory").

The `set -e` option should be enabled to exit on error.

nachocab:
Fixes #4501 

----
📚 Documentation preview 📚: https://pep-previews--4502.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ venv:
 		echo "venv already exists."; \
 		echo "To recreate it, remove it first with \`make clean-venv'."; \
 	else \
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

PR #4507, opened 2025-07-23
PEP 743: Omit ``PyList_GET_ITEM``

crusaderky:
* Change is either:
    * [x] To a Draft PEP

`PyList_GET_ITEM` should be treated akin to `PyList_GetItem` when it comes to free-threading considerations.
This PR matches https://github.com/python/cpython/pull/137042.

----
📚 Documentation preview 📚: https://pep-previews--4507.org.readthedocs.build/


encukou:
I think PEP 743 should say that `Py_COMPAT_API_VERSION` works with limited API (which already excludes `PyList_GET_ITEM`). @vstinner, would that match your intention?

vstinner:
Right, `Py_COMPAT_API_VERSION` can be used with the limited C API.

PR #4511, opened 2025-07-25
PEP 799: clarifications

hugovk:
This bit wasn't too clear to me:

> The code that as the time of writing this is in `profile.sampling` module no longer exist and will only exist in the `profilers` package.

This PR suggests:

> The code that, at the time of writing, is in the `profile.sampling` module will be moved to the `profilers` package.

Is that the idea?

Alternatively:

> The code that, at the time of writing, is in the `profile.sampling` module will no longer exist and will instead only exist in the `profilers` package.

----
📚 Documentation preview 📚: https://pep-previews--4511.org.readthedocs.build/


pablogsal (approved):

pablogsal:
LGTM

PR #4524, opened 2025-08-06
PEP 719: 3.13.6 was released on 6 August 2025

AA-Turner:
https://discuss.python.org/t/101482

A


AA-Turner:
```
[INFO] Using pre-commit with uv 0.8.5 via pre-commit-uv 4.1.4
An unexpected error has occurred: FileNotFoundError: /home/runner/.local/bin/uv
```

Lint [failing](https://github.com/AA-Turner/peps/actions/runs/16789122617) due to [the same pre-commit-uv problem](https://github.com/tox-dev/pre-commit-uv/issues/70). Edit: Passing on the PR itself, so we should be fine.

PR #4543, opened 2025-08-11
PEP 792: mark as Final

woodruffw:
Follows https://github.com/pypi/warehouse/pull/18422, which completes the PEP's implementation by enabling it on the index APIs 🙂


AA-Turner (approved):
Congratualations William!

A

PR #4555, opened 2025-08-18
PEP 800: Note Python 3.12 behavior change, note reference impl

JelleZijlstra:
----
📚 Documentation preview 📚: https://pep-previews--4555.org.readthedocs.build/

### DIFF
--- a/peps/pep-0800.rst
+++ b/peps/pep-0800.rst
@@ -191,11 +191,12 @@ add support for alternative implementations (for example, branching on the value
 stubs could condition the presence of the ``@disjoint_base`` decorator on the implementation where necessary.
 
 Although the concept of "disjoint bases" (referred to as "solid bases") in the CPython implementation has existed
-for decades, the rules for deciding which classes are disjoint bases have occasionally changed over time.
+for decades, the rules for deciding which classes are disjoint bases have occasionally changed.
 Before Python 3.12, adding a ``__dict__`` or support for weakrefs relative to the base class could make a
 class a disjoint base. In practice, this often meant that Python-implemented classes inheriting from
-classes implemented in C, such as namedtuple classes, were themselves disjoint bases.
-This behavior was changed in Python 3.12 by [PR #96028](https://github.com/python/cpython/pull/96028).
+classes implemented in C, such as :func:`~collections.namedtuple` classes, were themselves disjoint bases.
+This behavior was changed in Python 3.12 by
+`python/cpython#96028 <https://github.com/python/cpython/pull/96028>`__.
 This PEP focuses on supporting the behavior of Python 3.12 and later, which is simpler and easier to understand.
 Type checkers may choose to implement a version of the pre-3.12 behavior if they wish, but doing this correctly
 requires information that is not currently available in the type system.
@@ -372,8 +373,8 @@ Reference Implementation
 
 The runtime implementation of the ``@disjoint_base`` decorator will be available in
 typing-extensions 4.15.0.
-[Mypy PR #19678](https://github.com/python/mypy/pull/19678) implements support for disjoint bases
-in mypy and in the stubtest tool.
+`python/mypy#19678 <https://github.com/python/mypy/pull/19678>`__
+implements support for disjoint bases in mypy and in the stubtest tool.
 
 Appendix
 ========


peps/pep-0800.rst
```diff
@@ -190,7 +190,17 @@ use CPython's behavior to determine when to use the ``@disjoint_base`` decorator
 add support for alternative implementations (for example, branching on the value of :py:data:`sys.implementation.name <sys.implementation>`),
 stubs could condition the presence of the ``@disjoint_base`` decorator on the implementation where necessary.
 
-Similarly, the exact set of classes that are disjoint bases at runtime may change in future versions of Python.
+Although the concept of "disjoint bases" (referred to as "solid bases") in the CPython implementation has existed
+for decades, the rules for deciding which classes are disjoint bases have occasionally changed over time.
```
AA-Turner: "occasionally changed over time" is saying the same thing twice?

Perhaps:

```suggestion
for decades, the rules for deciding which classes are disjoint bases have occasionally changed.
```

(You'll need to manually wrap as the suggestion is over the max line length)

peps/pep-0800.rst
```diff
@@ -190,7 +190,17 @@ use CPython's behavior to determine when to use the ``@disjoint_base`` decorator
...
+Before Python 3.12, adding a ``__dict__`` or support for weakrefs relative to the base class could make a
+class a disjoint base. In practice, this often meant that Python-implemented classes inheriting from
+classes implemented in C, such as namedtuple classes, were themselves disjoint bases.
```
AA-Turner:
```suggestion
classes implemented in C, such as :func:`~collections.namedtuple` classes, were themselves disjoint bases.
```

peps/pep-0800.rst
```diff
@@ -190,7 +190,17 @@ use CPython's behavior to determine when to use the ``@disjoint_base`` decorator
...
+This behavior was changed in Python 3.12 by [PR #96028](https://github.com/python/cpython/pull/96028).
```
AA-Turner: This is invalid syntax. Perhaps:

```suggestion
This behavior was changed in Python 3.12 by `python/cpython#96028 <https://github.com/python/cpython/pull/96028>`__.
```

(though again over 79 chars, you could instead split the target and reference)

peps/pep-0800.rst
```diff
@@ -360,7 +370,10 @@ explain to users why type checkers treat certain branches as unreachable.
 Reference Implementation
 ========================
 
-None yet.
+The runtime implementation of the ``@disjoint_base`` decorator will be available in
+typing-extensions 4.15.0.
+[Mypy PR #19678](https://github.com/python/mypy/pull/19678) implements support for disjoint bases
```
AA-Turner: This also won't render as a link:
```suggestion
`python/mypy#19678 <https://github.com/python/mypy/pull/19678>`__ implements support for disjoint bases
```

(also over line length limit)

peps/pep-0800.rst
AlexWaygood: neither `collections.namedtuple` nor `typing.NamedTuple` is a class at runtime, FWIW! `typing.NamedTuple` is (perhaps questionably) documented as a class at https://docs.python.org/3/library/typing.html#typing.NamedTuple, but `collections.namedtuple` is documented as a function: https://docs.python.org/3/library/collections.html#collections.namedtuple

peps/pep-0800.rst
JelleZijlstra: Oops I was writing markdown :)

peps/pep-0800.rst
AA-Turner: Oops! I've fixed the suggestion.

The purist in me would probably document `typing.NamedTuple` as a function, but users would likely be surprised by this. Having the `collections` and `typing` versions differ is equally unhelpful, though!

AlexWaygood: yeah, having to go "uh, yeah... it's a function... but you can inherit from it..." in the docs is probably not the most user-friendly thing to do 😄

peps/pep-0800.rst
JelleZijlstra: Thanks, I've applied all your feedback.

PR #4592, opened 2025-09-17
PEP-8XX: Add timestamps to exception tracebacks

gpshead:
This draft PEP proposes adding optional timestamps to Python exception objects that can be displayed in tracebacks. Based on CPython PR #129337 by Gregory P. Smith.

The feature adds a `__timestamp_ns__` attribute to BaseException and allows configuration via environment variables or command-line options. This is particularly useful for debugging async applications with exception groups and distributed systems.

🤖 Generated with Claude Code (https://claude.ai/code)

_[to be discussed at the ongoing cambridge core team sprint]_ -gpshead

### DIFF
--- a/peps/pep-08XX.rst
+++ b/peps/pep-08XX.rst
@@ -1,0 +1,381 @@
+PEP: 8XX
+Title: Add timestamps to exception tracebacks
+Author: Gregory P. Smith <greg@krypto.org>
+PEP-Delegate:
+Discussions-To: TODO
+Status: Draft
+Type: Standards Track
+Created: 09-Sep-2025
+Python-Version: 3.15
+Post-History:
+
+
+Abstract
+========
+
+This PEP proposes adding optional timestamps to Python exception objects that
+can be displayed in tracebacks. When enabled, each exception will record the
+time it was instantiated (typically when raised), and this timestamp can be
+displayed alongside the exception message in formatted tracebacks. This feature
+is particularly useful for debugging complex asynchronous applications, exception
+groups, and distributed systems where understanding the temporal ordering of
+exceptions is critical.
+
+
+Motivation
+==========
+
+In modern Python applications, especially those using asynchronous programming
+patterns and exception groups introduced in :pep:`654`, understanding the
+temporal relationships between exceptions has become increasingly important.
+Consider the following scenarios:
+
+1. **Async servers with exception groups**: When multiple exceptions are caught
+   and grouped together, it's often difficult to understand which exception
+   occurred first and how they relate temporally to each other and to external
+   events in the system.
+
+2. **Distributed systems debugging**: When correlating exceptions across
+   different services, having precise timestamps allows developers to match
+   exceptions with logs from other systems, making root cause analysis more
+   straightforward.
+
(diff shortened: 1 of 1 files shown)


peps/pep-08XX.rst
```diff
@@ -0,0 +1,483 @@
...
+3. Allow for future use cases that might require sub-millisecond precision
+
+
+Open Issues
+===========
+
+1. Should there be a Python API to programmatically enable/disable timestamps
+   at runtime?
+
+2. Should there be a way to retroactively add timestamps to existing exception
```
gpshead: no.

rocky-d:
While this PEP is being drafted, there's already a PyPI package that implements this today: **[dttb](https://github.com/rocky-d/dttb)**.

```python
import dttb; dttb.apply()
```

It patches `sys.excepthook` and `threading.excepthook` to add timestamps to tracebacks, with timezone config and callback support. Could be useful as a real-world reference for the PEP's design decisions (e.g., format, timezone handling).

- GitHub: https://github.com/rocky-d/dttb
- PyPI: https://pypi.org/project/dttb/
- Docs: https://rocky-d.github.io/dttb/

rocky-d:
**Update on real-world timestamp implementation:** While my previous comment about dttb was marked as spam, I think it's worth clarifying the design differences, as they're relevant to the PEP's decisions.

**dttb (display-time approach):**
- Hooks `sys.excepthook` and `threading.excepthook` to add timestamps only when exceptions are displayed
- Lightweight, opt-in via `dttb.apply()`
- Cannot capture exceptions that are caught but not re-raised or logged
- No pickle/serialization concerns

**This PEP (creation-time approach):**
- Records `__timestamp_ns__` in the BaseException struct when the exception is instantiated
- Captures ALL exceptions, even those caught internally
- Enables correlation with exception groups and distributed tracing
- Supports serialization for inter-process communication

**Relevant observations from dttb's development:**

1. **Format choice:** dttb uses ISO 8601 with microsecond precision by default. The PEP's format (showing `<@timestamp>` inline) is good for readability while avoiding line breaks.

2. **Timezone handling:** dttb found that a single configurable timezone (UTC, local, or custom) is often sufficient. The PEP's approach of using environment variables aligns well with this.

3. **Thread safety:** dttb hooks both `sys.excepthook` and `threading.excepthook` independently. The PEP's struct-level approach is inherently thread-safe and doesn't require hook duplication.

4. **Performance:** dttb adds negligible overhead (a decorator call + formatted print). A struct field read should be even cheaper.

The PEP's creation-time approach is fundamentally more powerful for debugging async applications and exception groups, which is exactly where dttb's display-time approach has limitations. Both can coexist: the PEP provides the core data, and tools like dttb can layer additional formatting or filtering on top.

hugovk:
If you rename the file from `pep-08XX.rst` to `pep-9999.rst`, the docs build should succeed (or get further).

Or we can assign a number when if you're ready.

peps/pep-0829.rst
```diff
@@ -0,0 +1,343 @@
+PEP: 829
```
warsaw, line 1:
```suggestion
PEP: 830
```

Thanks @gpshead !

PR #4601, opened 2025-09-22
PEP 763: Mark as Withdrawn

woodruffw:
As the PEP's author, I'm marking PEP 763 as withdrawn -- I think that the overall approach defined in this PEP makes sense, but I don't think there's a clear route to acceptance here that can surmount procedural problems with this being a PEP versus a matter of PyPI policy (like account management, AUPs, malware, quotas, etc.).

----
📚 Documentation preview 📚: https://pep-previews--4601.org.readthedocs.build/


AA-Turner:
Please could you add a new section above _Abstract_ with a summary of the reasons for withdrawing? The text of this PR summary would be fine to use, etc. I think there are previous examples if you search for 'Withdrawal Notice' or similar.

It'd also be good to add the _Resolution_ header, if possible.

A

woodruffw:
Thanks @AA-Turner! I've linked to the DPO comment and added a withdrawal notice section (mirrored the other ones I found).

PR #4616, opened 2025-09-30
PEP 809: Add Discussions-To link

zooba:


peps/pep-0809.rst
AA-Turner, line 9: We can now comment on unchanged lines?!
```suggestion
Python-Version: 3.15
Post-History:
  `30-Sep-2025 <https://discuss.python.org/t/104073>`__,
```

AA-Turner (commented):
Please also add Post-History:

A

peps/pep-0809.rst
zooba: Yes, but I can't commit suggestions there, apparently...

AA-Turner (approved):

PR #4627, opened 2025-10-04
PEP 779: Mark as Final

AA-Turner:
* [X] Final implementation has been merged (including tests and docs)
* [X] PEP matches the final implementation
* [X] Any substantial changes since the accepted version approved by the SC/PEP delegate
* [X] Pull request title in appropriate format (``PEP 123: Mark as Final``)
* [X] ``Status`` changed to ``Final`` (and ``Python-Version`` is correct)

cc @Yhg1s @mpage @colesbury @hugovk 

I can't see any appropriate documentation to link to, as this is more of a 'policy' PEP, so I haven't added a `canonical-doc` directive. Please let me know if there's something that should be linked to otherwise!

A

### DIFF
--- a/peps/pep-0779.rst
+++ b/peps/pep-0779.rst
@@ -4,7 +4,7 @@ Author: Thomas Wouters <thomas@python.org>,
         Matt Page <mpage at python.org>,
         Sam Gross <colesbury at gmail.com>
 Discussions-To: https://discuss.python.org/t/84319
-Status: Accepted
+Status: Final
 Type: Standards Track
 Created: 13-Mar-2025
 Python-Version: 3.14


hugovk:
I guess we could link to https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-pep779 ?

Yhg1s:
> I guess we could link to https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-pep779 ?

I think you mean https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-free-threaded-now-supported? I'm not sure that really qualifies as 'canonical docs'. We are working on more docs but they're something that applies to *this PEP* so much as the concept of threading, which is much broader than the scope of this PEP... I think the PEP itself and the acceptance cover everything, at least for now. We may get a more canonical spot for this decision (other than the PEP) later.

hugovk:
> > I guess we could link to [docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-pep779](https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-pep779) ?
> 
> I think you mean [docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-free-threaded-now-supported](https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-free-threaded-now-supported)?

Yeah, Adam renamed the anchor in the two days between me writing it and you clicking it :)


> I'm not sure that really qualifies as 'canonical docs'. We are working on more docs but they're something that applies to _this PEP_ so much as the concept of threading, which is much broader than the scope of this PEP... I think the PEP itself and the acceptance cover everything, at least for now. We may get a more canonical spot for this decision (other than the PEP) later.

👍

PR #4633, opened 2025-10-06
PEP 810: Add section about performance

pablogsal:
----
📚 Documentation preview 📚: https://pep-previews--4633.org.readthedocs.build/

### DIFF
--- a/peps/pep-0810.rst
+++ b/peps/pep-0810.rst
@@ -761,11 +761,14 @@ Filter function performance
 
 The filter function (set via ``sys.set_lazy_imports_filter()``) is called for
 every *potentially lazy* import to determine whether it should actually be
-lazy. The filter has **almost no measurable performance cost**. To measure
-this, we benchmarked importing all 278 top-level importable modules from the
-Python standard library (which transitively loads 392 total modules including
-all submodules and dependencies), then forced reification of every loaded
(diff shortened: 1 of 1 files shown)


pablogsal:
CC @warsaw

peps/pep-0810.rst
```diff
@@ -734,6 +734,129 @@ Subinterpreters are supported. Each subinterpreter maintains its own
...
+The `pyperformance suite`_ confirms the implementation is performance-neutral.
+
+.. _pyperformance suite: https://github.com/facebookexperimental/
+    free-threading-benchmarking/blob/main/results/bm-20250922-3.15.0a0-27836e5/
+    bm-20250922-vultr-x86_64-DinoV-lazy_imports-3.15.0a0-27836e5-vs-base.svg
+
+Filter function performance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The filter function (set via ``sys.set_lazy_imports_filter()``) is called for
```
Yhg1s, line 762: Does this warrant an explanation about how the filter is not called if it is not set, so it's really just a NULL check at each potentially lazy import?

pablogsal: Done

warsaw (commented):
Post-mergeulously approved!

PR #4639, opened 2025-10-07
PEP 791: edits (address SC feedback)

skirpichev:
* Change is either:
    * [ ] To a Draft PEP


skirpichev:
N/B: This extends `c_div()` example and adds a trivial note that a separate namespace is good also to avoid name conflicts (with example).

I'll wait for more feedback in the d.p.o thread, then make this ready for review.

vstinner:
You should also update the PEP to use `math.integer` namespace, no?

skirpichev:
> You should also update the PEP to use math.integer namespace, no?

I'm not sure.  IMO, this should be postponed till final resolution of the SC.

On one hand, SC preference was expressed sufficiently clear.  On another hand, this will also require adjustment in the implementation.  That might be a waste of work.

I think it's clear that it's a minor point and that decision on module naming - left to SC.

vstinner:
SC expressed a strong preference for `math.integer` name. I suggest updating the PEP now to make it more likely to be accepted by the SC.

skirpichev:
> Tell me when it's ready to be merged.

Oh, forgot to reply.  BTW, when my pr is ready to review - it's ready to be merged.  On my taste, of course.

vstinner:
Ok, I merged your PR.

vstinner:
You may add a comment to https://github.com/python/steering-council/issues/310 and the discussion to say that the PEP has been updated to address SC's feedback.

PR #4643, opened 2025-10-07
PEP 810: Add lazy objects to the spec.

Yhg1s:
Add lazy objects to the spec, and also clarify what happens when reification fails and tweak some of the language.

----
📚 Documentation preview 📚: https://pep-previews--4643.org.readthedocs.build/


DinoV (approved):
Looks great!

PR #4645, opened 2025-10-07
Update pep for dir behavior

DinoV:


AA-Turner:
@DinoV Please prefix PRs with the PEP number in future, for a clearer git history. (The PR template reminds us to do this, but sometimes GitHub doesn't prompt to use the template, which is unfortunate)

A

PR #4647, opened 2025-10-08
Add Savannah for 64-bit RPi support

savannahostrowski:


AA-Turner:
Adding DO-NOT-MERGE as SC approval is required. (In this case I presume it'll be a formality.)

A

savannahostrowski:
Ah, was I supposed to file a ticket on the SC repo first? It's a little odd for RPi since @gpshead already signed up to support it for 32-bit.

AA-Turner:
As I say, the situation's a bit unique for Raspberry Pi, but strictly speaking, [per the PEP](https://peps.python.org/pep-0011/#support-tiers):

> To be promoted to a tier, steering council support is required and is expected to be driven by team consensus.

I don't know if there are substantial differences in the RPi platform between 32 & 64 bit, but the triple appears to be different at least!

A

willingc (approved):
🎉

brettcannon:
I would just ask the SC and they will probably quickly approve it.

savannahostrowski:
@python/steering-council Mind taking a look 🙏 ?

hugovk:
Open an issue at https://github.com/python/steering-council

For example: https://github.com/python/steering-council/issues?q=sort%3Aupdated-desc%20is%3Aissue%20state%3Aclosed%20pep%2011%20tier

PR #4648, opened 2025-10-08
PEP 731: Remove Michael Droettboom from C API WG member list

encukou:
@mdboom resigned.

Thank you for your service, Michael!

----
📚 Documentation preview 📚: https://pep-previews--4648.org.readthedocs.build/


vstinner:
> Thank you for your service, Michael!

Yeah, thanks for your service!

PR #4656, opened 2025-10-13
PEP 810: Update decisions on with blocks and __dict__ reification

pablogsal:
This commit updates PEP 810 based on team discussion and feedback:

1. Allow lazy imports inside with blocks
   - Removed with blocks from syntax restrictions
   - Added rejected ideas section explaining why restriction was considered

2. __dict__ does not automatically reify lazy imports
   - Both `globals()` and `__dict__` return raw dictionary without reification
   - Updated reification section, FAQ, and rejected ideas

----
📚 Documentation preview 📚: https://pep-previews--4656.org.readthedocs.build/

### DIFF
--- a/peps/pep-0810.rst
+++ b/peps/pep-0810.rst
@@ -1233,36 +1233,6 @@ Can I force reification of a lazy import without using it?
 Yes, individual lazy objects can be resolved by calling their ``resolve()``
 method.
 
-What's the difference between ``globals()`` and ``mod.__dict__`` for lazy imports?
-----------------------------------------------------------------------------------
-
-Both ``globals()`` and ``mod.__dict__`` return the module's dictionary without
-reifying lazy imports. Accessing lazy objects through either will yield lazy
-proxy objects. This provides a consistent low-level API for introspection:
-
-.. code-block:: python
-
-  # In your module:
-  lazy import json
-
-  g = globals()
-  print(type(g['json']))  # <class 'LazyImport'>
-
-  d = __dict__
-  print(type(d['json']))  # <class 'LazyImport'>
-
-  # From external code:
-  import sys
-  mod = sys.modules['your_module']
-  d = mod.__dict__
-  print(type(d['json']))  # <class 'LazyImport'>
-
-Both ``globals()`` and ``__dict__`` expose the raw namespace view without
-implicit side effects. This symmetry makes the behavior predictable: accessing
-the namespace dictionary never triggers imports. If you need to ensure an
-import is resolved, call the ``resolve()`` method explicitly or access the
-attribute directly (e.g., ``json.dumps``).
-
 Why not use ``importlib.util.LazyLoader`` instead?
 --------------------------------------------------
 
@@ -1664,13 +1634,6 @@ aligning with Python's "consenting adults" philosophy. For genuinely problematic
 cases like ``with suppress(ImportError): lazy import foo``, static analysis
 tools and linters are better suited to catch these patterns than hard language
 restrictions.
-
-Additionally, forbidding explicit ``lazy import`` in ``with`` blocks would
-create complex rules for how the global lazy imports flag should behave,
-leading to confusing inconsistencies between explicit and implicit laziness. By
-allowing ``lazy import`` in ``with`` blocks, the rule is simple: the global
-flag affects all module-level imports except those in ``try`` blocks and wild
-card imports, matching exactly what's allowed with explicit syntax.
 
 Forcing eager imports in ``with`` blocks under the global flag
 ---------------------------------------------------------------


peps/pep-0810.rst
```diff
@@ -1664,6 +1648,59 @@ From the discussion on :pep:`690` it is clear that this is a fairly
...
+about catching exceptions, ``with`` blocks are commonly used for resource
+management, temporary state changes, or scoping -- contexts where lazy imports
+work perfectly fine. The ``lazy import`` syntax is explicit enough that
+developers who write it inside a ``with`` block are making an intentional choice,
+aligning with Python's "consenting adults" philosophy. For genuinely problematic
+cases like ``with suppress(ImportError): lazy import foo``, static analysis
+tools and linters are better suited to catch these patterns than hard language
+restrictions.
+
+Additionally, forbidding explicit ``lazy import`` in ``with`` blocks would
```
pganssle: I don't think this is any more true than the `try`/`except` case, no? It's not really more complexity than the current PEP unless you also have the part where implicit lazy is forced eager in `with` blocks.

peps/pep-0810.rst
```diff
@@ -465,54 +461,37 @@ immediately resolve all lazy objects (e.g. ``lazy from`` statements) that
 referenced the module. It **only** resolves the lazy object being accessed.
 
 Accessing a lazy object (from a global variable or a module attribute) reifies
-the object. Accessing a module's ``__dict__`` reifies **all** lazy objects in
-that module. Calling ``dir()`` at the global scope will not reify the globals
-and calling ``dir(mod)`` will be special cased in ``mod.__dir__`` avoid
+the object. Calling ``dir()`` at the global scope will not reify the globals
+and calling ``dir(mod)`` will be special cased in ``mod.__dir__`` to avoid
 reification as well.
```
Yhg1s: Since this now talks about when reification *doesn't* happen, this should be in the next paragraph.
```suggestion
the object. 
```

peps/pep-0810.rst
```diff
@@ -465,54 +461,37 @@ immediately resolve all lazy objects (e.g. ``lazy from`` statements) that
...
-
-  print('json' in sys.modules)  # True - reified by __dict__ access
-  print(type(d['json']))  # <class 'module'>
+However, calling ``globals()`` or accessing a module's ``__dict__`` does **not**
+trigger reification -- they return the module's dictionary, and accessing lazy
+objects through that dictionary still returns lazy proxy objects that need to
+be manually reified upon use. A lazy object can be resolved explicitly by
+calling the ``resolve`` method. Other, more indirect ways of accessing
+arbitrary globals (e.g. inspecting ``frame.f_globals``) also do **not** reify
+all the objects.
```
Yhg1s: Moved from the previous paragraph.
```suggestion
However, calling ``globals()`` or accessing a module's ``__dict__`` does
**not** trigger reification -- they return the module's dictionary, and
accessing lazy objects through that dictionary still returns lazy proxy
objects that need to be manually reified upon use. A lazy object can be
resolved explicitly by calling the ``resolve`` method. Calling ``dir()`` at
the global scope will not reify the globals, nor will calling ``dir(mod)``
(through special-casing in ``mod.__dir__``.) Other, more indirect ways of
accessing arbitrary globals (e.g. inspecting ``frame.f_globals``) also do
**not** reify all the objects.
```

peps/pep-0810.rst
```diff
@@ -1251,35 +1230,40 @@ either the lazy proxy or the final resolved object.
 Can I force reification of a lazy import without using it?
 ----------------------------------------------------------
 
-Yes, accessing a module's ``__dict__`` will reify all lazy objects in that
-module. Individual lazy objects can be resolved by calling their ``resolve()``
-method.
+Yes, individual lazy objects can be resolved by calling their ``resolve()``
+method. Note that accessing a module's ``__dict__`` or calling ``globals()``
+does **not** automatically reify lazy imports -- you'll see the lazy proxy
+objects themselves, which you can then manually resolve if needed.
```
Yhg1s: I don't think we should reiterate when things aren't reified when talking about how to force reification.
```suggestion
method.
```

peps/pep-0810.rst
```diff
@@ -1251,35 +1230,40 @@ either the lazy proxy or the final resolved object.
...
 
 What's the difference between ``globals()`` and ``mod.__dict__`` for lazy imports?
```
Yhg1s: This entire question now doesn't make sense anymore. There isn't a difference between them, is there? Can we just remove this FAQ entry?

peps/pep-0810.rst
```diff
@@ -465,54 +461,37 @@ immediately resolve all lazy objects (e.g. ``lazy from`` statements) that
...
-
-  print('json' in sys.modules)  # False - still lazy
-
-  # main.py
-  import sys
-  import my_module
-
-  # Accessing __dict__ from external code DOES reify all lazy imports
-  d = my_module.__dict__
+the object.
```
hugovk: Lint fix
```suggestion
the object.
```

PR #4661, opened 2025-10-18
PEP 810: Clarify that no C-API will be added in this PEP

pablogsal:
----
📚 Documentation preview 📚: https://pep-previews--4661.org.readthedocs.build/


willingc (approved):
Thanks for the addition @pablogsal :smile:

PR #4667, opened 2025-10-22
PEP 810: Add clarification on the warnings for global mode

pablogsal:
----
📚 Documentation preview 📚: https://pep-previews--4667.org.readthedocs.build/


pablogsal:
CC @diegorusso

PR #4670, opened 2025-10-23
PEP 12: Allow Pending as Post-History

hugovk:
Follow on from https://github.com/python/peps/pull/4361 which allowed "Discussions-To: Pending".

It would be handy to also put "Post-History: Pending" as a placeholder for the initial draft.

----
📚 Documentation preview 📚: https://pep-previews--4670.org.readthedocs.build/


warsaw (approved):
Thank you!  This is a frequent mini-annoyance.

PR #4672, opened 2025-10-24
Poll required: PEP 13: Clarify tie resolution

zware:
This change requires a vote among the core team before it can be merged.  Once the poll is opened, the history of amendments note will be amended to include the correct date of the poll closure and a link to the poll.

----
📚 Documentation preview 📚: https://pep-previews--4672.org.readthedocs.build/

### DIFF
--- a/peps/pep-0013.rst
+++ b/peps/pep-0013.rst
@@ -358,7 +358,7 @@ History of amendments
   Adopted Multi-winner Bloc STAR voting for council elections.
 * `2024-12-10 <https://discuss.python.org/t/72293/4>`__:
   Added a one-week deadline for seconding a vote of no confidence.
-* `2025-11-11 <https://discuss.python.org/t/104607>`__:
+* `2025-11-12 <https://discuss.python.org/t/104607>`__:
   Clarified that the software used for elections may resolve ties
   automatically if possible.
 


tim-one (approved):
LGTM! Thank you 😄

peps/pep-0013.rst
```diff
@@ -358,6 +358,9 @@ History of amendments
   Adopted Multi-winner Bloc STAR voting for council elections.
 * `2024-12-10 <https://discuss.python.org/t/72293/4>`__:
   Added a one-week deadline for seconding a vote of no confidence.
+* `2025-10-XX <https://discuss.python.org/t/placeholder-replace-me-just-discussion-for-now-should-be-the-poll/73918/87>`__:
```
hugovk: Two weeks from now will be November:
```suggestion
* `2025-11-XX <https://discuss.python.org/t/placeholder-replace-me-just-discussion-for-now-should-be-the-poll/73918/87>`__:
```

hugovk (approved):

peps/pep-0013.rst
zware: :man_facepalming: so it is, and I had just been reminded of that fact earlier in the evening.  I'll get that fixed with the rest of the date and the link when available.

tim-one:
@warsaw 
> One thing that occurs to me is that the wording here means that all ties (even mid-resolution ties) can be resolved by the software. I think that's fine.

I believe it's essential. The BetterVoting software has no possibility to pause "in the middle" of scoring to wait for human input on how to resolve a tie.

For a 5-winner SC election, if the winner of the 5th round was picked via breaking a tie, then it would be reasonable to let the tied candidates decide among themselves who gets the seat, and ignore the software's pick. But I wouldn't want to complicate the PEP to cater to that.

There's no real need anyway. If the ballots themselves don't contain enough info to break a tie, "the community" has expressed no preference. "Random" is as good as it gets then.

tim-one:
This hasn't seen new activity in a few days, and nothing but approvals. Time for the poll? Note that Guido withdrew his offer to run the poll,, so someone else will need to step up to do that.

tim-one:
@zware, are you still working on this, or is it ready for a poll?

AA-Turner:
Is there a succinct phrasing to capture that the automatic resolution of ties by the voting system ought be 'fair', explicitly random, or etc? A valid solution to breaking ties would be simply to pick the candidate who's surname is first alphabetically, but obviously we'd not want to use anything like this.

A

zware:
I'm not sure we need to get that specific here.  Since we're here documenting that we expect the election service to break (some to all) ties, the service's tie-breaking scheme just becomes part of the implicit selection criteria for the service.  I don't think we'd choose to use a service with insane tie-breaking (though as Tim has said, a tie means there's actually no preference, so choosing the first alphabetically by middle initial isn't as insane as it looks at first glance...as long as it's choosing one of the tied candidates :) ).

> @zware, are you still working on this, or is it ready for a poll?

I think we're ready for a poll, which I'd prefer not to drive but can if I must.  Once it's open I'll update the amendment note with the correct date and link.

tim-one:
FYI, I posted a poll just now:

https://discuss.python.org/t/vote-changing-pep-13-to-clarify-tiebreaking-for-sc-elections/104607

peps/pep-0013.rst
```diff
@@ -358,6 +358,9 @@ History of amendments
   Adopted Multi-winner Bloc STAR voting for council elections.
 * `2024-12-10 <https://discuss.python.org/t/72293/4>`__:
   Added a one-week deadline for seconding a vote of no confidence.
+* `2025-11-11 <https://discuss.python.org/t/104607>`__:
```
hugovk:
```suggestion
* `2025-11-12 <https://discuss.python.org/t/104607>`__:
```
To match the [poll end date](https://discuss.python.org/t/vote-changing-pep-13-to-clarify-tiebreaking-for-sc-elections/104607):

> EDIT: the poll is configured to close at 2025-11-12T06:00:00.000Z. Why? “About two weeks” from its posting date. If people want to dispute that, please supply the precise string you’d like to see instead. Note that results won’t be visible (not even to me) until the poll closes.

PR #4684, opened 2025-11-03
PEP 810: Mark as accepted and address SC's feedback

pablogsal:


pablogsal:
CC @python/steering-council @warsaw

PR #4696, opened 2025-11-10
Mark Python 3.12 as security and 3.14 as bugfix

hugovk:
Couple more followups from https://github.com/python/peps/pull/4331.

----
📚 Documentation preview 📚: https://pep-previews--4696.org.readthedocs.build/

### DIFF
--- a/release_management/python-releases.toml
+++ b/release_management/python-releases.toml
@@ -3066,7 +3066,7 @@ date = 2025-10-09
 
 [metadata."3.12"]
 pep = 693
-status = "bugfix"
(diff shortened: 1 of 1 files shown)


hugovk:
The data at these URLs are now identical, except an intended "Georg Brandl, Ned Deily (3.3.7+)" -> "Georg Brandl & Ned Deily (3.3.7)".

* https://raw.githubusercontent.com/python/devguide/main/include/release-cycle.json
* https://peps.python.org/api/release-cycle.json

PR #4701, opened 2025-11-13
PEP 814: Add Reference Impl. and Discussions-To

vstinner:
----
📚 Documentation preview 📚: https://pep-previews--4701.org.readthedocs.build/

### DIFF
--- a/peps/pep-0814.rst
+++ b/peps/pep-0814.rst
@@ -1,6 +1,7 @@
 PEP: 814
 Title: Add frozendict built-in type
 Author: Victor Stinner <vstinner@python.org>, Donghee Na <donghee.na@python.org>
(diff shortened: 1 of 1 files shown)


peps/pep-0814.rst
```diff
@@ -1,6 +1,7 @@
 PEP: 814
 Title: Add frozendict built-in type
 Author: Victor Stinner <vstinner@python.org>, Donghee Na <donghee.na@python.org>
+Discussions-To: https://discuss.python.org/t/104854
```
hugovk, line 4: Also add Post-History after Python-Version:

```rst
Post-History: `13-Nov-2025 <https://discuss.python.org/t/104854>`__
```

vstinner: Oh ok. Will do in my next PR.

PR #4702, opened 2025-11-13
Infra: document `release-cycle.json` and `python-releases.json`

hugovk:
Documentation for https://github.com/python/peps/pull/4331.

----
📚 Documentation preview 📚: https://pep-previews--4702.org.readthedocs.build/api/

### DIFF
--- a/peps/api/index.rst
+++ b/peps/api/index.rst
@@ -109,7 +109,7 @@ illustrating some of the possible values for each field:
 release-cycle.json
 ------------------
 
-There is a read-only JSON document of Python releases since version 2.6
+There is a read-only JSON document of Python releases
 available at https://peps.python.org/api/release-cycle.json.
 
 Each feature version is represented as a JSON object,


peps/api/index.rst
```diff
@@ -102,3 +105,88 @@ illustrating some of the possible values for each field:
        "url": "https://peps.python.org/pep-3124/"
      }
    }
+
+release-cycle.json
+------------------
+
+There is a read-only JSON document of Python releases since version 2.6
+available at https://peps.python.org/api/release-cycle.json.
```
AA-Turner: I'd prefer not to commit to 'since 2.6', as at some point we should drop long-obsolete versions from the file (2.6 can now learn to drive!)

peps/api/index.rst
```diff
@@ -102,3 +105,88 @@ illustrating some of the possible values for each field:
...
+
+For example:
```
AA-Turner, line 132: Include the schema?

```suggestion
Each feature version is represented as a JSON object,
keyed by the minor version number ("X.Y").
The structure of each JSON object is as follows:

.. code-block:: typescript

   {
     "<language version number>": {
       "branch": string,
       "pep": integer,
       "status": 'feature' | 'prerelease' | 'bugfix' | 'security' | 'end-of-life',
       "first_release": string,  // Date formatted as YYYY-MM-DD
       "end_of_life": string,  // Date formatted as YYYY-MM-DD
       "release_manager": string
     },
}

For example:
```

peps/api/index.rst
hugovk: Sure, something like this?
```suggestion
There is a read-only JSON document of Python releases
available at https://peps.python.org/api/release-cycle.json.
```

hugovk:
@AA-Turner Any other suggestions?

PR #4704, opened 2025-11-13
Infra: generate an iCalendar file of the release dates

hugovk:
Now we have all the Python release dates in a single file (https://github.com/python/peps/blob/main/release_management/python-releases.toml) and generate some JSON files from it (https://github.com/python/peps/pull/4702, https://github.com/python/peps/pull/4331), it's easy to generate other useful files as well.

One useful file is an iCalendar `.ics` file that can be imported to calendars.

The Fedora Python team are generating one at https://github.com/fedora-python/python-release-schedule-ical through scraping the PEP HTML files. I'm using this and have found it extremely useful. If we generate one here, they could retire that repo if they like.

Differences between the two:

* python-release-schedule-ical generates a calendar for Python 3.5 onwards. I defaulted to doing it for all releases. We could also easily limit from a given version if needed, but the files are 55K and 98K respectively, both pretty small.
* python-release-schedule-ical generates UIDs like `python3.5.0alpha1@python.org`. I used a similar pattern, but added a hyphen to avoid conflicts: `python-3.5.0alpha1@python.org`
* I sorted this iCal by date, earliest releases come first. The order doesn't matter, but can make it easier to read/debug the file.

cc @hroncok @hrnciar from Fedora Python, would this be useful for you?

----
📚 Documentation preview 📚: https://pep-previews--4704.org.readthedocs.build/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,5 +10,8 @@ sphinx-notfound-page >= 1.0.2
 pytest
 pytest-cov
 
+# For python-releases.ical
+icalendar
+
 # For python-releases.toml
 tomli >= 1.1.0 ; python_version < "3.11"
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+omit =
+    release_management/__main__.py
+
+[report]
+exclude_also =
+    if __name__ == .__main__.:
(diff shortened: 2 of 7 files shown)


hroncok:
It would be useful indeed. Thanks.

requirements.txt
```diff
@@ -10,5 +10,8 @@ sphinx-notfound-page >= 1.0.2
 pytest
 pytest-cov
 
+# For python-releases.ical
+icalendar
```
encukou, line 14: Obligatory “supply chain” security question: do we trust `icalendar` maintainers and their operational security?
This could be tasty vector for running code on core dev boxes.
cc @sethmlarson 

---

Maybe consider generating “manually”? iCal is a simple, stable text-based format.

sethmlarson: @encukou I would recommend generating the ICS file manually within the code instead of relying on a library. The format is quite simple, especially for the use-case we need :)

AA-Turner: I've proposed a simpler alternative at #4705 following this suggestion.

A

hugovk:
Closed in favour of https://github.com/python/peps/pull/4705.

PR #4715, opened 2025-11-24
Tests: Improve tests and coverage configuration

AA-Turner:
cc @hugovk 

This also uses the new support in pytest 9.0 for `.pytest.toml`.

A

----
📚 Documentation preview 📚: https://pep-previews--4715.org.readthedocs.build/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,10 +15,8 @@ commands =
 
 [coverage:run]
 omit =
-    # Ignore all __main__.py files
-    */__main__.py
-    # Ignore all files in the PEPs folder
-    peps/*
+    */__main__.py  # Ignore all __main__.py files
+    peps/*         # Ignore all files in the PEPs folder
 
 [coverage:report]
 exclude_also =


tox.ini
```diff
@@ -12,3 +12,12 @@ pass_env =
     FORCE_COLOR
 commands =
     python -bb -X dev -W error -m pytest {posargs}
+
+[coverage:run]
+omit =
+    */__main__.py  # Ignore all __main__.py files
+    peps/*         # Ignore all files in the PEPs folder
```
AA-Turner, line 19: Comments must be on the same line due to https://github.com/tox-dev/tox-ini-fmt/issues/42

tox.ini
```diff
@@ -12,3 +12,12 @@ pass_env =
...
+
+[coverage:report]
+exclude_also =
+    if __name__ == .__main__.:
```
hugovk, line 23: And also skip/omit/exclude `release_management/__main__.py` as suggested in https://github.com/python/peps/pull/4704?

AA-Turner:
`pyproject.toml` isn't the right conceptual fit. Coverage only supports `.coveragerc`, and I'd prefer not to have more INI-style config files. 

A

tox.ini
AA-Turner: Isn't that already covered by the blanket exclusion?

A

AA-Turner: I merged this to pull it into #4705 to test, it appears that the blanket exclusion does work, but only if there are no comments on the same line. I can't find anything in coverage's docs or GH issues about comments either way. Sorry for my error here!

PR #4721, opened 2025-12-01
PEP 814: Type annotation

vstinner:
----
📚 Documentation preview 📚: https://pep-previews--4721.org.readthedocs.build/


vstinner:
cc @corona10

PR #4750, opened 2025-12-18
PEP 816: Add Discussions-To and Post-History links

hugovk:
----
📚 Documentation preview 📚: https://pep-previews--4750.org.readthedocs.build/


brettcannon:
@hugovk Thanks!

The two-step publication process still trips me up on occasion.

PR #4753, opened 2025-12-19
PEP 798: Mark as Accepted

hugovk:
* [x] SC/PEP Delegate has formally accepted/rejected the PEP and posted to the ``Discussions-To`` thread
* [x] Pull request title in appropriate format (``PEP 123: Mark as Accepted``)
* [x] ``Status`` changed to ``Accepted``/``Rejected``
* [x] ``Resolution`` field points directly to SC/PEP Delegate official acceptance/rejected post, including the date (e.g. `` `01-Jan-2000 <https://discuss.python.org/t/12345/100>`__ ``)
* [x] Acceptance/rejection notice added, if the SC/PEP delegate had major conditions or comments
* [x] ``Discussions-To``, ``Post-History`` and ``Python-Version`` up to date

@adqm @edemaine 

We forgot to mark this as accepted.

I added the modification required by the SC at https://discuss.python.org/t/pep-798-unpacking-in-comprehensions/99435/60 as a note.

Perhaps just the first paragraph is enough?

----
📚 Documentation preview 📚: https://pep-previews--4753.org.readthedocs.build/

### DIFF
--- a/peps/pep-0798.rst
+++ b/peps/pep-0798.rst
@@ -13,6 +13,26 @@ Post-History:
   `19-Jul-2025 <https://discuss.python.org/t/pep-798-unpacking-in-comprehensions/99435>`__,
 Resolution: `03-Nov-2025 <https://discuss.python.org/t/pep-798-unpacking-in-comprehensions/99435/60>`__
 
+.. note::
+   The Steering Council accepts this PEP with one modification: we require that
+   both synchronous and asynchronous generator expressions use explicit loops
+   rather than yield from for unpacking operations.
+
+   The Steering Council believes that simplicity and consistency are paramount
+   here. The delegation behaviour provided by ``yield from`` adds semantic
+   complexity for advanced use cases involving ``.send(),`` ``.throw()``, and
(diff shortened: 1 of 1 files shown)


adqm:
Thanks @hugovk!  I was actually going to put in a PR for this (and for the actual implementation) this weekend, so the timing is perfect :slightly_smiling_face: 

If you think it works just to add this clarifying paragraph at the top, that seems fine to me.  My updated version instead rewrites the relevant portions of the PEP, but I'm not sure which approach is preferable.

hugovk:
I don't mind which way, let's ask @JelleZijlstra.

JelleZijlstra:
I'd prefer changing the main text. It will be confusing for future readers of the PEP if the text of the PEP claims something different than the implementation.

hugovk:
Sounds good.

@adqm Would you like to open your own PR? Then we can close this one.

adqm:
@hugovk Sure thing!  I'll make another pass through my local changes and put in a PR, probably tomorrow morning.

adqm:
@hugovk I just opened #4758, so I think this PR can be closed and we can continue discussion there.  Thanks!

PR #4763, opened 2025-12-29
PEP 101: Remove three more manual steps

hugovk:
After https://github.com/python/pythondotorg/pull/2832, we no longer need to manually edit EOL releases to add retired notices, a big red banner is now automatically added:

image

https://www.python.org/downloads/release/python-3925/

----
📚 Documentation preview 📚: https://pep-previews--4763.org.readthedocs.build/

### DIFF
--- a/peps/pep-0101.rst
+++ b/peps/pep-0101.rst
@@ -743,13 +743,6 @@ else does them.  Some of those tasks include:
 
         git push upstream --delete 3.3  # or perform from GitHub Settings page
 
-- Remove the release from the list of "Active Python Releases" on the Downloads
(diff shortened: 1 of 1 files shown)


hugovk:
Also remove mention of the HTML Help, that's no longer around: https://github.com/python/release-tools/pull/148.

hugovk:
https://github.com/python/pythondotorg/pull/2862 has been merged (awaiting deploy) meaning this table of active Python releases https://www.python.org/downloads/ is also autogenerated, so no need to manually update that any more:

image

And because this is getting up-to-date data from the API, it's more accurate. For example, I had initially copied and pasted the wrong 3.15 first release date (2026-10-07) when it's really 2026-10-01.

PR #4766, opened 2026-01-02
Replace pre-commit with prek in CI

hugovk:
Like https://github.com/python/cpython/pull/143149.

----
📚 Documentation preview 📚: https://pep-previews--4766.org.readthedocs.build/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,8 +2,7 @@ name: Lint PEPs
 
 on: [push, pull_request, workflow_dispatch]
 
-permissions:
-  contents: read
+permissions: {}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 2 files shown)


CAM-Gerlach (approved):
LGTM, thanks @hugovk ! FWIW, while I've been a big pre-commit fan in the past, prek brings significant longstanding improvements to performance, features and community trust. While I know some other devs have concerns about moving too quickly, and I agree keeping compat with pre-commit for the time being is the way to go, FWIW in terms of trust and reliability that others were concerned about, pretty much anything is a net improvement over pre-commit in that department, as many other CPython maintainers and Python community members can attest from firsthand experience 😉 .

hugovk:
Yep, and like https://github.com/python/cpython/issues/143148, this only updates the CI and not the Makefile or docs (README.rst, CONTRIBUTING.rst).

PR #4785, opened 2026-01-16
PEP 686: Mark as Final

vstinner:
----
📚 Documentation preview 📚: https://pep-previews--4785.org.readthedocs.build/

### DIFF
--- a/peps/pep-0686.rst
+++ b/peps/pep-0686.rst
@@ -2,7 +2,7 @@ PEP: 686
 Title: Make UTF-8 mode default
 Author: Inada Naoki <songofacandy@gmail.com>
 Discussions-To: https://discuss.python.org/t/14737
-Status: Accepted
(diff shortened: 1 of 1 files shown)


vstinner:
cc @methane

hugovk:
Is there somewhere in the docs we could link to with `canonical-doc`?

Perhaps https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-utf8-default if nothing else?

Re: https://github.com/python/peps/blob/main/.github/PULL_REQUEST_TEMPLATE/Mark%20a%20PEP%20as%20Final.md

vstinner:
Is it mandatory to add `canonical-doc` in the PEP? The PEP remains a good documentation for that change.

scutuatua-crypto (approved):

methane (approved):

PR #4787, opened 2026-01-20
PEP 822: update d-string spec and rejected ideas.

methane:
----
📚 Documentation preview 📚: https://pep-previews--4787.org.readthedocs.build/

### DIFF
--- a/peps/pep-0822.rst
+++ b/peps/pep-0822.rst
@@ -18,25 +18,25 @@ multiline string literals.
 Dedented multiline strings use a new prefix "d" (shorthand for "dedent") before
 the opening quote of a multiline string literal.
 
-Example (spaces are visualized as ``_``):
+Example (spaces are visualized as ``.``):
 
 .. code-block:: python
 
    def hello_paragraph() -> str:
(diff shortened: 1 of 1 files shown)


bwoodsend:
Very subjective but am I the only one who wishes that the *spaces replaced with underscores for clarity* picked a character that's easier to count consecutive runs of? Every time I see `____` I need to highlight part of the string to figure out how many underscores it has.

hugovk:
Yes, perhaps `----` or `····` would be better?

methane:
Although it was easy to read in the text editor, it becomes difficult to read several characters when viewed in a browser because multiple underscores are connected.

Since we are using `--->` to represent TAB, how about using `.` to represent spaces?

encukou:
Or use [`␣` (Open Box/`&blank;`)](https://en.wikipedia.org/wiki/Whitespace_character#Substitute_images:~:text=9251-,Open%20box)

hugovk (approved):
A simple `.` is good, and it's easy for people to type in the DPO thread. Thanks!

PR #4811, opened 2026-02-07
Infra: Add version filter to PEP index

hugovk:
Adds a version filter to the PEP 0 index, allowing to only show PEPs relevant to the selected version.

This is especially useful for me right now, so I can check on the progress of PEPs for 3.15, and we can check for example if accepted PEPs have been implemented, and need marking as final.

Stores it in local storage, so the same filter applies the next time you visit (on that browser/machine). Not entirely sure about this, we could remove it so it resets on new visits/reloads?

----
📚 Documentation preview 📚: https://pep-previews--4811.org.readthedocs.build/

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -50,7 +50,7 @@ peps/pep-0020.rst  @tim-one
 peps/pep-0042.rst  @jeremyhylton
 # ...
 peps/pep-0100.rst  @malemburg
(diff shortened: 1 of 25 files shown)


hugovk:
I'm going to merge this, but happy to iterate or even revert based on feedback.

ZeroIntensity:
Woah, I just noticed this while looking at the PEPs page. This is super cool!

PR #4812, opened 2026-02-08
PEP 798: Mark as Final

adqm:
* [ ] Any substantial changes since the accepted version approved by the SC/PEP delegate
* [x] Pull request title in appropriate format (``PEP 123: Mark as Final``)

-----------

@hugovk, I'm leaving a couple of the boxes unchecked for right now, because:

* While there weren't any wording changes since the Accepted version, I'm not 100% sure we got direct feedback about the current wording from an SC member.
* I don't think the one about canonical docs is relevant here (but happy to add something if I'm wrong about that).

----
📚 Documentation preview 📚: https://pep-previews--4812.org.readthedocs.build/

### DIFF
--- a/peps/pep-0798.rst
+++ b/peps/pep-0798.rst
@@ -3,7 +3,7 @@ Title: Unpacking in Comprehensions
 Author: Adam Hartz <hz@mit.edu>, Erik Demaine <edemaine@mit.edu>
 Sponsor: Jelle Zijlstra <jelle.zijlstra at gmail.com>
 Discussions-To: https://discuss.python.org/t/99435
-Status: Accepted
+Status: Final
 Type: Standards Track
 Created: 19-Jul-2025
 Python-Version: 3.15
--- a/peps/pep-0798.rst
+++ b/peps/pep-0798.rst
@@ -11,6 +11,11 @@ Post-History: `16-Oct-2021 <https://mail.python.org/archives/list/python-ideas@p
 Resolution: `03-Nov-2025 <https://discuss.python.org/t/pep-798-unpacking-in-comprehensions/99435/60>`__
 
 
+.. canonical-doc::
+   :external+py3.15:ref:`Displays for lists, sets and dictionaries <comprehensions>`,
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
> I don't think the one about canonical docs is relevant here (but happy to add something if I'm wrong about that).

It should link to the docs for this feature in CPython.

> While there weren't any wording changes since the Accepted version, I'm not 100% sure we got direct feedback about the current wording from an SC member.

Yes, looks like we didn't directly get approval from the SC: #4758 implemented what Pablo asked for in https://discuss.python.org/t/pep-798-unpacking-in-comprehensions/99435/60 but we didn't go back and get explicit approval. @pablogsal (since you're the one who posted the acceptance), do you think this is OK, or should we explicitly ask the SC for approval of the final text?

adqm:
> It should link to the docs for this feature in CPython.

Right.  My confusion comes from the fact that the docs are somewhat scattered among the library reference; there isn't a single page that is about unpacking in comprehensions, per se (and I'm not sure where such a page would go if it did exist).  I suppose we could link to <https://docs.python.org/3/reference/expressions.html> if we really need a link, but that seems off to me since that isn't really a page about this feature (and since it has less detail than the PEP about the specifics).  Looking through some historical PEPs, it seems like not every final PEP has one of these links (e.g., [572](https://peps.python.org/pep-0572/)).

encukou:
You can add several links, see PEP-697 for an example.

hugovk:
@adqm Reminder for this :)

If nothing better, let's link to:

* https://docs.python.org/3.15/reference/expressions.html#displays-for-lists-sets-and-dictionaries
* https://docs.python.org/3.15/reference/expressions.html#dictionary-displays

Because they're the bits with "Changed in version 3.15: Unpacking with the [*/**] operator is now allowed in the expression."

JelleZijlstra:
Added canonical docs links.

JelleZijlstra (approved):

hugovk:
Thanks!

PR #4832, opened 2026-02-27
PEP 818: Reduce number of details in spec

hoodmane:
According to @Yhg1s there are too many implementation details here. This reduces them.

Does this look better to you @Yhg1s? 

----
📚 Documentation preview 📚: https://pep-previews--4832.org.readthedocs.build/


hugovk:
The 3.15 feature freeze has passed, would you also like to retarget `Python-Version` to 3.16?

PR #4905, opened 2026-04-14
PEP 615: Fix Sphinx warnings

Secrus:
Ref: #4087 

The reference was not used in the document, but I found it significant enough to keep it.

----
📚 Documentation preview 📚: https://pep-previews--4905.org.readthedocs.build/


hugovk (approved):
The refs use was deleted in https://github.com/python/peps/commit/cfe6639a6d005f40e1b116a89250efa54a715d8d

PR #4922, opened 2026-04-16
PEP 708: Mark as Rejected

pfmoore:
----
📚 Documentation preview 📚: https://pep-previews--4922.org.readthedocs.build/

### DIFF
--- a/peps/pep-0708.rst
+++ b/peps/pep-0708.rst
@@ -3,13 +3,22 @@ Title: Extending the Repository API to Mitigate Dependency Confusion Attacks
 Author: Donald Stufft <donald@stufft.io>
 PEP-Delegate: Paul Moore <p.f.moore@gmail.com>
 Discussions-To: https://discuss.python.org/t/24179
-Status: Provisional
+Status: Rejected
 Type: Standards Track
 Topic: Packaging
 Created: 20-Feb-2023
 Post-History: `01-Feb-2023 <https://discuss.python.org/t/23414/>`__,
-              `23-Feb-2023 <https://discuss.python.org/t/24179>`__
-Resolution: https://discuss.python.org/t/24179/72
+              `23-Feb-2023 <https://discuss.python.org/t/24179>`__,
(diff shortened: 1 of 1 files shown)


pfmoore:
How do I have two resolution headers? Separating the URLs with commas didn't work, nor did having two `Resolution:` lines. For a provisionally accepted, and then rejected, PEP I would like to keep a record of both decision posts, for the record.

Worst case scenario, I could move the URL for the provisional acceptance into the text of the rejection note, but that doesn't feel right.

peps/pep-0708.rst
```diff
@@ -3,13 +3,22 @@ Title: Extending the Repository API to Mitigate Dependency Confusion Attacks
...
-              `23-Feb-2023 <https://discuss.python.org/t/24179>`__
+              `23-Feb-2023 <https://discuss.python.org/t/24179>`__,
+              `02-Apr-2026 <https://discuss.python.org/t/106800>`__,
 Resolution: https://discuss.python.org/t/24179/72
+Resolution: https://discuss.python.org/t/106800/14
+
+Rejection
+=========
+
+After 3 years in the provisionally accepted state, the PEP has been
```
hugovk:
```suggestion
After three years in the provisionally accepted state, the PEP has been
```

hugovk (approved):

hugovk:
> How do I have two resolution headers? Separating the URLs with commas didn't work, nor did having two `Resolution:` lines. For a provisionally accepted, and then rejected, PEP I would like to keep a record of both decision posts, for the record.
> 
> Worst case scenario, I could move the URL for the provisional acceptance into the text of the rejection note, but that doesn't feel right.

We'd need to adjust the code to allow two resolutions.

I'd just point to the most recent one, we have the post history for older things.

pfmoore:
> I'd just point to the most recent one, we have the post history for older things.

OK, it's not ideal IMO, but it'll do.

pradyunsg (approved):

PR #4926, opened 2026-04-18
Infra: bump Sphinx to 9.0+, limit docutils to <0.22

Secrus:
~~Blocked by: #4908, #4906~~
Closes: #4087 

Bump Sphinx to 9.0+ (9.2 at the moment). `docutils` is upper-limited because of build errors listed in #4924.

----
📚 Documentation preview 📚: https://pep-previews--4926.org.readthedocs.build/

### DIFF
--- a/pep_sphinx_extensions/pep_processor/html/pep_html_builder.py
+++ b/pep_sphinx_extensions/pep_processor/html/pep_html_builder.py
@@ -48,4 +48,4 @@ class DirectoryBuilder(FileBuilder):
     # sync all overwritten things from DirectoryHTMLBuilder
     name = DirectoryHTMLBuilder.name
     get_target_uri = DirectoryHTMLBuilder.get_target_uri
-    get_outfilename = DirectoryHTMLBuilder.get_output_path
+    get_output_path = DirectoryHTMLBuilder.get_output_path
--- a/pep_sphinx_extensions/pep_processor/html/pep_html_builder.py
+++ b/pep_sphinx_extensions/pep_processor/html/pep_html_builder.py
@@ -48,4 +48,4 @@ class DirectoryBuilder(FileBuilder):
     # sync all overwritten things from DirectoryHTMLBuilder
     name = DirectoryHTMLBuilder.name
     get_target_uri = DirectoryHTMLBuilder.get_target_uri
-    get_outfilename = DirectoryHTMLBuilder.get_outfilename
+    get_outfilename = DirectoryHTMLBuilder.get_output_path
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,10 +3,9 @@
 # Sphinx requires >= 2.17. JSON5 support added in 2.19, `lazy` highlight added in 2.21
 pygments >= 2.21
 
-# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
-# See https://github.com/sphinx-doc/sphinx/pull/11100
-Sphinx >= 5.1.1, != 6.1.0, != 6.1.1, < 8.1.0
-docutils >= 0.19.0
+Sphinx >= 8.2
+# 0.22 causes build errors, see https://github.com/python/peps/issues/4924
+docutils < 0.22
 sphinx-notfound-page >= 1.0.2
 # For search
 pagefind[bin] >= 1.5.0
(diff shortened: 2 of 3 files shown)


Secrus:
@hugovk, since there is only one PEP left on the #4087 list, could you please review the proposed change? I am mostly interested in knowing what the policy for older Sphinx versions is, and whether I should go with a lower base for the version range.

hugovk:
Let's not set a higher version than needed, but if we have a good reason for bumping it, I think that's fine.

Something to note, Sphinx does drop older Python sooner than many projects. For example, Sphinx 9.1.0 requires Python 3.12 or newer.

But then people writing PEPs are targeting changes in 3.15 and soon 3.16 :)

Secrus:
Ok, I have set the Sphinx requirement to `> 8.1` to prevent the errors we were fixing. That seems like a reasonable middle ground.

requirements.txt
```diff
@@ -5,10 +5,8 @@
 # https://github.com/pygments/pygments/discussions/3145
 pygments @ git+https://github.com/pygments/pygments@2cad2642058441b59782a6a18f03c98c42d081f1
 
-# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
-# See https://github.com/sphinx-doc/sphinx/pull/11100
-Sphinx >= 5.1.1, != 6.1.0, != 6.1.1, < 8.1.0
-docutils >= 0.19.0
+Sphinx > 9.0
+docutils >= 0.19.0, < 0.22
```
hugovk: Only the upper bound is important here, Sphinx already sets `docutils>=0.21,<0.23`:

```suggestion
docutils < 0.22
```
And include a comment on why we need an upper bound? Something like pending https://github.com/sphinx-doc/sphinx/issues/14455.

requirements.txt
```diff
@@ -5,10 +5,8 @@
 # https://github.com/pygments/pygments/discussions/3145
 pygments @ git+https://github.com/pygments/pygments@2cad2642058441b59782a6a18f03c98c42d081f1
 
-# Sphinx 6.1.0 broke copying images in parallel builds; fixed in 6.1.2
-# See https://github.com/sphinx-doc/sphinx/pull/11100
-Sphinx >= 5.1.1, != 6.1.0, != 6.1.1, < 8.1.0
-docutils >= 0.19.0
+Sphinx > 9.0
```
hugovk: I think we can do Sphinx 8.2 or greater?

```suggestion
Sphinx >= 8.2
```

hugovk (commented):
[RTD](https://app.readthedocs.org/projects/pep-previews/builds/33740098/):

>  Index file is not present in HTML output directory Your documentation did not generate an index.html at its root directory. This is required for documentation serving at the root URL for this version.

Try this in `pep_html_builder.py`:

```diff
     name = DirectoryHTMLBuilder.name
     get_target_uri = DirectoryHTMLBuilder.get_target_uri
-    get_outfilename = DirectoryHTMLBuilder.get_outfilename
+    get_output_path = DirectoryHTMLBuilder.get_output_path
```

Related to https://github.com/sphinx-doc/sphinx/pull/13045.

Secrus:
@hugovk looks like the upgrade to `sphinx >= 8.2` would force us to drop Python 3.10 support. Is that ok? I could make a separate PR for that.

hugovk:
Yep, fine to drop 3.10, it's EOL in a few months.

Secrus:
@hugovk I have tried the `get_output_path` change, but as you can see, the error is the same. It seems like, for some reason, despite using `dirhtml`, we end up with the `html` build instead. I don't understand why that happens.

pep_sphinx_extensions/pep_processor/html/pep_html_builder.py
```diff
@@ -48,4 +48,4 @@ class DirectoryBuilder(FileBuilder):
     # sync all overwritten things from DirectoryHTMLBuilder
     name = DirectoryHTMLBuilder.name
     get_target_uri = DirectoryHTMLBuilder.get_target_uri
-    get_outfilename = DirectoryHTMLBuilder.get_outfilename
+    get_outfilename = DirectoryHTMLBuilder.get_output_path
```
hugovk: Missed a bit:
```suggestion
    get_output_path = DirectoryHTMLBuilder.get_output_path
```

By the way, please avoid force pushing, most @python repos squash merge at the end, and force pushing makes it harder to see incremental changes.

hugovk (approved):
Thanks!

And reminder for next time, please don't force push: https://devguide.python.org/getting-started/pull-request-lifecycle/#don-t-force-push

PR #4927, opened 2026-04-18
PEP 101: Generalize Android wording to include other platforms

mhsmith:
Automated iOS releases were added in https://github.com/python/release-tools/pull/365, and there may be other platforms in the future, so I've replaced the reference to Android with a more general statement.

----
📚 Documentation preview 📚: https://pep-previews--4927.org.readthedocs.build/


hugovk:
Thanks!

PR #4935, opened 2026-04-24
PEP 661: Mark as Accepted

JelleZijlstra:
----
📚 Documentation preview 📚: https://pep-previews--4935.org.readthedocs.build/


peps/pep-0661.rst
nschneid, line 265: These lines render as preformatted text rather than bullets: https://pep-previews--4935.org.readthedocs.build/pep-0661/#c-api

JelleZijlstra: Thanks, fixed

PR #4937, opened 2026-04-25
PEP 829: Accepted

warsaw:
Also, updates based on SC acceptance feedback

----
📚 Documentation preview 📚: https://pep-previews--4937.org.readthedocs.build/

### DIFF
--- a/peps/pep-0829.rst
+++ b/peps/pep-0829.rst
@@ -29,7 +29,7 @@ This PEP proposes:
   ignored.
 * A 5 year period after which a warning will be produced for the existence of
   ``import`` lines in ``.pth`` files.
-* Introducing a new ``.start`` file parallel to ``.pth`` files which list
+* Introducing a new ``.start`` file parallel to ``.pth`` files which lists
   entry points (as ``pkg.mod:callable``) to execute by the ``site.py`` module.
 * The presence of a matching ``.start`` file disables ``import`` line
   processing in ``.pth`` files.
--- a/peps/pep-0829.rst
+++ b/peps/pep-0829.rst
@@ -24,7 +24,7 @@ to the first line of user code.
 
 This PEP proposes:
 
-* A 3 year deprecating for the processing of ``import`` lines in ``.pth``
+* A 3 year deprecation period for the processing of ``import`` lines in ``.pth``
   files, after which, ``import`` lines in ``.pth`` files will be silently
   ignored.
 * A 5 year period after which a warning will be produced for the existence of
--- a/peps/pep-0829.rst
+++ b/peps/pep-0829.rst
@@ -2,7 +2,7 @@ PEP: 829
 Title: Package Startup Configuration Files
 Author: Barry Warsaw <barry@python.org>
 Discussions-To: https://discuss.python.org/t/pep-829-structured-startup-configuration-via-site-toml-files/106789
-Status: Draft
+Status: Accepted
 Type: Standards Track
 Created: 31-Mar-2026
 Python-Version: 3.15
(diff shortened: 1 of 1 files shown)


peps/pep-0829.rst
```diff
@@ -21,18 +22,18 @@ Previously controlled through legacy ``.pth`` files parsed and executed by the
...
-<https://securitylabs.datadoghq.com/articles/litellm-compromised-pypi-teampcp-supply-chain-campaign/>`_
-have used arbitrary code execution in ``.pth`` files as an attack vector.
+This PEP proposes:
 
-This PEP doesn't completely close this vector, but it does propose an
-important and useful improvement, by narrowing the attack surface and enabling
-a future policy mechanism for controlling which packages are allowed or
-prevented from extending the path and executing start up code.  See
-:ref:`security` for additional discussion.
+* A 3 year deprecating for the processing of ``import`` lines in ``.pth``
```
JelleZijlstra:
```suggestion
* A 3 year deprecation period for the processing of ``import`` lines in ``.pth``
```

peps/pep-0829.rst
```diff
@@ -21,18 +22,18 @@ Previously controlled through legacy ``.pth`` files parsed and executed by the
...
+  files, after which, ``import`` lines in ``.pth`` files will be silently
```
JelleZijlstra:
```suggestion
  files, after which ``import`` lines in ``.pth`` files will be silently
```

peps/pep-0829.rst
```diff
@@ -21,18 +22,18 @@ Previously controlled through legacy ``.pth`` files parsed and executed by the
...
+  ignored.
+* A 5 year period after which a warning will be produced for the existence of
```
JelleZijlstra: Wait, they're deprecated after 3 years, and then after 5 years we produce a warning?

peps/pep-0829.rst
```diff
@@ -21,18 +22,18 @@ Previously controlled through legacy ``.pth`` files parsed and executed by the
...
+  ``import`` lines in ``.pth`` files.
+* Introducing a new ``.start`` file parallel to ``.pth`` files which list
```
JelleZijlstra:
```suggestion
* Introducing a new ``.start`` file parallel to ``.pth`` files which lists
```

peps/pep-0829.rst
warsaw: The thinking behind this is that there's a strong desire to stop processing `import` lines asap (3 years) but not to annoy people with warnings until the last active Python that _doesn't_ support 829 is EOL (5 years).  Without this, you can't "roman ride" with both a legacy `.pth` file for older Pythons and a `.start` file for newer Pythons without warnings.

JelleZijlstra: So the earlier deprecation is silent? I know this has been decided on already, just want to make sure it's presented clearly. Perhaps something like this:

Support for `import` lines in `.pth` files will be gradually removed:
* For the first three years (expected to be Python 3.15, 3.16, and 3.17) the behavior will remain as it is now.
* For the next two years (expected to be Python 3.18 and 3.19), `import` lines will be silently ignored.
* Afterwards (expected to be Python 3.20 and higher), a warning will be produced for the existence of `import` lines in `.pth` files.

warsaw: Yes, that's much clearer, thanks!  Let me push an update.

warsaw: PTAL

JelleZijlstra (approved):

peps/pep-0829.rst
JelleZijlstra: Thanks, approved!

willingc (approved):
Thanks @warsaw.

PR #4950, opened 2026-05-02
PEP 811: Mark as 'Active'

StanFromIreland:
This is technically more suitable for this PEP, per [PEP 1](https://peps.python.org/pep-0001/):

> Process PEPs may also have a status of “Active” if they are never meant to be completed. E.g. [PEP 1](https://peps.python.org/pep-0001/) (this PEP).

Another option is marking it as 'Final,' and noting the page in the devguide as the up-to-date spec. I wasn't quite sure which is best, what do you think @sethmlarson?


sethmlarson (approved):
Thank you!!

PR #4952, opened 2026-05-04
PEP 3118: Clarify format string "unimplemented" note

seberg:
In order to wrap up the discussion around the choice of ``Zd`` vs. ``D`` a bit, I propose changing the "not implemented" note a bit.

Not sure if it comes quite across, but I think:
1. There is no good reason to break with any of the "additions" here (I.e. don't use `&` to mean something different).
2. Python should feel free to diverge, but I think the note pointing to NumPy (which is "just" an example) hopefully makes it clear that this isn't a void.

### DIFF
--- a/peps/pep-3118.rst
+++ b/peps/pep-3118.rst
@@ -14,7 +14,8 @@ Post-History: `09-Apr-2007 <https://mail.python.org/archives/list/python-dev@pyt
    Not all features proposed here were implemented. Specifically:
 
    - ``PyObject_CopyToObject`` was not added.
-   - :ref:`3118-struct-additions` were not added, except for ``?`` (``_Bool``).
+   - :ref:`3118-struct-additions` were only partially adopted by Python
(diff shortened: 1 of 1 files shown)


encukou:
IMO, this should go in a new PEP.

seberg:
> IMO, this should go in a new PEP.

Happy either way. I just opened it because it was a recurring discussion point that this note made it sounds like there are no strings attached to just using `D` rather than `Zd` here. And IMO that is incorrect as things _are_ implemented by downstream and that shouldn't be ignored.

warsaw:
> IMO, this should go in a new PEP.

Agreed.  At this point probably the only thing that should go into PEP 3118 is a `Superseded` header once a new PEP sorting all this out is written.

seberg:
OK, I'll just close this then if you don't think clarifying this additional note doesn't seem helpful.

PR #4958, opened 2026-05-06
PEP 748: tlslib - configuration

Julien00859:
First time contributor :tada:

A few suggestions to make the configuration a bit more explicit. I decided to leave most of the attributes undocumented as they are pretty explicit to me, and to instead only document the few attributes that are different from the client and the server.

### [PEP 748: ConfigurationError is only for unsupported features](https://github.com/python/peps/commit/b3229ec84b803a0b4240c1287b66eb3f9c33bf1e)

Discussed at https://github.com/trailofbits/tlslib.py/issues/72#issuecomment-3409443129

> ConfigurationError was intended for when specific implementations don't support certain behavior (e.g. adding a certificate by identifier). I think ValueError should be fine, probably raised when validating the configuration. *~Joop*

### [PEP 748: certificate_chain is mandatory server-side](https://github.com/python/peps/commit/ea0b73f34c78b034e0ace18b5d7738e5efc2d125)

Discussed at https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263/75

> This makes sense to me. We can allow empty server certificates in the insecure module. *~Joop*

### [PEP 748: disambiguate config trust_store=None](https://github.com/python/peps/commit/6e61445101d6657d90de403ed722395b6d2ec4c7)

Discussed at https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263/78

> I agree, this is better. *~Joop*

### DIFF
--- a/peps/pep-0748.rst
+++ b/peps/pep-0748.rst
@@ -1072,7 +1072,7 @@ The definitions of the errors are below:
 
 
     class ConfigurationError(TLSError):
-        """An special exception that implementations can use when the provided
+        """A special exception that implementations can use when the provided
         configuration uses features not supported by that implementation."""
 
 
--- a/peps/pep-0748.rst
+++ b/peps/pep-0748.rst
@@ -276,7 +276,7 @@ The ``TLSClientConfiguration`` class would be defined by the following code:
             inner_protocols: Sequence[NextProtocol | bytes] | None = None,
             lowest_supported_version: TLSVersion | None = None,
             highest_supported_version: TLSVersion | None = None,
-            trust_store: TrustStore | None = None,
+            trust_store: TrustStore = TrustStore.system(),
         ) -> None:
             if inner_protocols is None:
                 inner_protocols = []
@@ -309,8 +309,12 @@ The ``TLSClientConfiguration`` class would be defined by the following code:
             return self._highest_supported_version
 
         @property
-        def trust_store(self) -> TrustStore | None:
+        def trust_store(self) -> TrustStore:
             return self._trust_store
+
+A ``trust_store`` is mandatory and is used to validate the server certificates.
+It uses the system's trust store by default. Insecure connections are only
+possible via the :ref:`insecure` module.
 
 Server Configuration
 ^^^^^^^^^^^^^^^^^^^^
@@ -372,6 +376,9 @@ The ``TLSServerConfiguration`` class would be defined by the following code:
         def trust_store(self) -> TrustStore | None:
             return self._trust_store
 
+A ``trust_store`` is optional. Setting one enables client authentication and
+uses the trust store to validate the client certificates. Leaving it ``None``
+disables client authentication.
 
 Context
 ~~~~~~~
@@ -1502,6 +1509,8 @@ a trust store:
 Note that this function only needs to verify that supported constructors were
 used for the certificates, private keys, and trust store. It does not need to
 parse or retrieve the objects to validate them further.
+
+.. _insecure:
 
 Insecure Usage
 --------------
--- a/peps/pep-0748.rst
+++ b/peps/pep-0748.rst
@@ -316,6 +316,11 @@ A ``trust_store`` is mandatory and is used to validate the server certificates.
 It uses the system's trust store by default. Insecure connections are only
 possible via the :ref:`insecure` module.
 
+When ``certificate_chain`` is set, it is a single signing chain comprising a
+leaf client certificate including its corresponding private key and optionally a
+list of intermediate certificates. These certificates will be offered to the
+server during the handshake if required.
+
 Server Configuration
 ^^^^^^^^^^^^^^^^^^^^
 
@@ -335,7 +340,7 @@ The ``TLSServerConfiguration`` class would be defined by the following code:
 
         def __init__(
             self,
-            certificate_chain: Sequence[SigningChain] | None = None,
+            certificate_chain: Sequence[SigningChain],
             ciphers: Sequence[CipherSuite] | None = None,
             inner_protocols: Sequence[NextProtocol | bytes] | None = None,
             lowest_supported_version: TLSVersion | None = None,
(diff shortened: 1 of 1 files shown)


hugovk:
cc PEP authors @jvdprng and @woodruffw.

peps/pep-0748.rst
```diff
@@ -305,12 +308,87 @@ The ``TLSClientConfiguration`` class would be defined by the following code:
         def highest_supported_version(self) -> TLSVersion | None:
             return self._highest_supported_version
 
+        @property
+        def trust_store(self) -> TrustStore:
+            return self._trust_store
+
+A ``trust_store`` is mandatory and is used to validate the server certificates.
+It uses the system's trust store by default. Insecure connections are only
+possible via the :ref:`insecure` module.
```
jvdprng, line 317:
```suggestion
It uses the system's trust store by default. Insecure connections with
server authentication disabled are only possible via the :ref:`insecure` module.
```

peps/pep-0748.rst
```diff
@@ -305,12 +308,87 @@ The ``TLSClientConfiguration`` class would be defined by the following code:
...
         @property
         def trust_store(self) -> TrustStore | None:
             return self._trust_store
 
-The ``TLSServerConfiguration`` object is similar to the client one, except that
-it takes a ``Sequence[SigningChain]`` as the ``certificate_chain`` parameter.
+Server authentication is mandatory, at least one signing chain comprising a leaf
+server certificate including its corresponding private key and optionally a list
+of intermediate certificates. It is possible to set more than a single signing
+chain to support multiple virtual hosts.
```
jvdprng:
```suggestion
Server authentication is mandatory, so the configuration must include
at least one signing chain comprising a leaf server certificate including 
its corresponding private key and optionally a list of intermediate 
certificates. It is possible to set more than a single signing chain to 
support multiple virtual hosts. Insecure connections with server
authentication disabled are only possible via the :ref:`_insecure` module.
```

jvdprng (changes requested):
Thanks for the PR! I left two suggestions.

peps/pep-0748.rst
```diff
@@ -273,7 +276,7 @@ The ``TLSClientConfiguration`` class would be defined by the following code:
             inner_protocols: Sequence[NextProtocol | bytes] | None = None,
             lowest_supported_version: TLSVersion | None = None,
             highest_supported_version: TLSVersion | None = None,
-            trust_store: TrustStore | None = None,
+            trust_store: TrustStore = TrustStore.system(),
```
jvdprng, line 279: I remember now why we had `None` here before: we wanted to avoid creating a shared default object in the API which should be immutable/comparable. So I think we need something else here:
- A sentinel object (we deliberately chose not to do this IIRC)
- Do not define the default, and specify that a missing `TrustStore` should default to the system trust store. This is my preference.

peps/pep-0748.rst
jvdprng: Actually, we should specify that implementations should raise a `ValueError` when providing an empty `Sequence` for the signing chain.

jvdprng (changes requested):
On second thought, there is one additional thing we should change.

peps/pep-0748.rst
Julien00859: > Do not define the default, and specify that a missing TrustStore should default to the system trust store. This is my preference.

I don't understand, can you add an example?

Julien00859: > We wanted to avoid creating a shared default object in the API which should be immutable/comparable.

I think there is room for a `TrustStore.is_system(truststore)` method.

jvdprng: Yeah, what I was thinking of does not work. I think a larger problem is that `TrustStore` objects are currently not immutable. I would propose to make them immutable by freezing them (which means I need to fix some path handling in the `stdlib`, but that's probably fine) which would make it fine to assign the system trust store as a default parameter.

We could then consider making the configs frozen as well. What do you think?

Julien00859: Making the class immutable means that implementations that *require* one of `_path` or `_buffer` can't have the following:

```py
# example for an implementation that doesn't support paths
def load(trust_store):
    if trust_store._buffer:
        return
    if trust_store._path:
        with open(trust_store._path, ...) as file:
            trust_store._buffer = file.read()
        return
    raise ValueError("TrustStore.system() is not supported")
```

which is a pattern that is implemented in both tlslib.stdlib (buffer->path) and siotls (path->buffer).

I think we need to discuss the three `TrustStore`, `Certificate` and `PrivateKey` APIs more[^1][^2]. For the time being, I suggest we add a `TrustStore.is_system()` method to solve the problem at hand and have the discussion about immutability and path->buffer/buffer->path elsewhere.

```py
    def is_system(self):
        return self._buffer is None and self._path is None and self._id is None
```

+1 to have the configuration immutable. It already is a frozen dataclass in siotls. I'll push a commit to specify it in the spec.

[^1]: https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263/71: § Certificate Chain, Certificate Revocation List
[^2]: https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263/71: § Save DER in-place

Julien00859: No need to change the spec, it already says that the configuration objects are immutable (emphasis mine):

> For this reason, we split the responsibility of SSLContext into two separate objects, which are each split into server and client versions. The TLSServerConfiguration and TLSClientConfiguration objects act as containers for a TLS configuration: the ClientContext and ServerContext objects are instantiated with a TLSClientConfiguration and TLSServerConfiguration object, respectively, and are used to create buffers or sockets. **All four objects would be immutable.**

PR #4960, opened 2026-05-06
PEP 748: tlslib - context & socket

Julien00859:
### [PEP 748: create_connection and create_server instead of connect](https://github.com/python/peps/commit/785c0c4e0b4b78d4869c933cd28cce6a7273a349)

Prior discussions at:
* https://github.com/trailofbits/tlslib.py/pull/68
* https://github.com/trailofbits/tlslib.py/pull/75

That commit is the solution to my struggle binding an Ipv6 localhost address when I first tested tlslib. At the time the library was always binding the socket using `family=AF_INET`. The solution we implemented in tlslib is to perform a DNS request on the user-provided address to determine the family to use (`family=AF_INET6` in my case).

I'm still unhappy with this API server-side. "connect" is the wrong verb, it should be "bind". It is not obvious if other sockets are supported (e.g. Unix Domain Socket).

`create_connection` and `create_server` make so much more sense to me, at least as long as we don't have a `wrap_socket` function.

### [PEP 748: shutdown(show: 0|1|2) instead of close(force: bool)](https://github.com/python/peps/commit/3d40e2fe33caab5986a6a52c30d8c0126da157c5)

Prior discussions:
* https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263/72 (Close Notify part)
* https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263/76
* https://discuss.python.org/t/pre-pep-discussion-revival-of-pep-543-a-unified-tls-api-for-python/51263/79

Work in progress in siotls:
* https://codeberg.org/drlazor8/siotls/pulls/26

The messages I wrote on the forum explain the problems I have with `close(force: bool)`, which mainly boils down to "it feels too TLS 1.2-ish". I spent many days exploring ideas to have a clean API to terminate the TLS connection a way that is safe in both TLS 1.2 (synchronous termination) and TLS 1.3 (asynchronous and IMO better). Reusing `socket`'s `shutdown` terminology feels right to me. It works well for TLS 1.3 and is fine for TLS 1.2.

There are other things in the current Buffer & Socket spec that I find are leaking openssl's API, and that I expect won't work very well with other TLS libraries (including my own) but I'm not ready at this time to make new proposals.

### DIFF
--- a/peps/pep-0748.rst
+++ b/peps/pep-0748.rst
@@ -390,7 +390,7 @@ The ``ServerContext`` protocol class has the following class definition:
             ...
 
         @abstractmethod
-        def create_buffer(self, server_hostname: str) -> TLSBuffer:
+        def create_buffer(self) -> TLSBuffer:
             """Creates a TLSBuffer that acts as an in-memory channel,
             and contains information about the TLS exchange
             (cipher, negotiated_protocol, negotiated_tls_version, etc.)."""
--- a/peps/pep-0748.rst
+++ b/peps/pep-0748.rst
@@ -422,7 +422,7 @@ need to implement the following:
 
 * ``recv`` and ``send``
 * ``accept`` (server-side)
-* ``close``
+* ``shutdown`` and ``close``
 * ``getsockname``
 * ``getpeername``
 
@@ -437,6 +437,12 @@ connection, such as:
 The following code describes these functions in more detail:
 
 .. code-block:: python
+
+    class TLSShutdownMode(enum.IntEnum):
+        SHUT_RD = 0
+        SHUT_WR = 1
+        SHUT_RDWR = 2
+
 
     class TLSSocket(Protocol):
         """This class implements a socket.socket-like object that creates an OS
@@ -464,15 +470,41 @@ The following code describes these functions in more detail:
             ...
 
         @abstractmethod
-        def close(self, force: bool = False) -> None:
-            """Shuts down the connection and mark the socket closed.
-            If force is True, this method should send the close_notify alert and shut down
-            the socket without waiting for the other side.
-            If force is False, this method should send the close_notify alert and raise
-            the WantReadError exception until a corresponding close_notify alert has been
-            received from the other side.
-            In either case, this method should return WantWriteError if sending the
-            close_notify alert currently fails."""
+        def shutdown(self, how: TLSShutdownMode) -> None:
+            """
+            Shutdown TLS and the underlying socket.
+
+            * ``SHUT_RD`` (``0``) unilateraly close the receiving-side: discard
+              present and future unread messages.
+            * ``SHUT_WR`` (``1``) gracefully close the sending-side: send a
+              closing alert and prevent sending more messages.
+            * ``SHUT_RDWR`` (``2``): both ``SHUT_WR`` and ``SHUT_RD``.
+
+            .. danger::
+
+                Both ``shutdown(SHUT_RD)`` and ``shutdown(SHUT_RDWR)`` pose a
+                risk of data loss: they are unsafe unless the connection is
+                otherwise known to be over or when truncation is not an issue.
+
+                In TLS 1.2, the same risk applies also to ``shutdown(SHUT_WR)``.
+            """
+            ...
+
+        @abstractmethod
+        def close(self) -> None:
+            """
+            Close the underlying socket, but only when it is safe to do so.
+
+            When the sending-side of the connection is still open, it gracefully
+            closes the TLS sending-side before closing the socket, raising
+            ``WantWriteError`` when it fails.
+
+            When the receiving-side of the connection is still open, it always
+            raises ``WantReadError`` as there's a risk of data loss / truncation
+            attack. The user must first either: (safe) ``recv`` until the peer
+            closes its sending-side of the connection, or (unsafe) take the risk
+            and unilateraly ``shutdown`` the receiving-side of the connection.
+            """
             ...
 
         @abstractmethod
--- a/peps/pep-0748.rst
+++ b/peps/pep-0748.rst
@@ -351,7 +351,15 @@ The ``ClientContext`` protocol class has the following class definition:
             ...
 
         @abstractmethod
-        def connect(self, address: tuple[str | None, int]) -> TLSSocket:
+        def create_connection(
+            self,
+            address: tuple[str | None, bytes | str | int | None],
+            timeout: float | None = ...,
+            source_address: _Address | None = None,
+            *,
+            all_errors: bool = False,
+            sever_hostname: str | None = None,
+        ) -> TLSSocket:
             """Creates a TLSSocket that behaves like a socket.socket, and
             contains information about the TLS exchange
             (cipher, negotiated_protocol, negotiated_tls_version, etc.).
@@ -382,9 +390,18 @@ The ``ServerContext`` protocol class has the following class definition:
             ...
 
         @abstractmethod
-        def connect(self, address: tuple[str | None, int]) -> TLSSocket:
-            """Creates a TLSSocket that behaves like a socket.socket, and
-            contains information about the TLS exchange
+        def create_server(
+            self,
+            address: _Address,
+            *,
+            family: int = socket.AF_INET,
+            backlog: int | None = None,
+            reuse_port: bool = False,
+            dualstack_ipv6: bool = False,
+        ) -> TLSSocket:
+            """Creates a listening socket with an accept() function that returns
+            a TLSSocket that behaves like a socket.socket, and contains
+            information about the TLS exchange
             (cipher, negotiated_protocol, negotiated_tls_version, etc.).
             """
             ...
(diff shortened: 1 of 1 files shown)


hugovk:
cc PEP authors @jvdprng and @woodruffw.

peps/pep-0748.rst
```diff
@@ -351,7 +351,14 @@ The ``ClientContext`` protocol class has the following class definition:
...
 
         @abstractmethod
-        def connect(self, address: tuple[str | None, int]) -> TLSSocket:
+        def create_connection(
+            self,
+            address: tuple[str | None, int],
+            timeout=GLOBAL_DEFAULT,
+            source_address=None,
+            *,
+            all_errors=False,
```
jvdprng: Needs typing information

peps/pep-0748.rst
```diff
@@ -365,7 +372,45 @@ The ``ClientContext`` protocol class has the following class definition:
...
+
+        @abstractmethod
+        def create_server(
+            self,
+            address: tuple[str | None, int],
+            *,
+            family=AF_INET,
+            backlog=None,
+            reuse_port=False,
+            dualstack_ipv6=False,
```
jvdprng: Needs typing information

peps/pep-0748.rst
```diff
@@ -365,7 +372,45 @@ The ``ClientContext`` protocol class has the following class definition:
...
+        def configuration(self) -> TLSServerConfiguration:
+            """Returns the TLS server configuration that was used to create the server context."""
+            ...
+
+        @abstractmethod
+        def create_server(
+            self,
+            address: tuple[str | None, int],
+            *,
+            family=AF_INET,
```
jvdprng: Do we want to set `AF_INET` as the default here? We could have `None` that lets the backend choose based on the address?

peps/pep-0748.rst
```diff
@@ -351,7 +351,14 @@ The ``ClientContext`` protocol class has the following class definition:
             ...
 
         @abstractmethod
-        def connect(self, address: tuple[str | None, int]) -> TLSSocket:
+        def create_connection(
```
jvdprng, line 354: Should we add a `server_hostname` parameter here to support 'connect by IP, verify as DNS name'?

peps/pep-0748.rst
```diff
@@ -418,22 +463,64 @@ The following code describes these functions in more detail:
...
-        def close(self, force: bool = False) -> None:
-            """Shuts down the connection and mark the socket closed.
-            If force is True, this method should send the close_notify alert and shut down
-            the socket without waiting for the other side.
-            If force is False, this method should send the close_notify alert and raise
-            the WantReadError exception until a corresponding close_notify alert has been
-            received from the other side.
-            In either case, this method should return WantWriteError if sending the
-            close_notify alert currently fails."""
+        def shutdown(self, how: Literal[0, 1, 2]) -> None:
```
jvdprng: I'm worried that the current `shutdown` / `close` descriptions may be overspecified for the PEP. Should we focus on describing the observable behavior?
- `shutdown(SHUT_WR)` gracefully closes the TLS sending side
- `shutdown(SHUT_RD)` and `shutdown(SHUT_RDWR)` may discard unread peer data and are unsafe unless truncation is not an issue
- `close()` should only fully close the transport when safe, otherwise it should raise `WantReadError`/`WantWriteError` as appropriate

Also, should we define a specific `enum` for `how`? Using literals might be brittle.

peps/pep-0748.rst
```diff
@@ -365,7 +372,45 @@ The ``ClientContext`` protocol class has the following class definition:
...
+        @abstractmethod
+        def create_server(
+            self,
+            address: tuple[str | None, int],
+            *,
+            family=AF_INET,
+            backlog=None,
+            reuse_port=False,
+            dualstack_ipv6=False,
+        ) -> TLSSocket:
```
jvdprng, line 401: Do we want stronger typing for the TLS sockets? Both 'listening sockets' and 'connected sockets' are still the same type.

jvdprng (changes requested):
Thanks for writing this up! 

I feel like this pulls a lot of socket API surface into the PEP. Couldn't we stick to a minimum API surface that leaves a bunch of socket options to the backends? E.g., something like: 
```
  class ClientContext(Protocol):
      def connect(
          self,
          address: tuple[str | None, int],
          *,
          server_hostname: str | None = None,
          timeout: float | None = None,
      ) -> TLSSocket: ...

  class ServerContext(Protocol):
      def create_server(
          self,
          address: tuple[str | None, int],
          *,
      ) -> TLSListener: ...
```

I'm not sure whether splitting out 'listening sockets' and 'connected sockets' is a good idea, but it would solve some of the current blurriness around listen/accept. In general, I'm hesitant to pull in a bunch of socket stuff without discussing this on the PEP boards first.

If we do want to go down this route, we should probably also specify in the PEP that we want to focus on `INET` sockets. I've left some comments regarding cleaning up types in that case.

The shutdown changes are definitely needed, but I feel like the current PR overspecifies a little.

peps/pep-0748.rst
Julien00859: +1 about the over-specification, your shorter text is better indeed

Julien00859: The best would indeed to type `how` using an enum,  but I'm afraid such an enum doesn't exist. The right place for such an enum would be `socket`, but in `socket` there already are the three global integers `socket.SHUT_RD/WR/RDWR` and deprecating those in favor of an enum doesn't seem right to me.

Maybe we can type if using `Literal[socket.SHUT_RD, socket.SHUT_WR, socket.SHUT_RDWR]` instead of `[0, 1, 2]`. Not sure how it would render in a sphinx/pdoc documentation though :(

peps/pep-0748.rst
Julien00859: I decided to implement them using stronger types in siotls, but IMO we need not to enforce it in the spec.

peps/pep-0748.rst
Julien00859: I only replicated the `socket.create_server` API which uses `family=AF_INET` by default. To me it is either we stick precisely to the same signature as `socket.create_server`, or we use another function name.

peps/pep-0748.rst
Julien00859: Agreed, and we need to add the typing information in the original `socket.create_connection` as well.

peps/pep-0748.rst
Julien00859: Agreed, and we need to add the typing information in the original `socket.create_server` as well.

peps/pep-0748.rst
Julien00859: > Maybe we can type if using `Literal[socket.SHUT_RD, socket.SHUT_WR, socket.SHUT_RDWR]` instead of `[0, 1, 2]`. Not sure how it would render in a sphinx/pdoc documentation though :(

Doesn't work :(

```
$ uv run ty check src/siotls/socket.py
error[invalid-type-form]: Type arguments for `Literal` must be `None`, a literal value (int, bool, str, or bytes), or an enum member
   --> src/siotls/socket.py:178:44
    |
176 |             logger.debug('%s/%s bytes sent', out_data_sent, len(out_data))
177 |
178 |     def shutdown(self, how: typing.Literal[socket.SHUT_RDWR]) -> None:
    |                                            ^^^^^^^^^^^^^^^^
179 |         """
180 |         Shutdown TLS and the underlying socket.
    |
info: rule `invalid-type-form` is enabled by default

Found 1 diagnostic
```

jvdprng: Why should the enum be in socket? I think it's actually better to define something like a `TLSShutdownMode` enum here, because the first step is to close the TLS connection and the second is to close the socket. Each backend is then free to determine how to close the socket.

peps/pep-0748.rst
Julien00859: Used the typing from typeshed https://github.com/python/typeshed/blob/fcc86fc62ddc8439b5ec96cee787cd9de9e86bbc/stdlib/socket.pyi#L1505

peps/pep-0748.rst
Julien00859: Used the typing from typeshed https://github.com/python/typeshed/blob/fcc86fc62ddc8439b5ec96cee787cd9de9e86bbc/stdlib/socket.pyi#L1491

peps/pep-0748.rst
Julien00859: I can't help but think we are duplicating the information, would you agree to wait for a third opinion?

Julien00859: I've pushed a shorter docstring for `shutdown()`.

I decided to keep the docstring of `close()` mostly as it was. That `close()` raises exceptions in TLSSocket when it does not with regular sockets is gonna be a big surprise to users. I think it warrant a lengthier docstring, and the current one is IMO a good fit.

The commit is in a `fixup!` because I spent too many hours coming with the original one, that's my way of not loosing my work :smiling_face_with_tear:. It'll will eventually be squashed don't worry.

PR #4964, opened 2026-05-08
Add Diego as author of PEP 831

diegorusso:


hugovk:
Please also mark as Final and link to the canonical docs, see: https://github.com/python/peps/blob/main/.github/PULL_REQUEST_TEMPLATE/Mark%20a%20PEP%20as%20Final.md

PR #4968, opened 2026-05-10
PEP 661: Amend to allow two customizations

JelleZijlstra:
These are changes I'd like to make during the beta phase (I'll post asking for SC approval).

* Change is either:
    * [ ] To a Draft PEP
    * [x] To an Accepted or Final PEP, with Steering Council approval

### DIFF
--- a/peps/pep-0661.rst
+++ b/peps/pep-0661.rst
@@ -277,7 +277,9 @@ C API
 Sentinels can also be useful in C extensions. We propose two new
 C API functions:
 
-* ``PyObject *PySentinel_New(const char *name, const char *module_name)`` creates a new sentinel object.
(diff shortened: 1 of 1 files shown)


taleinat:
Both of these, customizable repr and setting `__module__`, were in earlier versions of the PEP and make sense.

Regarding making repr customizable by passing in an argument, LGTM.

Regarding making `__module__` writeable, I find that a bit surprising. Have you considered adding another argument to `sentinel()`? That would be more in line with namedtuple and Enum which do this for similar reasons, and be more in line with sentinels being immutable. (I see Kyle made a similar comment in the PEP discussion.)

JelleZijlstra:
The SC approved this change.

PR #4969, opened 2026-05-11
PEP 825: Various clarifications

konstin:
No functional changes, but clarifications to the variant ordering algorithm, index support and scope.

* Change is either:
    * [x] To a Draft PEP


peps/pep-0825.rst
```diff
@@ -308,11 +308,17 @@ there MUST exist a corresponding ``{name}-{version}-variants.json``
...
+
+The exact URL where the file is hosted is insignificant, but it MUST
+be provided in all the responses where the variant wheels are included.
+It should follow the rules for files in the
+:ref:`packaging:simple-repository-api`, except that the various metadata
+served by the index (such as ``core-metadata``, ``dist-info-metadata``,
+``requires-python`` or ``yanked``) are not meaningful for that file.
+Indexes MAY publish or skip these attributes, as long as the values do
+not prevent correct operation. Tools MAY either use or ignore these
+values.
```
pfmoore: If you're putting `variant.json` files in the `files` section of the JSON response, then you need to follow the existing rules, which say that for each `files` entry, `filename`, `url`, `hashes` and `size` are mandatory, and all other keys are optional. There's nothing in the existing spec that suggests that tools can ignore attributes if they are present, and I don't think we should be special-casing `variant.json` files.

I actually don't understand why you're allowing these fields to be present at all. What's wrong with simply saying that `variant.json` files must not include any of the optional fields defined in the index spec (with the exception of the `upload-time` field, which may be present)[^1]?

Honestly, this feels like a problem caused by trying to put `variant.json` files into a field that's intended to hold details of distribution files. I'd argue that a cleaner design would be to have a separate `variant-data` field in the project details response[^2].

[^1]: I'm only making an exception for upload time because I can see cases where it might be useful, in analysing index data.
[^2]: The HTML version of the index would still need to have all the links together, but that should be viewed as secondary, as per PEP 833 (which is not yet approved, but is likely to be).

peps/pep-0825.rst
```diff
@@ -379,17 +385,20 @@ like:
...
 For the purpose of ordering, variant properties are grouped into
 features, and features into namespaces. For every namespace, the tool
-MUST obtain an ordered list of compatible features, and for every
-feature, an ordered list of compatible values. The method of obtaining
-these lists will be defined in a subsequent PEP.
-
-The default ordering MUST be performed equivalent to the following
+MUST obtain a list of compatible features, and for every feature, a list
+of compatible values. The method of obtaining these lists will be
+defined in a subsequent PEP. The items in these lists will be provided
```
pfmoore, line 395: "The method of obtaining these lists is not defined (and hence is tool-specific)."

We should avoid making this PEP explicitly dependent on "future PEPs", as if we do so it's impossible to approve this PEP on its own merits.

peps/pep-0825.rst
pfmoore, line 406: This ordered list is not well defined if multiple indexes are involved.

peps/pep-0825.rst
```diff
@@ -402,9 +411,10 @@ algorithm:
       value of the respective ``default-priorities.feature.{namespace}``
       key.
```
pfmoore, line 412: Same problem here.

peps/pep-0825.rst
```diff
@@ -415,48 +425,49 @@ algorithm:
       of the respective
       ``default-priorities.property.{namespace}.{feature_name}`` key.
```
pfmoore, line 426: And here.

peps/pep-0825.rst
```diff
@@ -415,48 +425,49 @@ algorithm:
...
    every feature. This is ``value_order`` in the example.
 
-4. For every compatible variant, determine the most preferred value
-   corresponding to every feature in that variant. This is done by
-   finding among the values present in the variant properties the one
-   that has the lowest position in the ordered property value list.
-   After this step, a list of features along with their best values
-   is available for every variant. This is done in the
-   ``Variant.best_value_properties()`` method in the example.
+4. For every group, determine the most preferred value corresponding to
```
pfmoore, line 435: I don't understand this step. A group, as I understand it, is a set of wheels with the same variant label. So each wheel in a group has the *same* value for every property. So how can there be a "most preferred value" when there's only one value?

I suspect I'm misunderstanding here because the terminology still isn't clear to me. But I'm more concerned with making sure that the reader can follow what's being described than I am arguing about terms.

Or is this actually about trying to order the different *groups*, so you're trying to identify the most preferred *group* based on the values the labels denote?

peps/pep-0825.rst
```diff
@@ -737,6 +749,10 @@ Note that steps 4. through 8. are introduced specifically for variant
 wheels. The remaining steps correspond to the current installer
 behavior.
 
+When installing from a source that does not provide an `index-level
+metadata`_, the same algorithm can be used, except that the variant
+metadata needs to be read directly from the wheels.
```
pfmoore, line 754: This doesn't feel like it's true. If you have 2 wheels served from a local directory, with different `variant.json` files in the wheels, doesn't that put you in the same situation as if you had multiple indexes (which, as I note, the algorithm doesn't cover)?

peps/pep-0825.rst
```diff
@@ -817,15 +873,25 @@ variant may require a different minimal libc version, in which case the
...
-corresponding to their preference, much like Platform compatibility tags
-conventionally are. The variant metadata provides the ability to
-override this order at package level. However, namespaces are unordered
-by design (e.g. we will not decide upfront which GPU vendors take
-precedence) and therefore they always need to be ordered by the package
-maintainer.
+While a future PEP will define how variant properties are provided, a
+baseline assumption is made that the compatible properties will be
+provided in specific order corresponding to their preference. This makes
+it possible to use a generic sorting algorithm that, and later define
```
pfmoore: Typo? "a generic sorting algorithm that, and later define"...

peps/pep-0825.rst
```diff
@@ -817,15 +873,25 @@ variant may require a different minimal libc version, in which case the
...
-keep the specification easier to comprehend, a baseline assumption is
-made that the compatible properties will be provided in specific order
-corresponding to their preference, much like Platform compatibility tags
-conventionally are. The variant metadata provides the ability to
-override this order at package level. However, namespaces are unordered
-by design (e.g. we will not decide upfront which GPU vendors take
-precedence) and therefore they always need to be ordered by the package
-maintainer.
+While a future PEP will define how variant properties are provided, a
+baseline assumption is made that the compatible properties will be
```
pfmoore, line 877: This shouldn't be an assumption, it should be explicitly required in *this* PEP. Then this sentence can be reworded something along these lines:

This PEP simply requires that compatible properties are provided in a specific order corresponding to their preference. We do not state how tools will provide this ordered list, but a future PEP is planned which will standardise the mechanism.

peps/pep-0825.rst
```diff
@@ -817,15 +873,25 @@ variant may require a different minimal libc version, in which case the
...
+provided in specific order corresponding to their preference. This makes
+it possible to use a generic sorting algorithm that, and later define
+properties as data without having to change the algorithm.
+
+A future PEP will define how the ordering for features and values is
+provided. However, namespaces are governed independently and considered
```
pfmoore, line 883: Same here. You need to say what *this* PEP requires, so that tools can (in principle) implement their own mechanism for providing an ordering. The fact that we'll standardise that mechanism in a future PEP isn't important here, except as context.

peps/pep-0825.rst
```diff
@@ -817,15 +873,25 @@ variant may require a different minimal libc version, in which case the
...
+on equal footing, and therefore there will be no standard ordering for
+them. Instead, the package maintainer will decide which namespaces have
```
pfmoore: Conversely, here you're saying "the package maintainer will decide", but you haven't specified how they record those decisions in the variant metadata. Is this information recorded by manually editing the `package.json` files, or is it held somewhere else?

peps/pep-0825.rst
```diff
@@ -938,7 +1004,7 @@ Reference Implementation
 The `variantlib <https://github.com/wheelnext/variantlib>`__ project
 contains a reference implementation of a complete variant wheel
 solution. It is compliant with this PEP, but also goes beyond it,
-providing example solutions to `open issues`_.
+providing example solutions to some of deferred items.
```
pfmoore:
```suggestion
providing example solutions to some of the deferred items.
```

peps/pep-0825.rst
```diff
@@ -982,15 +1048,34 @@ would be incorrectly deemed compatible because of the
...
+------------
 
-The following problems are deferred to subsequent PEPs:
+The following problems are deferred to subsequent PEPs in the series:
 
 - governance of variant namespaces
 - determining which variant properties are compatible with the system
 - building variant wheels
 
+In addition to that, the following questions are left undefined:
```
pfmoore: The question is well-defined, what you are saying is that you aren't going to *answer* the question.

Which means that this PEP doesn't support installation from multiple sources of wheels. I don't think that's an acceptable position to take, as installing from multiple sources is an *extremely* common scenario, and if this PEP doesn't support it, that's a significant usability problem.

IMO, what you need to do is identify all of the places in the PEP where having wheels from multiple sources would cause an issue (I've noted a few of them above) and explicitly state in those places that the PEP doesn't define behaviour in that situation[^1]. That changes the position from "multiple indexes are out of scope" to "behaviour for multiple indexes is tool-defined".

[^1]: My gut feeling is that if you do that, the variant ordering algorithm will look a bit of a mess. If that's the case, then that's a good thing, honestly, as it will expose the fact that the PEP relies on some form of index ordering mechanism. Procedurally, that's a problem, as standardising an index ordering mechanism is a whole other can of worms. But I'd rather we faced that issue than simply ignoring it and ending up with a standard that can't be implemented properly by tools that don't support index priority.

pfmoore (changes requested):

peps/pep-0825.rst
konstin: > Which means that this PEP doesn't support installation from multiple sources of wheels. I don't think that's an acceptable position to take, as installing from multiple sources is an extremely common scenario, and if this PEP doesn't support it, that's a significant usability problem.

We're specifically say "undefined", not "not supported". Tools can support multiple index in any shape they want, by virtue of the spec saying nothing about it.

> That changes the position from "multiple indexes are out of scope" to "behaviour for multiple indexes is tool-defined".

Can you explain what the difference between those two is? Currently, there is no spec for multiple index support: Multiple indexes are out of scope in any other PEP, we're just writing it out explicitly because otherwise people complained about the absence of that, and it's still tool-defined, as tool make choices about multiple indexes right now, in the absence of a spec.

peps/pep-0825.rst
konstin: I don't really see that contradiction, I think we have to live with some of that tension if we want to avoid having one massive PEP 817. There's other PEPs too which specifically leave gaps for future specifications, and I don't see that as fundamentally different. If we say it's tool-specific, that would be rather misleading, as we don't want to open up the namespace to everyone beyond experimentation, we have the expectation that in the end we'll have a compatibility standard where everyone speak the same language.

peps/pep-0825.rst
konstin: See my comment about multiple indexes at the bottom.

peps/pep-0825.rst
pfmoore: "Multiple indexes are out of scope" means that an implementation that doesn't allow multiple sources for variant wheels is PEP compliant. That seems to me like it would be a severe limitation on the usability of variants, especially as it would effectively disallow installing from a local directory of wheels (because in the absence of an index, each wheel is its own independent source of variant data).

On the other hand, "behaviour for multiple indexes is tool-defined" means that PEP-compliant implementations need to support multiple indexes/sources, but they get to choose the behaviour in those specific places where the PEP says that behaviour is undefined.

I appreciate that it's a subtle distinction, but IMO it's important. The precise wording doesn't matter to me, but the difference in semantics *does*.

> Multiple indexes are out of scope in any other PEP

I don't know of any other PEP/standard where it matters whether multiple indexes are involved[^1]. This PEP is different precisely because the ordering algorithm is described in terms that assume a single `variant.json` file is applicable to all wheels being considered, and that means it *does* matter if the wheels come from multiple places.

[^1]: If you're aware of any, please let me know and I'll check how they handle the situation.

peps/pep-0825.rst
pfmoore: > I think we have to live with some of that tension if we want to avoid having one massive PEP 817.

Agreed (although we have to accept that if the tension is too great, a single PEP 817 may be a better approach). The key here is that we have a way to *implement* PEP 825 in the absence of those other PEPs. If we can't start implementation work once PEP 825 is (provisionally) accepted, there's no point in accepting it independently. So regardless of what you might prefer, anything not specified in PEP 825 *will be* tool-defined. When the follow-up PEP XXX comes along, people who correctly guessed what it would say will be ready, and everyone else will have to change their implementation, but we can't know what the answer will be in advance. I'd rather we made it clear that's the situation, rather than having people think they can't start implementation work because parts of the behaviour need a further PEP to specify them.

> There's other PEPs too which specifically leave gaps for future specifications

Maybe I'm getting too concerned about this. Can you give me some examples and I'll check how they handle the situation?

peps/pep-0825.rst
mgorny: Yes, "optional fields" is what we meant. We'll correct this.

For the rest, see rationale. The goal is to support indexes during the transitional period with minimal modifications necessary. Should the tools be rejecting projects if the index gives unnecessary-yet-harmless fields?

peps/pep-0825.rst
mgorny: See "variant properties". A single wheel can have multiple values corresponding to a single feature. You select the most preferred one from them for ordering.

If the wheel has:

```
nvidia :: sm_arch :: 120_real
nvidia :: sm_arch :: 110_real
```

you would sort only on:

```
nvidia :: sm_arch :: 120_real
```

peps/pep-0825.rst
mgorny: It's a single source, so the same rules of consistency as in "index-level metadata" apply, and the same merging algorithm described there.

peps/pep-0825.rst
mgorny: Shouldn't the rationale section be describing why were the specific decisions made in the specification part, rather than stating what is required (I.e. effectively repeating specification)? I dare say the "context" is the whole point of having a rationale in the first place.

peps/pep-0825.rst
mgorny: For the wheel file, it's "Variant metadata → Default priorities". For how the data ends up in that file, subsequent PEP concerning building packages (which should literally mirror the data in `pyproject.toml`, but saying that right now will only extend the PEP beyond its scope).

peps/pep-0825.rst
mgorny: Could you point us to a document explaining this difference?

I don't really understand how "tool-defined" implies that something has to be implemented. Isn't "reject as not supported" also a valid tool-defined behavior?

Also, does this mean that a tool that currently does not support multiple sources at all won't be able to be compliant with variant wheels without actually implementing support for multiple sources, i.e. something it currently didn't need to do when installing regular wheels?

pfmoore: No, I can't point you to a document. This isn't a technical term with a precise definition, it's a matter of how the PEP reads to me. And by extension, how I believe it could be read in the future by others - I'm basing this on my experience with *other* standards where we've had problems after the fact because people have interpreted imprecise wording in a way that we hadn't originally intended.

I don't think this side discussion is helping. It's a minor point in isolation, and it's threatening to overwhelm the main issue I'm trying to draw attention to, which is that I don't see how we can simply ignore (or "leave to tool implementers" if you prefer) the question of how to calculate ordering when faced with multiple sources of information to combine. At the end of the day, the PEP will need to address this issue if it's going to be accepted. I've tried to read the ordering algorithm as described in the PEP multiple times now, and every time I've got stuck, usually because I can't work out how I'd generalise the given algorithm to handle multiple sources (something I expect to have to do for pip, where "pick the first index" isn't a valid option unless we can assume all indexes have the same data).

To be very explicit here (please don't take this as any sort of threat, it's not intended that way, but I do want to be sure there's no misunderstanding), I'm currently not seeing a way I can accept the PEP if it doesn't *somehow* clarify how tools are supposed to handle multiple sources. I'm trying to help fix that problem, but if my comments are having the opposite effect, I should probably just state the issue and leave it to you to decide how to take things forward.

The PEP doesn't say who the sponsor is (that's something that should be fixed, by the way), but I'd recommend you speak to them as well - they should be able to help navigate the process around incorporating feedback like this, if it's a problem.

peps/pep-0825.rst
pfmoore: Sure, if you want to move this "baseline assumption" into the specification section as a requirement, that works for me.

peps/pep-0825.rst
pfmoore: OK, let's reword this as

> on equal footing, and therefore there will be no standard ordering for
> them. Instead, the ordering of namespaces will be explicitly stated in
> the `variants.json` file (which is provided by the package maintainer as
> part of the build process).

That clarifies that the "decision" is simply a matter of supplying the correct data in the appropriate file, and not some yet-to-be-defined process. You could reword the next sentence similarly, as *overriding* the ordering is actually just *amending* that file.

Basically, by being a bit more explicit, you can avoid some uncertainty. That's all, really.

peps/pep-0825.rst
hugovk: > The PEP doesn't say who the sponsor is (that's something that should be fixed, by the way), but I'd recommend you speak to them as well - they should be able to help navigate the process around incorporating feedback like this, if it's a problem.

A sponsor isn't needed here because Barry and Donald are co-authors and on the core team, so you can speak to them.

peps/pep-0825.rst
mgorny: Specifically note the provision:

> Indexes MAY publish or skip these attributes, as long as the values do not prevent correct operation. Tools MAY either use or ignore these values.

This is intended as "you don't have to go out of your way to special case that one file", as long as you don't do something really wrong there.

peps/pep-0825.rst
mgorny: The whole point is that you _can_ start the implementation (provided that you're ready to take the risk that the details will change). We say "subsequent PEP" in very specific places, telling you to leave specific gaps. Like "you can implement most of the sort algorithm, just leave the gap for functions that will provide these lists"; even the example below literally provides an implementation with gaps for these functions.

I don't think "guessing" really works here. I suppose you can infer what the implementation will be from PEP 817, or you can do your own thing. Whether this makes you PEP 825 compliant is unclear to me, since the PEP by design is part of the larger series, and therefore full compliance implies implementing all of them. A tool that implements PEP 825 but then diverges from subsequent PEPs does not really implement "variant wheels"; it implements a custom solution that is partially based on "variant wheels" but it is definitely not compliant with the design as a whole, and therefore it is not guaranteed to be interoperable.

peps/pep-0825.rst
mgorny: But it literally does that, in the fragment you've commented on above:

> The items in these lists will be provided in specific order that will impact variant wheel ordering.

We're merely trying to provide a bit more context of why this requirement is baked in there.

peps/pep-0825.rst
pfmoore: OK. Let me think about this some more. The idea was that we split PEP 817 into independent parts, but if it's simply not possible to make those parts independent, then maybe we're better accepting that and going back to the single unified approach with PEP 817. Or maybe this approach of having a series of PEPs that don't work independently, but which need to be accepted or rejected as a whole is the right approach - but I don't know of any prior occasions when this was done, so I'd need to do some research to decide how viable it is.

mgorny: I personally think split PEPs are better, but rather than independent, building one on top of another. If anything, they _are_ easier to read, and I found it easier to describe the concepts that way. It also keeps the discussion more focused.

pfmoore: I just looked back at the PEP 817 discussion, where splitting the PEP was first discussed. At the time, I said:

> I would strongly prefer that the mechanism for recording variant properties within wheels be separated out into its own PEP, leaving everything about how that data is used to other PEPs.

With that as context, I'll note that:

1. Ordering is arguably about *using* the data, and could reasonably be removed from PEP 825, along with all of the discussion about selecting wheels.
2. The index-level `variant.json` files, along with the need to merge those files from multiple sources, could reasonably be split into its own PEP - it *is* about how data is stored, but it's not fundamental, it's an optimisation (a *necessary* optimisation, because without the index-level data the mechanism would be too costly to be practical, but still an optimisation). If, on the other hand, you want to keep it here, I think you're going to have to tackle the issue of merging data from multiple sources, because that *is* part of "how the data is stored" if the index-level files are[^1]. I'd prefer to keep this as part of this PEP, but I'd understand if you prefer not to tackle index merging right now.

That would leave PEP 825 as basically just a set of terminology definitions, a JSON data schema, a wheel naming convention and a definition of where the variant data is stored in the wheel. That, IMO, would be a good foundation, and should[^2] be uncontroversial.

It doesn't do anything to address the issues that are blocking us here, all it does is put them into their own PEPs. So I'd completely understand if you don't want to do that (writing more PEPs is more work). But if what matters to you is getting *something* approved sooner rather than later, then it might help with that.

[^1]: If you have a set of wheels from 2 indexes, then "how the data is stored" is "in 2 files, one from each index", and "how do I access that data" involves combining the 2 files.
[^2]: Fingers crossed. Every time we've said that so far it's proved wrong. I hope I'm not jinxing things here!

> If anything, they are easier to read

I agree 100% that smaller PEPs make understanding the proposal easier - in a sense, they are acting like chapters in a larger document. What I'm not sure about is whether it makes sense to approve individual PEPs from a series structured like that (given that you've said you can't make them completely independent). If you're OK with writing the various PEPs and having them all remain in Draft status until they are submitted as a group for approval, I'd have no problem with that. But I got the distinct impression that you want an early approval of at least part of the work, for motivational reasons if nothing else.

Would it help if I said formally that I support the *idea* of wheel variants, but until the full set of PEPs is available and the proposal has been discussed and is documented in its full form (across multiple PEPs) I won't be approving individual parts?

mgorny: "Ordering" is a bit of unclear fit indeed. I've included it in PEP 825 since it can be defined independently of actual data (with the minimal gaps for input), and I've figured out that the next PEP will be big enough to justify keeping it out of it. On the other hand, I suppose moving it later would have the advantage of giving more context _why_ things were designed this way (not really relevant to specification but to rationale). That said, I'm not convinced that moving stuff at this point is really worth the effort.

pfmoore: I think that the *data structure* for ordering data is reasonable to include here, but manipulation of it (merging orderings, overriding, using it to select wheels) could be deferred. But I do agree that deferring just puts off the problem, it doesn't make it go away. So I'm happy to keep it here and address the issues now.

Also, as I noted on DPO, my thinking regarding the index-level files has changed somewhat. They are only an optimisation for the label->property mapping data. For ordering data, they aren't, because you don't define merging of ordering data from multiple sources. So a consumer *can't* go to the ordering data in the wheels, because they then can't merge that data, and so can't establish a single ordering for all of the wheels on the index.

As an aside, one point I'm not clear on - is it right that a project could legitimately provide *no* ordering data at all, relying solely on the default ordering that the providers give?

peps/pep-0825.rst
rgommers: > Yes, "optional fields" is what we meant. We'll correct this.

This was done in commit https://github.com/python/peps/pull/4969/changes/2dbbf45e463057361362cb3720cccd2dd81b83ab. Other parts of the review comment seem answered.

peps/pep-0825.rst
rgommers: > So I'm happy to keep it here and address the issues now.

This conversation seems resolved.

> As an aside, one point I'm not clear on - is it right that a project could legitimately provide no ordering data at all, relying solely on the default ordering that the providers give?

@mgorny answered this [here](https://discuss.python.org/t/pep-825-wheel-variants-package-format-split-from-pep-817/106196/128): `default-priorities.namespace` is required.

peps/pep-0825.rst
rgommers: Same comment as higher up; this PEP requires `default-priorities.namespace`.

peps/pep-0825.rst
rgommers: Seems resolved.

peps/pep-0825.rst
rgommers: Agreed, consistency rules apply.

peps/pep-0825.rst
rgommers: This question seems answered; the specification text seems fine. If a small rephrasing would have helped, a comment with the exact suggested rephrase could be applied. I'd say it's fine as is though; specification text doesn't get more clear by pre-emptively explaining questions like these (assuming the spec is indeed unambiguous, which it is here).

peps/pep-0825.rst
rgommers: Agreed with the comment; whether multiple indexes is defined as "out of scope" or "tool-specific" doesn't matter here, this comment can be resolved as not relevant either way.

peps/pep-0825.rst
rgommers: > To be very explicit here (please don't take this as any sort of threat, it's not intended that way, but I do want to be sure there's no misunderstanding), I'm currently not seeing a way I can accept the PEP if it doesn't somehow clarify how tools are supposed to handle multiple sources

I'll post about this on Discourse, because it's a content/process discussion either way. This PR just contains clarifications which all seem ready to go, and address some questions and confusion.

rgommers (approved):
I went through all open comments; it doesn't seem like there is anything to change right now. Given that this PR is meant as clarifications and textual improvements to the currently published PEP text, it looks to me like it can be merged as is. Open comments about textual details have been answered. More substantial content discussions belong on Discourse.

dstufft (approved):

PR #4972, opened 2026-05-16
PEP 821: Refinement from Feedback

Daraan:
Updated draft to V2


hugovk:
The 3.15 feature freeze has passed, I suggest updating to `Python-Version: 3.16`

PR #4973, opened 2026-05-18
PEP 783: Mark as final

hoodmane:

PR #4977, opened 2026-05-19
Infra: Move "Source" and "Last modified" from `<article>` to `<footer>`

hugovk:
Looking at PRs like https://github.com/python/peps/pull/4973, the Read the Docs comment says something like:

> 701 files changed · ± 701 modified

And lists _all the PEPs_. This is because they contain a "Last modified" timestamp in the footer. 

For the RTD build, we're not fetching the whole commit history, so it's setting the last modified time as a recent time rather than the real last modified time. This doesn't really matter for the previews, better to have faster builds.

Right now, the PEP html structure is like:

```html
<header></header>
<article>
  <section id="pep-content">/<section>
  <hr>
  <p>Source: ...</p>
  <p>Last modified: ...</p>
</article>
<nav id="pep-sidebar"></nav>
```

We could configure RTD to ignore the bit of HTML containing the timestamp, but RTD already ignores changes in a `<footer>`, and these things also semantically belong in a `<footer>`, which should also improve navigation for screen readers and so on.

Therefore this PR moves "Source" and "Last modified" to a `<footer>`:

```html
<header></header>
<article>
  <section id="pep-content">/<section>
</article>
<nav id="pep-sidebar"></nav>
<footer>
  <p>Source: ...</p>
  <p>Last modified: ...</p>
</footer>
```

Also show the timestamp as "UTC" not "GMT": https://www.timeanddate.com/time/gmt-utc-time.html

### DIFF
--- a/pep_sphinx_extensions/__init__.py
+++ b/pep_sphinx_extensions/__init__.py
@@ -23,6 +23,7 @@ from pep_sphinx_extensions.pep_processor.parsing import (
     pep_parser,
     pep_role,
 )
+from pep_sphinx_extensions.pep_processor.transforms import pep_footer
 from pep_sphinx_extensions.pep_processor.transforms import pep_references
 from pep_sphinx_extensions.pep_zero_generator.pep_index_generator import create_pep_zero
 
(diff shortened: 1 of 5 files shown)


hugovk:
> 735 files changed · ± 735 modified

Hmmm... 🤣

hugovk:
And of course, `main` still has timestamps in `<article>`, not `<footer>`, so we'll have to merge this first.

hugovk:
Merging this to check if it fixes the preview links.

PR #4979, opened 2026-05-20
PEP 827: Split whether a param has a default out from "quals"

msullivan:
Q/.quals becomes K/.kind, and we add D/.default to track whether there
is a default. It should be the type of the default if it exists,
otherwise `Never`. Tracking the type here allows specifying a
`Literal` in order to track a specific value, similar to what we
support for `Member`.

There is a possible follow-up to use an enum for the `ParamKind`
instead of string literals but I'm not certain about it yet.

* Change is either:
    * [x] To a Draft PEP

### DIFF
--- a/peps/pep-0827.rst
+++ b/peps/pep-0827.rst
@@ -8,7 +8,7 @@ Status: Draft
 Type: Standards Track
 Topic: Typing
 Created: 27-Feb-2026
-Python-Version: 3.15
+Python-Version: 3.16
 Post-History: 02-Mar-2026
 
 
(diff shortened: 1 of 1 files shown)


peps/pep-0827.rst
```diff
@@ -419,17 +438,18 @@ as::
...
+            Param[Literal["e"], int, Literal["keyword"], Literal[0]],
             Param[None, int, Literal["**"]],
         ],
         int,
     ]
 
 
-or, using the type abbreviations we provide::
+or, using the type abbreviations we provide (though this version will not track
+specific values for the defaults)::
```
dnwpark, line 454: This wording is a bit confusing. Anyways, maybe we should just add the default as a param with `Never` defaults.

dnwpark (approved):

peps/pep-0827.rst
```diff
@@ -376,18 +376,37 @@ this PEP.
...
-    type ArgsParam[T] = Param[Literal[None], T, Literal["*"]]
-    type KwargsParam[T] = Param[Literal[None], T, Literal["**"]]
+    type NamedDefaultParam[N: str, T] = Param[N, T, Literal["keyword"], T]
+    type ArgsParam[T] = Param[None, T, Literal["*"]]
+    type KwargsParam[T] = Param[None, T, Literal["**"]]
+
+
+The argument ``K``, of type ``ParamKind``, represents the parameter
+kind of the parameter. The default "positional or named" kind of
+argument is represented as ``Never``. It is an error to create
```
JelleZijlstra: seems like a strange use of Never. I would try to align these kinds with the ones in inspect: https://docs.python.org/3.15/library/inspect.html#inspect.Parameter.kind

msullivan: I have a planned follow-up to change the literals to actually be enums modeled directly after `Parameter.kind` that I wanted to keep separate from this. But I guess I can make t he default one "positional-or-named" now also while still strings

1st1 (approved):

PR #4980, opened 2026-05-20
PEP 808: Mark as Accepted

henryiii:
* [x] SC/PEP Delegate has formally accepted/rejected the PEP and posted to the ``Discussions-To`` thread
* [x] Pull request title in appropriate format (``PEP 123: Mark as Accepted``)
* [x] ``Status`` changed to ``Accepted``/``Rejected``
* [x] ``Resolution`` field points directly to SC/PEP Delegate official acceptance/rejected post, including the date (e.g. `` `01-Jan-2000 <https://discuss.python.org/t/12345/100>`__ ``)
* [ ] Acceptance/rejection notice added, if the SC/PEP delegate had major conditions or comments
* [x] ``Discussions-To``, ``Post-History`` and ``Python-Version`` up to date

### DIFF
--- a/peps/pep-0808.rst
+++ b/peps/pep-0808.rst
@@ -5,12 +5,13 @@ Author: Henry Schreiner <henryschreineriii@gmail.com>,
 Sponsor: Filipe Laíns <lains@python.org>
 PEP-Delegate: Paul Moore <p.f.moore@gmail.com>
 Discussions-To: https://discuss.python.org/t/104883
-Status: Draft
+Status: Accepted
 Type: Standards Track
 Topic: Packaging
 Created: 19-Sep-2025
 Post-History: `17-Apr-2025 <https://discuss.python.org/t/88608>`__,
               `14-Nov-2025 <https://discuss.python.org/t/104883>`__
+Resolution: `19-May-2026 <https://discuss.python.org/t/104883/39>`__
 
 
 Abstract


henryiii:
There was one minor comment:

> There is one relatively minor point that should be clarified when transferring the specification into the official standards documentation. The PEP states:
> 
> > If a field is present in both places, then the build backend is allowed to insert entries into the list or table, but not remove entries, reorder entries, or modify the entries.
> 
> This does not put any restrictions on _where_ entries may be inserted - at the start, at the end, or in the middle. In practice, none of the existing fields affected by this PEP assign any meaning to ordering, so this doesn’t matter, but in order to avoid confusion should a future PEP add an order-dependent field, the specification should be explicit. Allowing insertions either anywhere, or just at the end, is fine with me - I’ll leave it to the PEP author to choose.

I think that means the PEP can be accepted as is, then when updating the spec at packaging.python.org this needs to be clarified, and a note is not needed here, but happy to add one if preferred. I _think_ we have picked one, but I rather want to look at what it affects over the next couple days when preparing implementations before settling one one vs. the other.

hugovk:
Let's ask @pfmoore to clarify.

pfmoore:
I'm fine with accepting the PEP as it stands. As long as the resolution is added in packaging.python.org, that's fine with me.

hugovk:
Thanks both!

PR #4986, opened 2026-05-31
Mark various accepted PEPs as Final

ZeroIntensity:
I didn't touch PEPs that already have a PR marking them as final. For the packaging PEPs, I only finalized those that already have a `.. canonical-pypa-spec` marker. I don't know enough about packaging to dig through the spec and find links for the others.

### DIFF
--- a/peps/pep-0703.rst
+++ b/peps/pep-0703.rst
@@ -3,7 +3,7 @@ Title: Making the Global Interpreter Lock Optional in CPython
 Author: Sam Gross <colesbury at gmail.com>
 Sponsor: Łukasz Langa <lukasz at python.org>
 Discussions-To: https://discuss.python.org/t/22606
-Status: Accepted
+Status: Final
 Type: Standards Track
 Created: 09-Jan-2023
 Python-Version: 3.13
@@ -17,6 +17,8 @@ Resolution: `24-Oct-2023 <https://discuss.python.org/t/pep-703-making-the-global
    back any changes that turn out to be too disruptive – which includes
    potentially rolling back all of PEP 703 entirely if necessary
    (however unlikely or undesirable we expect that to be).
+
+.. canonical-doc:: :ref:`freethreading-python-howto`
 
 
 Abstract
--- a/peps/pep-0770.rst
+++ b/peps/pep-0770.rst
@@ -13,7 +13,7 @@ Post-History:
   `06-Jan-2025 <https://discuss.python.org/t/76308>`__,
 Resolution: `11-Apr-2025 <https://discuss.python.org/t/76308/112>`__
 
-.. canonical-pypa-spec:: https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-sboms-directory
+.. canonical-pypa-spec:: `The .dist-info/sboms/ directory <https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-sboms-directory>`
 
 Abstract
 ========
--- a/peps/pep-0770.rst
+++ b/peps/pep-0770.rst
@@ -4,7 +4,7 @@ Author: Seth Larson <seth@python.org>
 Sponsor: Brett Cannon <brett@python.org>
 PEP-Delegate: Brett Cannon <brett@python.org>
 Discussions-To: https://discuss.python.org/t/76308
-Status: Accepted
+Status: Final
 Type: Standards Track
 Topic: Packaging
 Created: 02-Jan-2025
--- a/peps/pep-0770.rst
+++ b/peps/pep-0770.rst
@@ -13,7 +13,7 @@ Post-History:
   `06-Jan-2025 <https://discuss.python.org/t/76308>`__,
 Resolution: `11-Apr-2025 <https://discuss.python.org/t/76308/112>`__
 
-.. canonical-pypa-spec:: `The .dist-info/sboms/ directory <https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-sboms-directory>`
+.. canonical-pypa-spec:: `The .dist-info/sboms/ directory <https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-sboms-directory>`__
 
 Abstract
 ========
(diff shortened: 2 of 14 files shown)


peps/pep-0703.rst
```diff
@@ -3,7 +3,7 @@ Title: Making the Global Interpreter Lock Optional in CPython
 Author: Sam Gross <colesbury at gmail.com>
 Sponsor: Łukasz Langa <lukasz at python.org>
 Discussions-To: https://discuss.python.org/t/22606
-Status: Accepted
+Status: Final
```
JelleZijlstra, line 6: Not sure if this was meant to stay non-final until free-threading is the default.

ZeroIntensity: Not sure if there was any prior discussion, but personally, I think it makes more sense to mark PEP 703 as final (especially considering PEP 779 is marked final as well). There will almost certainly be a separate PEP that makes FT the default.

encukou:
Thanks for the nudge!

For my 3.15 PEPs (793, 803, 820), I want to go through them once again and note any surprises and changes found during implementation. Since there will be some of that, I want to get SC approval before marking them Final.
(But before that, [docs](https://github.com/python/cpython/pull/150580)!)

peps/pep-0794.rst
```diff
@@ -2,7 +2,7 @@ PEP: 794
 Title: Import Name Metadata
 Author: Brett Cannon <brett@python.org>
 Discussions-To: https://discuss.python.org/t/94567
-Status: Accepted
+Status: Final
```
brettcannon: This is not final yet; tools have not rolled out support.

brettcannon (changes requested):

ZeroIntensity:
> For my 3.15 PEPs (793, 803, 820), I want to go through them once again and note any surprises and changes found during implementation. Since there will be some of that, I want to get SC approval before marking them Final.

No problem, reverted it. I think as long as there's a clear path to "Final", it's no big deal -- I got the impression that Hugo was getting a little frustrated by the lack of communication from PEP authors, so thanks for clarifying :)

peps/pep-0794.rst
ZeroIntensity: Okay, I reverted it.

Out of curiosity, why does packaging have a different definition of "Final" than standards? For standards, "Final" means "implemented", not "used by users".

brettcannon (approved):
Approving because my PEP was removed. 😉

peps/pep-0770.rst
hugovk: Please update this to use a Sphinx reference, rather than a plain URL.

(https://pep-previews--4986.org.readthedocs.build/pep-0770/)

hugovk (commented):
For the packaging PEPs, do they need to be implemented by tools before marking Final?

ZeroIntensity:
> For the packaging PEPs, do they need to be implemented by tools before marking Final?

To be blunt: I have no idea. Brett's comment suggests that they're supposed to wait, but I don't think it makes much sense to do so -- for example, would you consider the C23 standard as "not done" simply because there's not full compiler support for it yet?

peps/pep-0794.rst
merwok: Brett means this is not implemented by tools.

ZeroIntensity: I know -- my question was why we have a process difference between packaging and standards PEPs. Standards PEPs are marked as final when they're officialized in CPython, not when downstream users start using it. I guess it's because there's no CPython equivalent for packaging?

hugovk:
Let's ask @pfmoore: are these packaging PEPs ready to be marked Final?

* PEP 668: Marking Python base environments as “externally managed”
* PEP 691: JSON-based Simple API for Python Package Indexes
* PEP 753: Uniform project URLs in core metadata
* PEP 770: Improving measurability of Python packages with Software Bill-of-Materials

pfmoore (approved):

pfmoore:
Yeah, those 4 packaging PEPs can be marked final. We only need implementations in the case of provisional acceptance where that's an explicit condition. And all 4 of those have been implemented in any case, I believe.

hugovk (approved):
Thanks!

peps/pep-0794.rst
merwok: That’s what I am trying to say: this is not about a percentage of end users using the feature described in the PEP, but initial implementation by relevant projects, like the packaging library for core formats, pip/uv for installer things, pypi for index standards.

peps/pep-0770.rst
```diff
@@ -13,7 +13,7 @@ Post-History:
   `06-Jan-2025 <https://discuss.python.org/t/76308>`__,
 Resolution: `11-Apr-2025 <https://discuss.python.org/t/76308/112>`__
 
-.. canonical-pypa-spec:: https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-sboms-directory
+.. canonical-pypa-spec:: `The .dist-info/sboms/ directory <https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-sboms-directory>`__
```
merwok, line 16: (Why was it impossible to use `:ref:some text <dist-info-sbom-ref>` here?)

ZeroIntensity: There is no ref: https://github.com/pypa/packaging.python.org/blob/c31a69a5a951d3e0ba0200fe011c5fd90aff8279/source/specifications/binary-distribution-format.rst#L278-L280

merwok: Ah! Headings always create a ref, but its value derives from the heading content (so I can see why your tries didn’t find it, given inline markup in the title), and are prone to breaking for any change. Hopefully the doc will remain stable, and/or a linkcheck job in this repo can catch breakage.

ZeroIntensity: Who controls the PyPA spec page (or is it us)? We could just ask nicely to add a ref there.

hugovk: The PyPA folk do, you can open a PR at https://github.com/pypa/packaging.python.org.

ZeroIntensity: https://github.com/pypa/packaging.python.org/pull/2075

PR #4993, opened 2026-06-04
PEP 753: mark as Final

woodruffw:
This PEP has been final for a while, per its canonical PyPA spec (which has also been linked in the PEP for a while):

https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-project-urls

(This was my own fault, I probably forgot to update this.)

CC @facutuesca as co-editor


willingc (approved):
Thanks @woodruffw.

PR #4996, opened 2026-06-11
Infra: change `<article>` to `<main>`

hugovk:
Follow on from https://github.com/python/peps/pull/4977.

We still get "697 files changed · ± 697 modified" on the Read the Docs previews: https://github.com/python/peps/pull/4994#issuecomment-4670664344.

The [detection logic](https://docs.readthedocs.com/platform/stable/reference/main-content-detection.html#detection-logic) is:

> 1. **Elements with `role="main"` attribute:** This ARIA role is used by many static site generators and themes to indicate the main content area.
> 2. **The `<main>` HTML tag:** The semantic HTML5 element for main content.
> 3. **Parent of the first `<h1>` tag:** If no explicit main content markers are found, the system assumes all sections are siblings under a common parent, and uses the parent of the first heading as the main content container.
> 4. **The `<body>` tag:** As a last resort, if none of the above are found, the entire body is used.

Let's go for 2 and use `<main>` instead of `<article role="main">`:

Both have good semantics: there can be multiple `article`s in a page (we have just one), or a single `main`. But `main` is also a "landmark", which helps accessibility:

> The `<main>` element behaves like a [main landmark](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/main_role) role. [Landmarks](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Techniques#landmark_roles) can be used by assistive technology to quickly identify and navigate to large sections of the document. Prefer using the `<main>` element over declaring `role="main"`, unless there are [legacy browser support concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/main#browser_compatibility).

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/main

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/article

### DIFF
--- a/pep_sphinx_extensions/pep_theme/static/mq.css
+++ b/pep_sphinx_extensions/pep_theme/static/mq.css
@@ -38,7 +38,7 @@
         padding: 0.5rem 1rem 0;
         width: 100%;
     }
-    section#pep-page-section > article,
+    section#pep-page-section > main,
     section#pep-page-section > footer#pep-page-footer {
         max-width: 37em;
         width: 74%;
@@ -65,7 +65,7 @@
   }
 }
 @media (min-width: 60em) {
-    section#pep-page-section > article,
+    section#pep-page-section > main,
     section#pep-page-section > footer#pep-page-footer {
         max-width: 56em;
         padding-left: 3.2%;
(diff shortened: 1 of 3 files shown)


encukou (approved):
Looks good, thank you!

PR #4998, opened 2026-06-19
PEP 835: Address Discourse feedback from encukou and tjreedy

till-varoquaux:
This PR incorporates recent feedback from the Discourse discussion for PEP 835. Specifically, it:

- Clarifies `type(None)` support
- Promotes the `FORWARDREF_STRUCTURAL` mechanism to the Specification section
- Bumps the deprecation timeline for the compatibility shim to Python 3.21 to align with PEP 387
- Clarifies the phrasing around stringified annotations under PEP 563

### DIFF
--- a/peps/pep-0835.rst
+++ b/peps/pep-0835.rst
@@ -228,13 +228,64 @@ origin:
 Supported Left-Hand Operands
 -----------------------------
 
-The ``@`` operator is implemented by adding ``nb_matrix_multiply`` to the
-metatype (``type``) and to several typing-related types. The operator is
-supported for any left-hand operand that currently supports the ``|``
-operator for making a union.
+The ``@`` operator is implemented by adding ``nb_matrix_multiply`` to the metatype (``type``) and to several typing-related types. The operator is supported for instances of ``type`` and other typing constructs that currently support the ``|`` operator for unions. The specific handling of ``None`` is currently unsettled (see Open Issues).
 
 For all other left-hand operands, the operator returns ``NotImplemented``,
 allowing normal ``__matmul__`` dispatch to proceed.
+
+Forward References and Deferred Evaluation
+-------------------------------------------
+
+Under :pep:`749`, annotations are lazily evaluated. The ``annotationlib``
+module provides several formats for retrieving annotations:
+
+- ``Format.VALUE``: Fully evaluates the annotation. Raises ``NameError``
+  if any name is unresolvable.
+- ``Format.FORWARDREF``: Wraps unresolvable names in ``ForwardRef`` objects.
+  However, when operators like ``@`` or ``|`` fail to evaluate because of an
+  unresolvable name, this format falls back to returning the entire expression
+  as an opaque string wrapped in a single ``ForwardRef`` object.
+- ``Format.STRING``: Returns the raw source text with no evaluation.
+
+For the ``@`` operator, ``Format.FORWARDREF`` is insufficient. Consider::
+
+    class Model:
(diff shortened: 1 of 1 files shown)


peps/pep-0835.rst
```diff
@@ -229,45 +229,15 @@ Supported Left-Hand Operands
 -----------------------------
 
 The ``@`` operator is implemented by adding ``nb_matrix_multiply`` to the
-metatype (``type``) and to several typing-related types. The operator is
+metatype (``type``) and to several types used in type hinting (including
+``NoneType``). The operator is
 supported for any left-hand operand that currently supports the ``|``
-operator for making a union.
+operator for making a union (including ``type(None)``, which acts as the
+canonical equivalent to ``None`` in type hints).
```
JelleZijlstra: This is a questionable claim; some parts of typing.py turn None into NoneType but not all typing APIs do.

Also the mention of NoneType is a bit confusing. NoneType is a type like any other, so it inherits `__or__` (currently) and `__matmul__` (if this PEP is accepted) from `type`. The more interesting question is what happens to None itself. Currently, it does not in fact support `__or__`; `None | int` works, but that's through the latter's `__ror__`. Therefore, unions of None with a stringified type (`None | "int"`) fail at runtime. You are proposing to make matmul work for None.

peps/pep-0835.rst
```diff
@@ -304,7 +275,10 @@ operators are still evaluated. The example above produces::
 The metadata is immediately accessible. This format also resolves the
 pre-existing issue with ``|`` unions, where ``"Foo" | int`` under
 ``Format.FORWARDREF`` produces ``ForwardRef('Foo | int')`` instead of the
-structural ``ForwardRef('Foo') | int``.
+structural ``ForwardRef('Foo') | int``. Notably, if ``FORWARDREF_STRUCTURAL``
```
JelleZijlstra: Seems very speculative; there are lots of problems that would need to be solved before this can be entertained.

peps/pep-0835.rst
till-varoquaux: Great point. You're right that `__matmul__` on `NoneType` is a special case with potential runtime risks since `None` relies on `__ror__`. I've removed the claim and moved the `None` debate down to a new "Open Issues" section.

peps/pep-0835.rst
till-varoquaux: Apologies, I got carried away! I agree that changing the default evaluation strategy is a massive separate undertaking. I've completely removed the sentence.

ilevkivskyi:
@till-varoquaux I don't think you need to update the PEP often. It makes sense to only update the parts where a certain consensus has been reached.

Btw, since @JelleZijlstra is here, I was thinking about an alternative syntax for annotated metadata using `&`. I guess it is planned to be added for intersection types, right? And conceptually annotated metadata is similar to an intersection. This idea has other drawbacks (e.g. it is not clear how to specify in a simple way that a given expression is an annotation rather than a proper type), but just wanted to share it here in a more quiet place as the main discussion thread is already busy.

till-varoquaux:
Thanks for the advice on PEP management, Ivan. 
I lean towards keeping & reserved for intersection types. I wouldn't want to conflict with that work.
I am interested in your and @JelleZijlstra ’s thoughts on the future of static tracking for some subset of `Annotated`. If the community sees value in tracking metadata for new typing extensions, finding a clean syntax becomes more justified.

till-varoquaux:
This is superseded by https://github.com/python/peps/pull/5027

PR #5000, opened 2026-06-19
PEP 545: Update language tag guidelines

m-aciek:
Cover usage of script subtags in IETF language tags.

Previous (current) version mentions only region subtags, whereas script subtags can cover wider audience, while still being correct. For example Django documentation [use](https://github.com/django/django-docs-translations/blob/b0e07ee0d116a22405e03e6e21dda28bf1726d15/manage_translations.py#L22) `zh-hans` for Chinese written in simplified script.

cc @JulienPalard @methane @vstinner 

* Change is either:
    * [x] To an Active Process PEP

### DIFF
--- a/peps/pep-0545.rst
+++ b/peps/pep-0545.rst
@@ -187,15 +187,25 @@ Language Tag
 ''''''''''''
 
 A common notation for language tags is the :rfc:`IETF Language Tag <5646>`
-[4]_ based on ISO 639, although gettext uses ISO 639 tags with
-underscores (ex: ``pt_BR``) instead of dashes to join tags [5]_
-(ex: ``pt-BR``).  Examples of IETF Language Tags: ``fr`` (French),
-``ja`` (Japanese), ``pt-BR`` (Orthographic formulation of 1943 -
-Official in Brazil).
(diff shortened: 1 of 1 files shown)


StanFromIreland:
I think we’ve generally decided to leave the PEP for now (as it also needs many other changes), and update the devguide instead.

hugovk:
This should be discussed in another place first, such as https://github.com/python/docs-community/issues or https://discuss.python.org/c/documentation/26.

willingc:
@m-aciek It would be better to start discussing it as a PR to the devguide and in the community meeting first. Thanks!

vstinner:
I'm no longer working on documentation translation, so I wrote https://github.com/python/peps/pull/5001 to remove myself from code owners.

m-aciek:
I'm going to close this for now rather than move the discussion elsewhere. The devguide is neutral regarding language subtags. It's the PEP that favors region subtags somewhat over script subtags.

For what it's worth, we've settled on using plain `pa` for Punjabi in the Gurmukhi script, since in BCP 47 `pa` has the `Suppress-Script: Guru` annotation.

PR #5006, opened 2026-06-24
PEP 793, 803, 820: Update & mark as Final

encukou:
  - I think the changes turned out to not be substantial; see below
* [x] Pull request title in appropriate format (``PEP 123: Mark as Final``)
  - I hope combining them is fine :)
* [x] Canonical docs/spec linked with a ``canonical-doc`` directive

Besides typos, minor inaccuracies, and links to current docs and implementation-tracking issues, the changes are:

PEP 793:
- Updated with changes specified in PEP 820

PEP 803: 
- If ``Py_TARGET_ABI3T`` is defined, then ``Python.h`` will make sure that ``Py_GIL_DISABLED`` is defined as well.
  Users may check this macro to enable free-threading-specific code like extra locking.

  (Users do this in practice.)
- Add a ``PyABIInfo_FREETHREADING_AGNOSTIC`` flag for ``PyABIInfo.flags``

  (I forgot to mention it both here and in https://github.com/capi-workgroup/decisions/issues/72.)
- Mention that `PyCriticalSection` was added too.

PEP 820:
- Clarify that the `PySlot_STATIC` flag is required for data that needs to be static (``Py_mod_methods``, ``Py_tp_methods``, ``Py_tp_members``, ``Py_tp_getset``) when using the new-style slots.
  (The original reference implementation didn't have this.)

### DIFF
--- a/peps/pep-0793.rst
+++ b/peps/pep-0793.rst
@@ -2,13 +2,15 @@ PEP: 793
 Title: PyModExport: A new entry point for C extension modules
 Author: Petr Viktorin <encukou@gmail.com>
 Discussions-To: https://discuss.python.org/t/93444
-Status: Accepted
+Status: Final
 Type: Standards Track
 Created: 23-May-2025
 Python-Version: 3.15
 Post-History: `14-Mar-2025 <https://discuss.python.org/t/84498/>`__,
               `27-May-2025 <https://discuss.python.org/t/93444/>`__,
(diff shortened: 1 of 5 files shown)


encukou:
I'll merge next week if there are no objections.

encukou:
*almost* next week

hugovk:
@zooba Do you have plans for [PEP 809](https://peps.python.org/pep-0809/) for 3.16, or would you prefer to withdraw it?

PR #5011, opened 2026-06-29
Infra: Link breadcrumbs back to homepage not /pep-0000/

hugovk:
Right now:

1. Go to https://peps.python.org/
2. Click on a PEP: https://peps.python.org/pep-0008/
3. In the breadcrumbs at the top: "Python » PEP Index » PEP 8", click "PEP Index"

Expected result:

* Taken back to https://peps.python.org/

Actual result:

* Taken to https://peps.python.org/pep-0000/

That `/pep-0000/` is confusing, let's go back to the usual `/` instead.


willingc (approved):
Thanks @hugovk.

PR #5013, opened 2026-07-01
PEP 828: add synchonous delegation recipe

graingert:
It's technically incorrect to yield multiple times from inside a structured concurrency context: it's only correct to yield once in that context inside an async generator function decorated with `@asynccontextmanager` (See PEP 789), therefore the issues as stated with synchronous delegation are incorrect so I removed that text.

Instead keep the note about the complexity and include a recipe on how users can achieve this themselves without adding extra complexity to the PEP 828 implementation, with the hope this recipe can be added to the python docs in the implementation PR


peps/pep-0828.rst
```diff
@@ -379,16 +379,46 @@ asynchronous generator methods: :meth:`~agen.asend` to :meth:`~generator.send`,
...
+    class AsAsyncGenerator(AsAsyncIterable):
+        async def asend(self, value):
+            try:
+                return self._wrapped.send(value)
+            except StopIteration as e:
+                raise StopAsyncIteration(e.value) from e
+
+        async def athrow(self, exc):
+            try:
+                return self._wrapped.throw(exc):
```
graingert:
```suggestion
                return self._wrapped.throw(exc)
```

peps/pep-0828.rst
```diff
@@ -379,16 +379,46 @@ asynchronous generator methods: :meth:`~agen.asend` to :meth:`~generator.send`,
...
+            return self
+
+        async def __anext__(self):
+            try:
+                return self._wrapped.__next__()
+            except StopIteration as e:
+                raise StopAsyncIteration(e.value) from e
+
+
+    class AsAsyncGenerator(AsAsyncIterable):
```
graingert:
```suggestion
    class AsAsyncGenerator(AsAsyncIterator):
```

peps/pep-0828.rst
```diff
@@ -379,16 +379,46 @@ asynchronous generator methods: :meth:`~agen.asend` to :meth:`~generator.send`,
 :meth:`~agen.athrow` to :meth:`~generator.throw`, and :meth:`~agen.aclose`
 to :meth:`~generator.close`.
 
-For example, asynchronous exceptions could be injected into synchronous
-generators:
+It's trivial for anyone that needs to delegate to a subgenerator to write
+the wrapper class to upgrade a synchronous Iterable or Generator to an
+async one beffore calling ``async yield from``
```
graingert:
```suggestion
async one before calling ``async yield from``
```

peps/pep-0828.rst
```diff
@@ -379,16 +379,46 @@ asynchronous generator methods: :meth:`~agen.asend` to :meth:`~generator.send`,
 :meth:`~agen.athrow` to :meth:`~generator.throw`, and :meth:`~agen.aclose`
 to :meth:`~generator.close`.
 
-For example, asynchronous exceptions could be injected into synchronous
-generators:
+It's trivial for anyone that needs to delegate to a subgenerator to write
+the wrapper class to upgrade a synchronous Iterable or Generator to an
+async one before calling ``async yield from``
```
graingert:
```suggestion
async one before calling ``async yield from``.
```

peps/pep-0828.rst
```diff
@@ -379,16 +379,46 @@ asynchronous generator methods: :meth:`~agen.asend` to :meth:`~generator.send`,
 :meth:`~agen.athrow` to :meth:`~generator.throw`, and :meth:`~agen.aclose`
 to :meth:`~generator.close`.
 
-For example, asynchronous exceptions could be injected into synchronous
-generators:
+It's trivial for anyone that needs to delegate to a subgenerator to write
+the wrapper class to upgrade a synchronous Iterable or Generator to an
```
ZeroIntensity: Could you add Sphinxs refs to the `collections.abc` classes here? It's nice to have links.

peps/pep-0828.rst
```diff
@@ -379,16 +379,46 @@ asynchronous generator methods: :meth:`~agen.asend` to :meth:`~generator.send`,
...
+            self._wrapped = iter(wrapped)
+
+        def __aiter__(self):
+            return self
+
+        async def __anext__(self):
+            try:
+                return self._wrapped.__next__()
+            except StopIteration as e:
+                raise StopAsyncIteration(e.value) from e
```
ZeroIntensity, line 400: Does this really need to be a class? I think this could just be a generator function.

ZeroIntensity (commented):
Thanks for doing this

peps/pep-0828.rst
graingert: I think it's cleaner as a class then it works with the inheritance with AsAsyncGenerator

ZeroIntensity (approved):
LGTM, thanks!

PR #5024, opened 2026-07-06
Infra: Pin Pygments to a development version

ZeroIntensity:
This allows the PEP 810 examples to be highlighted correctly. See also https://github.com/python/cpython/issues/153227.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # Requirements for building PEPs with Sphinx
 # Sphinx requires >= 2.17. JSON5 support added in 2.19
-Pygments >= 2.19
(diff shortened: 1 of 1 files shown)


hugovk:
Please can you add a comment similar to the one I suggestion in the CPython PR?

requirements.txt
```diff
@@ -1,6 +1,10 @@
 # Requirements for building PEPs with Sphinx
+
 # Sphinx requires >= 2.17. JSON5 support added in 2.19
-Pygments >= 2.19
+# Temporary direct requirement, pending release of Pygments > 2.20.0
```
hugovk:
```suggestion
# Temporary archive install, pending release of Pygments > 2.20.0
```

hugovk (approved):
Thanks!

PR #5029, opened 2026-07-10
Infra: Pin actions to SHAs and add a defensive persist-credentials in workflows

willingc:


hugovk (approved):
Thanks! I'll follow up by adding zizmor to pre-commit/CI: https://github.com/python/peps/pull/5030.

PR #5053, opened 2026-07-21
PEP 838: Move to standards track

konstin:
PEP 838 affects the stdlib `venv` module, so it goes into the standards track instead of the packaging track.

* Change is either:
    * [x] To a Draft PEP

### DIFF
--- a/peps/pep-0838.rst
+++ b/peps/pep-0838.rst
@@ -2,11 +2,9 @@ PEP: 838
 Title: Adding python-version to pyvenv.cfg
 Author: Konstantin Schütze <konstin@mailbox.org>
 Sponsor: Alex Waygood <alex.waygood@gmail.com>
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
To me this fits in better as a packaging PEP, since the only part of the stdlib that it touches is a very-much packaging-related part of the stdlib (the `venv` module), and since it is proposing a standard that we are asking external packaging tools that write `pyvenv.cfg` files (uv, virtualenv) to comply with as well as the stdlib `venv` module.

But I don't have a particularly strong opinion, and have very little experience on packaging PEPs; very happy to defer to @ncoghlan and @pfmoore :-)

AlexWaygood:
I hear this has been discussed with @warsaw as well, so I'll go ahead and land this -- thanks!

PR #5070, opened 2026-08-03
PEP 694: Amend scanning in staged releases and add legacy API changes to use staged releases

cjames23:
Amendments based on DPO discussion https://discuss.python.org/t/pre-pep-staged-releases-separated-from-pep-694/107804/59 

@warsaw  - here are my proprosed amendments which I would like your sign off on  as well.


peps/pep-0694.rst
```diff
@@ -285,6 +288,14 @@ The unguessable :ref:`stage preview URL <staged-preview>` is a separate capabili
 governed by this authorization check; it grants read-only preview access to any client that holds the token,
 so that (for example) a CI job can install-test a staged release without project upload credentials.
 
+As one such stricter policy, an index **MAY** require *additional* authorization, beyond upload permission, to
+:ref:`publish <publishing-session-completion>` or :ref:`cancel <publishing-session-cancellation>` a session,
```
warsaw: +1 for publishing, but can you provide some additional scenarios about why cancelation should require additional authorization?

peps/pep-0694.rst
```diff
@@ -743,6 +754,20 @@ change: deferred processing resolves to either ``published`` on success or ``err
 resolves to ``error``, the session remains editable and the reason is reported in the session's ``notices``,
 as described in :ref:`publishing-session-states`.
 
+The ``processing`` state **MAY** be used to run asynchronous review of a session's files before it is
+published, such as malware scanning.  If the review completes without an adverse result, the session resolves
+to ``published`` as normal.  If the review does not complete within a period chosen by the index, the server
+**MAY** treat it as though it had completed without an adverse result and publish the session, so that a
+backlogged or unavailable review system does not indefinitely prevent publication.  If the review produces an
+adverse result, the session resolves to ``error`` with the reason reported in the session's ``notices``; the
+server **MAY** decline to publish such a session on any subsequent retry, in which case it is eventually
```
warsaw, line 823: This section (if not covered below), should provide details on what happens if files in the session are deleted and reuploaded.  It could be:

* Continue to await human review
* Rescan newly uploaded files
* Reset the entire malware scan state

It probably would also be helpful to say something about denial of service attacks on the scanner.  E.g. bad actor starts an upload session, uploads known malware knowing it will get flagged.  Bad Actor deletes the files and reuploads them to trigger another scan.  Rinse and repeat until the scanning service is overwhelmed.

peps/pep-0694.rst
```diff
@@ -1348,6 +1373,42 @@ changes to the installer tool required.
 The details of this user experience are left to installer tool maintainers.
 
 
+.. _legacy-interop:
+
+Legacy Upload API Interoperability
```
warsaw: I wonder if it makes sense to be more explicit about the "control plane" API for managing the session (and file uploads), and/or lump this and the explicit session creation together?   Maybe add a top-level "Session Creation" section, putting this and the original session creation request together, then the rest of it would be the same control plane API.

It also occurs to me that we *might* want to update the state diagrams so that it's obvious that there are two entry points "open" state.  However, if I'm reading this section correctly, the index can also auto-publish if the conditions allow (no adverse scan or scan timeout).   Maybe that'll make the diagram too cluttered in which case we could skip it or add a second diagram.  LMK what you think and I can take a crack at updating the diagram once this PR lands.

The other thing this section should discuss is the interaction between legacy and 2.0 sessions.  E.g. what happens if there's an open 2.0 session?  Can someone use the legacy API to add a file to that existing session?  I would think that would get rejected with a 409.  Also vice versa (meaning: do we allow mixing of session creation requests?).  We just need to be explicit about how these mix.

warsaw (changes requested):
This is a great addition to the PEP.  I have some comments for a few things that need clarification, but otherwise +1.  And welcome aboard as a co-author!

peps/pep-0694.rst
cjames23: IN terms of a top level Session Creation section I had that in another branch but was not sure, this is the signal it should be a separated out top level section which would then be the same control plane API. 

I will be explicit about the mixing of session creation requests which I am aligned on an attempt to mix would be rejected with a 409.

peps/pep-0694.rst
cjames23: I see how this reads a little weird. I was trying to state that an index MAY require a separate set of permissions for moving publish state to either completed or cancelled.

peps/pep-0694.rst
```diff
@@ -29,6 +29,9 @@ Along with standardization, the upload API provides additional useful features s
 * "staging" a release, which can be used to test uploads before publicly publishing them,
   without the need for `test.pypi.org <https://test.pypi.org/>`__;
 
+* entering the publishing session workflow from the existing legacy upload API, so that staging is
+  available to publishers before their tooling adopts this API;
```
warsaw, line 33:
```suggestion
* entering the publishing session workflow from the existing legacy upload API, so that staging is
  available to publishers before their tooling fully adopts this API;
```

I struggled a little bit with this text, since publishers _will_ have to update their tooling either way.  Even twine will have to grow the inclusion of a `staged` field.  I think what you're trying to get at is that we can get the staging benefits without the full new control plane protocol.  I couldn't find an elegant way to say that so I just added the word "fully".

peps/pep-0694.rst
```diff
@@ -285,6 +288,21 @@ The unguessable :ref:`stage preview URL <staged-preview>` is a separate capabili
...
+As one such stricter policy, an index **MAY** require *additional* authorization, beyond upload permission, for
+the actions that decide a session's terminal outcome, namely :ref:`publishing <publishing-session-completion>`
+it and :ref:`canceling <publishing-session-cancellation>` it, while still allowing session creation and file
+upload with upload permission alone.  This separates the authority to assemble a release from the authority to
+publish or discard it: an automated system can hold a credential that creates a session and uploads files to it
+but can neither publish nor cancel it, with that authority held elsewhere, so that compromise of the automated
+system alone can neither publish a release nor discard a pending one.
+
+Cancellation is protected because it is destructive: it discards the staged files and frees the name-version
+pair.  An index that left it at the upload-permission level would let a compromised upload credential delete a
```
warsaw, line 300: Cancelation also frees any temporary new project name reservation, which would allow an attacker with the compromised credential to "steal" the name.  So that's yet another reason why an index might want to separate these permissions.

peps/pep-0694.rst
```diff
@@ -585,6 +608,54 @@ sub-mapping with the following keys:
     these notices are specific to the referenced file.
 
 
+.. _legacy-interop:
+
+Create a Publishing Session from a Legacy Upload
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Publishers cannot use the features of this API until their upload tooling adopts it, and the legacy API is
+expected to remain available as this pep does not propose a deprecatioin schedule.  To make staging available to those publishers sooner, an index
```
warsaw, line 617:
```suggestion
expected to remain available anyway, as this PEP does not propose a deprecation schedule.  To make staging available to those publishers sooner, an index
```

The presence of a `staged` field would technically require tooling to change, but minimally.  This section is also proposing that an index can create a staged release even without the `staged` field, which is interesting because then no tools actually need to be changed.

I wonder if that distinction should be called out more explicitly?

peps/pep-0694.rst
```diff
@@ -585,6 +608,54 @@ sub-mapping with the following keys:
...
+PEP to :ref:`preview <staged-preview>`, :ref:`publish <publishing-session-completion>`, or :ref:`cancel
+<publishing-session-cancellation>` the session.  An index **MAY** also create a session for an upload based on
+its own policy or the project's configuration, without the field being present; this allows a project to
+require that its releases are staged in a way that an upload client cannot bypass.
+
+Because the legacy API uploads a single file per request, subsequent legacy uploads for the same project and
+version **SHOULD** be added to the same open session when that session was itself created through a legacy
+upload, so that the release is still published as a unit.
+
+A session's creation path is fixed when it is created, and the two paths are not mixed.  If a non-terminal
```
warsaw, line 636:
```suggestion
A session's creation path is fixed when it is created, and the two paths **MUST NOT** be mixed.  If a non-terminal
```

peps/pep-0694.rst
```diff
@@ -585,6 +608,54 @@ sub-mapping with the following keys:
...
+session already exists for a name-version pair (see :ref:`publishing-session-multiple`), a request to
+contribute to it through the *other* path **MUST** be rejected with a ``409 Conflict``: a legacy
+``staged=true`` upload for a pair that already has an open session created through the Upload 2.0 API is
+rejected, and an Upload 2.0 request that would add to a session created through a legacy upload is likewise
+rejected.
```
warsaw, line 641: I'm confused by the requirements here.  I would totally expect rejection for a path where a legacy upload *without* the `staged=true` flag when an Upload 2.0 session is already created for the name-version pair, because that's mixing legacy-legacy with upload-2.0.  But legacy+staged into an Upload 2.0 stage *could* be allowed without any semantic confusion -- I think!  Or is that also a problem?  And vice versa of course (legacy+staged followed by Upload 2.0 ... okay or not okay?).  If I'm confused maybe some additional rationale is warranted?

warsaw (changes requested):
I really like this addition to the PEP.  It provides a fantastic transition period from legacy to upload-2.0.  I have a few questions and suggestions.

peps/pep-0694.rst
cjames23: I agree. I struggled with a way to get this to be elegant to say we can get there without the full new control plane protocol.

peps/pep-0694.rst
cjames23: I think I covered this in the latest update

peps/pep-0694.rst
cjames23: Updated in the latest commit. I think it should be more clear now hopefully.

Issue #5073, opened 2026-08-04
PEP 831: Clarify build backend flag propagation


pablogsal:
PEP 831 currently overstates how CPython build flags propagate through generic build backends. Update the PEP to distinguish Setuptools and deliberate sysconfig consumers from other PEP 517 backends, and document build-details.json as the appropriate future standardized interface.

Context: https://discuss.python.org/t/pep-831-frame-pointers-everywhere-enabling-system-level-observability-for-python/106958/7

PR #5074, opened 2026-08-04
gh-5073: PEP 831: Clarify build backend flag propagation

pablogsal:


peps/pep-0831.rst
```diff
@@ -524,35 +533,59 @@ level and wish to avoid double-specification, analogous to Fedora's per-package
...
+complete, useful flame graphs out of the box.
 
 Extension authors who observe an unacceptable regression in a specific module
-can opt out per-extension via ``extra_compile_args`` (see `Extension Build
+can opt out through backend-specific compiler options (see `Extension Build
 Impact`_).  Distributions that already enable frame pointers system-wide
 (Fedora, Ubuntu, Arch Linux) need take no action.
 
+Build Backend Guidance
+----------------------
```
hugovk, line 567: In addition to guidance to build backends, should there be guidance for tools such as cibuildwheel which specify config to build backends?

For example, should they add an option to enable frame pointers in backends, and perhaps make it the default at some point?

savannahostrowski (commented):
The clarification here looks right to me. I left a comment on python/cpython#155172; depending on the intended scope there, we may want to reflect its relationship to the future standardized field here.

We'll probably also want to update corresponding CPython docs, since `Doc/using/configure.rst` still says that the flags "propagate to third-party C extensions through sysconfig".

PR #5078, opened 2026-08-05
PEP 843: Export Statement for DRY Re-exports

NeilGirdhar:

* [x] Read and followed [PEP 1](https://peps.python.org/1) & [PEP 12](https://peps.python.org/12)
* [x] File created from the [latest PEP template](https://github.com/python/peps/blob/main/peps/pep-0012/pep-NNNN.rst?plain=1)
* [x] PEP has next available number, & set in filename (``pep-NNNN.rst``), PR title (``PEP 123: <Title of PEP>``) and ``PEP`` header
    * Tip: find the next available number with [pepotron](https://github.com/hugovk/pepotron) — run ``uvx pepotron next`` (or ``pipx install pepotron`` then ``pep next``)
* [x] Core dev/PEP editor listed as ``Author`` or ``Sponsor``, and formally confirmed their approval
* [x] ``Author``, ``Status`` (``Draft``), ``Type`` and ``Created`` headers filled out correctly
* [x] ``PEP-Delegate``, ``Topic``, ``Requires`` and ``Replaces`` headers completed if appropriate
* [x] Code is well-formatted (PEP 7/PEP 8) and is in [code blocks, with the right lexer names](https://peps.python.org/pep-0012/#literal-blocks) if non-Python
* [x] Authors/sponsor added to ``.github/CODEOWNERS`` for the PEP

    * [x] Motivation
    * [x] Specification
    * [x] Rationale
    * [x] Backwards Compatibility
    * [x] Security Implications
    * [x] How to Teach This
    * [ ] Reference Implementation
    * [x] Rejected Ideas
    * [x] Open Issues
    * [x] Acknowledgements
    * [ ] Footnotes
    * [x] Change History
* [x] ``Python-Version`` set to valid (pre-beta) future Python version, if relevant
* [x] Any project stated in the PEP as supporting/endorsing/benefiting from the PEP formally confirmed such

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -717,6 +717,7 @@ peps/pep-0839.rst  @corona10
 peps/pep-0840.rst  @jeremyhylton @gvanrossum
 peps/pep-0841.rst  @corona10 @sobolevn
 peps/pep-0842.rst  @ZeroIntensity
+peps/pep-0843.rst  @ZeroIntensity
 peps/pep-0844.rst  @warsaw
 # ...
 peps/pep-2026.rst  @hugovk
--- a/peps/pep-0843.rst
+++ b/peps/pep-0843.rst
@@ -60,7 +60,7 @@ PEP 842 broadens the mechanism into a keyword usable in five forms:
 * A module re-export statement, ``from MODULE export NAME``
 
 All five populate ``__export__``, which also becomes ``__all__`` and
-triggers an ``ExportWarning`` on access to anything left out.
+triggers an ``ExportError`` on access to anything left out.
 
 PEP 842's version has no wildcard equivalent to `Wildcard form`_. This
 PEP takes only the module re-export statement, deliberately leaving out
@@ -75,7 +75,16 @@ Plotly almost universally export their public API using a common
 pattern. A top-level ``__init__.py`` carries a wall of imports from
 private submodules, followed by (or interleaved with) an ``__all__``
 list that repeats the same names as strings, or the reflexive-alias
-idiom used throughout instead. The first looks like this:
+idiom used throughout instead. `pandas
+<https://github.com/pandas-dev/pandas/blob/c79d638e27a10fd2359491bbf7ac1effe8a8b3ff/pandas/__init__.py#L34-L346>`__
+and `polars
+<https://github.com/pola-rs/polars/blob/df599052daf96e7a9cc30a3b0c6bd25d6947e3c0/py-polars/src/polars/__init__.py#L58-L515>`__
+both carry the doubled-list form; `FastAPI
+<https://github.com/fastapi/fastapi/blob/95f8322ee1dcda7ceace7b1c4f6c9915b36d748f/fastapi/__init__.py#L7-L25>`__
+and `Typer
+<https://github.com/fastapi/typer/blob/fe2aa0e2f9c853de378e60ca24ec3b256144decf/typer/__init__.py#L7-L32>`__
+both use the reflexive-alias form throughout. The first looks like
+this:
 
 .. code-block:: python
 
@@ -97,8 +106,12 @@ reorganized into submodules for the maintainer's convenience, while the
 public layout stays stable for users.
 
 Something has to do the flattening. Today that something is a
-hand-maintained, doubly-written list, and such lists routinely drift out
-of sync as the code changes.
+hand-maintained, doubly-written list: the export list (in ``__all__``)
+and the import list (of ``import`` statements) say the same thing
+twice. Any rename, addition, or removal has to be made in two places by
+hand, and the two can silently drift apart. That's the DRY violation
+this PEP removes by folding both into one ``from <module> export
+<name>`` statement.
 
 The alternative, ``import x as x``, is a workaround for the language's
 missing export concept, and it still trips up some auto-formatters, which
--- a/peps/pep-0843.rst
+++ b/peps/pep-0843.rst
@@ -1,0 +1,635 @@
+PEP: 843
+Title: Export Statement for DRY Re-exports
+Author: Neil Girdhar <mistersheik@gmail.com>
+Discussions-To: Pending
+Status: Draft
+Type: Standards Track
+Created: 05-Aug-2026
+Python-Version: 3.16
+Post-History: `05-Aug-2026 <https://discuss.python.org/t/108353>`__
+
+
+Abstract
+========
+
+Large libraries separate their **implementation layout** (the tree of
+modules convenient for maintainers) from their **public layout** (the
+flat, curated tree they present to users). Building that public layout
+today means choosing between two imperfect options.
+
+The first is writing every exported name twice: once in an import
+statement, again as a string in ``__all__``. The two lists must be kept
+in sync by hand every time the public layout changes.
+
+The second is the reflexive-alias idiom, ``from x import y as y``. This
+is part of the type system: type checkers treat it as a signal that the
+import is an intentional re-export. It still reads like a typo to anyone
+who doesn't know the convention. Because the module has no curated
+``__all__``, it loses the ``dir()`` cleanliness and
+wildcard-import control a real ``__all__`` gives.
+
+This PEP adds a statement form that avoids both problems:
+
+.. code-block:: python
+
+   # spam/__init__.py
+   from ._internal.core export PublicAPI
+   from ._internal.widgets export Widget as PublicWidget
+
+It imports the name, optionally under an alias exactly as ``from ...
+import ... as ...`` does, and appends it to ``__all__`` in the same
+statement. Nothing is left to sync by hand, and no alias needs decoding.
+
+
+Relationship to PEP 842
+========================
+
+Both PEPs start from the same discomfort with ``__all__``, and agree on
+the same core mechanism for re-exports: a statement of the shape ``from
+<module> export <name>``, with a lazy variant (see `Lazy exports`_).
+That agreement, reached independently, is confirmation that this is the
+right shape for re-exports.
+
+PEP 842 broadens the mechanism into a keyword usable in five forms:
+
+* Standalone ``export NAME``
+* ``export NAME = VALUE`` assignments
+* ``export def``
+* ``export class``
+* A module re-export statement, ``from MODULE export NAME``
+
+All five populate ``__export__``, which also becomes ``__all__`` and
+triggers an ``ExportWarning`` on access to anything left out.
+
+PEP 842's version has no wildcard equivalent to `Wildcard form`_. This
+PEP takes only the module re-export statement, deliberately leaving out
+the rest; see `Non-goals`_ for what and why.
+
+
+Motivation
+==========
+
+Widely-used libraries like NumPy, pandas, polars, Typer, FastAPI, and
+Plotly almost universally export their public API using a common
+pattern. A top-level ``__init__.py`` carries a wall of imports from
+private submodules, followed by (or interleaved with) an ``__all__``
+list that repeats the same names as strings, or the reflexive-alias
+idiom used throughout instead. The first looks like this:
+
+.. code-block:: python
+
+   from ._internal.core import PublicAPI as PublicAPI
+   from ._internal.widgets import Widget as Widget
+   from ._internal.errors import SpamError as SpamError
+   # ... often hundreds of lines like this ...
+
+   __all__ = [
+       "PublicAPI",
+       "Widget",
+       "SpamError",
+       # ... the same names again ...
+   ]
+
+This is the file where a library flattens its implementation layout into
+its public layout. As a library grows, the two diverge: code gets
+reorganized into submodules for the maintainer's convenience, while the
+public layout stays stable for users.
+
+Something has to do the flattening. Today that something is a
+hand-maintained, doubly-written list, and such lists routinely drift out
+of sync as the code changes.
+
+The alternative, ``import x as x``, is a workaround for the language's
+missing export concept, and it still trips up some auto-formatters, which
+see a bare ``import x`` as unused and remove it.
+
+
+``__all__`` conflates two concerns
+-------------------------------------
+
+Hand-maintained ``__all__`` also mixes two distinct concerns in one
+file: the list of imports (an implementation detail of how the
+flattening is wired up) and the declaration of the public API (a promise
+to users). The two live in separate statements at different places in
+the file, and nothing keeps them in sync except a reviewer checking them
+against each other by eye, or a linter rule built for exactly this case.
+The more common "unused import" checks don't help: if a name is
+imported but never added to ``__all__``, that name reads as unused, and
+autofixers routinely delete it rather than surface the omission.
+
+
+Underscores solve a different problem
+--------------------------------------
+
+A natural response is: "just prefix internal names with an underscore."
+But the privacy this PEP cares about lives at the package level, not the
+name level: which parts of a large, multi-module package belong in the
+**public layout**. Underscore-prefixing already marks a name private
+within a module.
+
+The problem shows up in "hub" modules (usually ``__init__.py`` files)
+whose only job is gathering names from internal modules and presenting
+them under a stable public name. Every name that reaches a hub module is
+already meant to be public: the underscore convention has done its job
+by the time the name enters the hub. Hub modules need a non-repetitive
+way to say "this is also part of the package's public layout."
+
+
+Non-goals
+=========
+
+This PEP does not aim to:
+
+* Restrict runtime attribute access to non-exported names, or change
+  ``__getattr__`` semantics. See `Why no runtime enforcement`_.
+* Mark a freshly written ``def``, ``class``, or assignment as exported at
+  its definition site, the way the third-party ``atpublic`` package does
+  with ``@public``/``@private`` decorators. See `Why only re-exports`_.
+
+
+Specification
+=============
+
+``export`` is a soft keyword that replaces ``import`` in a ``from`` import
+statement:
+
+.. code-block:: python
+
+   from ._internal.core export PublicAPI
+   from ._internal.widgets export Widget as PublicWidget
+   from numpy.typing export NDArray
+
+A ``from <module> export <name> [as <alias>]`` statement does what ``from
+<module> import <name> [as <alias>]`` does: it binds ``<name>``, or
+``<alias>`` if given, in the current namespace, and also appends that name
+to the module's ``__all__``, creating ``__all__`` if it doesn't already
+exist.
+
+Because it desugars to an ordinary import plus an append to ``__all__``,
+``export`` composes with control flow exactly as ``import`` does:
+
+.. code-block:: python
+
+   if sys.platform == "win32":
+       from ._internal.windows export WindowsThing
+   else:
+       from ._internal.posix export PosixThing
+
+Each branch runs its own import and its own ``__all__`` append, so the
+name that ends up exported depends on which branch ran, with no separate
+``__all__`` bookkeeping required.
+
+``<module>`` may be relative (``from .core export Thing``, ``from
+..sub.core export Thing``) or absolute (``from numpy.typing export
+NDArray``), exactly as in an ordinary ``from ... import ...`` statement.
+A single statement may export multiple names, using the same syntax as a
+regular multi-name ``from`` import, including a parenthesized, multi-line
+list for long ones:
+
+.. code-block:: python
+
+   from ._internal.widgets export Widget, Gadget as PublicGadget
+
+   from ._internal.widgets export (
+       Widget,
+       Gadget,
+       Doohickey,
+   )
+
+Exporting a name is itself a use of it, so tools that flag "imported but
+unused" names (linters, formatters) should treat every name bound by a
+``from <module> export ...`` statement as used, the same way they already
+special-case ``from module import Thing as Thing``. This PEP doesn't
+change what those tools decide. It follows from what ``export`` means: the
+export is the use.
+
+
+Wildcard form
+-----------------
+
+This proposal also includes a wildcard form, ``from <module> export *``.
+It binds every name that ``from <module> import *`` would bind, using the
+same rule (``<module>``'s own ``__all__`` if it defines one, otherwise
+every top-level name that doesn't start with an underscore), and appends
+all of those names to the current module's ``__all__``:
+
+.. code-block:: python
+
+   # spam/_internal/core.py
+   __all__ = ["PublicAPI", "Helper"]  # curated by the internal module itself
+   ...
+
+   # spam/__init__.py
+   from ._internal.core export *
+   # binds PublicAPI and Helper, and adds both to spam.__all__
+
+This supports a common two-tier layout: an internal module curates its own
+``__all__`` as it's written, and the hub re-exports that whole list in one
+statement, instead of naming each item again.
+
+The wildcard form is equivalent to:
+
+.. code-block:: python
+
+   # from ._internal.core export *
+   from ._internal.core import *
+   __all__ = list(globals().get("__all__", [])) + _names_bound_by_star_import
+
+where ``_names_bound_by_star_import`` is the list of names ``from
+._internal.core import *`` just bound, the same list Python's import
+machinery already computes to execute a wildcard import.
+
+
+Lazy exports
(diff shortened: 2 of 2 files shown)


peps/pep-0843.rst
```diff
@@ -0,0 +1,622 @@
...
+
+Non-goals
+=========
+
+This PEP does not aim to:
+
+* Restrict runtime attribute access to non-exported names, or change
+  ``__getattr__`` semantics. See `Why no runtime enforcement`_.
+* Mark a freshly written ``def``, ``class``, or assignment as exported at
+  its definition site, the way :pep:`829`'s ``atpublic`` package does with
```
warsaw: PEP 829 and `atpublic` are really orthogonal.  The latter is a package that has decades long use in the real world.  The connection is that with the next version (likely `atpublic 8.0.0` with an alpha already released), you'll be able to `pip install atpublic[install]` to get the `public()` and `private()` decorators automatically installed into builtins at interpreter startup time.

NeilGirdhar: Thanks for the correction, I'll remove mention of PEP 829!

peps/pep-0843.rst
```diff
@@ -0,0 +1,623 @@
+PEP: 843
+Title: Export Statement for DRY Re-exports
+Author: Neil Girdhar <mistersheik@gmail.com>
+Discussions-To: https://discuss.python.org/t/108353
+Status: Draft
+Type: Standards Track
+Created: 05-Aug-2026
+Python-Version: 3.16
+Post-History: `05-Aug-2026 <https://discuss.python.org/t/108353>`__
+
```
ZeroIntensity, line 11: You should mark this as `Replaces: 842`. If this PEP is accepted, PEP 842 will be marked as "Superseded".

peps/pep-0843.rst
```diff
@@ -0,0 +1,623 @@
+PEP: 843
+Title: Export Statement for DRY Re-exports
+Author: Neil Girdhar <mistersheik@gmail.com>
+Discussions-To: https://discuss.python.org/t/108353
```
ZeroIntensity: Using the 842 thread is fine for the post-history, but you're going to want your own thread for this PEP. Mark this as "Pending" for now; you'll create the thread after merging.

peps/pep-0843.rst
```diff
@@ -0,0 +1,623 @@
...
+Relationship to PEP 842
+========================
+
+Both PEPs start from the same discomfort with ``__all__``. PEP 842 has
+grown that into a broader ``export`` keyword with four forms: standalone
+``export NAME``, ``export NAME = VALUE`` assignments, ``export
+def``/``export class``, and a module re-export statement, ``from MODULE
+export NAME``. All four populate ``__export__``, which also becomes
+``__all__`` and triggers an ``ExportWarning`` on access to anything left
+out.
```
ZeroIntensity: You should probably clarify that `ExportWarning` is specific to 842.

peps/pep-0843.rst
```diff
@@ -0,0 +1,623 @@
...
+
+The last of those forms is, syntactically, almost exactly this PEP's
+proposal: independent confirmation that ``from <module> export <name>``
+is the right shape for re-exports, though PEP 842's version has no
+wildcard equivalent to `Wildcard form`_ and no lazy equivalent to `Lazy
+exports`_. This PEP takes only that one slice, deliberately leaving out
```
ZeroIntensity: PEP 842 allows lazy imports, as it is in the grammar (though I haven't tested the implementation for it). I clarified that in my draft.

.github/CODEOWNERS
```diff
@@ -715,6 +715,7 @@ peps/pep-0837.rst  @serhiy-storchaka
 peps/pep-0838.rst  @AlexWaygood
 peps/pep-0840.rst  @jeremyhylton @gvanrossum
 peps/pep-0842.rst  @ZeroIntensity
+peps/pep-0843.rst  @NeilGirdhar
```
ZeroIntensity: Our (GitHub's?) rule is that you need to be a member of the `python` organization on GitHub to be listed as a code owner. The core dev sponsor is listed here instead.

peps/pep-0843.rst
```diff
@@ -0,0 +1,623 @@
+PEP: 843
```
ZeroIntensity, line 1: For the record, you typically need a core dev sponsor before assigning a number. I'm open to sponsoring, but I want to see what happens with the new 842 revision first.

peps/pep-0843.rst
NeilGirdhar: Thanks for the review, that's a great catch to double check! I looked it up and PEP 12 actually calls this out directly:

▎ Only add this header if the older PEP is in "final" form, i.e. is either Accepted, Final, or Rejected. You aren't replacing an older open PEP if you're submitting a competing idea.

Since PEP 842 is still Draft, and this PEP is a competing/narrower proposal rather than an obsoleting one, it sounds like Replaces isn't the right fit just yet, that header seems meant for after the editors settle which proposal moves forward, at which point the other PEP gets marked Superseded and this one would pick up Replaces. Happy to add it the moment that's warranted! Thanks again for keeping me honest on process here.

peps/pep-0843.rst
NeilGirdhar: That makes perfect sense, thank you!

peps/pep-0843.rst
NeilGirdhar: Awesome!!  No rush on my end. Happy to wait and see how the new PEP 842 revision goes before asking anyone to commit. Thanks for taking the time to look over my PEP so carefully.

.github/CODEOWNERS
NeilGirdhar: Oh, of course!  Removed, thanks!

peps/pep-0843.rst
NeilGirdhar: I agree: the paragraph structure was confusing.  I completely separated the commonalities from the PEP 842-specific points for clarity.

peps/pep-0843.rst
NeilGirdhar: Thanks for clarifying; I've corrected my exposition.

peps/pep-0843.rst
ZeroIntensity: I think "Withdrawn" is missing from that list. If the new revision doesn't go over well, then I'm going to withdraw the PEP and (potentially, depending on feedback) sponsor this one. My suggestion is just to plan for that timeline early. There's not much point in having this PEP at all while 842 is still being actively discussed.

JelleZijlstra: Neil is correct about the formal use for the header, so I think no change is needed here.

NeilGirdhar: I see it a bit differently, and I think it's actually saving everyone time to have 843 around while 842 is still active. Without it, most of my feedback on 842 would probably show up as a string of individual "why not do X instead" comments scattered across the thread. Having them collected into a full proposal lets readers compare two well-argued, fully worked-out approaches side by side, instead of piecing together scattered objections. I'd love to keep it visible and let it keep doing that job while 842 evolves.  What do you think?

peps/pep-0843.rst
JelleZijlstra: Yes, we can keep this open as a courtesy for a bit, but PEPs from non-core devs do need a sponsor.

peps/pep-0843.rst
```diff
@@ -0,0 +1,629 @@
...
+modules convenient for maintainers) from their **public layout** (the
+flat, curated tree they present to users). Building that public layout
+today means choosing between two imperfect options.
+
+The first is writing every exported name twice: once in an import
+statement, again as a string in ``__all__``. The two lists must be kept
+in sync by hand every time the public layout changes.
+
+The second is the reflexive-alias idiom, ``from x import y as y``. Type
+checkers (like mypy even in strict mode) treat it as a signal that the
```
JelleZijlstra: Why "even"? It's a documented part of the type system.

peps/pep-0843.rst
```diff
@@ -0,0 +1,629 @@
...
+PEP 842's version has no wildcard equivalent to `Wildcard form`_. This
+PEP takes only the module re-export statement, deliberately leaving out
+the rest; see `Non-goals`_ for what and why.
+
+
+Motivation
+==========
+
+This section is grounded in existing, widely-used libraries rather than
+a hypothetical ``spam`` module.
```
JelleZijlstra: Strange wording

peps/pep-0843.rst
```diff
@@ -0,0 +1,629 @@
...
+
+
+``__all__`` conflates two concerns
+-------------------------------------
+
+Hand-maintained ``__all__`` also mixes two distinct concerns in one
+file: the list of imports (an implementation detail of how the
+flattening is wired up) and the declaration of the public API (a promise
+to users). The two live in separate statements at different places in
+the file, and a reviewer must check them against each other by eye.
```
JelleZijlstra: Or rely on a linter that warns when they're out of sync.

peps/pep-0843.rst
```diff
@@ -0,0 +1,629 @@
...
+
+A ``@public``-style decorator, as in the third-party ``atpublic``
+package, works neatly for individually defined functions and classes,
+but it doesn't compose with ``import`` statements: there's no object to
+decorate when the "definition" is just a name entering the module
+through an import.
+``atpublic`` works around this with a function-call form,
+``public(some_imported_name)``, but that reintroduces the double-write
+this PEP removes: the name is written once in the import and again as an
+argument to ``public()``. A statement-level ``export`` keyword avoids
```
JelleZijlstra: It also doesn't work well with aliases

peps/pep-0843.rst
ZeroIntensity: I'd be fine with that if this PEP were a different solution. Functionally, this PEP is just a subset of 842, so I'm not sure there'll be many "why not do X instead" comments.

peps/pep-0843.rst
NeilGirdhar: I was trying to emphasize the universality of the rule, but it didn't come across.  Thanks for catching this.

peps/pep-0843.rst
NeilGirdhar: Yeah, this was a weak start.  Grounded in affirmative statements instead.

peps/pep-0843.rst
NeilGirdhar: Good point.  I softened the wording. But often linters do the wrong thing and wipe out the import rather than adding to `__all__`, so linters aren't a perfect fix.

peps/pep-0843.rst
NeilGirdhar: Thank you!!

hugovk:
Please add the new PEP checklist to the top post, and start checking things off:

https://github.com/python/peps/blob/main/.github/PULL_REQUEST_TEMPLATE/Add%20a%20new%20PEP.md

peps/pep-0843.rst
hugovk: We normally only assign numbers when there's a sponsor. There's quite a bit of uncertainty whether this will get one, and some time until we'll know the answer. 

Do we want to assign 843 already to this? Or use the usual 9999 placeholder for now?

peps/pep-0843.rst
hugovk: Yes, discussion should be in another topic than 842, as that one is long enough already!

As there's no sponsor for this one yet, it should be in the Ideas category.

@ZeroIntensity Relatedly, as you're working on a new revision, you may want to lock the old thread (by flagging and asking mods) until you post the new topic. This is something Brett has done to manage his long PEP discussions.

peps/pep-0843.rst
NeilGirdhar: > I'd be fine with that if this PEP were a different solution. Functionally, this PEP is just a subset of 842, so I'm not sure there'll be many "why not do X instead" comments.

It's true that 842 is a much larger change than 843 with significantly different argument supporting it.  That's exactly why people reading 842 might ask "why do we need to do *all* of this?".  That's why I think it's helpful to have 843 around: it formulates a much smaller change we could make that IMO solves the problem.

peps/pep-0843.rst
ZeroIntensity: Good idea -- [request for locking sent](https://discuss.python.org/t/pep-842-module-exports/108353/167).

peps/pep-0843.rst
ZeroIntensity: Perhaps not "assign" 843 to this, but I think that it would be good to "reserve" it for now, solely because if we want a follow-up PEP to 842, 843 is the perfect number for it.

vstinner:
It would be better to get a sponsor for your PEP and merge the PR before mentioning your PEP on a public discussion: https://discuss.python.org/t/export-statement-for-dry-re-exports/108444.

NeilGirdhar:
> It would be better to get a sponsor for your PEP and merge the PR before mentioning your PEP on a public discussion: https://discuss.python.org/t/export-statement-for-dry-re-exports/108444.

Fair point, I'm definitely looking for a sponsor!

peps/pep-0843.rst
warsaw: And soon 844, so it makes sense to just reserve 843 for this PEP so there's some sequence to the three clearly related PEPs.  Since @ZeroIntensity already offered to sponsor I won't need to, but I'd also be happy to.

warsaw:
I can't find the comment in the review now, but via email I saw this comment from @ZeroIntensity:

> You should mark this as Replaces: 842. If this PEP is accepted, PEP 842 will be marked as "Superseded".

I don't think that's quite right though.  `Replaces` and `Superseded` are really there to relate a new PEP that replaces an _existing approved_ PEP.  Since 842 isn't approved (yet), I wouldn't expect those headers to be added to this one.

Of 842, 843, and 844, I would expect whatever combination gets approved would just be marked `Approved` and the others (if any) would be marked `Rejected`.  There's nothing to replace or supersede.

~Maybe we need to clarify this in PEP 12?~

ZeroIntensity:
Yes, Jelle disagreed with me as well.

I still think `Replaces`/`Superseded` is the best choice for 842 if we merge this, because this PEP is just a subset of 842. If it becomes clear that 842 is doing too much, then I'll withdraw it and sponsor this -- hence why I think it "replaces" 842.

hugovk:
Yeah, let's be careful about sticking to the boring old processes for stuff like this, better to have a broader discussion elsewhere if we want to change it.

[PEP 12](https://peps.python.org/pep-0001/#pep-review-resolution) says:

> Add a Replaces header if your PEP obsoletes an earlier PEP. The value of this header is the number of the PEP that your new PEP is replacing. Only add this header if the older PEP is in “final” form, i.e. is either Accepted, Final, or Rejected. You aren’t replacing an older open PEP if you’re submitting a competing idea.

(Now I remember I've also [suggested](https://github.com/python/peps/pull/4819#discussion_r2818712583) Replaced/Superseded-by on a non-final PEP before, so it's good the authors didn't take me up on the idea! 😅)

If PEP 842 is withdrawn: a withdrawal notice at the top is useful, and this can include some prose with reasoning and link to 843.

ZeroIntensity:
Given the accumulating contention surrounding PEP 842, I would like to confirm my sponsorship of PEP 843.

peps/pep-0843.rst
```diff
@@ -0,0 +1,635 @@
...
+Python-Version: 3.16
+Post-History: `05-Aug-2026 <https://discuss.python.org/t/108353>`__
+
+
+Abstract
+========
+
+Large libraries separate their **implementation layout** (the tree of
+modules convenient for maintainers) from their **public layout** (the
+flat, curated tree they present to users). Building that public layout
```
gvanrossum: What's a flat tree? Isn't that an oxymoron? :-)

peps/pep-0843.rst
```diff
@@ -0,0 +1,635 @@
...
+
+The first is writing every exported name twice: once in an import
+statement, again as a string in ``__all__``. The two lists must be kept
+in sync by hand every time the public layout changes.
+
+The second is the reflexive-alias idiom, ``from x import y as y``. This
+is part of the type system: type checkers treat it as a signal that the
+import is an intentional re-export. It still reads like a typo to anyone
+who doesn't know the convention. Because the module has no curated
+``__all__``, it loses the ``dir()`` cleanliness and
```
gvanrossum: AFAIK `dir()` does not use `__all__`. Are you proposing that it should?

peps/pep-0843.rst
```diff
@@ -0,0 +1,635 @@
...
+wildcard-import control a real ``__all__`` gives.
+
+This PEP adds a statement form that avoids both problems:
+
+.. code-block:: python
+
+   # spam/__init__.py
+   from ._internal.core export PublicAPI
+   from ._internal.widgets export Widget as PublicWidget
```
gvanrossum, line 38: If that's all, this is indeed a very short PEP. Since PEP 842 proposes the same syntax (but affecting `__export__` instead) I'm not sure that we need this as a separate PEP, *unless* PEP 842 is dead.

peps/pep-0843.rst
NeilGirdhar: Lol, you got me!  I just meant that it's flattened relative to the implementation tree (fewer levels, curated).  I'll reword to "shallow, curated tree".

peps/pep-0843.rst
NeilGirdhar: You're right, that was wrong — `dir()` just walks `__dict__`, it never consults `__all__`.   Fixed the sentence.

peps/pep-0843.rst
NeilGirdhar: Fair question. Even with 842 very much alive, I think there's value in the narrow, hub-only version standing as its own PEP for a bit. It turns "why not just do what 843 does" from a comment buried in the 842 thread into something people can evaluate directly, syntax and rationale together, without the rest of 842's grammar attached. I laid this out in more detail [here](https://discuss.python.org/t/pep-842-module-exports-new-revision/108460/90). If it turns out 842 is the right vehicle for this syntax, that's a good outcome too. The point was the comparison, not competing for the same slot.

gvanrossum:
Please don’t force-push. It will be squashed when merged.

peps/pep-0843.rst
ZeroIntensity: I will say that PEP 842 is looking very unpromising at the moment.

peps/pep-0843.rst
```diff
@@ -0,0 +1,646 @@
+PEP: 843
+Title: Export Statement for DRY Re-exports
+Author: Neil Girdhar <mistersheik@gmail.com>
+Discussions-To: Pending
```
ZeroIntensity, line 5: List me as a sponsor here and also add me as the codeowner for this file.

peps/pep-0843.rst
```diff
@@ -0,0 +1,646 @@
...
+hand-maintained, doubly-written list, and such lists routinely drift out
+of sync as the code changes.
+
+The alternative, ``import x as x``, is a workaround for the language's
+missing export concept, and it still trips up some auto-formatters, which
+see a bare ``import x`` as unused and remove it.
+
+
+``__all__`` conflates two concerns
+-------------------------------------
```
ZeroIntensity: A bunch of these headers are missing proper underline alignment.

peps/pep-0843.rst
```diff
@@ -0,0 +1,646 @@
...
+form's in `Lazy exports`_.
+
+Like other soft keywords, ``export`` remains a valid identifier everywhere
+except immediately after ``from <module>`` in an import statement.
+
+
+Rationale
+=========
+
+Why a keyword and not a decorator
```
ZeroIntensity, line 418:
```suggestion
Why a keyword and not a decorator?
```

NeilGirdhar: I didn't know this, but it's actually called "an elliptical noun-phrase heading".  It's not supposed to be punctuated with a question mark.

peps/pep-0843.rst
NeilGirdhar: Fixed, thank you!

peps/pep-0843.rst
NeilGirdhar: Done

peps/pep-0843.rst
ZeroIntensity: Up to you. I think it reads better.

peps/pep-0843.rst
```diff
@@ -0,0 +1,647 @@
+PEP: 843
+Title: Export Statement for DRY Re-exports
```
merwok, line 2: Would you consider a different adjective to use before _Re-exports_ here?

Don’t Repeat Yourself is really a principle that _specific logic_ should leave in one place, not be duplicated in different places with inevitable drift causing bugs later on. It does not mean _never repeat a name twice ever_.

NeilGirdhar: Sure, and good point

peps/pep-0843.rst
```diff
@@ -0,0 +1,647 @@
...
+PEP 842 broadens the mechanism into a keyword usable in five forms:
+
+* Standalone ``export NAME``
+* ``export NAME = VALUE`` assignments
+* ``export def``
+* ``export class``
+* A module re-export statement, ``from MODULE export NAME``
+
+All five populate ``__export__``, which also becomes ``__all__`` and
+triggers an ``ExportWarning`` on access to anything left out.
```
ZeroIntensity: PEP 842 in its final revision uses `ExportError`.

peps/pep-0843.rst
```diff
@@ -0,0 +1,647 @@
...
+PEP takes only the module re-export statement, deliberately leaving out
+the rest; see `Non-goals`_ for what and why.
+
+
+Motivation
+==========
+
+Widely-used libraries like NumPy, pandas, polars, Typer, FastAPI, and
+Plotly almost universally export their public API using a common
+pattern. A top-level ``__init__.py`` carries a wall of imports from
```
ZeroIntensity, line 75: It would be nice if you could provide some GitHub links to support this.

ZeroIntensity (commented):
This generally looks good. Let's try to get this up soon, since some people have been confused about why PEP 844 references a PEP that doesn't seem to exist.

peps/pep-0843.rst
ZeroIntensity: This doesn't look like it has been addressed yet. Instead of "DRY", maybe we could say "convenient"? I think "Export Statement for Convenient Re-exports" sounds good.

peps/pep-0843.rst
NeilGirdhar: Fixed

peps/pep-0843.rst
NeilGirdhar: I've been thinking about this, and I think DRY might be the right term:

The export list (in `__all__`) and the import list (of `import` statements) say the same thing twice.  So any rename, addition, or removal has to be made in two places by hand, and the two can silently drift apart. That's the DRY violation this PEP removes by folding both into one `from X export Y` statement.

I've worked this reasoning into the PEP.

peps/pep-0843.rst
NeilGirdhar: Done.

ZeroIntensity:
Please also fix the merge conflicts.

ZeroIntensity (approved):
LGTM. If you're ready to start the discussion, I can merge this.

NeilGirdhar:
Sounds great, thanks for speedy review

ZeroIntensity:
If any of the other reviewers see something they'd like changed, feel free to send a follow-up.

peps/pep-0843.rst
merwok: I still disagree with that, but it’s your PEP!

NeilGirdhar: Fair enough, I figured I'd track down the canonical citation, which is Andrew Hunt and David Thomas, _The Pragmatic Programmer_, where they define DRY as:

“Every piece of **knowledge** must have a single, unambiguous, authoritative representation within a system.”

I guess it depends on whether you think of names as a piece of knowledge.  I do :smile:

merwok: I will concede the point! 🙂

PR #5081, opened 2026-08-06
PEP 1: Correct minor issues in 'PEP Editor Responsibilities & Workflow'

StanFromIreland:
I noticed a little typo ("Alternately") when reading the section. Additionally I fixed a few little other things I noticed.


warsaw (approved):
Thanks!

PR #5082, opened 2026-08-06
PEP 1, 12 and template: update style

hugovk:
We usually don't review PEPs to the same level as the main docs following our [style guide](https://devguide.python.org/documentation/style-guide/), as each PEP is more of a standalone document compared to the docs' editorial whole; and also not to delay publishing and discussing PEPs: authoring and shepherding a PEP is a big task as it is.

However, let's update PEP 1, 12 and especially the template to set an example.

### DIFF
--- a/peps/pep-0001.rst
+++ b/peps/pep-0001.rst
@@ -107,7 +107,7 @@ PEP editors
 -----------
 
 The PEP editors are individuals responsible for managing the administrative
-and editorial aspects of the PEP workflow (e.g. assigning PEP numbers and
+and editorial aspects of the PEP workflow (for example, assigning PEP numbers and
 changing their status).  See `PEP editor responsibilities & workflow`_ for
 details.
 
@@ -163,7 +163,7 @@ Submitting a PEP
 Following the above initial discussion, the workflow varies based on whether
 any of the PEP's co-authors are core developers. If one or more of the PEP's
 co-authors are core developers, they are responsible for following the process
-outlined below. Otherwise (i.e. none of the co-authors are core developers),
+outlined below. Otherwise (when none of the co-authors are core developers),
 then the PEP author(s) will need to find a sponsor for the PEP.
 
 Ideally, a core developer sponsor is identified, but non-core sponsors may also
@@ -210,7 +210,7 @@ The standard PEP workflow is:
   * It is sound and complete.  The ideas must make technical sense.  The
     editors do not consider whether they seem likely to be accepted.
   * The title accurately describes the content.
-  * The PEP's language (spelling, grammar, sentence structure, etc.)
+  * The PEP's language (spelling, grammar, sentence structure, and so on).
     and code style (examples should match :pep:`7` & :pep:`8`) should be
     correct and conformant.  The PEP text will be automatically checked for
     correct reStructuredText formatting when the pull request is submitted.
@@ -331,7 +331,7 @@ responsibility of the Steering Council, which is formally initiated by
 opening a `Steering Council issue`_ once the authors (and sponsor, if any)
 determine the PEP is ready for final review and resolution.
 
-To expedite the process in selected cases (e.g. when a change is clearly
+To expedite the process in selected cases (for example, when a change is clearly
 beneficial and ready to be accepted, but the PEP hasn't been formally submitted
 for review yet), the Steering Council may also initiate a PEP review, first
 notifying the PEP author(s) and giving them a chance to make revisions.
@@ -367,7 +367,7 @@ The Steering Council will generally approve such self-nominations by default,
 but may choose to decline them.
 Possible reasons for the Steering Council declining a
 self-nomination as PEP-Delegate include, but are not limited to, perceptions of
-a potential conflict of interest (e.g. working for the same organisation as the
+a potential conflict of interest (for example, working for the same organisation as the
 PEP submitter), or simply considering another potential PEP-Delegate to be
 more appropriate. If core developers (or other community members) have concerns
 regarding the suitability of a PEP-Delegate for any given PEP, they may ask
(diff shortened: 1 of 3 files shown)


peps/pep-0012/pep-NNNN.rst
```diff
@@ -40,37 +40,37 @@ Rationale
 [Describe why particular design decisions were made.]
 
 
-Backwards Compatibility
+Backwards compatibility
```
warsaw, line 43: I don't think we should change the template.  There's nothing wrong with the existing all-caps headers except that it doesn't conform to our current standard.  Consistency is useful, but what does PEP 8 say about consistency? 😄 

Since we already have so many PEPs that are using this template, changing it does introduce a larger inconsistency!  All those PEPs using the all-caps headers will be inconsistent and I don't think we should go back and change them.

Yes, we have to accept drift when we *add* a new section (such as `Change History`) or re-order them (like `Rationale` moved to after `Specification`), but there were some substantive reasons behind those changes.

peps/pep-0001.rst
```diff
@@ -30,7 +30,7 @@ commands for retrieving older revisions, and can also be browsed
 `on GitHub <https://github.com/python/peps>`__.
 
 
-PEP Audience
+PEP audience
```
warsaw, line 33: Changing PEP 1's headers is fine in isolation since it's not a template, but ... is it worth it?  We have so many all-caps headers in the PEPs, I'm not so sure.

warsaw (changes requested):
I'm -1 on changing the PEP 12 template away from all-caps headers, for the reasons given below.  -0 for changing PEP 1 headers.  The other improvements (e.g. not using .e.g. 😄 ) and fixing the case for `reStructuredText` seem fine though!

peps/pep-0012/pep-NNNN.rst
hugovk: If anything, I'd prefer to drop the changes in PEP 1/12 and only change the template :D

PEP 8 has hobgoblins and also:

> Consistency within a project is more important. Consistency within one module or function is the most important.

Indeed, individual PEPs are more like standalone documents than our main docs, which form an editorial whole. 

And knowing when to be inconsistent:

* We don't apply the style guide as rigorously to PEPs as the docs, and I usually don't bother with the same style nits in PEP reviews.

* Changing the template is for consistency within single PEPs, and for future ones, and not worrying about all the historical ones.

The [docs style guide](https://devguide.python.org/documentation/style-guide/#capitalization) mirrors the PEP 8 text above about consistency in smaller units:

> In the Python documentation, the use of sentence case in section titles is preferable, but consistency within a unit is more important than following this rule. If you add a section to a chapter where most sections are in title case, you can either convert all titles to sentence case or use the dominant style in the new section title.

## PEP 84x

Looking at the 84x PEPs (skipping [840](https://peps.python.org/pep-0840/) that doesn't use the usual template and anyway only has inconclusive single-word headers!):

<table>
<tr>
<th><a href="https://peps.python.org/pep-0841/">841</a></th>
<th><a href="https://peps.python.org/pep-0842/">842</a></th>
<th><a href="https://pep-previews--5078.org.readthedocs.build/pep-0843/">843</a></th>
<th><a href="https://pep-previews--5079.org.readthedocs.build/pep-0844/">844</a></th>
</tr>
<td>
<tr>
 <td valign=top>image
 <td valign=top>image
 <td valign=top>image
 <td valign=top>image
</table>

These authors added their own subheadings in Sentence case, so are now inconsistent with the Title Case main headings. If the main template headings were Sentence case, these would be consistent.

## PEP 83x

Looking at the 83x:

* Title Case headings + Title Case subheadings: [830](https://peps.python.org/pep-0830/), [831](https://pep-previews--5074.org.readthedocs.build/pep-0831/index.html), [835](https://peps.python.org/pep-0835/), [836](https://peps.python.org/pep-0836/)

* Title Case headings + Sentence case subheadings: [832](https://peps.python.org/pep-0832/), [833](https://peps.python.org/pep-0833/), [834](https://pep-previews--4974.org.readthedocs.build/pep-0834/), [837](https://peps.python.org/pep-0837/), [838](https://peps.python.org/pep-0838/), [839](https://pep-previews--5041.org.readthedocs.build/pep-0839/)

Altogether 83x+84x have 10 with Sentence case subheadings, 4 with Title Case.

## PEP xxx

Getting an LLM to categorise the whole bunch:

| Headings      | Subheadings   | PEPs |
|---------------|---------------|-----:|
| Title Case    | Sentence case |  223 |
| Title Case    | Title Case    |  140 |
| Sentence case | Sentence case |   95 |
| Sentence case | Title Case    |    0 |
| Unknown / mixed / no subheadings | | 274 |
| **Total**     |               |  732 |

The majority with Title Case headings use Sentence case subheadings. Changing the default would have improved internal consistency.

I'd argue many of the TC + TC authors _were_ being consistent and just matched the main heading style, so if we flipped the template, they would have remained consistent with Sc + Sc, not Sc + TC.

And interestingly, there are none with Sc + TC.

Flipping the template would match both our style, _and_ authors' own preference.

peps/pep-0001.rst
```diff
@@ -210,7 +210,7 @@ The standard PEP workflow is:
   * It is sound and complete.  The ideas must make technical sense.  The
     editors do not consider whether they seem likely to be accepted.
   * The title accurately describes the content.
-  * The PEP's language (spelling, grammar, sentence structure, etc.)
+  * The PEP's language (spelling, grammar, sentence structure, and so on).
     and code style (examples should match :pep:`7` & :pep:`8`) should be
```
StanFromIreland: Along with Latin abbreviations, I suggest expanding ampersands in prose.

peps/pep-0012/pep-NNNN.rst
willingc: @warsaw @hugovk Here's my perspective:
- I would leave existing PEPs as is and not update their style to match the current style.
- Going forward with the template, I think, at minimum we should recommend all section headings below the top level to be sentence case. 
- As for the top-level headings in the template, I would go ahead and change to sentence case as well, but I don't feel strongly on this point.

My rationale is to give the best experience to the reader. The top-level headings are fairly short so it isn't too bad to leave them as caps. I do think below the top levels the titles sometimes get long and those long titles would have better readability if in sentence case.

PR #5084, opened 2026-08-07
Infra: Add a Sphinx extension to highlight custom keywords

ZeroIntensity:
Follow-up to #5077.

In PEPs that change syntax, new keywords aren't highlighted as such in code blocks. This makes examples look very foreign, which arguably affects the reception of the PEP.

As a solution, this PR introduces a Sphinx extension that adds special code blocks that allow adding new keywords. For example:

```rst
.. code-block:: python+soft-keywords:new_keyword,other_new_keyword

    use the new keyword(s) here
```

### DIFF
--- a/pep_sphinx_extensions/pep_processor/parsing/pep_soft_keyword.py
+++ b/pep_sphinx_extensions/pep_processor/parsing/pep_soft_keyword.py
@@ -1,0 +1,88 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from docutils import nodes
+from pygments.lexers.python import PythonLexer
+from pygments.token import Keyword, Name
+from sphinx import addnodes
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from sphinx.application import Sphinx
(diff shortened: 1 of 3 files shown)


pep_sphinx_extensions/pep_processor/parsing/pep_soft_keyword.py
```diff
@@ -0,0 +1,88 @@
...
+from pygments.token import Keyword, Name
+from sphinx import addnodes
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from sphinx.application import Sphinx
+    from sphinx.environment import BuildEnvironment
+
+_LANG_PREFIX = "python+soft-keywords:"
```
encukou, line 16: Why “soft”? Aren't they always highlighted as keywords?

ZeroIntensity: Yes, they're always highlighted as keywords. I added "soft" to make it clearer that the implementation will be with a soft keyword, but I'm happy to change it.

PR #5093, opened 2026-08-13
Infra: drop Python 3.10

Secrus:
Python 3.10 support is blocking the work on #4926. Drop discussed with and approved by Hugo [here](https://github.com/python/peps/pull/4926#issuecomment-5092537932).


hugovk (approved):
Thanks!

PR #5098, opened 2026-08-16
PEP 835: Refined rationale and ecosystem statistics

till-varoquaux:
This PR updates PEP 835 based on discourse feedback, adding metrics from the ecosystem migration.

### DIFF
--- a/peps/pep-0835.rst
+++ b/peps/pep-0835.rst
@@ -15,7 +15,7 @@ Abstract
 ========
 
 This PEP proposes overloading the ``@`` operator on types to allow writing
-``Annotated[T, M]`` as ``T @M``.
+``Annotated[T, M]`` as ``T @ M``.
 
 Motivation
 ==========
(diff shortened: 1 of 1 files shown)


till-varoquaux:
After this, I think we'll be done for a review by the typing council.

peps/pep-0835.rst
```diff
@@ -442,24 +414,45 @@ syntax for specifying test boundaries::
...
+   to the ``@`` operator.
+
+The required modifications to add support were:
+
+* **FastAPI:** 0 LOCs (inherited from Pydantic)
+* **Pydantic:** 42 LOCs
+* **SQLAlchemy:** 280 LOCs
+* **cattrs:** 153 LOCs
+* **Hypothesis:** 13 LOCs
+* **Beartype:** 10 LOCs
```
ilevkivskyi: Does these number include adding `__rmatmul__` so that `None @ <smth>` will work?

ilevkivskyi (approved):
LG, thanks! I have just one question to clarify.

peps/pep-0835.rst
till-varoquaux: I've updated the PEP to include the `__rmatmul__` cost in the totals. Adding the fallback requires exactly two lines of code per metadata base class:

```python
def __rmatmul__(self, typ: Any) -> Any:
    return Annotated[typ, self]
```

Here is the exact breakdown of the added lines:
- **Pydantic**: +8 LOCs (added to 4 metadata classes)
- **SQLAlchemy, Hypothesis, Beartype, cattrs**: +2 LOCs each
- **FastAPI**: +0 LOCs (inherits from Pydantic)

The ecosystem totals in the PEP now fully reflect this. Thanks for catching this.

ilevkivskyi: Great, thanks!

PR #5100, opened 2026-08-19
PEP 825: address further feedback from DPO

mgorny:
* Change is either:
    * [x] To a Draft PEP

Key changes (from Change History):

- Strengthened the index rules to require that the irrelevant optional
  attributes are not published by the index, and that they are ignored
  by tools.
- Added a suggestion for dealing with missing variants in index-level
  metadata file.
- Indicated that the index-level metadata file can also be used in
  non-standard sources of wheels.


rgommers (approved):
Thanks @mgorny. All LGTM.

PR #5102, opened 2026-08-21
PEP 11: Add `riscv64-unknown-linux-gnu` to Tier 3

StanFromIreland:
It was accepted by the Steering Council in https://github.com/python/steering-council/issues/360.

### DIFF
--- a/peps/pep-0011.rst
+++ b/peps/pep-0011.rst
@@ -135,6 +135,7 @@ powerpc64le-unknown-linux-gnu    glibc, clang                Victor Stinner
 
                                  glibc, gcc                  Victor Stinner
 riscv64-unknown-linux-gnu        glibc, clang                Stan Ulbrych, Emma Smith
+
                                  glibc, gcc                  Stan Ulbrych, Emma Smith
 s390x-unknown-linux-gnu          glibc, gcc                  Victor Stinner
 wasm32-unknown-emscripten        emcc                        Russell Keith-Magee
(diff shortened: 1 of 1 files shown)


peps/pep-0011.rst
```diff
@@ -134,6 +134,7 @@ aarch64-unknown-linux-gnu        64-bit Raspberry Pi OS, gcc Savannah Ostrowski,
 powerpc64le-unknown-linux-gnu    glibc, clang                Victor Stinner
 
                                  glibc, gcc                  Victor Stinner
+riscv64-unknown-linux-gnu        glibc, clang, gcc           Stan Ulbrych, Emma Smith
```
encukou: Elsewhere the comma means *and* or *with*; variants are on sepearate rows.
```suggestion
riscv64-unknown-linux-gnu        glibc, clang                Stan Ulbrych, Emma Smith
                                 glibc, gcc                  Stan Ulbrych, Emma Smith
```

encukou:
ReST, as usual, needs a blank line :(
Other than that, +1 :)

brettcannon (approved):

emmatyping (approved):
We should probably eventually set a baseline of what architectural features we require (rv64gc or rva23 seem like good options) but we can defer that until we have more experience with the platform.

StanFromIreland:
> We should probably eventually set a baseline of what architectural features we require (rv64gc or rva23 seem like good options) but we can defer that until we have more experience with the platform.

Indeed. Currently, all of our buildbots are RV64GC (although we may be able to get some RVA23 machines in the near future), but I'd rather not impose such restrictions until we've had more time to better understand what baseline makes sense.

vstinner:
Note: configure.ac was updated by https://github.com/python/cpython/pull/156277.

PR #5110, opened 2026-08-31
PEP 825: Collected updates

konstin:
This PR collates several updates to PEP 825 (https://discuss.python.org/t/pep-825-wheel-variants-package-format-split-from-pep-817/106196).

## Change History

**26-Aug-2026**

- Added `pep825-metadata-consistency` as an appendix, setting out
  why the `metadata consistency`_ requirements are justified and what
  they cost. Summarized the metadata consistency argument in the
  Rationale.
- Closed the open issue on the use of variant environment markers.
  The investigation in the appendix settles the question in favour of
  markers, so the Open Issues section has been removed and the
  reasoning recorded in the Rationale instead.
- Removed grouping by build numbers, instead relegating them to
  tie-breaking along with other wheel properties. This reverts a
  potential change in behavior for non-variant wheels.
- Added a high-level overview to "variant ordering and selection",
  clarified that the ordering of the lists used in the algorithm is
  from the most preferable to the least preferable and that the
  ordering by platform compatibility tags is not changed.

### DIFF
--- a/peps/pep-0825/appendix-metadata-consistency.rst
+++ b/peps/pep-0825/appendix-metadata-consistency.rst
@@ -1,0 +1,523 @@
+:orphan:
+
+.. _pep825-metadata-consistency:
+
+Appendix: Rationale for the Metadata Consistency Requirements
+=============================================================
+
+This appendix supplements the `Metadata consistency
+<https://peps.python.org/pep-0825/#metadata-consistency>`__ section of
+:pep:`825`. That section states what is required; this one argues that
+the requirement is justified, sets out what it costs, and is explicit
+about the motivation behind it and the trade-offs involved.
+
+It also sets out the reasoning behind the `variant environment markers
+<https://peps.python.org/pep-0825/#variant-environment-markers>`__,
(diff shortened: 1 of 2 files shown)


rgommers (approved):

peps/pep-0825/appendix-metadata-consistency.rst
```diff
@@ -0,0 +1,523 @@
...
+=============================================================
+
+This appendix supplements the `Metadata consistency
+<https://peps.python.org/pep-0825/#metadata-consistency>`__ section of
+:pep:`825`. That section states what is required; this one argues that
+the requirement is justified, sets out what it costs, and is explicit
+about the motivation behind it and the trade-offs involved.
+
+It also sets out the reasoning behind the `variant environment markers
+<https://peps.python.org/pep-0825/#variant-environment-markers>`__,
```
hugovk: Can you replace these absolute links with regular Sphinx references?

mgorny: Done.

PR #5117, opened 2026-09-07
PEP 420: add Note of out date ``pkg_resources``.

azzayou:
add Note for removing ``pkg_resources`` from ``setuptools`` and add refrence to it.


JelleZijlstra:
PEPs are historical documents; they are not kept up to date with later changes.

PR #5126, opened 2026-09-16
PEP 609: Mark as superseded by PEP 772

hugovk:
[PEP 772](https://peps.python.org/pep-0772/) ("Packaging Council governance process") is already marked as "Replaces: 609" ("Python Packaging Authority (PyPA) Governance").

772 is already accepted, the first Packaging Council election concluded yesterday, so this PR applies the matching change for 609.

cc 772 authors: @warsaw @eximious @pradyunsg 

cc new council: @brettcannon @dstufft @henryiii @pradyunsg @rgommers 

Draft for now, in case the authors/council want to do something about the PyPA stuff in 609 first.


brettcannon:
PEP 772 might actually be wrong in replacing PEP 609 as the latter is the PyPA while former while the latter is not specifically about the PyPA (e.g., PEP 772 talks about working with the PyPA, not taking it over).

hugovk:
Yep, in the short term we could instead delete "Replaces: 609" from PEP 772, and then any PyPA changes can be considered independently and in their own time?

warsaw:
> PEP 772 might actually be wrong in replacing PEP 609 as the latter is the PyPA while former while the latter is not specifically about the PyPA (e.g., PEP 772 talks about working with the PyPA, not taking it over).

Agreed

> Yep, in the short term we could instead delete "Replaces: 609" from PEP 772, and then any PyPA changes can be considered independently and in their own time?

Agreed.  Let's let the PPC decide.

PR #5132, opened 2026-09-20
PEP 849: "More Expressive Type Expressions"

ImogenBits:

* [x] Read and followed [PEP 1](https://peps.python.org/1) & [PEP 12](https://peps.python.org/12)
* [x] File created from the [latest PEP template](https://github.com/python/peps/blob/main/peps/pep-0012/pep-NNNN.rst?plain=1)
* [ ] PEP has next available number, & set in filename (``pep-NNNN.rst``), PR title (``PEP 123: <Title of PEP>``) and ``PEP`` header
    * Tip: find the next available number with [pepotron](https://github.com/hugovk/pepotron) — run ``uvx pepotron next`` (or ``pipx install pepotron`` then ``pep next``)
* [x] Core dev/PEP editor listed as ``Author`` or ``Sponsor``, and formally confirmed their approval
* [x] ``Author``, ``Status`` (``Draft``), ``Type`` and ``Created`` headers filled out correctly
* [x] ``PEP-Delegate``, ``Topic``, ``Requires`` and ``Replaces`` headers completed if appropriate
* [x] Code is well-formatted (PEP 7/PEP 8) and is in [code blocks, with the right lexer names](https://peps.python.org/pep-0012/#literal-blocks) if non-Python
* [x] Authors/sponsor added to ``.github/CODEOWNERS`` for the PEP

    * [x] Motivation
    * [x] Specification
    * [x] Rationale
    * [x] Backwards Compatibility
    * [x] Security Implications
    * [x] How to Teach This
    * [x] Reference Implementation
    * [x] Rejected Ideas
    * [ ] Open Issues
    * [ ] Acknowledgements
    * [ ] Footnotes
    * [ ] Change History
* [x] ``Python-Version`` set to valid (pre-beta) future Python version, if relevant
* [x] Any project stated in the PEP as supporting/endorsing/benefiting from the PEP formally confirmed such

I've used 849 since this is a continuation of PEPs 649 and 749 and it's almost the next available number. I'm happy to change it to whatever is next (currently 848) if that is more appropriate.


peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
...
+              `02-Sep-2026 <https://discuss.python.org/t/draft-pep-more-expressive-type-expressions/108856/>`__
+
+
+********
+Abstract
+********
+
+Currently, only a small subset of Python expressions can be used in type
+annotations, which greatly limits the expressiveness of the Python type system
+and its ease of use. This PEP proposes to expand annotate functions such that
```
JelleZijlstra: This framing is a bit strange to me. You can already use any expressions, it's just that many are not introspectable.

peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
+PEP: 849
+Title: More Expressive Type Expressions
+Author: Imogen Hergeth <python at imogen.tech>
+Sponsor: Jelle Zijlstra <jelle.zijlstra at gmail.com>
```
JelleZijlstra, line 4: Confirming my sponsorship

peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
...
+:pep:`586` introduced literal types, which enumerate a concrete list of possible
+literal values, e.g. the numbers ``1``, ``2`` and ``3``. This is currently spelled
+as ``Literal[1, 2, 3]``. Many users intuitively want to instead spell this as
+``1 | 2 | 3``, writing out the bare literals and using the union operator to join
+them, which also is what other languages such as TypeScript use. This is
+currently not possible in Python because the expression ``1 | 2 | 3`` evaluates
+simply to ``3`` because ``|`` is interpreted as the binary-or operation rather than
+the union of types. Further, constant folding completely eliminates this
+expression from appearing anywhere in the generated bytecode, and it thus is
+completely impossible to retrieve the actual type annotation at runtime.
```
JelleZijlstra:
```suggestion
the union of types. Further, constant folding eliminates this
expression from appearing anywhere in the generated bytecode, and it thus is
impossible to retrieve the actual type annotation at runtime.
```

peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
...
+                value = eval_type_annotation_AST(ast.value, namespace)
+                return getattr(value, ast.attr)
+        ...
+
+
+Performance Impact
+==================
+
+While every new feature has to be weighed against its impact on performance and
+complexity, typing related features deserve additional scrutiy because type
```
JelleZijlstra:
```suggestion
complexity, typing related features deserve additional scrutiny because type
```

peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
...
+use annotations, import was moderately faster and the memory footprint slightly
+smaller using the proposed annotate functions, both by a few percent. But
+unfortunately, evaluation time increases by a lot, about seven times as long.
+
+In total, since the negative performance impacts only affect the smallest group
+of users, who also benefit from the newly possible type annotations, we consider
+this a worthwhile trade off. The specification of the proposed format also is open
+enough that many optimizations are possible should they be deemed necessary in
+the future. For many users, this proposal will even be a slight performance
+increase since their code never evaluates any annotate functions.
```
JelleZijlstra, line 238: What about the effect on e.g. dataclasses, which need to evaluate their annotations? Will this make dataclass creation slower?

peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
...
+
+    type _TargetType = <some complex expression>
+    cast(_TargetType, value)
+
+This lets you use any new type expression within ``<some complex expression>``
+since type aliases also are implemented using annotate functions. While this
+workaround is somewhat annoying, the fact that it is needed relatively
+infrequently makes us believe that it is the better solution at this moment.
+We recommend that this usage of intermediate type aliases is observed in the
+future and new syntax to avoid it considered should it be necessary.
```
JelleZijlstra: This sentence is hard to parse

peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
...
+an annotate function that supports all annotation formats by converting the
+passed in values appropriately.
+
+
+***********************
+Backwards Compatibility
+***********************
+
+Since this PEP only adds new functionality, there are no direct backwards
+compatibility concerns. However, we also want to point out that future additions
```
JelleZijlstra: Deprecating get_type_hints is a pretty big compatibility concern

peps/pep-0849.rst
```diff
@@ -0,0 +1,426 @@
...
+format. These approaches were rejected because they force the type-specific
+semantics to be defined in the interpreter itself. This not only limits the
+semantics to e.g. not require namespace lookups and also locks the usage of
+typing features to the Python version that is being used, rather than allowing
+the currently possible backporting via ``typing_extensions``.
+
+Storing Annotation Source Code
+==============================
+
+Instead of storing binary data that defines the annotations' ASTs an alternative
```
JelleZijlstra, line 433: The current approach means that the STRING format will still not faithfully represent the original string though, right? I think that's a minor issue but worth noting.

JelleZijlstra (commented):
Thank you!

peps/pep-0849.rst
ImogenBits: I'll update the wording to clarify that while every expression is in principle usable, many cannot be assigned a meaning in the typing spec because of the introspectability.

peps/pep-0849.rst
ImogenBits: I've added a paragraph explaining the impact of this PEP on these kinds of introspection tools.

peps/pep-0849.rst
ImogenBits: Do you mean that I should add a part in the backwards compatibility section talking about this or drop the deprecation entirely? I don't have any strong opinions either way, I just assumed that since get_type_hints would be largely useless it is better to deprecate it than have a somewhat broken helper function in the standard library.

peps/pep-0849.rst
ImogenBits: I've thought about this but I'm not sure what the best approach is. The AST data can be used to recreate the actual source code for the STRING format (same with FORWARDREF). We wouldn't even have to add much since the current approach already is to construct an AST and then unparse it. We could either make the annotate functions directly do that to support the format or implement another case in get_annotations. I haven't added that to the PEP so far just because it isn't directly related to the main idea behind it, but if it would be better to include it I'd be happy to.

peps/pep-0849.rst
JelleZijlstra: I think deprecating it is fine, but it deserves explicit discussion here.

peps/pep-0849.rst
JelleZijlstra: Oh, I was thinking of the fact that we're losing formatting details (e.g. whitespace), which I assume would still be the case with your proposal.

I definitely think that we should implement STRING on top of AST now. It would be much simpler (essentially just calling `ast.unparse`) and work better in edge cases.

ImogenBits: That's a good point, I've now explicitly pointed this out in the PEP.

peps/pep-0849.rst
```diff
@@ -315,6 +315,14 @@ annotations will need to call ``get_type_annotations`` instead to properly
 resolve any type annotations that contain new typing features. Leaving this
 function as-is will only create confusion about which function should be used.
 
+The AST format also provides a simpler and more reliable method to create
+annotations in the ``STRING`` and ``FORWARDREF`` formats. Currently, compiler
+generated annotate functions do not support these directly, rather the helper
+functions in :py:mod:`annotationlib` try to create a best-effort AST and then
+uparse it into the requested format. Under this PEP these helper functions will
```
JelleZijlstra: unparse

ImogenBits: I've fixed this and some more spelling and grammar mistakes.

peps/pep-0849.rst
```diff
@@ -0,0 +1,462 @@
...
+
+Storing Annotation Source Code
+==============================
+
+Instead of storing binary data that defines the annotations' ASTs an alternative
+is to simply store the annotations source code directly. This also was presented
+as a possibility all the way back in :pep:`649`. The two approaches are largely
+equivalent since one can create the AST from the source code and vice versa.
+While the unparsed AST is not necessarily the exact string that occurred in
+the source code since it can contain different whitespace, it is semantically
```
JelleZijlstra: Nit: not just whitespace, the AST loses a number of other details in the original source code, such as string quoting style, style of string escapes, hexadecimal vs. decimal integers. None of that is likely to matter in type annotations and we don't have to enumerate all the possibilities, better to avoid the incorrect narrowing:

```suggestion
the source code since the AST does not preserve the precise formatting of the code, it is semantically
```

JelleZijlstra:
Seems like there's a CLA issue?

ImogenBits:
Ah right, I figured out the signing, it should work now

